PR #4, opened 2023-03-09
Update BYD-MODBUS-2-LEAF-CAN.ino

JesperSOGT:
Idea to do the scaling safe and keep it in the range.


dalathegreat:
Closing this one, it became obsolete since the software is now split up into different files!

PR #29, opened 2023-07-26
Contactor improvements

dalathegreat:
Improve the code for contactor handling. Make BYD CAN safer

- Contactor startup sequence is now triggered when the LEAF BMS deems it OK to start
- Contactor opening (shutdown) is now handled incase the BMS sends a critical FAULT signal. LilyGo will have to be reset in order to re-engage the contactors. This is added as a safety feature. Only really critical errors result in FAULT state (over/undervoltage, loss of CAN, BMS sending emergency opening of contactors)
- BYD CAN message sending will be stoppped incase a FAULT is detected. This is done to signal to the inverter that some error state has occured. For now we don't know any error bits, so best to just stop sending CAN and force shutdown that way.

### DIFF
--- a/Software/BYD-CAN.cpp
+++ b/Software/BYD-CAN.cpp
@@ -87,6 +87,16 @@ void update_values_can_byd()
   //Temperature min
   BYD_210.data.u8[2] = (temperature_min >> 8);
   BYD_210.data.u8[3] = (temperature_min & 0x00FF);
+
+  //Indicate to inverter if BMS has encountered issues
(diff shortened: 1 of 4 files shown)


dalathegreat:
Direct control of contactors via LilyGo tested on bench setup.

- Verified startup sequence working like it should (Precharge ON, Negative ON, Positive ON, Precharge OFF)
- Verified shutdown sequence working when bms_status == FAULT. Contactors open after 15seconds in fault state

bild

PR #45, opened 2023-08-15
- Added support for Dual CAN Bus Using MCP2515

rjsc:
- Added support for PWM on the contactors
- Added contactor on/off conrol from the inverter
- Moved #defines from Software.ino to config.h to be able to use #ifdefs in SolAX-CAN.cpp


dalathegreat:
Merged into main, massive thanks @rjsc 🥳

PR #56, opened 2023-09-14
Rewrite with central user settings file

dalathegreat:
This rewrite creates a `USER_SETTINGS.h` file that contains all the options and settings. This makes it faster to update to a newer version, without having to re-make all the customizations that you've previously made. It also makes it safer to use the software.

This PR also fixes a bug, SOLAX_CAN could not be compiled without DUAL_CAN option enabled. This is now fixed.

This PR also fixes some Tesla model 3 bugs, contactor not closing and bms_state getting stuck


dalathegreat:
Merging this, has been tested on 3x installs 🙌

PR #61, opened 2023-09-27
Update KIA-HYUNDAI-64-BATTERY.cpp

LevinSwe:
Kona 64 fixes:
- Add PID polling
- Add more vaues from CAN messages


dalathegreat:
Nice work @LevinSwe 🙌

PR #76, opened 2023-11-05
add GitHub Action that compiles Arduino sketch for ESP32

lenvm:
This pull request adds a GitHub Action that runs on each push and each pull request, to check that the Arduino sketch compiles.

It does this by means of the [arduino/compile-sketches](https://github.com/arduino/compile-sketches) action.


dalathegreat (approved):
Looks like a great feature! Let's merge it! :)

PR #79, opened 2023-11-07
Tesla Model3 debug improvements

dalathegreat:
This PR adds the following new functionality/fixes for the Tesla batteries
- Autodetect LFP or NCA/NCM chemistry. 🔋  These two battery types have different voltage limits, and the code now can select the correct limits automatically
- 50 new fault codes added! ⚠️ The error matrix from 0x3AA is now fully decoded, and the software will alert you if one of them is triggered. Example, `"ERROR: Pack discharge current is above the safe max discharge current limit!"` and another example, `"ERROR: shunt current reading is not available"`
- RealSOC now visible in the USB printout. Previously this value was overwritten instantly by the rescaled value heading towards inverter


dalathegreat:
Merging this, tested on 2x Tesla installs!

PR #90, opened 2023-11-11
Add SOC% plausibility detection

dalathegreat:
### What
This PR adds plausibility checks for SOC% on Nissan LEAFs batteries. 

### Why
Using a Nissan LEAF battery below 1A for extended periods of times will cause the BMS to not calculate SOC% correctly. This can happen for instance when using a Checkwatt profile. When this happens, the BMS needs to be rebooted.

### How
We check that if the battery voltage is close to max, and the SOC% is low, we raise a FAULT(red LED) to stop charging/discharging. We also send a message on the USB to inform the user what has happened, so they can take action.

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -242,7 +242,7 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
   }
 
   //Check if SOC% is plausible
-  if (battery_voltage > (ABSOLUTE_MAX_VOLTAGE - 100)) {
+  if (battery_voltage > (ABSOLUTE_MAX_VOLTAGE - 100)) { // When pack voltage is close to max, and SOC% is still low, raise FAULT
     if (LB_SOC < 650) {
       bms_status = FAULT;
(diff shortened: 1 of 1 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -241,6 +241,14 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
     max_target_discharge_power = 0;
   }
 
+  //Check if SOC% is plausible
+  if (battery_voltage > (ABSOLUTE_MAX_VOLTAGE - 100)) {
```
matvey83, line 245: Why deduct 10V from `ABSOLUTE_MAX_VOLTAGE`?

A comment explaining that would be nice.

matvey83: Oh I see, so if the battery voltage is almost 100% but SOC is under 65% we report an error.   Perhaps a comment explaining this, and why this range was chosen would be helpful.

dalathegreat: @matvey83 comment added in 3545c3d !

rjsc (approved):

PR #91, opened 2023-11-11
Rewrite max/min voltage to save memory

dalathegreat:
### What
This PR removes all inverter specific max/min voltage limits, they all use the common one set at startup instead.

### How
All inverter types now use the same `max_voltage `and `min_voltage`. Previously I didn't get this to work due to casting them as const.

### Why
This both saves memory, and makes it easier to add new inverter protocols in the future


Newevg:
Confirming min/max voltage settings with Sungrow SH10RT

Newevg (approved):

PR #92, opened 2023-11-12
Refactor Software.ino

lenvm:
## What
This pull request refactors the `Software.ino` file.

## How
It does this by:
- Reordering the functions at the bottom of the file to reflect the order in `main()`
- Splitting `handle_can()` into `receive_can()` and `send_can()`
- Adding comment sections "Input", "Process", "Output" to the main loop
- Changing the order of declarations in the preamble to match the order in `setup()`
- Adding initialization functions
- Updating comments (while we are at it)

Please find the individual commits to see all consequent changes.

## Why
This refactoring is done to improve the readability of the code. It should be particularly helpful for someone that is not familiar yet with this project, but also existing users may find it helpful.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -14,13 +14,13 @@
 #include "src/lib/miwagner-ESP32-Arduino-CAN/CAN_config.h"
 #include "src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.h"
 
-//Interval settings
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Looks good to me! Will definately help newcomers to the project. Hopefully in the future we can get away from all ifdefs, and have a webui, but now I'm getting ahead of myself :)

PR #93, opened 2023-11-13
Tesla: Increase decimal on Amperage value

dalathegreat:
Noticed by @Newevg , that the scaling on the current reported by Tesla batteries does not contain decimals. This makes the current value used by inverters not match up with reality.

Example, Nissan LEAF pack reports (130) when 13.0A of current is pulled from the battery. The Tesla Model 3 pack report 13 when 13.0A of current is pulled from the battery.

The fix is to multiply the current value with *10 to get the value to match up with the rest of the system


matvey83 (approved):
LGTM

Newevg (commented):
Confirmed, issue is now fixed

PR #94, opened 2023-11-13
New feature: Test fake battery

dalathegreat:
### What
This pull request adds a new feature, which lets you test your system with a fake battery!

### Why
This new test mode can be very useful when testing a new inverter, or testing the inverter before the battery setup is complete, or doing isolation testing for troubleshooting. This means you no longer have to have a working battery on the CAN bus. The user "f1racr" on the Discord had great progress with this test mode!

### How
There is now a new define, `#define TEST_FAKE_BATTERY` that can be enabled in the `USER_SETTINGS.h` file. When enabled, the values set in `TEST-FAKE-BATTERY.cpp` will be sent towards your inverter of choice. While in this test mode, the LED on the board will cycle thru all RGB values, letting you know you are in a special mode.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -79,6 +79,7 @@ uint16_t cell_min_voltage = 3700;  // Stores the minimum cell voltage value in t
 #define YELLOW 1
 #define RED 2
 #define BLUE 3
+#define TEST_ALL_COLORS 10
 
 Adafruit_NeoPixel pixels(1, WS2812_PIN, NEO_GRB + NEO_KHZ800);
 static uint8_t brightness = 0;
@@ -450,6 +451,9 @@ void handle_LED_state() {
       break;
     case RED:
       pixels.setPixelColor(0, pixels.Color(150, 0, 0));  // Red LED full brightness
+      break;
+    case TEST_ALL_COLORS:
+      pixels.setPixelColor(0, pixels.Color(brightness, abs((100 - brightness)), abs((50 - brightness))));  // RGB
       break;
     default:
       break;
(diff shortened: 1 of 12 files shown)


lenvm:
@dalathegreat, to ensure that someone doesn't accidentally break the code for the fake battery in the future, please consider adding a line with TEST_FAKE_BATTERY to the workflows compiling the batteries: https://github.com/dalathegreat/BYD-Battery-Emulator-For-Gen24/blob/271fc8744cedab3b54da1df2180d5f0f33c2cdb9/.github/workflows/compile-all-batteries.yml#L42 and https://github.com/dalathegreat/BYD-Battery-Emulator-For-Gen24/blob/271fc8744cedab3b54da1df2180d5f0f33c2cdb9/.github/workflows/compile-all-combinations.yml#L45

Software/Software.ino
```diff
@@ -79,6 +79,7 @@ uint16_t cell_min_voltage = 3700;  // Stores the minimum cell voltage value in t
 #define YELLOW 1
 #define RED 2
 #define BLUE 3
+#define TEST_ALL_COLORS 10
```
lenvm, line 82: To me it seems a bit strange that the led states are defined twice: once in Software.ino, and once in the battery file. Perhaps not blocking for this PR, but for the future it may be useful to define these states once in a separate header file, which can then be included by both Software.ino and the battery file.

dalathegreat: Great suggestion, I refactored this and put it in a header in commit 2621e29

lenvm (approved):

PR #98, opened 2023-11-16
Add webserver

lenvm:
## What
This pull request adds webserver functionality.
The devboard creates a WiFi Access Point named `Battery Emulator` that can be connected to using the password "123456789". The IP address of the board as Access Point is printed via Serial. The IP address allows you to connect to the webserver via a webbrowser.
The devboard simultaneously connects to an existing network using the `ssid` and `password` settings provided in `webserver.cpp`. If the connection is successful, the IP address obtained is printed via Serial. This IP address you can connect to via a webbrowser, while you are connected to your local network.

The webserver currently displays:
- LED status
- WiFi SSID it is trying to connect to
- WiFi status: connected or not connected
- IP address obtained from the WiFi it has connected to

The webserver also allows for updating over the air. Please see the `README.md` file for details.

## How
The webserver is implemented by using the existing ESP32 `Wifi.h` library.
The over-the-air capability is implemented by adding 3 new libraries:
- me-no-dev/AsyncTCP
- me-no-dev/ESPAsyncWebServer
- ayushsharma82/ElegantOTA

The specific commit of each library is added to the `README.md` file of the respective library. Adding the libraries as a git submodule would be an alternative, however some specific changes to make the libraries work in the directory that they are stored in are needed, therefore this approach of storing the commit used in the `README.md` file is taken, for traceability.

## Why
It is useful to be able to know the state of the devboard without connecting to it using Serial. Now you are able to check the state of the battery emulation from your couch 😎

This PR opens the possibility to add more webserver functionalities, such as:
- displaying more variables, such as state of health, power, etc.
- configuring which battery is used and which inverter protocol shall be emulated
- configuring which WiFi network shall be connected to
- displaying current CAN state
- displaying detailed diagnosis messages for the battery and the inverter, helping users to debug potential issues
- etc.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -40,7 +40,7 @@ static ACAN2515_Buffer16 gBuffer;
 #define MB_RTU_NUM_VALUES 30000
 #endif
 #if defined(LUNA2000_MODBUS)
-#define MB_RTU_NUM_VALUES 50000
+#define MB_RTU_NUM_VALUES 30000
 #endif
 #if defined(BYD_MODBUS) || defined(LUNA2000_MODBUS)
 uint16_t mbPV[MB_RTU_NUM_VALUES];  // Process variable memory
--- a/Software/src/inverter/LUNA2000-MODBUS.h
+++ b/Software/src/inverter/LUNA2000-MODBUS.h
@@ -2,7 +2,7 @@
 #define LUNA2000_MODBUS_H
 #include <Arduino.h>
 
-#define MB_RTU_NUM_VALUES 50000
+#define MB_RTU_NUM_VALUES 30000
 // Definitions for BMS status
 #define STANDBY 0
 #define INACTIVE 1


rjsc:
You must store the string in the program memory explicitly, or else, every string gets loaded to RAM at boot. The RAM is getting filled up with strings plus the many mosbus registers from LUNA protocol.
Check [https://arduino-esp8266.readthedocs.io/en/latest/PROGMEM.html](url)

I would also recommend using a asynchronous webserver implentacion such as [https://github.com/me-no-dev/ESPAsyncWebServer](url)

lenvm:
Thanks @rjsc!

I’ll have a go at implementing the webserver as an asynchronous webserver that saves the HTML and CSS files in SPIFFS, as described in [this tutorial](https://randomnerdtutorials.com/esp32-web-server-spiffs-spi-flash-file-system/).

dalathegreat:
@lenvm The Luna2000 registers are not confirmed to need 50k space. For now we can reduce this line in `LUNA2000-MODBUS.h` and `Software.ino`

#define MB_RTU_NUM_VALUES 30000

This will make it compile.

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -0,0 +1,180 @@
...
+  init_WiFi_AP();
+  init_WiFi_STA(ssid, password);
+
+  // Start webserver
+  server.begin();
+}
+
+void init_WiFi_AP() {
+  const char* ssidAP = "Battery Emulator";  // maximum of 63 characters;
+  const char* passwordAP = "123456789";  // minimum of 8 characters; set to NULL if you want the access point to be open
```
matvey83, line 36: Should probably move these to USER_SETTINGS.h

As well as a boolean flag to enable/disable the web server.

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -0,0 +1,180 @@
...
+        char c = client.read();  // read a byte, then
+        Serial.write(c);         // print it out the serial monitor
+        header += c;
+        if (c == '\n') {  // if the byte is a newline character
+          // if the current line is blank, you got two newline characters in a row.
+          // that's the end of the client HTTP request, so send a response:
+          if (current_line.length() == 0) {
+            // HTTP headers always start with a response code (e.g. HTTP/1.1 200 OK)
+            // and a content-type so the client knows what's coming, then a blank line:
+            client.println("HTTP/1.1 200 OK");
```
matvey83, line 105: Would be nice to pull the code that generates the response HTML into a separate file, as I imagine it will grow to be quite big eventually.

And also just would be nice to have logic for this separated from the rest of the web server logic.

Software/src/devboard/webserver/webserver.cpp
lenvm: Yes, I fully agree with you on both points:

- The `ssid`, `password`, `ssidAP` and `passwordAP` should be user settings.
- There should be a `#define` to determine whether WiFi and the webserver should be active.

I had not implemented these points yet, as I first wanted to figure out how to resolve the issue with the code not compiling for `LUNA2000_MODBUS`.

Software/src/devboard/webserver/webserver.cpp
lenvm: Yes, I agree with you on this as well. The folder structure as suggested by the [ESP32 Web Server using SPIFFS tutorial](https://randomnerdtutorials.com/esp32-web-server-spiffs-spi-flash-file-system/) uses a `data` folder and `index.html` and `ccs.style` files. This seems a good file structure to me.

lenvm:
> @lenvm The Luna2000 registers are not confirmed to need 50k space. For now we can reduce this line in `LUNA2000-MODBUS.h` and `Software.ino`
> 
> #define MB_RTU_NUM_VALUES 30000
> 
> This will make it compile.

Thanks @dalathegreat, that is great to know! In order to also address the other comments in this PR, I am of the opinion that the best way forward is to try to implement the webserver as an asynchronous webserver, with the `HTML` and `CSS` files stored in SPIFFS. I'll give this a go and update this PR accordingly, before modifying the DRAM space defined for Luna2000.

Software/src/devboard/webserver/webserver.cpp
lenvm: @matvey83, I have implemented the two points above.

lenvm:
@dalathegreat I have implemented the code as asynchronous webserver, but I have not implemented the code using SPIFFS, as this is currently [not supported](https://forum.arduino.cc/t/using-the-filesystem-spiffs-with-arduino-ide-2-0-x-is-problematic/1162671/2) in Arduino 2.0.

I have therefore followed your suggestion and changed the value for the Luna2000 registers from 50k to 30k.

dalathegreat:
@lenvm Should we mention that an external antenna should be fitted to the board?
bild
Maybe we could open a Wiki page for the Webserver side of this

StarkJohan:
It's not necessary to connect an external antenna as the board includes a ceramic antenna which is connected by default (the red component marked C3). If an external antenna is to be used, the zero ohm resistor needs to be moved 90 degrees to the alternate position.

lenvm:
@dalathegreat, can you please review this pull request? In my opinion it is ready to be merged.

PR #102, opened 2023-11-20
Solax: Fix cell max volt mapping

dalathegreat:
The Solax protocol, min/max cell voltage had a silly copy paste error, this PR fixes that!


StarkJohan (approved):
Looks good.

Issue #111, opened 2023-11-26
Fronius Gen24 stops working at -10°C


dalathegreat:
### Problem description
The Fronius Gen24 (and other Fronius inverters also affected), will stop charge/discharge if the battery gets colder than -10°C. This is due to the original battery pack (BYD HVM), is a lithium iron phosphate battery, that cannot be charged in cold weather. When using NCM/LMO/NCA batteries, this is not a problem, since these chemistries are OK for outdoor cold use.

![bild](https://github.com/dalathegreat/BYD-Battery-Emulator-For-Gen24/assets/26695010/c3f62291-1dd9-4528-bbef-b8c3b18fe43f)

### Suggested solution
We should freeze (pun intended), the values between -10 to -20°C, and fool the inverter that the battery is warmer than it actually is. Most EV packs have built in heaters, for instance the LEAF pack will turn the heat on at -17°C. This way we can still have a too cold shutdown at -20, which is a far more reasonable limit than -10. Incase you actually see -20°C with a LEAF pack, it means your heater has failed to kick in (and you should look into why)

lenvm:
@dalathegreat, this issue should be fixed in #112. Shall we close this issue?

PR #112, opened 2023-11-26
BYD Modbus: Add freezing temperature overwriting

dalathegreat:
### What
This PR adds temperature checks for BYD Modbus protocol. Incase we are getting too cold, we fake the temperature value for a while

### Why
The Fronius Gen24 (and other Fronius inverters also affected), will stop charge/discharge if the battery gets colder than -10°C. This is due to the original battery pack (BYD HVM), is a lithium iron phosphate battery, that cannot be charged in cold weather. When using NCM/LMO/NCA batteries, this is not a problem, since these chemistries are OK for outdoor cold use. #111

### How
Incase the real temperature is between -9.0 and -20.0C degrees, we fake the value to -9.0C during this value range. When the battery goes below -20.0C , the real value is again forwarded. By doing this we gain an additional 10 degrees of operating range.

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -2,6 +2,7 @@
 
 void update_modbus_registers_byd() {
   //Updata for ModbusRTU Server for BYD
+  verify_temperature_modbus();
   handle_update_data_modbusp201_byd();
   handle_update_data_modbusp301_byd();
 }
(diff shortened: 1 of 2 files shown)


Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -106,3 +107,20 @@ void handle_update_data_modbusp301_byd() {
   static uint16_t i = 300;
   memcpy(&mbPV[i], battery_data, sizeof(battery_data));
 }
+
+void verify_temperature_modbus() {
+  // This section checks if the battery temperature is negative, and incase it falls between -9.0 and -20.0C degrees
+  // The Fronius Gen24 (and other Fronius inverters also affected), will stop charge/discharge if the battery gets colder than -10°C.
+  // This is due to the original battery pack (BYD HVM), is a lithium iron phosphate battery, that cannot be charged in cold weather.
+  // When using EV packs with NCM/LMO/NCA chemsitry, this is not a problem, since these chemistries are OK for outdoor cold use.
```
lenvm, line 115: Perhaps add a check for `LFP_Chemistry`, as it seems that you don’t want to fake the battery temperature when an LFP pack is used, as could be the case for Tesla packs.

dalathegreat: Great suggestion, added in ced6963

lenvm (approved):

PR #113, opened 2023-12-06
Fix update data to sofar

Dimonlipko:
Sofar not get new data from battery, because not call function "update_values_can_sofar()". I add it.


dalathegreat (approved):

dalathegreat:
Thanks @Dimonlipko , merging this!

PR #119, opened 2024-01-02
Tesla: Fix power W calculation

dalathegreat:
### What
Noticed by @MarkoBolt that the stat_bat_power variable is not formatted correctly. This affects all inverters getting wrong values, and also the webserver will show wrong values.

### Why
See how the W is shows as -21313W instead of −8685W
bild

After the fix is applied:
bild

### How
This PR fixes the calculation of the value. It is tricky since the signed16 value should be put in an unsigned register (due to legacy modbus reasons)


dalathegreat:
Forcing this as merged, due to being tested in the Issue it was spawned from!

PR #120, opened 2024-01-03
Nissan LEAF: Heated battery handling

dalathegreat:
# WIP: LEAF Testers welcome!

## What
This PR adds handling for internal battery heaters on Nissan LEAF packs. 

## Why
When the LEAF battery gets cold, it does not want to accept charge. Nissan installed high voltage heating elements inside the battery on cold weather equipped vehicles, that turn on when the temperature goes below -17*C. Unfortunately, this info needs to be acknowledged by the vehicle in order to turn on. This PR adds the commands needed to ack the request to start/stop heating.

## How
We now monitor the variables sent by battery
- LB_HeatExist (Specifies if battery pack is equipped with heating elements)
- LB_Heating_Stop (When transitioning from 0->1, signals a STOP heat request)
- LB_Heating_Start (When transitioning from 1->0, signals a START heat request)
- Batt_Heater_Mail_Send_Request (Stores info when a heat request is happening)

These values are printed out to the USB serial, so that you can check if your battery has the cold weather heaters present. It will also output to USB serial when a request to start/stop heating is happening.

When the start/stop request is sent, we then modify the reply message 0x50B , normally sent by the LEAFs VehicleControlModule. We set the flag "Batt_Heater_Mail_Send_OK" to TRUE when a heating request is coming in. This *SHOULD* allow the battery to work as intended, we unfortunately have no official documentation on how this process works. So testers welcome :)


dalathegreat:
Merging this, installed on my own system!

PR #121, opened 2024-01-04
Tesla S/3/X/Y: Fix Wh max/remain values

dalathegreat:
### What
This PR fixes the WH max and WH remaining value sent towards inverter. It also adds SOC% plausibility detection, same as LEAF has.

### Why
Noticed by f1racer, the Wh value max would sometimes display as 0, and the remaining value would be incorrect. This would mess up Webserver info, along with sending incorrect info towards inverters.
bild

### How
We now use the configured WH number to avoid any u16 overflows. The WH number set in `USER_SETTINGS.h` can be between 0-65000Wh


Newevg (approved):
Perfect, Thank you

dalathegreat:
Thanks for testing! Merging this!
bild

PR #123, opened 2024-01-04
Add way to disable AP

dalathegreat:
### What
This PR adds the ability to disable the Access Point Wifi broadcasting. 📶❌

### Why
Due to cybersecurity reasons, broadcasting an AP might not always be desired

### How
The `USER_SETTINGS.cpp` file now contains a row that controls wheter the AP should be enabled or not

`#define ENABLE_AP //Comment out this line to turn off the broadcasted AP`

This row can be commented out incase you want to disable AP


Newevg (approved):
all working

PR #126, opened 2024-01-07
Bugfix: Charge/Discharge status bits (+refactoring)

dalathegreat:
### What
Noticed by @Cabooman , the `bms_char_dis_status` parameter would never flip correctly between Charging and Discharging, due to u16 / s16 incompability 

### Why
The BYD Modbus handling of this value has been incorrect, which could lead to unwanted desire on inverter side.

### How
This PR fixes the following
- `bms_char_dis_status` parameter now correct depending on current direction flow
- Enumerations added to `bms_char_dis_status` parameter
- Slight memory optimization for `bms_char_dis_status` and `bms_status` variable
- Inverter defines are no longer duplicated in each .header file. Instead all defines to go "../devboard/config.h"

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -33,10 +33,10 @@ void handle_update_data_modbusp201_byd() {
   system_data[0] = 0;  // Id.: p201 Value.: 0 Scaled value.: 0 Comment.: Always 0
   system_data[1] = 0;  // Id.: p202 Value.: 0 Scaled value.: 0 Comment.: Always 0
   system_data[2] =
-      (capacity_Wh_startup);  // Id.: p203 Value.: 32000 Scaled value.: 32kWh Comment.: Capacity rated, maximum value is 60000 (60kWh)
-  system_data[3] = (max_power);  // Id.: p204 Value.: 32000 Scaled value.: 32kWh Comment.: Nominal capacity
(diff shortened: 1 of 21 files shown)


Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -56,12 +56,12 @@ void handle_update_data_modbusp201_byd() {
...
   static uint16_t battery_data[24];
-  if (battery_current > 0) {         //Positive value = Charging
-    bms_char_dis_status = 2;         //Charging
-  } else if (battery_current < 0) {  //Negative value = Discharging
-    bms_char_dis_status = 1;         //Discharging
-  } else {                           //battery_current == 0
-    bms_char_dis_status = 0;         //idle
+  if (battery_current == 0) {
+    bms_char_dis_status = STANDBY;
+  } else if (battery_current > 32768) {  //Negative value = Discharging
```
Cabooman, line 61: 32767 is the maximum positive representation, so
```suggestion
  } else if (battery_current > 32767) {  //Negative value = Discharging
```
would be preferable

Cabooman (changes requested):
Nice with some cleanup!

Only thing I see that needs attention:
"> 32768" in BYD-MODBUS.cpp should be "> 32767"

My comment is pending for some reason, maybe I need to submit my review or something... testing testing

Cabooman (approved):

PR #127, opened 2024-01-07
Bugfix/stable ota

Cabooman:
# Bug: Unstable OTA update in live setting

## Description
OTA updates work well when not connected to RS485/CAN
 ## Root cause
The CAN functionality is the likely culprit from testing outcomes. Likely, the CAN ISR messes with the OTA routine
## Solution
Disabling the CAN peripheral at OTA-start seems to do the trick in a limited testing setup. Disabling the interrupt could also work but this is easy and enough

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -316,6 +316,7 @@ String processor(const String& var) {
 void onOTAStart() {
   // Log when OTA has started
   Serial.println("OTA update started!");
+  ESP32Can.CANStop();
   bms_status = 5;  //Inform inverter that we are updating
   LEDcolor = BLUE;
 }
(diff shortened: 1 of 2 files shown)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -316,6 +316,7 @@ String processor(const String& var) {
 void onOTAStart() {
   // Log when OTA has started
   Serial.println("OTA update started!");
+  ESP32Can.CANStop();
```
dalathegreat, line 319: Very good that Modbus RTU can stay running! The BYD Modbus RTU protocol takes the bms_status value, no other protocol so far is interested in the update status. So this solution is approved! ✅

dalathegreat (approved):

Software/src/devboard/webserver/webserver.cpp
Cabooman: Thanks!

PR #137, opened 2024-01-17
Bugfix: Nissan LEAF, increase precision on values

dalathegreat:
### What
This PR improves the accuracy on both Voltage, Amperage and Power values. Thanks to Eriksson25 over on Discord for noticing! 🙌 

### Why
To get more realistic values towards inverter, and to see more realistic values in logs

### How
Instead of downsampling the values by dividing them by 2 on each CAN read, we instead do more math when it is time to use the values. Here's an example:

Ex1: Real values from battery: Voltage = 370.5V , Amperage 0.5A
- Previous implementation, 371V*1A=371W
- New implementation, 370,5V*0.5A=186W

Ex2: Real values from battery: Voltage = 400.5V , Amperage 5.5A
- Previous implementation, 401V*6A=2406W
- New implementation, 400,5V*5.5A=2203W


eriksson25 (approved):

eriksson25 (approved):
Works perfect!

PR #142, opened 2024-01-24
Feature/k wscaling

dalathegreat:
### What
This PR adds automatic switchover between W and kW visualization in Webserver. The Webserver also shows if contactor closing is allowed by battery/inverter. This info can be useful when troubleshooting why automatic control over contactors is not working.

### Why
Some users were confused by large W numbers. This PR makes it more user friendly

### How
When values are smaller than 1000, they are shown in W. When they exceed 1000, it switches to kW
bild


rjsc (approved):
Tested and working.

PR #143, opened 2024-01-25
Tesla: Add ERROR debug message when inverter blocks contactor closing

dalathegreat:
### What
This PR adds a new fault code via USB to inform user that contactor closing is blocked

### Why
Some people were confused when using double LilyGos or Solax protocol that the contactor would not close. The `inverterAllowsContactorClosing` parameter needs to be true to allow contactor closing.

### How
A USB Serial debug message is added to the Tesla.cpp file


kharnt0x:
LGTM 👍

PR #145, opened 2024-01-28
Bugfix: Tesla / LEAF / BYD CAN

dalathegreat:
### What
This PR improves the following:
- Nissan LEAF: CAN sending optimized for more speed. Safety added for empty battery signal. All debug printouts are now ifdef guarded.
- Tesla: Modified SOH% Warning. Fixed decimal formatting on temperatures.
- BYD CAN: Added USB printout for inverter brand

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -162,8 +162,6 @@ void print_with_units(char* header, int value, char* units) {
 }
 
 void update_values_leaf_battery() { /* This function maps all the values fetched via CAN to the correct parameters used for modbus */
-  bms_status = ACTIVE;  //Startout in active mode
-
   /* Start with mapping all values */
 
   StateOfHealth = (LB_StateOfHealth * 100);  //Increase range from 99% -> 99.00%
@@ -239,6 +237,8 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
     max_target_charge_power = 0;  //No need to charge further, set max power to 0
   }
 
+  bms_status = ACTIVE;  //Startout in active mode
+
   /*Extra safety functions below*/
   if (LB_GIDS < 10)  //800Wh left in battery
   {                  //Battery is running abnormally low, some discharge logic might have failed. Zero it all out.
@@ -251,7 +251,9 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
       (ABSOLUTE_MAX_VOLTAGE - 100)) {  // When pack voltage is close to max, and SOC% is still low, raise FAULT
     if (LB_SOC < 650) {
       bms_status = FAULT;
+#ifdef DEBUG_VIA_USB
       Serial.println("ERROR: SOC% reported by battery not plausible. Restart battery!");
+#endif
     }
   }
 
@@ -259,8 +261,14 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
     max_target_charge_power = 0;
   }
 
+  if (LB_Capacity_Empty) {  //Battery reports that it is fully discharged. Stop all further discharging incase it hasn't already
+    max_target_discharge_power = 0;
+  }
+
   if (LB_Relay_Cut_Request) {  //LB_FAIL, BMS requesting shutdown and contactors to be opened
+#ifdef DEBUG_VIA_USB
     Serial.println("Battery requesting immediate shutdown and contactors to be opened!");
+#endif
     //Note, this is sometimes triggered during the night while idle, and the BMS recovers after a while. Removed latching from this scenario
     errorCode = 1;
     max_target_discharge_power = 0;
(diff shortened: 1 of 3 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -251,16 +251,24 @@ void update_values_leaf_battery() { /* This function maps all the values fetched
       (ABSOLUTE_MAX_VOLTAGE - 100)) {  // When pack voltage is close to max, and SOC% is still low, raise FAULT
     if (LB_SOC < 650) {
       bms_status = FAULT;
+#ifdef DEBUG_VIA_USB
       Serial.println("ERROR: SOC% reported by battery not plausible. Restart battery!");
+#endif
```
matvey83, line 256: Would be nice to create a new function that takes in a string and prints it to serial out if `DEBUG_VIA_USB` is defined, or be a no-op if it isn't.  That would save a bit of duplication.

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
```diff
@@ -167,6 +167,10 @@ void update_values_tesla_model_3_battery() {  //This function maps all the value
     StateOfHealth =
         static_cast<uint16_t>((static_cast<double>(nominal_full_pack_energy) / bat_beginning_of_life) * 10000.0);
   }
+  //If the value is unavailable, set SOH to 99%
+  if (nominal_full_pack_energy < 20) {
```
matvey83, line 171: Why 20?  Any significance to it?   A comment would be nice.

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
```diff
@@ -259,7 +263,7 @@ void update_values_tesla_model_3_battery() {  //This function maps all the value
 
   //Check if BMS is in need of recalibration
   if (nominal_full_pack_energy < 20) {
```
matvey83, line 265: Looks like the `20` can be pulled out into a constant, as its used in 2 places now.

Software/src/inverter/BYD-CAN.cpp
```diff
@@ -172,13 +179,34 @@ void update_values_can_byd() {  //This function maps all the values fetched from
   //Temperature min
   BYD_210.data.u8[2] = (temperature_min >> 8);
   BYD_210.data.u8[3] = (temperature_min & 0x00FF);
+
+#ifdef DEBUG_VIA_USB
+  if (char1_151 != 0) {
+    Serial.print("Detected inverter: ");
```
matvey83, line 185: This is great!

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
rjsc: I'd suggest renaming DEBUG_VIA_USB to DEBUG, and Serial.println to debug.println.
Later this would allow to output debug via USB, MQTT, Web interface, etc.

kharnt0x: I think we should just get this one moving and do this as a future optimization when mqtt merged

It's a great idea though

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
kharnt0x: Let's add a debug/console message here for when SOH is faked

Software/src/inverter/BYD-CAN.cpp
kharnt0x: this is good but just fyi it will print on every can message right now, so it clogs up the output

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
dalathegreat: I agree with @kharnt0x , it is a nice idea, but the scope is growing too much for this small PR!

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
dalathegreat: @matvey83 , good idea, added in [feb45c3](https://github.com/dalathegreat/Battery-Emulator/pull/145/commits/feb45c36057490403fcb37fbd686ad17114228f0)

@kharnt0x , we already have the `"Warning: kWh remaining reported by battery not plausible. Battery needs cycling."` in this scenario, does not make sense to add more that confuses users!

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
dalathegreat: Good idea, added in [feb45c3](https://github.com/dalathegreat/Battery-Emulator/pull/145/commits/feb45c36057490403fcb37fbd686ad17114228f0)

Software/src/inverter/BYD-CAN.cpp
dalathegreat: @kharnt0x , no it prints it only when update_values_can_byd(); is run, every 5 seconds!

kharnt0x (approved):

PR #146, opened 2024-01-28
Feature: Add ability to save parameters to flash

dalathegreat:
### What
This PR adds the ability to permanently store battery limit settings from the webpage. This means that even if the hardware is rebooted or updated via OTA/USB, the settings will continue to exist!

bild

### Why
Users requested this feature

### How
The Preferences.h library is used to store data to flash memory on the ESP32.

To skip reading from flash memory, there is an option in `USER_SETTINGS.h` that you can comment out
`#define LOAD_SAVED_SETTINGS_ON_BOOT  //Enable this line to read settings stored via the webserver on boot`

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -19,7 +19,7 @@
 #include "src/devboard/webserver/webserver.h"
 #endif
 
-Preferences preferences;  // Parameter storage
+Preferences settings;  // Store user settings
 
 // Interval settings
 int intervalUpdateValues = 4800;  // Interval at which to update inverter values / Modbus registers
@@ -105,7 +105,7 @@ bool inverterAllowsContactorClosing = true;
 void setup() {
   init_serial();
 
-  init_storage();
+  init_stored_settings();
 
 #ifdef WEBSERVER
   init_webserver();
@@ -174,35 +174,35 @@ void init_serial() {
   Serial.println("__ OK __");
 }
 
-void init_storage() {
-  preferences.begin("batterySettings", false);
+void init_stored_settings() {
+  settings.begin("batterySettings", false);
 
-#ifdef CLEAR_SAVED_SETTINGS
-  preferences.clear();  // If this clear function is executed, no parameters will be read from storage
+#ifndef LOAD_SAVED_SETTINGS_ON_BOOT
+  settings.clear();  // If this clear function is executed, no settings will be read from storage
 #endif
 
   static uint16_t temp = 0;
-  temp = preferences.getUInt("BATTERY_WH_MAX", false);
+  temp = settings.getUInt("BATTERY_WH_MAX", false);
   if (temp != 0) {
     BATTERY_WH_MAX = temp;
   }
-  temp = preferences.getUInt("MAXPERCENTAGE", false);
+  temp = settings.getUInt("MAXPERCENTAGE", false);
   if (temp != 0) {
     MAXPERCENTAGE = temp;
   }
-  temp = preferences.getUInt("MINPERCENTAGE", false);
+  temp = settings.getUInt("MINPERCENTAGE", false);
   if (temp != 0) {
     MINPERCENTAGE = temp;
   }
-  temp = preferences.getUInt("MAXCHARGEAMP", false);
+  temp = settings.getUInt("MAXCHARGEAMP", false);
   if (temp != 0) {
     MAXCHARGEAMP = temp;
   }
-  temp = preferences.getUInt("MAXDISCHARGEAMP", false);
+  temp = settings.getUInt("MAXDISCHARGEAMP", false);
   if (temp != 0) {
     MAXDISCHARGEAMP = temp;
   }
-  preferences.end();
+  settings.end();
 }
 
 void init_CAN() {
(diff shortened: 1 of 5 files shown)


matvey83 (commented):
Can we also add ability to change battery type and inverter type on the fly?

Software/Software.ino
```diff
@@ -169,6 +174,37 @@ void init_serial() {
   Serial.println("__ OK __");
 }
 
+void init_storage() {
```
matvey83, line 177: A more consistent name could be `load_stored_settings()` or `load_previous_settings()` or "parameters" instead of "settings".

Software/USER_SETTINGS.h
```diff
@@ -36,7 +36,8 @@
 //#define DUAL_CAN              //Enable this line to activate an isolated secondary CAN Bus using add-on MCP2515 controller (Needed for FoxESS inverters)
 //#define SERIAL_LINK_RECEIVER  //Enable this line to receive battery data over RS485 pins from another Lilygo (This LilyGo interfaces with inverter)
 //#define SERIAL_LINK_TRANSMITTER  //Enable this line to send battery data over RS485 pins to another Lilygo (This LilyGo interfaces with battery)
-//#define WEBSERVER  //Enable this line to enable WiFi, and to run the webserver. See USER_SETTINGS.cpp for the Wifi settings.
+#define WEBSERVER  //Enable this line to enable WiFi, and to run the webserver. See USER_SETTINGS.cpp for the Wifi settings.
+//#define CLEAR_SAVED_SETTINGS //Enable this line to clear all data that has been saved via the webserver page
```
matvey83, line 40: How about inverting logic and renaming this to `LOAD_SAVED_PARAMETERS` or `LOAD_SAVED_PARAMETERS_ON_BOOT` or something similar, eg "SETTINGS instead of "PARAMETERS".  If it's set, on boot, saved settings will be loaded, if commented out, saved settings will be ignored.

Software/Software.ino
```diff
@@ -18,6 +19,8 @@
 #include "src/devboard/webserver/webserver.h"
 #endif
 
+Preferences preferences;  // Parameter storage
```
matvey83, line 22: Consider renaming `preferences` to `settings` or `setting_preferences` or something, for consistent terminology.

And lets use same term throughout the code, either call these "settings" or "parameters" or "preferences" but not all 3.

Software/Software.ino
```diff
@@ -669,3 +705,13 @@ void init_serialDataLink() {
   Serial2.begin(9600, SERIAL_8N1, RS485_RX_PIN, RS485_TX_PIN);
 #endif
 }
+
+void storeParameters() {
```
matvey83, line 709: For consistency rename to `store_parameters()`.   Unless you decide to use "settings" as a term instead of "parameters" in which case "store_settings()".

matvey83 (commented):
Would be good to use consistent terminology and either use "settings" everywhere or "parameters".

dalathegreat:
> Can we also add ability to change battery type and inverter type on the fly?

No there is not enough memory on the board! (also a massive risk that software would suddenly switch to wrong protocols)

Software/Software.ino
dalathegreat: Fixed in [93b6a39](https://github.com/dalathegreat/Battery-Emulator/pull/146/commits/93b6a3941eb01b88149da9e26e594ee23970e926)

Software/Software.ino
dalathegreat: compromise to keep init_ , init_stored_settings(); 
[93b6a39](https://github.com/dalathegreat/Battery-Emulator/pull/146/commits/93b6a3941eb01b88149da9e26e594ee23970e926)

matvey83:
> > Can we also add ability to change battery type and inverter type on the fly?
> 
> No there is not enough memory on the board! (also a massive risk that software would suddenly switch to wrong protocols)

I see.  I was mostly interested in changing to "fake battery" from UI, would be helpful when diagnosing issues.  But I can see how that too can be dangerous (a sudden switch to fake battery may cause inverter to over-charge or discharge the battery, in theory).    Maybe just add a comment explaining why battery type and inverter type are not editable from UI?

Software/src/devboard/webserver/webserver.h
```diff
@@ -128,4 +128,7 @@ void onOTAEnd(bool success);
 template <typename T>
 String formatPowerValue(String label, T value, String unit, int precision);
 
+extern void storeSettings();
+extern void restoreSettings();
```
matvey83, line 132: I see no `restoreSettings()` defined; should this be `init_stored_settings()`?

dalathegreat: Whoops, that was a leftover from when I experimented with restoring settings from WebUI. I decided against it, removed the restoreSettings row in [f06a979](https://github.com/dalathegreat/Battery-Emulator/pull/146/commits/f06a9799eaa484216529a496e8aa57c7cdb4b170)

matvey83 (approved):
Looks good, thanks.

PR #149, opened 2024-01-31
Tesla: Fix for high voltage value

dalathegreat:
### What
This PR attempts to fix the "High Voltage Output Pins: 1556V" that some users have been getting

### Why
Misleading information in USB output

### How
Fixed the bitshifting in the CAN reading for 0x2b4, the upper two bits in frame2 were accidentally included!


kharnt0x:
@dalathegreat I've tested this one and it's working! LGTM

PR #151, opened 2024-02-01
Feature: LEAF charger + Cell monitoring

dalathegreat:
### What
This PR adds the following features
- Cell monitoring page in webui (All cellvoltages for LEAF + Tesla)
bild

- Test fake battery now has editable voltage in webui

- LEAF PDM support, for emergency generator charging ⛽  
bild

### Why
Popularly requested features

### How
This is a mess of a PR, coded in a 7h marathon. It will have bugs in the PDM support, but I will squish those later when I test the generator charging at home.

PR #152, opened 2024-02-02
Feature/mqtt on main

Cabooman:
### What
Addition of MQTT subscribe/publish functionality, using the popular Pubsub library, with general data publishing and other examples

### Why
Because we're all Home Assistant junkies whether we admit it (to ourselves) or not ❤️ 
On a more serious note, this allows the publication of data in a common format that can be stored/supervised/visualized or used for various home automation purposes. It also allows control of battery emulator functionality from various home automation frameworks if implemented.

### How
- The PubSub library is added to the library stack.
- Settings for MQTT are added
- MQTT connection/initialization/loop/functions added

Global variables for data communication between inverter and battery are published every 5 s to the 'battery/info' topic (topic name can be changed to something better): SOC, SOH, MaxTemp, MinTemp, MaxCell, MinCell in JSON format, for example:
```
{
  "SOC": 68.500,
  "StateOfHealth": 92.000,
  "temperature_min": 5.600,
  "temperature_max": 7.200,
  "cell_max_voltage": 3812,
  "cell_min_voltage": 3782
}
```

Specific modules can publish whatever they want whenever they want to whatever topic they want. An example is included in 
 NISSAN-LEAF-BATTERY.cpp in the publish_data() function where 96 cell voltages are published every 5 seconds.

MQTT discovery for Home Assistant is available using the approach in NISSAN-LEAF-BATTERY.cpp. It's a bit more involved and is not necessary for Home Assistant integration, but it's really neat if done right


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -917,3 +931,70 @@ uint16_t Temp_fromRAW_to_F(uint16_t temperature) {  //This function feels horrib
...
+
+  // Every 5-ish seconds, build the JSON payload for the state topic. This requires
+  // some annoying formatting due to C++ not having nice Python-like string formatting.
+  // msg_length is a cumulative variable to track start position (param 1) and for
+  // modifying the maxiumum amount of characters to write (param 2). The third parameter
+  // is the string content
+  size_t msg_length = snprintf(mqtt_msg, sizeof(mqtt_msg), "{\n\"cell_voltages\":[");
+  for (size_t i = 0; i < 97; ++i) {
+    msg_length +=
+        snprintf(mqtt_msg + msg_length, sizeof(mqtt_msg) - msg_length, "%s%d", (i == 0) ? "" : ", ", cell_voltages[i]);
```
dalathegreat, line 991: @Cabooman , the cell_voltages[%d] is now mapped to a globally available cellvoltages[] array in this PR: https://github.com/dalathegreat/Battery-Emulator/pull/151/files

Just thinking, maybe we can get rid of this battery specific implementation?

Cabooman: Yeah, that makes sense. I have a few thoughts.

1. I'd like to keep some battery specifics since there's still some interesting data to look at and it might vary from battery to battery, like interlock status or pyrofuse this-and-that
2. Cell voltages should move out, definitely. The issue is that the cell number really should be known at startup, otherwise we don't know how many sensors to set up via discovery. That basically means creating a separate variable that can be set by the battery part, shouldn't be an issue
3. 151 should go in first, then I rebase with main and take it from there

Software/src/battery/BATTERIES.h
```diff
@@ -19,6 +21,7 @@
 
 #ifdef NISSAN_LEAF_BATTERY
 #include "NISSAN-LEAF-BATTERY.h"  //See this file for more LEAF battery settings
+#define BATTERY_HAS_INIT
```
dalathegreat, line 24: This name BATTERY_HAS_INIT is not telling what is being done. Maybe INIT_CELL_AMOUNT?

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -936,3 +939,7 @@ uint16_t Temp_fromRAW_to_F(uint16_t temperature) {  //This function feels horrib
   }
   return static_cast<uint16_t>(1094 + (309 - temperature) * 2.5714285714285715);
 }
+
+void init_battery(void) {
```
dalathegreat, line 943: same with this function name, init_battery -> init_amount_of_cells?

Software/src/battery/BATTERIES.h
Cabooman: Given that we're getting a lot of different functionality in that affects different batteries differently, I didn't want to specify it as "do_specific_thing_at_startup" but rather "do_startup_stuff". That way each battery would typically have a generic startup function, a can receive function, and so on. Basically closing in on a more common interface

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
Cabooman: See other comment

Software/src/devboard/mqtt/mqtt.h
```diff
@@ -0,0 +1,62 @@
...
+#ifndef __MQTT_H__
+#define __MQTT_H__
+
+#include <Arduino.h>
+#include "../../../USER_SETTINGS.h"
+
+#define MSG_BUFFER_SIZE (1024)
+#define MQTT_SUBSCRIPTIONS \
+  { "my/topic/abc", "my/other/topic" }
+#define MQTT_SERVER "192.168.xxx.yyy"
```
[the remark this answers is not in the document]
Cabooman: Moved!

dalathegreat (approved):

PR #153, opened 2024-02-02
add wifi monitoring loop

kharnt0x:
This PR adds a WiFi monitoring loop using a state machine to be non-blocking. 

This ensures that if wifi goes down or is flakey, the Battery Emulator will try to reconnect without blocking the main features. 

Currently will run every 30 seconds

Also improves the wifi initial connection in the same way meaning we can boot faster and start talking to the battery/inverter straight away

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -148,6 +148,7 @@ void loop() {
 #ifdef WEBSERVER
   // Over-the-air updates by ElegantOTA
   ElegantOTA.loop();
+  WiFi_monitor_loop();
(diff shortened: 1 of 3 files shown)


dalathegreat:
Checked out and verified working. Turned off WifiRouter for a few minutes, and then powered it back on. The board managed to reconnect 🙌 

Also, getting rid of the `delay(500);` block was just icing on the cake! Nice work, approved! ✅

dalathegreat (approved):

PR #154, opened 2024-02-04
Pylon CAN: Fix Sofar voltages

dalathegreat:
### What
This PR adds a toggle for Sofar inverters using Pylon CAN. Thanks to Pazdzierz over on the Discord for discovering this bug!

### Why
The Min/Max voltages sent on CAN were not interpreted correctly by the inverter, blocking charging/discharging

bild

### How
The PYLON-CAN.cpp file now has a toggle for this inverter

`#define INVERT_VOLTAGE  //If defined, the min/max voltage frames will be inverted, useful for some inverters like Sofar that report the voltages incorrect otherwise`


dalathegreat:
Confirmed working! This PR fixes the Sofar inverters! ✅ 

bild

PR #158, opened 2024-02-06
BYD CAN: Fix average temperature calculation

dalathegreat:
### What
User brennar10 noticed a bug in the temperature calculation when using BYD CAN. When values go negative, there is a risk that the signed bit is lost during the calculation, causing temperature to appear incorrect

### Why
bild

### How
The bug is fixed by converting the uint16 values to local int16 values before performing any calculations.


dalathegreat:
Confirmed working by brennar10 🥳 
bild

matvey83 (commented):
(Forgot to "submit" review earlier, but better late than never.)

PR #160, opened 2024-02-07
Nissan LEAF: Fix temperature value so its human readable

dalathegreat:
### What
This PR makes negative temperature values display OK on LEAF batteries

### Why
Users did not like to see 6553*C in the USB log

### How
Value is casted to int16 before showing it


dalathegreat:
Tested by Hinken, confirmed working!

PR #166, opened 2024-02-09
Fix Wifi loop more, Event log, more mqtt autoconfig for homeassistant

kharnt0x:
- Improve Wifi monitor loop and made it solid (tested on my flakey wifi). Use inbuilt esp32 logic for reconnects. Only intervene as necessary
- Add ability to set wifi channel if known. This reduces connect time by 3-4 seconds according to ESP32 docs
- Add wifi channel, signal strength and hostname to webui as these are useful for debugging connection issues
- Add event log (thanks to @Cabooman ) and webui endpoint behind ifdef
- Add MQTT autoconfig for homeassistant for other data than just cell voltages
- Use global battery emulator version number for mqtt (move it to software settings)
- Remove unit tests from @Cabooman 's stuff here for now as i could not get them working in github actions or locally. When we gets them working maybe they can be put in a separate PR
- (MODEL 3) don't set light to yellow for remaining capacity warning as this seems to occur for majority of users (capacity reported as 1). Still log an INFO message and something to the new event log for now


Software/USER_SETTINGS.h
```diff
@@ -46,6 +46,10 @@
 #define MQTT_SERVER "192.168.xxx.yyy"
 #define MQTT_PORT 1883
 
+/* Event options*/
+#define EVENTLOGGING               //Enable this line to log events to the event log
```
dalathegreat, line 50: Is there a reason why one would want to turn off eventlogging? I think we could remove this define

Software/USER_SETTINGS.h
```diff
@@ -46,6 +46,10 @@
...
+#define DUMMY_EVENT_ENABLED false  //Enable this line to have a dummy event that gets logged to test the interface
```
dalathegreat, line 51: Let's remember to remove this before the next big release 😅 But good for testing now!

dalathegreat (approved):

Software/USER_SETTINGS.h
kharnt0x: not really if it's tested and working. I just put it there as it's a brand new feature and I'm unsure how it could impact stability

kharnt0x: for me it works fine, the only concern I had is over time iterating the event log counts. but I guess eventually it will just go back to 0 if it hits max?

kharnt0x: actually as occurances is only a u_int_8 I just saw it overrun and reset to 0. all working fine

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -111,12 +95,101 @@ static void publish_cell_voltages(void) {
...
+     "Battery Emulator Stat Batt Power", "{{ value_json.stat_batt_power }}", "W", "power"},
+    {"battery_current", "homeassistant/sensor/battery-emulator/battery_current/config",
+     "Battery Emulator Battery Current", "{{ value_json.battery_current }}", "A", "current"},
+    {"cell_max_voltage", "homeassistant/sensor/battery-emulator/cell_max_voltage/config",
+     "Battery Emulator Cell Max Voltage", "{{ value_json.cell_max_voltage }}", "V", "voltage"},
+    {"cell_min_voltage", "homeassistant/sensor/battery-emulator/cell_min_voltage/config",
+     "Battery Emulator Cell Min Voltage", "{{ value_json.cell_min_voltage }}", "V", "voltage"},
+    {"battery_voltage", "homeassistant/sensor/battery-emulator/battery_voltage/config",
+     "Battery Emulator Battery Voltage", "{{ value_json.battery_voltage }}", "V", "voltage"},
+};
```
Cabooman, line 132: This is a really nice way of templating discovery! One request though, move the declarations and definitions to before the start of the function definitions, to keep types in one continuous block, variables in another and functions in another

Cabooman (approved):

PR #169, opened 2024-02-11
Feature: Webserver cleanup + Event timestamps

dalathegreat:
### What
This PR does some refactoring on the Webserver code. Instead of one long file over 1000s lines long, each page now has its own html.cpp file.

The PR also adds timestamps to events logged, making for easier troubleshooting

The PR also minifies some HTML code, to increase performance

### Why
Making code simpler to maintain

### How
Timestamp calculation is performed on client side, since Battery-Emulator does not know time:

bild

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -130,9 +130,7 @@ void setup() {
   init_webserver();
 #endif
 
-#ifdef EVENTLOGGING
   init_events();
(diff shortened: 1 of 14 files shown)


Cabooman (approved):
Nice step forward for the web UI 👍

PR #176, opened 2024-02-12
Polishing the EEPROM handling

Cabooman:
### WHAT
Improved EEPROM handling, event log handling in memory and minimizing the number of writes to the flash-emulated EEPROM

### WHY
To clear and re-initialize the event log if it seems incompatible/weird and to reduce the number of writes to EEPROM

### HOW
Reducing writes: A counter counts the number of events that are logged. A timer keeps an XX minute timeout. Logged events are only written if the defined time has passed AND we have logged any events.
Log init: Check for a magic number, clear the log if the number differs from what you expect. Could be an older version, junk, etc.

## TESTING
Tested on the dining room table, currently live in my setup. Appears to work as intended, i.e. only write to the log if there are events to write.

Serial print at startup (i.e. showing the event log):
image

View of events set since startup in web UI (NOT fetched from the event log, those are currently only visible in the message at boot):
image


dalathegreat (approved):
LGTM!

PR #177, opened 2024-02-12
Bugfix: Clear battery events properly

dalathegreat:
### What
With the new introduction of Events, there are some scenarios that can never be recovered from. This PR fixes recovery for minor things, but keeps the latching for the critical stuff like under/overvoltage. 

We also now zero out allowed charge/discharge power for LEAF batteries incase we enter a critical fault state. Previously it was possible to continue operation even though a critical fault was active 😱 

### Why
To improve safety

### How
New info events are added for when battery is full/empty, and latching verified for critical safeties.


Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
```diff
@@ -269,23 +272,27 @@ void update_values_tesla_model_3_battery() {  //This function maps all the value
 
   if (LFP_Chemistry) {  //LFP limits used for voltage safeties
     if (cell_max_v >= MAX_CELL_VOLTAGE_LFP) {
-      set_event(EVENT_CELL_OVER_VOLTAGE, 0);
+      set_event(EVENT_CELL_OVER_VOLTAGE, cell_max_v);
```
Cabooman, line 275: for all the `set_event(xxx, cell_min/max_v);`, the data field is just 1 byte so this will wrap. We can expand the data field which will cause some side effects, or instead log mV above max, like `cell_max_v - MAX_CELL_VOLTAGE_LFP`

Cabooman (commented):
Good stuff!

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
kharnt0x: all looks good, just this one to address that I could see

dalathegreat: Fixed in [bc9a879](https://github.com/dalathegreat/Battery-Emulator/pull/177/commits/bc9a879606829758321506658d3fc21896f4f54a)

dalathegreat:
Latest main + this PR now running on my 30kWh LEAF battery! It feels much safer already, caught a cell undervoltage at -17*C much better than the old software 🙌

PR #188, opened 2024-02-16
Battery interface streamlining

Cabooman:
### WHAT
Common interfaces of batteries

### WHY
To enable cleanup of Software.ino with tons of #ifdefs, and for the sake of having clearly defined interfaces

### HOW
All battery files implement/define 4 functions of which 3 were previously common for all batteries with different names. The function declarations are done in BATTERIES.h. This enforces the interface definition of new batteries (somewhat). The functions currently required by the common interface are
``` 
void update_values_battery();
void receive_can_battery(CAN_frame_t rx_frame);
void send_can_battery();
void announce_battery(void);
```
The same can later be done for inverters

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -364,7 +364,7 @@ void inform_user_on_battery() {
 #ifdef SERIAL_LINK_RECEIVER
   Serial.println("SERIAL_DATA_LINK_RECEIVER selected");
 #endif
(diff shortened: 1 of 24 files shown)


dalathegreat (approved):
Anything that shaves 126 lines out of the Software.ino is a massive win! Looks like it will be easier to maintain this way :)

PR #190, opened 2024-02-17
All batteries: Voltage limits as variables

dalathegreat:
### What
This PR makes the minimum/maximum voltages variables instead of defines. Tesla batteries benefit from this.

### Why
Some batteries have multiple variations in chemistry, so we need to edit the limits at runtime. For instance LFP batteries have different voltage range compared to NCM.

### How
The newly combined init_battery function sets the values once at startup. It can be later changed in runtime by writing max_voltage and min_voltage to the desired values.


Newevg (approved):
Works for LFP Tesla

PR #191, opened 2024-02-17
Tesla: Tweak plausable SOC% limits

dalathegreat:
### What
This PR fixes a bug where LFP Tesla packs would trigger the SOC% not plausable

### Why
Due to the changes in PR #190 , the LFP pack would be just on the limit when it comes to autodetecting SOC% not plausable. LFP has a very flat voltage curve, so the check no longer makes sense for LFP

### How
We fix this by lowering the SOC% check, and also raising the voltage requirement.

### DIFF
--- a/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
+++ b/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
@@ -264,8 +264,8 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   //Once cell chemistry is determined, set maximum and minimum total pack voltage safety limits
   if (LFP_Chemistry) {
-    max_voltage = 3640;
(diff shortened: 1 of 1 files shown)


Newevg (changes requested):
Incorect MAX/MIN was preventing LFP 108pack from Charging:
 
//Once cell chemistry is determined, set maximum and minimum total pack voltage safety limits
  if (LFP_Chemistry) {
    max_voltage = 3880;
    min_voltage = 2968;
  } else {  // NCM/A chemistry
    max_voltage = 4030;
    min_voltage = 3100;
  }

dalathegreat:
@Newevg Revised the limits in [8f932e1](https://github.com/dalathegreat/Battery-Emulator/pull/191/commits/8f932e189e25b2a4168241e3df002b73ed013924)

PR #195, opened 2024-02-19
Refactor: Variable naming + RealSOC

dalathegreat:
### What
This PR makes the code easier for beginners to join in. It also adds "Real SOC%" as a parameter.

bild

⚠️ This PR changes 1300 lines of code. Bugs might have been introduced ⚠️

### Why
Reduce software complexity to make Battery-Emulator easier to maintain.

### How
- SMA_TRIPOWER_CAN inverter protocol added to Github workflows
- SERIAL_LINK_TRANSMITTER protocol added to Github workflows
- All global variables now have prefix system_
- Postfix added to all global variables to signal units
- LEAF/Kangoo batteries can now push 60kW instead of 30kW
- Removed mentions of Fronius in battery files
- Larger than 65kWh packs can now be configured (uint16 -> uint32)
- No more 65535-val fiddling with negative values. All values can be entered as normal human readable (-15)
- Batteries only write REAL soc. Scaling removed from each battery implementation.
- Webserver shows real and scaled SOC. Also added parameter where you can configure if you want to scale or not.
- Event added for when Solax requests contactor opening


Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -170,20 +170,21 @@ static void publish_common_info(void) {
...
-             "  \"battery_voltage\": %d\n"
+             "  \"system_scaled_SOC_pptt\": %.3f,\n"
+             "  \"system_SOH_pptt\": %.3f,\n"
+             "  \"system_temperature_min_dC\": %.3f,\n"
+             "  \"system_temperature_max_dC\": %.3f,\n"
+             "  \"system_active_power_W\": %.3f,\n"
+             "  \"system_battery_current_dA\": %.3f,\n"
+             "  \"system_cell_max_voltage_mV\": %.3f,\n"
+             "  \"system_cell_min_voltage_mV\": %.3f,\n"
+             "  \"system_battery_voltage_dV\": %d\n"
```
Cabooman, line 181: This string is a mapping for Home Assistant and not visible to the user, changing it will break users' configurations and requires more changes. I suggest keeping it as is (was?)

Cabooman (approved):

Software/Software.ino
```diff
@@ -239,7 +241,10 @@ void init_stored_settings() {
   temp = settings.getUInt("MAXDISCHARGEAMP", false);
   if (temp != 0) {
     MAXDISCHARGEAMP = temp;
-  }
+    temp = settings.getBool("USE_SCALED_SOC", false);
+    USE_SCALED_SOC = temp;  //This bool needs to be checked inside the temp!= block
+  }                         // No way to know if it wasnt reset otherwise
```
Cabooman, line 246: Is this because we don't know if the settings have been written? In that case the settings could include some non-trivial flag/magic number to flag that proper settings have been written. I can imagine that someone will at some point in time set MAXDISCHARGEAMP to 0 "because of reasons" and then not get the SOC scaling settings applied

Cabooman (approved):

Software/src/devboard/mqtt/mqtt.cpp
dalathegreat: Reverted in [3fb362a](https://github.com/dalathegreat/Battery-Emulator/pull/195/commits/3fb362af09b0ebc318b8f68ee7c2050a3f97c4d5)

Software/Software.ino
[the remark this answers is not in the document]
dalathegreat: Good idea, BUT, now peoples preferences are already populated by these names, so a bit late to change it now! (same as MQTT which I reverted :) )

Software/Software.ino
dalathegreat: Yes. I think we can re-think the settings in a separate PR!

PR #199, opened 2024-02-21
Feature: Add top balancing for Tesla batteries

dalathegreat:
### What
This PR adds top balancing feature for teslas, that reduce allowed charge power incase a cell starts to get too high

### Why
Charging with high power when nearing 100% causes some cells to go into overvoltage protection mode, and prevent the battery from balancing. 

### How
Three new parameters are added
- RAMPDOWNSOC 900 // 90.0 SOC% to start ramping down from max charge power towards 0 at 100.00%
- FLOATPOWERMAX 200 // W, what power to allow for top balancing battery
- MILLIVOLTFLOAT 20  // mV, how many mV under cell max voltage to start float charging

With these three parameters, the charge curve should look something more like this:
bild

### DIFF
--- a/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
+++ b/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
@@ -202,8 +202,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
   //The allowed charge power behaves strangely. We instead estimate this value
   if (system_scaled_SOC_pptt == 10000) {  // When scaled SOC is 100.00%, set allowed charge power to 0
     system_max_charge_power_W = 0;
-  }
-  if (soc_vi > 990) {
+  } else if (soc_vi > 990) {
     system_max_charge_power_W = FLOAT_MAX_POWER_W;
   } else if (soc_vi > RAMPDOWN_SOC) {  // When real SOC is between RAMPDOWN_SOC-99%, ramp the value between Max<->0
     system_max_charge_power_W = MAXCHARGEPOWERALLOWED * (1 - (soc_vi - RAMPDOWN_SOC) / (1000.0 - RAMPDOWN_SOC));
--- a/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
+++ b/Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
@@ -156,9 +156,9 @@ static const char* hvilStatusState[] = {"NOT OK",
 #define MIN_CELL_VOLTAGE_NCA_NCM 2950   //Battery is put into emergency stop if one cell goes below this value
 #define MAX_CELL_DEVIATION_NCA_NCM 500  //LED turns yellow on the board if mv delta exceeds this value
 
-#define MAX_CELL_VOLTAGE_LFP 3520   //Battery is put into emergency stop if one cell goes over this value
+#define MAX_CELL_VOLTAGE_LFP 3550   //Battery is put into emergency stop if one cell goes over this value
 #define MIN_CELL_VOLTAGE_LFP 2800   //Battery is put into emergency stop if one cell goes below this value
-#define MAX_CELL_DEVIATION_LFP 150  //LED turns yellow on the board if mv delta exceeds this value
+#define MAX_CELL_DEVIATION_LFP 200  //LED turns yellow on the board if mv delta exceeds this value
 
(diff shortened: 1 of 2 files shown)


Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
```diff
@@ -200,10 +200,23 @@ void update_values_battery() {  //This function maps all the values fetched via
   }
 
   //The allowed charge power behaves strangely. We instead estimate this value
-  if (system_scaled_SOC_pptt == 10000) {  // When scaled SOC is 100%, set allowed charge power to 0
+  if (system_scaled_SOC_pptt == 10000) {  // When scaled SOC is 100.00%, set allowed charge power to 0
     system_max_charge_power_W = 0;
-  } else if (soc_vi > 950) {  // When real SOC is between 95-99.99%, ramp the value between Max<->0
-    system_max_charge_power_W = MAXCHARGEPOWERALLOWED * (1 - (soc_vi - 950) / 50.0);
+  }
+  if (soc_vi > 990) {
```
matvey83, line 206: Should this be `else if`, to prevent the `system_max_charge_power_W = 0` from being overwritten below?

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
```diff
@@ -200,10 +200,23 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+    //If the cellvoltages start to reach overvoltage, only allow a small amount of power in
+    if (system_LFP_Chemistry) {
+      if (cell_max_v > (MAX_CELL_VOLTAGE_LFP - MILLIVOLTFLOAT)) {
+        system_max_charge_power_W = FLOATPOWERMAX;
+      }
+    } else {  //NCM/A
+      if (cell_max_v > (MAX_CELL_VOLTAGE_NCA_NCM - MILLIVOLTFLOAT)) {
+        system_max_charge_power_W = FLOATPOWERMAX;
+      }
+    }
```
matvey83, line 219: Lets refactor this as:

```
max_allowed_cell_voltage = system_LFP_Chemistry ? MAX_CELL_VOLTAGE_LFP : MAX_CELL_VOLTAGE_NCA_NCM

if (cell_max_v > (max_allowed_cell_voltage - MILLIVOLTFLOAT)) {
         system_max_charge_power_W = FLOATPOWERMAX;
}
```

Software/src/battery/TESLA-MODEL-3-BATTERY.h
```diff
@@ -7,6 +7,9 @@
 
 #define BATTERY_SELECTED
 
+#define RAMPDOWNSOC 900              // 90.0 SOC% to start ramping down from max charge power towards 0 at 100.00%
+#define FLOATPOWERMAX 200            // W, what power to allow for top balancing battery
+#define MILLIVOLTFLOAT 20            // mV, how many mV under overvoltage to start float charging
```
matvey83, line 12: Let's name constants so that units are added to the end of the name?  Perhaps FLOAT_START_MV would be a better name for this constant?
And FLOATPOWERMAX can be renamed to FLOAT_MAX_POWER_W perhaps?

matvey83 (commented):
It's concerning to me that this "float" mode is not actually really "float" charging the battery (which typically means charging is done at constant voltage).   Instead what we are doing here is reducing charge power to a smaller constant level.

Can we either remove mention of "float" from this code, OR convert it to an actual float mode, which means we will need to adjust charge power dynamically based on the voltage of the top cell and the SOC.

Software/src/battery/TESLA-MODEL-3-BATTERY.cpp
dalathegreat: Good catch, fixed in [c4a78d1](https://github.com/dalathegreat/Battery-Emulator/pull/199/commits/c4a78d19cff3264b0f33ed861a715fafd011b0d2)

Software/src/battery/TESLA-MODEL-3-BATTERY.h
dalathegreat: Agreed, fixed in [be662e4](https://github.com/dalathegreat/Battery-Emulator/pull/199/commits/be662e4fe0f3e368fe85456ed8eee97b996e78f8)

PR #200, opened 2024-02-21
Fix u32 Wh values getting corrupted

dalathegreat:
### What
This PR fixes large battery pack size values getting corrupted when sent to a Dual LilyGo setup

### Why
Self explanatory:
bild

### How
Example
75123wh gets rounded down to 7512 wh at sender to fit 16bits, and rescaled to 75120 at reciever. At most 9Wh inaccuracy, we can live with that :)


Patsy555 (approved):
Thanks, tested and is working as should.

Patsy555 (approved):

kharnt0x (approved):
lgtm!

PR #206, opened 2024-02-26
Feature/cell bars

Cabooman:
### WHAT
Really cool and awesome graphical display of cell voltages, in ultra modern 90's style. The cell monitor page now tells you if it doesn't have proper cell data, and auto updates the page every 10 seconds.
Also a minor tweak related to event handling, mainly the EEPROM storage which is probably not used actively by anyone at the moment.

### WHY
Because it's cool... and awesome. Also nice for a more digestible overview of the pack

### HOW
By the unholy trinity of C++, HTML and JS. Basically do more data processing client side, and now the battery emulator only has to transfer the data and not the (visible) html content

### TESTING
First tested offline in HTML for a proof of concept, then on a spare lily with fake battery, and now live in a FroLeaf setup

image


Cabooman:
One thought: Do all batteries populate individual cell voltages, or do batteries have other formats for the cell voltage info like just min/max?

dalathegreat (approved):

PR #213, opened 2024-03-06
Kona 40kWh: Add handling for 90S cellvoltages

dalathegreat:
### What
This PR improves support for 40kWh Kona batteries

### Why
90S battery showed garbage values on three cells out of range:
bild

### How
This PR checks those three cells, and make sure they are valid (over 1000mV) before writing value to array


Cabooman (approved):
It would be nicer without the extra cell array and it could be done by using the original code, but with a cleaning loop step that zeroes values below 1000, instead of only transferring those with > 1000 to the 'real' array. No practical difference though, since we seem to have enough static memory to play with :)

dalathegreat:
@Cabooman Yes you can see the first two commits where I tried to do that. But it failed for some reason, and when switching to a separate array it started to work perfectly. Good enough for me :)

PR #217, opened 2024-03-10
Kia/Hyundai: Swap sign on current value

dalathegreat:
### What
This PR inverts the current value. -10A becomes 10A, and 10A will become -10A

### Why
The value sent to inverter is wrong way around, making any UI confusing:
bild

### How
We invert the sign by using the unary negation operator (-)


Tinyman7613:
I can comfirmed that it is working.

PR #218, opened 2024-03-10
Renault Kangoo: Rescale SOC, current and power

dalathegreat:
### What
This PR fixes scaling on a few values, from user testing

### Why
User grafitix on the Discord server noticed that some Kangoo values where incorrect.

- SOC% had one decimal missing
- Current value had one decimal missing
- Power value was too large (due to current being wrong)

This PR fixes these three values for Kangoo packs


lenvm (approved):
Looks good to me

dalathegreat:
This one is still WIP, waiting for confirmation if SOC rescaling is needed or not

PR #219, opened 2024-03-10
Kia/Hyundai: Increase cell counter to accomodate 98S

dalathegreat:
### What
This PR adds back the last cell in the webserver for a 98S pack

### Why
The 90S patch previously commited broke support for 98S

### How
Increase i++ counter from 97 to 98!


Dibblah1 (approved):
Tested on Kona 98 cell pack and approved

PR #226, opened 2024-03-17
Bugfix: Make Solax no longer get stuck with open contactors at 0%SOC

dalathegreat:
### What
This PR makes Solax inverters more robust on double LilyGo setups

### Why
Solax inverters opens the contactors if the battery is at the minimum charge level but there is no PV power available or grid power allowed  to charge it. This is a problem, since we have been treating it as an ERROR, requiring reboot to get out of this state.

### How
- Inverter contactor opening changed from Error -> Info event
- CAN interface is now restarted incase Solax has timed out

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -123,8 +123,10 @@ void update_values_can_solax() {  //This function maps all the values fetched fr
   if (millis() - LastFrameTime >= SolaxTimeout) {
     inverterAllowsContactorClosing = false;
     STATE = BATTERY_ANNOUNCE;
+#ifndef DUAL_CAN
     ESP32Can.CANStop();  // Baud rate switching might have taken down the interface. Reboot it!
     ESP32Can.CANInit();  // TODO: Incase this gets implemented in ESP32Can.cpp, remove these two lines!
+#endif
   }
   //Calculate the required values
   temperature_average = ((system_temperature_max_dC + system_temperature_min_dC) / 2);
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -123,6 +123,8 @@ void update_values_can_solax() {  //This function maps all the values fetched fr
   if (millis() - LastFrameTime >= SolaxTimeout) {
     inverterAllowsContactorClosing = false;
     STATE = BATTERY_ANNOUNCE;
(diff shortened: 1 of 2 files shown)


Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -123,6 +123,8 @@ void update_values_can_solax() {  //This function maps all the values fetched fr
   if (millis() - LastFrameTime >= SolaxTimeout) {
     inverterAllowsContactorClosing = false;
     STATE = BATTERY_ANNOUNCE;
+    ESP32Can.CANStop();  // Baud rate switching might have taken down the interface. Reboot it!
```
rjsc, line 126: This only works with dual lilygo setups (inverter connected to the ESP32 CAN)
For Dual CAN (MCP2515) setups its not needed and it simply resets the battery communications.
Please add ans `#ifndef DUAL_CAN` to this block.

Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -123,6 +123,8 @@ void update_values_can_solax() {  //This function maps all the values fetched fr
...
+    ESP32Can.CANInit();  // TODO: Incase this gets implemented in ESP32Can.cpp, remove these two lines!
```
rjsc, line 127: This only works with dual lilygo setups (inverter connected to the ESP32 CAN)
For Dual CAN (MCP2515) setups its not needed and it simply resets the battery communications.
Please add ans #ifndef DUAL_CAN to this block.

rjsc (changes requested):
This only works with dual lilygo setups (inverter connected to the ESP32 CAN)
For Dual CAN (MCP2515) setups its not needed and it simply resets the battery communications.
Please add ans #ifndef DUAL_CAN to this block.

Software/src/inverter/SOLAX-CAN.cpp
dalathegreat: Fixed in [e9f004c](https://github.com/dalathegreat/Battery-Emulator/pull/226/commits/e9f004c47b17c3aaa9ffb968914742b6b657f7d8)

rjsc (approved):

dalathegreat:
@rjsc did you test this change already, or just merge it? 😅

rjsc:
Just reviewed the code and merged it.
My inverter is a FoxESS H1, also uses the Solax protocol but behaves a little differently.

PR #229, opened 2024-03-21
Triplet battery: Revise safety limits

dalathegreat:
### What
This PR changes the max allowed voltages for triplet batteries

### Why
Noticed by Ruudi S on Discord server, the values allow overcharging of the 88S battery in a worst case scenario

### How
We now limit to 360V max (4.1V/cell on 88s) , and 316V min (3.6V/cell on 88s)


lenvm (approved):
Looks good to me

PR #235, opened 2024-03-27
Bugfix: BMW i3 values

dalathegreat:
### What
A few values is scaled wrong / glitches on i3 batteries
- Current value
- Power value
- Max allowed charge/discharge overflows
- CRC check is broken on 2BD message

bild

### Why
Mistakes were made in the first implementation

### How
Iterative fixes

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -376,7 +376,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   system_battery_voltage_dV = battery_volts;  //Unit V+1 (5000 = 500.0V)
 
-  system_battery_current_dA = battery_current * 10;
+  system_battery_current_dA = battery_current;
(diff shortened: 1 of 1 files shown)


odlg:
There is a division by 10 in receive_can_battery() - so I guess we are loosing precision by this.
In update_values_battery() battery_power is calculated from system_battery_current_dA which is now 10 times higher, so that will be wrong then.

dalathegreat:
> There is a division by 10 in receive_can_battery() - so I guess we are loosing precision by this. In update_values_battery() battery_power is calculated from system_battery_current_dA which is now 10 times higher, so that will be wrong then.

Fix in [d31c349](https://github.com/dalathegreat/Battery-Emulator/pull/235/commits/d31c3490a1799aadc4944ccc0fca06ca4f3602a1) , hope I got the decimal right, it's -819.2 offset!

odlg (approved):
Looks good. 
Tested, and works.

PR #236, opened 2024-03-27
Make mainLoop run as background task with same priority as modbus

odlg:
**Why**
BMW i3 battery requires can messages every 20ms. BYD modbus spends over 200ms sending some messages and main arduino loop currently has lowest priority. So sending modbus works fine, but can bus messages are sent too late. The battery opens contactors due to this.

**How**
Move the code out of loop() method, make new method and make it loop itself, invoke it as part of setup()

**What**
Make main arduino loop run in a background task to be able to set priority.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -131,7 +131,7 @@ unsigned long timeSpentInFaultedMode = 0;
 bool batteryAllowsContactorClosing = false;
 bool inverterAllowsContactorClosing = true;
 
-TaskHandle_t mainLoopTask;               
+TaskHandle_t mainLoopTask;
 
 // Initialization
 void setup() {
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -18,6 +18,12 @@
 #include "src/lib/miwagner-ESP32-Arduino-CAN/CAN_config.h"
 #include "src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.h"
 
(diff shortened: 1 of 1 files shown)


lenvm:
@odlg, please run pre-commit to fix the clang-format issues on your branch.

Software/Software.ino
```diff
@@ -155,60 +163,71 @@ void setup() {
...
+  xTaskCreatePinnedToCore((TaskFunction_t)&mainLoop, "mainLoop", 4096, NULL, 8, &mainLoopTask, 1);
 }
 
 // Perform main program functions
 void loop() {
+  ;
+}
+
+void mainLoop(void* pvParameters) {
+  while (true) {
```
lenvm, line 178: Is this `while (true)` loop absolutely necessary?

lenvm:
@odlg, please add a bit more detail to the description of your PR.

Please use the "Why, how, what" methodology, as is used in the other PRs on this repository. This will improve traceability what was changed precisely in which PR, and for what reasons.

dalathegreat:
Just tested this in my CPU load testing branch, it reduces the delta-t for sending 10ms CAN messages from 13ms -> 11ms. What's even better, performing webserver tasks no longer cause this to spike to 15ms, it stays rock solid on 11ms 😎 Very impressive @odlg

dalathegreat (approved):

Issue #234, opened 2024-03-27
Modbus eats too much resources, causes CAN to deadlock


dalathegreat:
### What
The Modbus protocol is taking too much CPU from time to time, causing timing of CAN messages to be incorrect. 

### Why is this an issue?
Worst case scenario a 10ms message can be delayed with almost 200ms, totally unacceptable on an embedded real-time device. BMW i3 opens contactors due to this.

### How to fix
We need to:
- Get Modbus to lower the CPU load
- Implement some warning signs of this
- More?

StarkJohan:
- Deactivate modbus entirely when only CAN is needed/used.

dalathegreat:
https://github.com/dalathegreat/Battery-Emulator/pull/236 is now merged, good first step in lowering CPU load!

dalathegreat:
Situation is much better with the latest v5.6.0 release, but could always be improved further!

dalathegreat:
Setting this one as resolved now, 5.7.dev is getting really fast! 🐎

PR #237, opened 2024-03-27
Feature/Bugfix: CPU usage + Events

dalathegreat:
### What
This PR focuses on reducing CPU usage, to make the system more reliable. It also adds mechanism to alerting user incase performance deteriorates

### Why
See https://github.com/dalathegreat/Battery-Emulator/issues/234

### How
Features implemented
- Added a new event type, to signal that periodic CAN sending has taken too long time:
bild

Fixes implemented
- Webserver loads 30% faster. Previously a periodic 10ms CAN message could get delayed up to 18ms when F5ing the webserver, this is now cut down to 13ms max.
- Pressing "Event" button in webserver caused the sending of 10ms CAN messages to skyrocket to 185ms. This is now fixed, all serialwrites wrapped in #ifdef blocks
- Every single USB print is wrapped in #ifdef DEBUG_VIA_USB , unintended printing no longer causes CPU load increase
- Memory save: Periodic CAN times now set as defines, simplifies code and saves memory


odlg (approved):
Looks good.

PR #241, opened 2024-04-02
Bugfix: Further CPU load reductions

dalathegreat:
### What
This PR focuses on reducing CPU load further, to reduce the amount of CAN_OVERRUN events

### Why
Keeping the CAN messages in time is essential for good control over batteries. Delayed messages could cause issues in form of contactors opening under load

### How
- Webserver filesize further reduced, reducing load when accessing pages
- Webserver set to refresh mainpage every 15 seconds instead of 10 seconds
- Nissan LEAF: 10ms messages now sent before 100ms are queued
- AsyncTCP moved to Core 0 to reduce load on Core 1
- Scheduler delay reduced from 2ms to 1ms
- LED code made less resource intensive. 
   - digitalRead moved into LED block, which makes it run less often.
   - Amount of if statements reduced

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -29,7 +29,7 @@
 
 Preferences settings;  // Store user settings
 // The current software version, shown on webserver
-const char* version_number = "5.6.0";
+const char* version_number = "5.7.dev";
 
 // Interval settings
 uint16_t intervalUpdateValues = INTERVAL_5_S;  // Interval at which to update inverter values / Modbus registers
@@ -98,7 +98,6 @@ static uint8_t brightness = 0;
 static bool rampUp = true;
 const uint8_t maxBrightness = 100;
 uint8_t LEDcolor = GREEN;
-bool test_all_colors = false;
 
 // Contactor parameters
 #ifdef CONTACTOR_CONTROL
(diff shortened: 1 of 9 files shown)


Software/Software.ino
```diff
@@ -541,29 +534,31 @@ void handle_LED_state() {
   } else if (!rampUp && brightness == 0) {
     rampUp = true;
   }
-  if (test_all_colors == false) {
```
Cabooman, line 544: This could technically be kept, including the else-clause, but the initial condition changed to
`if (digitalRead(0 != LOW)`

Previously there was an if and then a switch, and now it's just the other way around but with two calls to setPixelColor if the button is pressed.

This is incredibly minor though, and it can be argued that flatter code is nicer then nesting so I have no real opinion either way

Cabooman (approved):
Nothing critical changed and I assume the leaf and web server changes have been tested, LGTM

Software/Software.ino
dalathegreat: Yup, simply moving the digitalRead into the LED function will make way less CPU calls. But I like it this way, so let's keep it :)

PR #247, opened 2024-04-07
Improvement: Increase size for data

dalathegreat:
### What
This PR allows for storing larger values;
- Instant Watt values can now go over +-60kW ( +-200kW)
- Max charge/discharge can now go over 60kW (-> 200kW)
- Battery size comment updated to 500kWh
- Battery voltage comments updated from 500V -> 1000V

### Why
To prepare for double battery support, and larger batteries

### How
uint16->uint32 mostly!


Cabooman (approved):
LGTM

Using unstructured global variables really doesn't scale lol

PR #249, opened 2024-04-09
Feature/bmw 3 additional data

odlg:
**What**
Additional data fetch from battery. Found additional method to fetch data that was not available before.

**Why**
Having min/max/avg cell voltage and SOH is needed to know the state and health of the battery. 

**How**
Sending specific CAN messages makes it respond a little later with requested information.


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -529,6 +529,13 @@ String processor(const String& var) {
     } else {  // > 0
       content += "<h4>Battery charging!</h4>";
     }
+#ifdef BMW_I3_BATTERY
```
dalathegreat, line 532: We should instead have separate pages for each battery size, where additional info is displayed. I would suggest removing these lines in webserver.cpp , and we can rework this later. We want to avoid this file growing into a monster!

Software/src/devboard/webserver/webserver.h
```diff
@@ -40,6 +40,13 @@ extern uint8_t LEDcolor;                                   //Enum, 0-10
 extern bool batteryAllowsContactorClosing;                 //Bool, 1=true, 0=false
 extern bool inverterAllowsContactorClosing;                //Bool, 1=true, 0=false
 
+#ifdef BMW_I3_BATTERY
```
dalathegreat, line 43: See the comment about growing webserver

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -16,6 +16,7 @@ static unsigned long previousMillis5000 = 0;   // will store last time a 5000ms
 static unsigned long previousMillis10000 = 0;  // will store last time a 10000ms CAN Message was send
 static uint8_t CANstillAlive = 12;             // counter for checking if CAN is still alive
 static uint16_t CANerror = 0;                  // counter on how many CAN errors encountered
+#define MAX_CAN_FAILURES 500                   // Amount of malformed CAN messages to allow before raising a warning
```
dalathegreat, line 19: This define should not be needed, it is duplicated, and placed in config.h (so include that file like done in Nissan LEAF battery)

dalathegreat (changes requested):
Very nice work! Some minor thoughts about webserver!

Software/src/devboard/webserver/webserver.cpp
odlg: ✔️

Software/src/battery/BMW-I3-BATTERY.cpp
odlg: Aarh, yes, a merge bug. Fixed.

dalathegreat (approved):

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -306,6 +348,12 @@ static uint16_t battery_prediction_voltage_longterm_charge = 0;
 static uint16_t battery_prediction_voltage_longterm_discharge = 0;
 static uint16_t battery_prediction_duration_charging_minutes = 0;
 static uint16_t battery_target_voltage_in_CV_mode = 0;
+
+uint16_t battery_soc = 0;
```
dalathegreat, line 352: consider making these `static` so no other battery tries to use the same name and get conflicts! battery_soc is quite generic :)

odlg: Fixed

PR #251, opened 2024-04-10
Add platformio serial logging

kukumagi:
What
When using platformio serial monitor it automatically creates log file with serial data in logs folder.

Why
Copying serial monitor data is tricky. This makes it a lot easier to debug even if there is lots of data in serial monitor. And also sending serial data to others for debugging is easier with this.

How
In VSCode using platformio serial monitor it automatically saves serial monitor output to log file in logs folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 .clang_complete
 .gcc-flags.json
 .pio
+logs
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):
LGTM! 👍

PR #252, opened 2024-04-12
Refactoring: Scalability and performance

Cabooman:
### WHAT

 * "Improved" LED handling - CLASSIC, FLOW and HEARTBEAT animations with lower frequency (50ms)
 * Introduction of a rudimentary hardware abstraction layer (HAL) to handle other boards
 * Untangling of settings - config.h and various other settings have moved to any of: system_settings.h hw_lilygo.h or related implementation files (e.g. WUP_PIN -> BMW i3, for now)
 * Implementation of micro seconds timing macros for debugging performance - displayed in the web UI if enabled
 * MQTT handling moved to its own task with slightly lower priority
 * Include handling improved, most files can now include "src/include.h" to get the central settings/definitions
 * An initial data layer is introduced in a moderate fashion, for Nissan Leaf power values and time measurements
 * A central SOC scaling component added, not in use yet
 * Mapping functions supporting mapping of some value in one range to the corresponding value in a different range
 *  Slightly improved compilation times by having preprocessor conditions on inverter code and charger code, which isn't very nice looking but basically the only reasonable way. It also matches what we previously introduced for batteries
 * Various bits and bobs for structuring stuff

### WHY
 * General niceness
 * Better performance
 * To move closer towards a more scalable code base

### TODO
 * #ifdef the timing information in the web UI
 * Rename wifiTaskto mqttTask


Software/USER_SETTINGS.h
```diff
@@ -7,6 +7,9 @@
 /* There are also some options for battery limits and extra functionality */
 /* To edit battery specific limits, see also the USER_SETTINGS.cpp file*/
 
+/* Select HW - DONT TOUCH */
```
dalathegreat, line 10: Since there is yet no other variant than HW_LILYGO, can we hide this from the USER_SETTINGS.h file and have it permadefined?

Software/Software.ino
```diff
@@ -96,12 +95,13 @@ float charger_stat_ACvol = 0;
...
-static bool rampUp = true;
-const uint8_t maxBrightness = 100;
-uint8_t LEDcolor = GREEN;
+// Task time measurement for debugging and for setting CPU load events
+int64_t core_task_time_us;
+MyTimer core_task_timer_10s(INTERVAL_10_S);
+
+int64_t mqtt_task_time_us;
+MyTimer mqtt_task_timer_10s(INTERVAL_10_S);
+// MyTimer task_timer_200ms(INTERVAL_200_MS);
```
dalathegreat, line 104: Remove this uncommented row?

Software/USER_SETTINGS.h
Cabooman: Absolutely, I was thinking the same thing. Whenever you feel the need to write "DONT TOUCH" next to something it should probably be less visible :)

Software/Software.ino
Cabooman: Yes!

Software/Software.ino
```diff
@@ -211,15 +238,30 @@ void mainLoop(void* pvParameters) {
...
     send_can();  // Send CAN messages
 #ifdef DUAL_CAN
     send_can2();
 #endif
-    run_event_handling();
+    END_TIME_MEASUREMENT_MAX(cantx, datalayer.system.status.time_cantx_us);
 
-    delay(1);  // Allow the scheduler to start other tasks on other cores
+    START_TIME_MEASUREMENT(events);
+    run_event_handling();
```
smaresca, line 252: Multiple functions invoked before this line in core_loop() may set an event and placement of run_event_handling() **here** therefore makes sense; however, is there any argument to be made for running event handling sooner?  

I think the semantics of set_event and clear_event (e.g., with the setting of bms status) mean that it's not necessary.  Just seeking confirmation.

Software/src/devboard/utils/value_mapping.h
```diff
@@ -0,0 +1,45 @@
+#ifndef __MAPPING_H__
+#define __MAPPING_H__
+
+#include "../../include.h"
+
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
+#define CONSTRAIN(val, min, max) (MIN(max, MAX(min, val)))
```
smaresca, line 8: probably needs a warning about side effects for sanity-preservation purposes. I can see why this is locally defined rather than rely upon platform-provided items.

Usage needs to be 
int x = function();
CONSTRAIN(x, 5, 10);
rather than
CONSTRAIN(function(), 5, 10)

This is not a complaint about current usage in this PR...just proactive acknowledgement of subtle bugs that could result.  Official Arduino docs make note for this very reason.

Software/src/devboard/utils/types.h
```diff
@@ -0,0 +1,63 @@
...
+#define INTERVAL_20_MS_DELAYED 30
+#define INTERVAL_30_MS_DELAYED 40
+#define INTERVAL_100_MS_DELAYED 120
+
+#define MAX_CAN_FAILURES 500  // Amount of malformed CAN messages to allow before raising a warning
+
+/** Start time measurement in microseconds
+ * Input parameter must be a unique "tag", e.g: START_TIME_MEASUREMENT(wifi);
+ */
+#ifdef FUNCTION_TIME_MEASUREMENT
```
smaresca, line 43: Opinion: performance/time measurement macros should exist outside of a types.h

Software/src/devboard/datalayer/datalayer.h
smaresca, line 1: I suggest moving datalayer up one level (rather than nesting in devboard/ ), placing it alongside equally important include.h, system_settings.h, etc.

smaresca (changes requested):
Just finished an initial read-through. I may have a comment or two as I rebase my branch against this.

Software/src/devboard/utils/types.h
Cabooman: Good point, they were some last-minute scope creep so I didn't think it over properly. Moving to utils!

Software/src/devboard/utils/value_mapping.h
Cabooman: I agree, I'll add a note. Not that I expect anyone to read it but then we've at least tried ;)

Software/Software.ino
Cabooman: Yes, placement of this technically doesn't matter, as it only updates the "system time", checks if the EE log should be written and such. Event reactions are always immediate

Software/USER_SETTINGS.h
```diff
@@ -42,6 +42,7 @@
 //#define SERIAL_LINK_TRANSMITTER  //Enable this line to send battery data over RS485 pins to another Lilygo (This LilyGo interfaces with battery)
 #define WEBSERVER  //Enable this line to enable WiFi, and to run the webserver. See USER_SETTINGS.cpp for the Wifi settings.
 //#define LOAD_SAVED_SETTINGS_ON_BOOT  //Enable this line to read settings stored via the webserver on boot (overrides any battery settings set in USER_SETTINGS.cpp)
+//#define FUNCTION_TIME_MEASUREMENT  // Enable this to record execution times and present them in the web UI
```
dalathegreat, line 45: I would add similar addition like on DEBUG_USB option `(WARNING, raises CPU load, do not use for production)`

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -497,9 +511,9 @@ String processor(const String& var) {
     float sohFloat = static_cast<float>(system_SOH_pptt) / 100.0;               // Convert to float and divide by 100
     float voltageFloat = static_cast<float>(system_battery_voltage_dV) / 10.0;  // Convert to float and divide by 10
     float currentFloat = static_cast<float>(system_battery_current_dA) / 10.0;  // Convert to float and divide by 10
-    float powerFloat = static_cast<float>(system_active_power_W);               // Convert to float
-    float tempMaxFloat = static_cast<float>(system_temperature_max_dC) / 10.0;  // Convert to float
-    float tempMinFloat = static_cast<float>(system_temperature_min_dC) / 10.0;  // Convert to float
+    float powerFloat = static_cast<float>(datalayer.battery.status.active_power_W);  // Convert to float
```
dalathegreat, line 514: Doesn't this mean that all other batteries other than LEAF will show 0W here?

Software/src/devboard/utils/soc_scaling.cpp
```diff
@@ -0,0 +1,36 @@
+#include "soc_scaling.h"
```
dalathegreat, line 1: I feel like the SOC scaling (and SOC handling in general) might change quite a bit with the introduction of Double-Battery, so is it really worth it to add two unused files in this PR (that might be outdated quickly?)

Software/USER_SETTINGS.h
Cabooman: Will do!

Software/src/devboard/webserver/webserver.cpp
Cabooman: Yes it does, and that's of course not intended. It could get some condition for value selection but that wouldn't be reasonable. I suggest I revert the data layer "test" and make a complete transition in the next branch

Software/src/devboard/utils/soc_scaling.cpp
Cabooman: Similar as the data layer, I'd like to keep the files and "plug them in" in the next branch. It's nicely packaged so it would be easy to extend the files with a double-SOC class. We probably shouldn't mix double- and single-battery functionality in the same "location" anyway

dalathegreat (approved):
Tested out the build on my desk-setup; Holding F5 in the webserver could tank the occasional 10ms message towards 13ms, but now the furthest drift is 11ms 📉 🥳

I am approving this, but would like to test it on some batteries before we merge! 🙌

PR #253, opened 2024-04-14
Bugfix: Remove Volvo battery writing to bms_status

dalathegreat:
### What
This PR removes a few lines where Volvo batteries were writing `system_bms_status = FAULT;`

### Why
This is redundant, event is already called which sets the status. Writing to same variable from multiple places can cause issues.

### How
Events are responsible for setting the system_bms_status


Cabooman (approved):

smaresca:
Thanks for this, I've had some of these items on my "hmm list of things to revisit"

PR #255, opened 2024-04-15
USB debug removal and improvements

dalathegreat:
### What
This PR cleans up some unnecessary USB debug handling

### Why
Reduces complexity

### How
* Volvo SPA: Removed USB printouts for cellvoltage errors (Events are better and already in place)
* Nissan LEAF: Removed USB printouts for inverter values (Can be viewed in Webserver instead, and data is 100% known working)
* Triplets: Removed USB printouts for inverter values (Can be viewed in Webserver instead)
* Chademo: Removed obsolete errorCode, and inverter values (Can be viewed in Webserver instead)


Cabooman (approved):
For the TODO's, I guess everything that affects battery power should have an event of some sort

PR #257, opened 2024-04-15
Improvement: Add new events for LEAF

dalathegreat:
### What
This PR adds new events for LEAF batteries, that previously were USB serial printouts

### Why
The serial printouts were impossible to see in practice. We now log battery warming and stopping heat as an info event.

### How
3 new events added (EVENT_BATTERY_CAUTION , EVENT_BATTERY_REQUESTS_HEAT and EVENT_BATTERY_WARMED_UP)
Also removed the unused Battery_current_X variables from the code, as these were unusable


Software/src/devboard/utils/events.h
Cabooman, line 1: Update EE_MAGIC_HEADER_VALUE to e.g. 0x0002 so that EE event logs are cleared when this new event order is used. Might seem like something basically unused but every user is sitting on a decent event log that survives a reset :) It needs to be more accessible though...

Cabooman (approved):

PR #261, opened 2024-04-20
Feature: Double battery 🔋 🔋 🥈 

dalathegreat:
### WIP
This requires extensive testing before it can be merged. See Wiki page being built for [Double-Battery](https://github.com/dalathegreat/Battery-Emulator/wiki/Double-Battery)

### What
This PR adds support for running 2x batteries at the same time. To start with, the following combinations are supported:

- 2xTesla S/3/X/Y
- 2xNissan LEAF
- 2xBMW i3

bild

### How
The second battery connects to the DualCAN add-on MCP chip. Incase the pack voltages read via CAN are within 3.0V between the two packs, the secondary battery can join the mix.

Two new events added, CAN2 failure and too large voltage diff:
bild


smaresca:
I think this PR should be rebased atop the main when Safety.cpp is merged. I noted the introduction of the event comparing voltage difference between batteries to keep current flow down when they are connected in parallel. Maybe-but-not-necessarily consider that they must have a similar SOH, SOC, etc.

dalathegreat:
> I think this PR should be rebased atop the main when Safety.cpp is merged. I noted the introduction of the event comparing voltage difference between batteries to keep current flow down when they are connected in parallel. Maybe-but-not-necessarily consider that they must have a similar SOH, SOC, etc.

This has now been rebased, and the Safety.cpp has a section specifically with Double-Battery checks. This is ready for testing (and ready for review!)

dalathegreat:
Feature has been tested, and in use for a few days. First startup can be seen here: https://youtu.be/skBhH_EwBUE

Merging this PR to main, to get more feedback on it! :raised_hands:

PR #263, opened 2024-04-23
Bugfix: Make faulted LED solid red again 🔴 

dalathegreat:
### What
This PR makes the LED on the board stay on with solid red color incase we enter faulted state

### Why
Having the fault state solid red without any animations helps the user figure out that something is wrong. It is also better for colorblind users that may mistake the green from red.

### How
The rewrite for the LED handling introduced in software 5.7.1 added new animations. The fault state accidentally got animated, it should be static.

### DIFF
--- a/Software/src/devboard/utils/led_handler.cpp
+++ b/Software/src/devboard/utils/led_handler.cpp
@@ -70,7 +70,7 @@ void LED::exe(void) {
           break;
         case EVENT_LEVEL_ERROR:
           color = led_color::RED;
-          pixels.setPixelColor(0, COLOR_RED(brightness));  // Red LED full brightness
+          pixels.setPixelColor(0, COLOR_RED(150));  // Red LED full brightness
           break;
         default:
           break;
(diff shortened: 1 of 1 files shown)


Software/src/devboard/utils/led_handler.cpp
```diff
@@ -70,7 +70,7 @@ void LED::exe(void) {
           break;
         case EVENT_LEVEL_ERROR:
           color = led_color::RED;
-          pixels.setPixelColor(0, COLOR_RED(brightness));  // Red LED full brightness
+          pixels.setPixelColor(0, COLOR_RED(150));  // Red LED full brightness
```
Cabooman, line 73: Here we should use LED_MAX_BRIGHTNESS as it will/might depend on the specific hw

Cabooman (changes requested):

Software/src/devboard/utils/led_handler.cpp
dalathegreat: Fixed in [3581465](https://github.com/dalathegreat/Battery-Emulator/pull/263/commits/35814656cd7fadfd326a385726b1cf600b7f5b1e)

Cabooman (approved):

PR #264, opened 2024-04-23
Bugfix: Remake logic how Kia64 startup is handled

dalathegreat:
### What
In this PR we edit the logic on how CAN sending towards Kia/Hyundai batteries work

### Why
Due to recent CPU load optimizations, it was noticed that the startup process had become less stable. This PR attempts to make starting the batteries up more reliable.

### How
We now wait until a CAN message has arrived from the battery, before we start to send stuff towards the battery.

### DIFF
--- a/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
+++ b/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
@@ -301,21 +301,21 @@ void update_values_battery() {  //This function maps all the values fetched via 
 void receive_can_battery(CAN_frame_t rx_frame) {
   switch (rx_frame.MsgID) {
     case 0x4DE:
-      startedUp = 1;
+      startedUp = true;
       break;
     case 0x542:  //BMS SOC
-      startedUp = 1;
+      startedUp = true;
       CANstillAlive = 12;                     //We use this message to verify that BMS is still alive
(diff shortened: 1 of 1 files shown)


dalathegreat:
bild

Fix verified working by tiny_man over on the Discord server 🥳

Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
```diff
@@ -301,17 +301,21 @@ void update_values_battery() {  //This function maps all the values fetched via
 void receive_can_battery(CAN_frame_t rx_frame) {
   switch (rx_frame.MsgID) {
     case 0x4DE:
+      startedUp = 1;
```
Cabooman, line 304: Since it's a bool, we should use true/false

Cabooman (changes requested):

Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
dalathegreat: True dat! Fixed in [78dfbef](https://github.com/dalathegreat/Battery-Emulator/pull/264/commits/78dfbef3e26458a0f1f66c98d5ae06433811b94f)

Tinyman7613:
Works great :)

Tinyman7613 (approved):
Works great

PR #267, opened 2024-04-23
Feature: Add more events for BMW i3 to aid troubleshooting

dalathegreat:
### What
This PR adds two more events to the i3 battery code

### Why
To assist installation troubleshooting

### How
We now alert the user via the webui via events, one for high voltage interlock not seated, and another one for capacitor missing on HV side.


odlg (approved):
Looks good.

PR #268, opened 2024-04-24
Microfeature: Add uptime tracking and display

smaresca:
**What**

Adds overflow-tolerant uptime tracking and reporting via webserver UI

**How**

By incorporating the very useful Uptime_Library by Yiannis Bourkelis (license compatibility verified). Header guard added locally within this project.

**Notes**
Feature requested by f1racr.

Multiple other time-oriented refinements are likely possible as benefit of having this in place (already noted by @Cabooman in discussion). This branch may be revised to account for some of those.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -14,6 +14,8 @@
 #include "src/devboard/utils/events.h"
 #include "src/devboard/utils/led_handler.h"
 #include "src/devboard/utils/value_mapping.h"
+#include "src/lib/Uptime_Library/src/uptime.h"
+#include "src/lib/Uptime_Library/src/uptime_formatter.h"
 #include "src/lib/bblanchon-ArduinoJson/ArduinoJson.h"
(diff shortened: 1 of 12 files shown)


Cabooman (changes requested):
Current implementation is OK, but I suggest:
- Integrating uptime for event timestamps
- Removing the example code and other things not needed in this context

dalathegreat:
The integration of timestamps for events is a bit tricker, could be done in a separate PR. I would be OK with merging this as soon as the examples folder is removed @smaresca , nice work 👍

smaresca:
> The integration of timestamps for events is a bit tricker, could be done in a separate PR. I would be OK with merging this as soon as the examples folder is removed @smaresca , nice work 👍

The examples folder is part of the upstream lib @dalathegreat . I was hesitant to make too many local modifications during import..

Cabooman (approved):

PR #269, opened 2024-04-25
Fix regression in SERIAL-LINK-RECEIVER

smaresca:
A small bugfix to permit successful compilation for SERIAL-LINK-RECEIVER (search/replace issue from PR#254).


Cabooman (approved):
Ah, an obvious replace-all bug :)

PR #270, opened 2024-04-25
Improvement/build serial link rx

Cabooman:
### WHAT
Add build of SERIAL_LINK_RECEIVER at commit or PR

### WHY
Currently, we don't build SERIAL_LINK_RECEIVER and this causes a risk of undetected regressions

### HOW
Add SERIAL_LINK_RECEIVER to workflows. Change the target "inverter" to BYD_CAN since BYD_MODBUS and SERIAL_LINK_RECEIVER are incompatible. Compile All Combinations cannot be updated as it will attempt to build the incompatible configuration.

### DIFF
--- a/.github/workflows/compile-all-batteries.yml
+++ b/.github/workflows/compile-all-batteries.yml
@@ -46,8 +46,8 @@ jobs:
           - SERIAL_LINK_RECEIVER
         # These are the emulated inverter communication protocols for which the code will be compiled.
         inverter:
-#         - BYD_CAN
-          - BYD_MODBUS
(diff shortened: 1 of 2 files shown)


smaresca:
Any coverage concerns with BYD_MODBUS -> BYD_CAN for tests?

Cabooman:
> Any coverage concerns with BYD_MODBUS -> BYD_CAN for tests?

No, I have no concerns. This specific workflow is only a compile check, so there's nothing actually tested. That's not great in itself but that's a different story :)

PR #271, opened 2024-04-26
Feature: Add Event when automatic contactor opening happens

dalathegreat:
### What
This PR adds an event for when automatic contactor control (via GPIO) decides to open contactors

### Why
For easier troubleshooting

### How
- An event is added "Info: Too much time spent in error state. Opening contactors, not safe to continue charging. Check other error code for reason!"
- We also increase ticks from 500->1000, to avoid making the control too rapid.


smaresca (commented):
Looks ok to me, but not able to test currently.

Cabooman (approved):
No obvious reason why this shouldn't work, LGTM

PR #272, opened 2024-04-26
Improvement/event uptimestamp

Cabooman:
### WHAT
Event timestamp update

### WHY
@smaresca added a proper library for tracking uptime. Previously, the event timestamp was relative time since last second, basically a 1s timer that incremented a counter. This introduces lots of inaccuracy, and since uptime is precisely what we need for the timestamp, uptime is what we should use

### HOW
uptime::calculateUptime() is moved from the core task and a 5 second period to the loop() task and a 600 ms period to increase uptime updates while keeping that load in the lowest priority task. It gets plenty of run time so there's no risk of errors. In events.cpp:update_event_time() we now use the uptime library to update the event timestamp.

In addition, events_html.cpp is updated with a leaner javascript function.

### TESTING
The update is tested live and is showing no-to-little jitter, and no drift over time

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -226,7 +226,7 @@ void core_loop(void* task_time_us) {
     if (millis() - previousMillisUpdateVal >= intervalUpdateValues)  // Every 5s normally
     {
       previousMillisUpdateVal = millis();
-      update_SOC();               // Check if real or calculated SOC% value should be sent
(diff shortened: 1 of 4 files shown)


dalathegreat:
Tested 👍 Love that the events_html.cpp is getting less bytes to transfer 😎

dalathegreat (approved):

PR #273, opened 2024-04-27
Improvement: Rename BMS to System in webserver

dalathegreat:
### What
This PR makes the Webserver say `System` instead of `BMS`

### Why
Some users were confused, and thought the BMS was bad. System is a better name, since we can go into Faulted mode even if the BMS still thinks everything is fine.

### How
Renamed the BMS -> System


Cabooman (approved):
Good short term solution :)

Might want to change the name internally as well later on, so that it's consistent

PR #274, opened 2024-04-27
Bugfix: Make chargers compile and add to workflow 🤖 

dalathegreat:
### What
This PR makes it possible to compile with Charger support enabled

### Why
Charger support was broken when the refactoring for v5.7.1 was done

### How
We now prevent this from ever happening again, `NISSANLEAF_CHARGER` is now added to the Github workflow

### DIFF
--- a/Software/src/charger/CHEVY-VOLT-CHARGER.cpp
+++ b/Software/src/charger/CHEVY-VOLT-CHARGER.cpp
@@ -1,6 +1,7 @@
 #include "../include.h"
 #ifdef CHEVYVOLT_CHARGER
 #include "../datalayer/datalayer.h"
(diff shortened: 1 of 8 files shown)


Software/src/charger/CHEVY-VOLT-CHARGER.cpp
```diff
@@ -115,7 +116,7 @@ void receive_can_chevyvolt_charger(CAN_frame_t rx_frame) {
   }
 }
 
-void send_can_chevyvolt_charger() {
+void send_can_charger() {
```
smaresca, line 119: plus net improvement re: API structuring for the win!

smaresca (approved):
Excellent!

PR #279, opened 2024-04-30
BMW i3: Add cellvoltage safeties and pack size autodetect

dalathegreat:
### What
This PR adds automatic pack size detection, which is used to set designvoltages and check safeties

### Why
Safer operation of BMW i3 batteries. Added:
* Cell deviation warning event
* Cell overvoltage fault event
* Cell undervoltage fault event
* Total pack voltage design voltage limits

### How
We use kWh reported amount to switch between 60/94/120AH cellvoltage/packvoltage limits. Note, values needs testing, this is a good starting point!

bild

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -18,13 +18,7 @@ static unsigned long previousMillis10000 = 0;  // will store last time a 10000ms
 static uint8_t CANstillAlive = 12;             // counter for checking if CAN is still alive
 static uint16_t CANerror = 0;                  // counter on how many CAN errors encountered
 #define ALIVE_MAX_VALUE 14                     // BMW CAN messages contain alive counter, goes from 0...14
-#define MAX_CELL_VOLTAGE_60AH 4110             // Battery is put into emergency stop if one cell goes over this value
-#define MIN_CELL_VOLTAGE_60AH 2700             // Battery is put into emergency stop if one cell goes below this value
-#define MAX_CELL_VOLTAGE_94AH 4140             // Battery is put into emergency stop if one cell goes over this value
-#define MIN_CELL_VOLTAGE_94AH 2700             // Battery is put into emergency stop if one cell goes below this value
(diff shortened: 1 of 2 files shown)


Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -447,6 +456,44 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+      (datalayer.battery.status.cell_max_voltage_mV - datalayer.battery.status.cell_min_voltage_mV);
+
+  // Start checking safeties. First up, cellvoltages!
+  if (battery_cell_deviation_mV > MAX_CELL_DEVIATION) {
+    set_event(EVENT_CELL_DEVIATION_HIGH, 0);
+  } else {
+    clear_event(EVENT_CELL_DEVIATION_HIGH);
+  }
+  if (detectedBattery == BATTERY_60AH) {
+    datalayer.battery.info.max_design_voltage_dV = 3950;
```
odlg, line 469: I can not see the reason for mixing defines with values in the code. I think it should be either only defines or values in the code.

odlg (changes requested):

Software/src/battery/BMW-I3-BATTERY.cpp
dalathegreat: @odlg Thanks, fixed in e9e65de

odlg (approved):
Looks fine now :-)

havrla:
I'll just add how it behaves in the vehicle:

If the electricity meter reaches 0kWh, consumption by the converter drops to a minimum and driving is only possible with a step, after a while it is not even possible to drive (approx. 320V). So a dead battery - it will end the energy counter softly, a lower cell voltage hard.

If the battery is charging, it will stop when the maximum battery cell voltage is reached (4.07/4.15/4.2V). Charged is then insured by the maximum battery voltage of 392V(60Ah), 398V(94Ah), 403V(120Ah). After reaching this value, the energy counter is synchronized from above.

The size of the energy counter is continuously calibrated, but here it will be shown in practice how it will work on PV plants. Large calibrations in the vehicle require the vehicle to be put to sleep for at least 15 minutes.

PR #282, opened 2024-05-01
Update KIA-E-GMP-BATTERY.cpp

JorgenSeemann:
Removed cells with voltage below 1000mv in the set_cell_voltages routine

### DIFF
--- a/Software/src/battery/KIA-E-GMP-BATTERY.cpp
+++ b/Software/src/battery/KIA-E-GMP-BATTERY.cpp
@@ -51,7 +51,9 @@ CANFDMessage EGMP_7E4_ack;
 
 void set_cell_voltages(CANFDMessage frame, int start, int length, int startCell) {
   for (size_t i = 0; i < length; i++) {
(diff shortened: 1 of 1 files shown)


dalathegreat:
Looks good @JorgenSeemann , but one of the build checks failed (we have very strict formatting rules)

I suggest you install the pre-commit check, it handles all formatting for you!

_The pre-commit git hook shall be installed locally by running the command line command `pre-commit install`, for instance using git-bash, in the location where the .pre-commit-config.yaml file is located._

Incase you do not want to do this setup, you can simply add two spaces here to get the check to pass
bild

JorgenSeemann (commented):
spaces added

JorgenSeemann (commented):
Added section to close contractor

dalathegreat:
The addition caused more formatting issues. I really recommend installing the pre-commit git hook, it makes development so much easier! `pre-commit install` in command line

dalathegreat (approved):

PR #287, opened 2024-05-04
Update Pylon-CAN.cpp

LevinSwe:
- Add function to byte-swap all 16bit values
- Add function to enable 30k offset for currents
- Add function to limit charge/discharge currents from User settings 
- Add status bits
- Set default value for 12V lead-acid battery in KIA64 battery (yellow LED on startup sometimes)


LevinSwe:
Battery Status might be wrong, its calling for "Balance-Request" in pylon.cpp, correct value;
row 219 --> 0x01
row 221 --> 0x02
row 223 --> 0x03
-Then the comment in row 217 also is wrong

smaresca:
I will be in a position to test this in a day or two. @LevinSwe  could you rebase, squash, and fix conflicts?

dalathegreat:
I tried to fix the conflict, think I did it! Now it's just code formatting next @LevinSwe !

PR #289, opened 2024-05-05
Add debug functions for can fd

kukumagi:
What
This PR adds settings to enable can fd debugging.

Why
To improve ease of access to can fd data in serial monitor.

How
Enabling #define DEBUG_CANFD_DATA in USER_SETTINGS.h then it starts to print out frames from can fd data stream.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -457,10 +457,60 @@ void init_battery() {
 
 #ifdef CAN_FD
 // Functions
+#ifdef DEBUG_CANFD_DATA
+void print_canfd_frame(CANFDMessage rx_frame) {
+  int i = 0;
+  Serial.print(rx_frame.id, HEX);
+  Serial.print(" ");
+  for (i = 0; i < rx_frame.len; i++) {
+    Serial.print((b >> 4) & rx_frame.data[i], HEX);
+    Serial.print(" ");
+  }
+  Serial.println(" ");
+}
+
+void debug_canfd_frame(CANFDMessage frame) {
+  // Frame ID-s that battery transmits. For debugging and development.
+  // switch (frame.id)
+  // {
+  // case 0x7EC:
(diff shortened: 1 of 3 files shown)


Software/src/battery/KIA-E-GMP-BATTERY.cpp
```diff
@@ -195,17 +195,79 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+  const bool ok = canfd.tryToSend(frame);
+  if (ok) {
+  } else {
+    Serial.println("Send canfd failure.");
+  }
+#else
+  canfd.tryToSend(frame);
+#endif
+}
+#ifdef DEBUG_CANFD_DATA
```
dalathegreat, line 209: Could I suggest moving this out to Software.ino instead? That way it could be used for all CAN-FD protocols. I have been thinking about doing the same for the normal CAN debug mode

dalathegreat (changes requested):

Software/Software.ino
```diff
@@ -457,10 +457,61 @@ void init_battery() {
...
+  Serial.print(" ");
+  for (i = 0; i < rx_frame.len; i++) {
+    Serial.print(rx_frame.data[i] < 16 ? "0" : "");
+    Serial.print(rx_frame.data[i], HEX);
+    Serial.print(" ");
+  }
+  Serial.println(" ");
+}
+
+void debug_canfd_frame(CANFDMessage frame) {
```
dalathegreat, line 473: One more suggestion, would be better if we only kept the `print_canfd_frame` function, and skipped the `debug_canfd_frame` in Software.ino!

kukumagi: Merged the two functions into one.

dalathegreat (approved):

PR #290, opened 2024-05-06
Feature: Add configurable limit to charge/discharge on BYD Modbus

dalathegreat:
### What
This PR does two things:
- New feature: Values set in webserver for max charge/discharge in amps is now read and used by the BYD Modbus inverter protocol
- Performance improvements for BYD Modbus

### Why
To be able to tune peak power from the webui on Fronius inverters

### How
Using the std::min function to see which is smaller, User set limit OR battery value? We use the smallest one! This PR also removes all memcopy handling, and simplifies the register writing significantly

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -29,28 +29,25 @@ void handle_static_data_modbus_byd() {
     memcpy(&mbPV[i], data_array_pointers[arr_idx], data_size);
     i += data_size / sizeof(uint16_t);
   }
+  static uint16_t init_p201[13] = {0, 0, 0, MAX_POWER, MAX_POWER, 0, 0, 53248, 10, 53248, 10, 0, 0};
+  memcpy(&mbPV[200], init_p201, sizeof(init_p201));
+  static uint16_t init_p301[24] = {0,  0,  128, 0, 0,  0,     0, 0, 0,  2000,  0,   2000,
+                                   75, 95, 0,   0, 16, 22741, 0, 0, 13, 52064, 230, 9900};
+  memcpy(&mbPV[300], init_p301, sizeof(init_p301));
 }
 
 void handle_update_data_modbusp201_byd() {
   // Store the data into the array
-  static uint16_t system_data[13];
-  system_data[0] = 0;  // Id: p201 Value: 0 Scaled: 0 Comment: Always 0
-  system_data[1] = 0;  // Id: p202 Value: 0 Scaled: 0 Comment: Always 0
-  system_data[2] = std::min(datalayer.battery.info.total_capacity_Wh, static_cast<uint32_t>(60000));
+  static uint16_t system_data[5];
+  system_data[1] = std::min(datalayer.battery.info.total_capacity_Wh, static_cast<uint32_t>(60000));
   // Id: p203 Value: 32000 Scaled: 32kWh Comment: Capacity rated, maximum value is 60000 (60kWh)
-  system_data[3] = MAX_POWER;  // Id: p204 Value: 32000 Scaled: 32kWh Comment: Nominal capacity
-  system_data[4] = MAX_POWER;  // Id: p205 Value: 14500 Scaled: 30,42kW Comment: Max Charge/Discharge Power=10.24kW
-  system_data[5] = (datalayer.battery.info.max_design_voltage_dV);
+  system_data[2] = MAX_POWER;  // Id: p204 Value: 32000 Scaled: 32kWh Comment: Nominal capacity
+  system_data[3] = MAX_POWER;  // Id: p205 Value: 14500 Scaled: 30,42kW Comment: Max Charge/Discharge Power=10.24kW
+  system_data[4] = (datalayer.battery.info.max_design_voltage_dV);
   // Id: p206 Value: 3667 Scaled: 362,7VDC Comment: Max Voltage, if higher charging is not possible (goes into forced discharge)
-  system_data[6] = (datalayer.battery.info.min_design_voltage_dV);
+  system_data[5] = (datalayer.battery.info.min_design_voltage_dV);
   // Id: p207 Value: 2776 Scaled: 277,6VDC Comment: Min Voltage, if lower Gen24 disables battery
-  system_data[7] = 53248;  // Id: p208 Value: 53248 Scaled: 53248 Comment: Always 53248 for BYD
-  system_data[8] = 10;     // Id: p209 Value: 10 Scaled: 10 Comment: Always 10
-  system_data[9] = 53248;  // Id: p210 Value: 53248 Scaled: 53248 Comment: Always 53248 for BYD
-  system_data[10] = 10;    // Id: p211 Value: 10 Scaled: 10 Comment: Always 10
-  system_data[11] = 0;     // Id: p212 Value: 0 Scaled: 0 Comment: Always 0
-  system_data[12] = 0;     // Id: p213 Value: 0 Scaled: 0 Comment: Always 0
-  memcpy(&mbPV[200], system_data, sizeof(system_data));
+  memcpy(&mbPV[202], system_data, sizeof(system_data));
 }
 
 void handle_update_data_modbusp301_byd() {
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -3,6 +3,8 @@
 #include "../datalayer/datalayer.h"
 #include "BYD-MODBUS.h"
 
+// For modbus register definitions, see https://gitlab.com/pelle8/inverter_resources/-/blob/main/byd_registers_modbus_rtu.md
+
 void update_modbus_registers_inverter() {
   verify_temperature_modbus();
   handle_update_data_modbusp201_byd();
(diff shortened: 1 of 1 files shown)


Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -32,40 +34,28 @@ void handle_static_data_modbus_byd() {
...
-  system_data[9] =
-      53248;  // Id.: p210 Value.: 53248 Scaled value.: 53248 Comment.: Always 53248 for this BYD, Peak Discharge power?
-  system_data[10] = 10;  // Id.: p211 Value.: 10 Scaled value.: 10 Comment.: Always 10
-  system_data[11] = 0;   // Id.: p212 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  system_data[12] = 0;   // Id.: p213 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  static uint16_t i = 200;
-  memcpy(&mbPV[i], system_data, sizeof(system_data));
+  //system_data[0] = 0;  // Id: p201 Value: 0 Scaled: 0 Comment: Always 0
+  //system_data[1] = 0;  // Id: p202 Value: 0 Scaled: 0 Comment: Always 0
+  system_data[2] = std::min(static_cast<int>(datalayer.battery.info.total_capacity_Wh), 60000);
```
Cabooman, line 39: It would be better to cast "60000" to uint32_t since assigning an int to a uint16_t isn't preferred. Assigning a uint32_t with a maximum value of 60000 is totally OK tho

Same for the other casts in other functions

Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -32,40 +34,28 @@ void handle_static_data_modbus_byd() {
...
-  system_data[8] = 10;  // Id.: p209 Value.: 10 Scaled value.: 10 Comment.: Always 10
-  system_data[9] =
-      53248;  // Id.: p210 Value.: 53248 Scaled value.: 53248 Comment.: Always 53248 for this BYD, Peak Discharge power?
-  system_data[10] = 10;  // Id.: p211 Value.: 10 Scaled value.: 10 Comment.: Always 10
-  system_data[11] = 0;   // Id.: p212 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  system_data[12] = 0;   // Id.: p213 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  static uint16_t i = 200;
-  memcpy(&mbPV[i], system_data, sizeof(system_data));
+  //system_data[0] = 0;  // Id: p201 Value: 0 Scaled: 0 Comment: Always 0
+  //system_data[1] = 0;  // Id: p202 Value: 0 Scaled: 0 Comment: Always 0
```
Cabooman, line 38: Maybe these could be moved and initialized to 0 somewhere? Static variables are in most cases initialized to 0 but it could be risky to rely on compiler behavior

Same for the rest of the 0's, and also for the non-zero constant values

Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -76,74 +66,56 @@ void handle_update_data_modbusp301_byd() {
...
-  battery_data[19] = 0;   // Id.: p320 Value.: 0 Scaled value.: 0 Comment.: always 0
-  battery_data[20] = 13;  // Id.: p321 Value.: 0 Scaled value.: 0 Comment.: counter discharge hi
-  battery_data[21] =
-      52064;  // Id.: p322 Value.: 0 Scaled value.: 0 Comment.: counter discharge lo....65536*92+7448 = 6036760 Wh?
-  battery_data[22] = 230;  // Id.: p323 Value.: 0 Scaled value.: 0 Comment.: device temperature (23 degrees)
-  battery_data[23] = datalayer.battery.status.soh_pptt;  // Id.: p324 Value.: 9900 Scaled value.: 99% Comment.: SOH
-  static uint16_t i = 300;
-  memcpy(&mbPV[i], battery_data, sizeof(battery_data));
+  // Convert max discharge Amp value to max Watt
+  static int user_configured_max_discharge_W =
```
Cabooman, line 85: Is user_configured_max_discharge_W intended to be constant? Using "static" will initialize it once and never again in subsequent function calls since there's no assignment outside of the declaration. I also prefer using unsigned types with unsigned types and not involving ints :)

Same for other variables

Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -76,74 +66,56 @@ void handle_update_data_modbusp301_byd() {
...
+  //battery_data[15] = 0;   // Id: p316 Value: 0 Scaled: 0 Comment: always 0
+  battery_data[16] = 16;     // Id: p317 Value: 0 Scaled: 0 Comment: counter charge hi
+  battery_data[17] = 22741;  // Id: p318 Value: 0 Scaled: 0 Comment: counter charge lo
+  //battery_data[18] = 0;   // Id: p319 Value: 0 Scaled: 0 Comment: always 0
+  //battery_data[19] = 0;   // Id: p320 Value: 0 Scaled: 0 Comment: always 0
+  battery_data[20] = 13;     // Id: p321 Value: 0 Scaled: 0 Comment: counter discharge hi
+  battery_data[21] = 52064;  // Id: p322 Value: 0 Scaled: 0 Comment: counter discharge lo
+  battery_data[22] = 230;    // Id: p323 Value: 0 Scaled: 0 Comment: device temperature (23 degrees)
+  battery_data[23] = datalayer.battery.status.soh_pptt;  // Id: p324 Value: 9900 Scaled: 99% Comment: SOH
+  memcpy(&mbPV[300], battery_data, sizeof(battery_data));
```
Cabooman, line 118: Same comment as earlier but a little different:
Perhaps the constant values including 0's could be initialized at startup? It would require some rearranging but would make this part slimmer and would also get rid of the commented assignments

Cabooman (changes requested):

Software/src/inverter/BYD-MODBUS.cpp
dalathegreat: Agreed, fixed in [293ef97](https://github.com/dalathegreat/Battery-Emulator/pull/290/commits/293ef97c7e85fbe4a5f997a9d774f568880b7dd8)

Software/src/inverter/BYD-MODBUS.cpp
dalathegreat: Let's save this for a bigger modbus refactoring. Restored the 0 writes in [57b580f](https://github.com/dalathegreat/Battery-Emulator/pull/290/commits/57b580f4c2e5c18d75133c0cbf499b5cfeef0d69)

Software/src/inverter/BYD-MODBUS.cpp
dalathegreat: Same comment as above, let's save this for a bigger refactoring session. Restored the uncommented 0 writes in [57b580f](https://github.com/dalathegreat/Battery-Emulator/pull/290/commits/57b580f4c2e5c18d75133c0cbf499b5cfeef0d69)

dalathegreat: I changed my mind! Did the first p201 refactoring here: [3a640bb](https://github.com/dalathegreat/Battery-Emulator/pull/290/commits/3a640bb2b81fc0987332cdb0515b6863948d0414)

dalathegreat:
Converting to draft, I want to investigate more on how to increase performance

dalathegreat:
Ready for review, this is now extremely sped up! :)

smaresca:
Looks good to me given the PR scope. I agree with Caboom on the need/suggestions for initialization refactoring; I had the same impulses. Glad to see reduced reliance upon memcpy. For the future: configurable limit for charge/discharge being expanded elsewhere.

Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -27,123 +35,55 @@ void handle_static_data_modbus_byd() {
...
-      53248;  // Id.: p208 Value.: 53248 Scaled value.: 53248 Comment.: Always 53248 for this BYD, Peak Charge power?
-  system_data[8] = 10;  // Id.: p209 Value.: 10 Scaled value.: 10 Comment.: Always 10
-  system_data[9] =
-      53248;  // Id.: p210 Value.: 53248 Scaled value.: 53248 Comment.: Always 53248 for this BYD, Peak Discharge power?
-  system_data[10] = 10;  // Id.: p211 Value.: 10 Scaled value.: 10 Comment.: Always 10
-  system_data[11] = 0;   // Id.: p212 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  system_data[12] = 0;   // Id.: p213 Value.: 0 Scaled value.: 0 Comment.: Always 0
-  static uint16_t i = 200;
-  memcpy(&mbPV[i], system_data, sizeof(system_data));
+  mbPV[202] = std::min(datalayer.battery.info.total_capacity_Wh, static_cast<uint32_t>(60000));
```
Cabooman, line 46: Here and in other places: 
Do you get a compile error if you don't use the static cast? A solution that's more easy on the eyes could be to simply try stating the constant value as "60000u" instead

Cabooman (approved):
LGTM! There's always small stuff to tweak but iterations are king

Software/src/inverter/BYD-MODBUS.cpp
dalathegreat: That works! Fixed in [6eb518a](https://github.com/dalathegreat/Battery-Emulator/pull/290/commits/6eb518ae77063ae0598b3b27c12d1dc4ae5ddd5c)

PR #294, opened 2024-05-12
Bugfix: Fix overwritten cell voltage value

smaresca:
Remove 0x426 block from BMW i3 support as it is not currently known how to provoke the SME into responding with a frame with that ID.

Originally: a fix for an overwritten index of datalayer.battery.status.cell_voltages_mV in the BMW i3 code for 0x426 frames.


dalathegreat:
@smaresca , can you remove the entire 0x426 block? :)

smaresca:
amended to remove the 0x426 case per @odlg  and @dalathegreat

odlg (approved):

PR #297, opened 2024-05-12
Bugfix: Correct scaling on kWh for Solax

dalathegreat:
### What
This PR fixes kWh remaning on Solax 

### Why
Decimal point was wrong;
bild

### How
Instead of dividing value with 100, divide with 10


josiahhiggs:
This works just right, lilygo and solax are now reading within 0.1kWh.

PR #298, opened 2024-05-12
Improvement: Send cellmV when cell events are raised

dalathegreat:
### What
This PR makes the cell over/under/deviation events easier to troubleshoot. Added to all battery types that were missing this.

### Why
Now when for instance a cell overvoltage event is raised, it is hard to know what the value of the cell was when it occured.

### How
Since the event can only get an u8 parsed (0-255), we shrink the cellvoltage by 20, and when reviewing the events you can simply add *20 to the value. (4250mV sent as 4250/20 = 212)

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -455,7 +455,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   // Start checking safeties. First up, cellvoltages!
   if (battery_cell_deviation_mV > MAX_CELL_DEVIATION_MV) {
-    set_event(EVENT_CELL_DEVIATION_HIGH, 0);
+    set_event(EVENT_CELL_DEVIATION_HIGH, battery_cell_deviation_mV);
   } else {
     clear_event(EVENT_CELL_DEVIATION_HIGH);
   }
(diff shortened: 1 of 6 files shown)


Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -455,36 +455,36 @@ void update_values_battery() {  //This function maps all the values fetched via
 
   // Start checking safeties. First up, cellvoltages!
   if (battery_cell_deviation_mV > MAX_CELL_DEVIATION_MV) {
-    set_event(EVENT_CELL_DEVIATION_HIGH, 0);
+    set_event(EVENT_CELL_DEVIATION_HIGH, battery_cell_deviation_mV);
```
odlg, line 458: If (for some erroneous reason) min-cell voltage is 0, you will not see it here as the value will wrap around.
Perhaps some function to make value here somewhat logarithmic. Like 0-240 is mV difference and from 241 and up it is increasing value with 1 means 20mV for each. Another option is to make event values 16 bit?

odlg (changes requested):

Software/src/battery/BMW-I3-BATTERY.cpp
dalathegreat: I think this is good enough, until we get the event value 16bit. Better than a 0, and sure, the edge cases wont be handled perfectly, but better than before atleast :)

dalathegreat:
Scrapping this PR, will rewrite it as a general safety handler to get rid of battery specific implementations!

PR #299, opened 2024-05-12
Bugfix/bmw i3 messages

odlg:
**What**
Fixed fetching extra data and wait until data available before setting low voltage event. Use max cell voltage instead of average voltage. Reduced risk of making unneeded events. 

**Why**
Make using BMW i3 batteries more robust.

**How**
Send different defined 0x6F1 messages and parse them as the type requested. Made logic regarding raising events more robust.

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -451,8 +451,10 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   datalayer.battery.status.temperature_max_dC = battery_temperature_max * 10;  // Add a decimal
 
-  datalayer.battery.status.cell_min_voltage_mV = datalayer.battery.status.cell_voltages_mV[0];
-  datalayer.battery.status.cell_max_voltage_mV = datalayer.battery.status.cell_voltages_mV[2];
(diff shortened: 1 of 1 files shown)


Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -447,44 +451,57 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+    // Start checking safeties. First up, cellvoltages!
+    if (battery_cell_deviation_mV > MAX_CELL_DEVIATION_MV) {
+      uint8_t report_value = 0;
+      if (battery_cell_deviation_mV <= 20 * 254) {
+        report_value = battery_cell_deviation_mV / 20;
+      } else {
+        report_value = 255;
+      }
+
+      set_event(EVENT_CELL_DEVIATION_HIGH, report_value);
```
dalathegreat, line 474: Like I mentioned, It would be good enough to just do
set_event(EVENT_CELL_DEVIATION_HIGH, battery_cell_deviation_mV);

here, since we raise an alarm if it goes over 150mv. And 150 fits into the u8!

dalathegreat: @odlg Anyways, I approved this PR, we can handle that discussion in the other PR!

dalathegreat (approved):

PR #301, opened 2024-05-13
New feature: Safety.cpp

dalathegreat:
### What
This PR improves the overall safety of the Battery-Emulator project, by moving out most common machinery protection events from each battery handler, and unifying it in a single file (Safety.cpp)

### Why
Having a dedicated spot to verify common safety checks improves safety. Less chance of mistakes in individual implementations. This also makes it easier for newcomers to add features, or new battery types

### How
Two new parameters added to datalayer (`CAN_error_counter` and `CAN_battery_still_alive`)
The following events are now checked globally in the Safety.cpp file:

- Battery overheating (New event!)
- Battery frozen (New event!)
- SOH% too low
- Cell deviation
- CAN still alive sending
- CAN CRC corrupted message counter


dalathegreat:
Tested on bench, working great! 
bild

Software/src/devboard/utils/events.h
```diff
@@ -38,10 +38,12 @@
   XX(EVENT_KWH_PLAUSIBILITY_ERROR)      \
   XX(EVENT_BATTERY_EMPTY)               \
   XX(EVENT_BATTERY_FULL)                \
+  XX(EVENT_BATTERY_FROZEN)              \
   XX(EVENT_BATTERY_CAUTION)             \
   XX(EVENT_BATTERY_CHG_STOP_REQ)        \
   XX(EVENT_BATTERY_DISCHG_STOP_REQ)     \
   XX(EVENT_BATTERY_CHG_DISCHG_STOP_REQ) \
+  XX(EVENT_BATTERY_OVERHEAT)            \
```
smaresca, line 46: I'd like to see events for other catastrophic errors we need to be careful about like contacts being welded (both Tesla model 3 and CHAdeMO can convey that state, for example), or pack under/over voltage (whether measured or computed relative to a battery/inverter specific threshold). These could be subsequent to this PR, but I wanted to mention while reviewing.

Software/src/devboard/utils/types.h
```diff
@@ -28,6 +28,6 @@ enum led_color { GREEN, YELLOW, RED, BLUE, RGB };
 #define INTERVAL_100_MS_DELAYED 120
 #define INTERVAL_500_MS_DELAYED 550
 
-#define MAX_CAN_FAILURES 500  // Amount of malformed CAN messages to allow before raising a warning
+#define CAN_STILL_ALIVE 12
```
smaresca, line 31: Please add a comment for context - e.g., explain that this is intended for use as a decrementing counter,  why the value of 12 was chosen, etc

Software/src/datalayer/datalayer.h
```diff
@@ -68,6 +68,11 @@ typedef struct {
    * battery.settings.soc_scaling_active
    */
   uint16_t reported_soc;
+  /** A counter that increases incase a CAN CRC read error occurs */
+  uint16_t CAN_error_counter;
+  /** uint8_t */
+  /** A counter set each time a new message comes from battery.*/
+  uint8_t CAN_battery_still_alive = 12;
```
smaresca, line 75: The define should be used here not the raw scalar. Same feedback re: comment to provide context around decrementing nature and why the value was selected

smaresca (changes requested):
Definitely a good introduction. Some minor feedback, requests, and observations/suggestions for future revisions.

Software/src/datalayer/datalayer.h
dalathegreat: Added in [b4773d8](https://github.com/dalathegreat/Battery-Emulator/pull/301/commits/b4773d89d8c9da48e760ade0c253ed39ded1753b)

Software/src/devboard/utils/types.h
dalathegreat: Comment updated in [b4773d8](https://github.com/dalathegreat/Battery-Emulator/pull/301/commits/b4773d89d8c9da48e760ade0c253ed39ded1753b)

Software/src/devboard/utils/events.h
dalathegreat: Good idea, I added a common under and overvoltage event in [841c144](https://github.com/dalathegreat/Battery-Emulator/pull/301/commits/841c144ff6b530a869ad5492573f0b8ebdd0d4a1)

smaresca (approved):
LGTM!

PR #303, opened 2024-05-14
Add SimpleISA library for use in measuring current/voltage

smaresca:
Note: lightly modified from source material to support ESP32-Arduino-CAN. 

Intended to help facilitate enhanced CHAdeMO support.

There's room for improvement in the upstream SimpleISA library, but modifications were kept light to avoid deviating too much from the variant being used in separate EV-oriented projects by damienmaguire, isaac96, and arber333.


dalathegreat (approved):

dalathegreat:
Remember to just add to the main README that this lib is used!

PR #304, opened 2024-05-14
Rework CHAdeMO support to more fully implement standard

smaresca:
**What**
This enhancement includes the appropriate CHAdeMO sequence for negotiation, connector locking and discharge support. Testing and further refinement is needed before this feature is entirely usable. Charging support is not yet present, though this PR lays the groundwork.

**Why**
To enable CHAdeMO vehicles to function as a battery, while retaining the safety afforded by a connector with locking solenoid, contactor control, and similar.

**How**
By implementing as much of the CHAdeMO standard as feasible given hardware limitation and access to closely-held specs.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -97,6 +97,9 @@ enum State { DISCONNECTED, PRECHARGE, NEGATIVE, POSITIVE, PRECHARGE_OFF, COMPLET
 State contactorStatus = DISCONNECTED;
 
 #define MAX_ALLOWED_FAULT_TICKS 1000
+/* NOTE: modify the precharge time constant below to account for the resistance and capacitance of the target system.
+ *	t=3RC at minimum, t=5RC ideally 
+ */
 #define PRECHARGE_TIME_MS 160
 #define NEGATIVE_CONTACTOR_TIME_MS 1000
 #define POSITIVE_CONTACTOR_TIME_MS 2000
@@ -459,6 +462,10 @@ void init_inverter() {
 void init_battery() {
   // Inform user what battery is used and perform setup
   setup_battery();
+
+#ifdef CHADEMO_BATTERY
+  intervalUpdateValues = 800;  // This mode requires the values to be updated faster
+#endif
 }
 
 #ifdef CAN_FD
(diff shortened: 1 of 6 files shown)


Software/src/battery/CHADEMO-BATTERY.h
```diff
@@ -5,6 +5,9 @@
 #include "../lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.h"
 
 #define BATTERY_SELECTED
+#ifndef CONTACTOR_CONTROL
+#error Contactor control is required for CHADEMO support
```
smaresca, line 9: This is causing a test to fail. Should the preprocessor #error be retained, hardcode the define for CONTACTOR_CONTROL, or some alternative?

dalathegreat: Maybe code in a define for ContactorControl incase Chademo is selected? So users dont miss it?

Software/Software.ino
```diff
@@ -453,6 +456,14 @@ void init_inverter() {
 void init_battery() {
   // Inform user what battery is used and perform setup
   setup_battery();
+
+#ifdef CHADEMO_BATTERY
+  intervalUpdateValues = 800;  // This mode requires the values to be updated faster
+#endif
+
+#ifndef BATTERY_SELECTED
```
dalathegreat, line 464: This #ifndef BATTERY_SELECTED warning can be removed, it has been moved to the other file in another PR

Software/src/battery/CHADEMO-BATTERY.h
smaresca: I interpreted your comment as a suggestion to a #define CONTACTOR_CONTROL here in lieu of a test. Let me know if you meant something different

Software/Software.ino
smaresca: Removed

Software/src/battery/CHADEMO-BATTERY.h
dalathegreat: Yes, I meant that if a user has selected Chademo, the Chademo file would automatically also  define the CONTACTOR_CONTROL which is needed!

dalathegreat (approved):

PR #310, opened 2024-05-18
Improvement: Move charge/discharge limiting to Safety file

dalathegreat:
### What
This PR moves the charge/discharge W cap due to high/low SOC% out from each battery file, into the Safety.cpp file

### Why
Safer operation of battery. Less chance of overcharging / underdischarging. Simpler code, easier to maintain and add new batteries.

### How
- Duplicate code chunks moved to the Safety.cpp file
- Plausible SOC% check is now global instead of only for LEAF
- Event for battery full/empty added for all batteries in global safety file
- Battery max charge values no longer capped to uint16 (forgotten from datalayer intro)
- Bonus: Make BMW i3 not warn about SOH% on startup
- Bonus: Skip evaluating same safeties on double lilygo


Software/src/devboard/safety/safety.cpp
```diff
@@ -34,13 +34,43 @@ void update_machineryprotection() {
     clear_event(EVENT_BATTERY_UNDERVOLTAGE);
   }
 
+  // Battery is fully charged. Dont allow any more power into it
+  // Normally the BMS will send 0W allowed, but this acts as an additional layer of safety
+  if (datalayer.battery.status.reported_soc == 10000)  //Scaled SOC% value is 100.00%
+  {
+    set_event(EVENT_BATTERY_FULL, 0);
+    datalayer.battery.status.max_charge_power_W = 0;
```
odlg, line 42: I guess in cases like BYD_MODBUS where modbus communication runs on one cpu core and this runs on the other - then this would flip-flap between 0 and the value set in the battery code.

dalathegreat: No it will not. The max_charge_power_W  gets updated in battery, but safeties modify value before inverter sees the value!

bild

odlg: @dalathegreat That is not sufficient. The other cpu-core runs in its own world (loop) and reads from and writes to the datalayer independently.

odlg: @dalathegreat One approach to would be to let the battery code update one subpart of the datalayer, then let the safety code either copy or set different values in a to-inverter subpart. Then all inverters should read from that subpart.

dalathegreat: I gave this some more thought, thanks @odlg , Removed value writing for now, so we can get this merged [e2ffb69](https://github.com/dalathegreat/Battery-Emulator/pull/310/commits/e2ffb695ca7a850f4cb9e244e2e6ce2262124247)

odlg (approved):

Software/src/devboard/safety/safety.cpp
dalathegreat: @odlg , I have been testing this extensively, with adding back the writing (in a bit different way), on the latest 6.0.RC branch. There is no value toggling what so ever, the data sent to inverter is locked to 0W when an absolute end is reached. Tested it with real CAN data, and also printing the value to serial as fast as possible, without it having any toggling. So I am calling this PR done for now, it simplifies adding batteries by quite much!

bild

odlg: @dalathegreat Which inverter did you use for the testing? It needs to be one of those who run their own background loop like BYD_MODBUS.
I guess measurements could be made on what percentage of the main loop time the value is wrong. It might be less than one percent, but still then one percent of the inverter readings will be wrong.

dalathegreat: BYD_MODBUS is precisely the one I tested! I will be trending the values from inverter now before we release the 6.0.RC as a standard release!

PR #311, opened 2024-05-19
Improvement - CHAdeMO: Add measured current and voltage via IVT Shunt

smaresca:
**What**

This PR updates EVSE data structures with measured current and voltage information in lieu of hardcoding values or using vehicle-supplied limits as the asserted values.

**Why**

Measured current and voltage are required for various aspects of CHAdeMO protocol compliance, safety checks, and dynamic power adjustments.

**How**

This functionality is achieved by using the SimpleISA adaptation added in a previous PR. When the ISA shunt is not available, there is an attempt to simulate protocol expectations..these simulations may be removed for safety purposes.


dalathegreat (approved):
LGTM!

PR #317, opened 2024-05-24
Fix canfd startup. Now when INT pin is connected then everything shou…

kukumagi:
…ld work as intended and no need to fiddle with disconnecting INT pin

What
This PR fixes canfd startup with INT pin connected.

Why
For some reason canfd did not start with INT pin connected and it required disconnecting INT and then reconnecting for it to work.

How
After many trial and error and debugging the ACAN2517FD library it turns out that the fix was to call canfd.poll() after starting. Dont know why but it works for me.


dalathegreat (approved):

dalathegreat:
Nice work @kukumagi , merging this!

PR #321, opened 2024-05-26
Prevent MQTT to publish cell voltage values before they are populated

amarofarinha:
In addition to checking the value of the variable number_of_cells, we also test the value of the last cell, as the value of the cells can be filled some time after initializing the number of cells. This test was also added before publishing the values cell_max_voltage and cell_min_voltage since they are ​​calculated from the cell data.


dalathegreat (approved):

dalathegreat:
Looks like a good fix! 👍

PR #324, opened 2024-05-27
LEAF - Setting cell voltage values after last cell frame values

amarofarinha:
Ensure that the values ​​in the cell voltage datalayer are only updated when all information from all cells is received from battery. Changed the maximum cell reference in cellmonitor to use the specific datalayer information of the battery in use instead of the MAX_AMOUNT_CELLS which is the maximum possible value.


dalathegreat (approved):
LGTM! 👍

PR #328, opened 2024-05-29
Bugfix: Nissan LEAF, Contactor closing allowed

dalathegreat:
### What
This PR fixes a bug, that sometimes the LEAF battery would not set the "Allowed contactor closing" ✅ when starting up

### Why
This prevented people using automatic contactor control from having a reliable startup

### How
We wait 500ms until we start sending CAN messages towards the battery, allowing it to fully start


StarkJohan (approved):
Tested and working.

PR #332, opened 2024-06-02
Performance on Leaf CAN receive + possible fix for KIA cell bateries values spikes

amarofarinha:
### What

- Performance on Leaf CAN receive
- possible fix for KIA cell bateries values spikes

### Why
To improve code quality and to have coherent cell voltage values ​​published via MQTT

### How
In the LEAF code, replace the "for" voltage array  copy with memcopy (which is more efficient). 
In the KIA, change the location where the voltage array is copied, similar to what has already been done in the LEAF.

### DIFF
--- a/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
+++ b/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
@@ -25,7 +25,7 @@ static uint16_t batteryVoltage = 0;
 static uint16_t inverterVoltageFrameHigh = 0;
 static uint16_t inverterVoltage = 0;
 static uint16_t cellvoltages_mv[98];
-static int16_t leadAcidBatteryVoltage = 0;
+static int16_t leadAcidBatteryVoltage = 120;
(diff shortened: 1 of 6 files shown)


Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
```diff
@@ -275,6 +276,10 @@ void receive_can_battery(CAN_frame_t rx_frame) {
       startedUp = true;
       datalayer.battery.status.CAN_battery_still_alive = CAN_STILL_ALIVE;
       SOC_Display = rx_frame.data.u8[0] * 5;  //100% = 200 ( 200 * 5 = 1000 )
+      //Map all cell voltages to the global array
+      memcpy(datalayer.battery.status.cell_voltages_mV, cellvoltages_mv, 98 * sizeof(uint16_t));
```
dalathegreat, line 280: Was this tested on a 90S battery? Just wondering if it works also there

dalathegreat (approved):

PR #341, opened 2024-06-09
BugFix  - calculate number of cell in KIA pack

amarofarinha:
Changed code location where number of cells are calculated. bug introduced on PR 332


dalathegreat (approved):
Approved, waiting for someone to test it!

PR #345, opened 2024-06-13
CPU load improvement: Apply less filtering on CAN signals

dalathegreat:
### What
This PR attempts to reduce CPU load, by filtering the CAN signals less

### Why
The code occasionally spikes in CPU load, this should help!

### How
Instead of triple checking that the CAN signals are going high/low, we instead directly set the values to high/low once they are sampled. This reduces CPU load in two ways;
- Less Processing: Sampling once requires fewer computations and less time compared to sampling three times.
- Reduced Interrupt Load: With fewer samples, there are fewer interrupts generated for processing the CAN messages.

:warning: After making this change, we need to monitor CAN communication performance to ensure that it has the desired effect of reducing CPU load, and does not introduce any new CAN issues


dalathegreat:
Putting this on hold until we have more CRC checks in place on battery protocols! Not sure we need this performance boost, might be too risky. PR #346 improved performance enough for now!

dalathegreat:
On second thought, let's continue looking into getting this merged!

PR #346, opened 2024-06-14
Move WIFI out from the core task

Cabooman:
### WHAT
A relocation of wifi_monitor() to a separate task dedicated to various connectivity

### WHY
The WIFI functionality sometimes takes a looong time to finish handling reconnections and other shenanigans, which we don't want to affect core functions

### HOW
Move wifi_monitor() to the mqtt task, rename everything properly enough, test

### TESTING
Tested on my personal FroLeaf setup. WIFI load now doesn't affect the core task


Software/Software.ino
```diff
@@ -128,8 +128,8 @@ void setup() {
   init_webserver();
   init_mDNS();
```
Cabooman, line 129: This should be moved into the connectivity task!

dalathegreat (approved):

PR #358, opened 2024-06-26
Removed MQTT subscription functionality

fspijkerman:
### What

Removed the code that allows subscriptions to MQTT topics

### Why
It has no functional application (currently), only a debug callback. 
We could easily implement it back once a requirement/feature requires it, in the mean time we save some cpu cycles and ram.

### How

Removed the code, after that tested if publishing MQTT data is not affected.


dalathegreat (approved):
LGTM! 👍

PR #360, opened 2024-06-28
Support platformIO builds and visual studio code

MartinVerges:
This branch allows the use of Visual Studio Code and PlatformIO.

Unfortunately, I could not figure out how to solve the problem with `ino` vs `cpp` files so this will break Arduino IDE. 
In order to work with Visual Studio Code the file needs to be named `Software.cpp` and for Arduino `Software.ino` otherwise you will see compile errors.

### DIFF
--- a/Software/Software.h
+++ b/Software/Software.h
@@ -1,0 +1,20 @@
+
+void init_serial();
+void init_stored_settings();
+void init_CAN();
+void init_contactors();
+void init_rs485();
+void init_inverter();
+void init_battery();
+void receive_can();
+void send_can();
+void send_can2();
+void update_SOC();
+void update_values_inverter();
+void init_serialDataLink();
+void storeSettings();
+void check_reset_reason();
+void connectivity_loop(void* task_time_us);
+void core_loop(void* task_time_us);
+void init_mDNS();
+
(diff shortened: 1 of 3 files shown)


puffnfresh:
Coincidentally I had a got at this tonight and found this PR when raising my own. We don't need to change from the `ino` format to `cpp` - the functions get hoisted so the solution is to just `typedef` out the class, or similar. You can see this in my PR.

MartinVerges:
Thank you @puffnfresh, i adopted that and it seems to work with visual studio code and Arduino IDE.

dalathegreat:
Shall we close this one, and merge #361 , seems less complex to do so? :) Nice work you both!

MartinVerges:
If it works without the header file, than #361 is fine for me. My visual
studio was complaining and that's why I created that file.
Best regards
Martin Verges
+49 174 9335695 (Mobile, Telegram, WhatsApp,..)

Daniel Öster ***@***.***> schrieb am So., 30. Juni 2024,
12:53:

puffnfresh:
It does work without the header file.

dalathegreat:
Thanks, closing this and merging PR #361 ! 🙌

Issue #333, opened 2024-06-04
Build fails in platformio


obbardc:
Still works in Arduino IDE with ESP32 v3.0.0 so I guess this is low-priority.

Doesn't seem to build when opening the project in PlatformIO (vscode on Linux):
```
Processing esp32dev (platform: espressif32; board: esp32dev; framework: arduino)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/espressif32/esp32dev.html
PLATFORM: Espressif 32 (2024.6.10+sha.dcfe2a3) > Espressif ESP32 Dev Module
HARDWARE: ESP32 240MHz, 320KB RAM, 4MB Flash
DEBUG: Current (cmsis-dap) External (cmsis-dap, esp-bridge, esp-prog, iot-bus-jtag, jlink, minimodule, olimex-arm-usb-ocd, olimex-arm-usb-ocd-h, olimex-arm-usb-tiny-h, olimex-jtag-tiny, tumpa)
PACKAGES: 
 - framework-arduinoespressif32 @ 3.0.1+sha.371d83e 
 - tool-esptoolpy @ 4.7.3 
 - tool-mklittlefs @ 3.2.0 
 - tool-riscv32-esp-elf-gdb @ 12.1.0+20221002 
 - tool-xtensa-esp-elf-gdb @ 12.1.0+20221002 
 - toolchain-xtensa-esp32 @ 12.2.0+20230208
Converting Software.ino
LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ chain, Compatibility ~ soft
Found 25 compatible libraries
Scanning dependencies...
Dependency Graph
|-- ESPmDNS @ 2.0.0
|-- WiFi @ 2.0.0
|-- EEPROM @ 2.0.0
|-- Preferences @ 2.0.0
|-- FS @ 2.0.0
|-- SPI @ 2.0.0
|-- LittleFS @ 2.0.0
|-- Update @ 2.0.0
|-- WebServer @ 2.0.0
Building in release mode
Compiling .pio/build/esp32dev/src/Software.ino.cpp.o
Compiling .pio/build/esp32dev/src/USER_SETTINGS.cpp.o
[...]

PR #361, opened 2024-06-28
Use unsupported but working ESP32 version in PlatformIO

puffnfresh:
Alternative to #360.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -53,6 +53,8 @@ static ACAN2515_Buffer16 gBuffer;
 #ifdef CAN_FD
 #include "src/lib/pierremolinaro-ACAN2517FD/ACAN2517FD.h"
 ACAN2517FD canfd(MCP2517_CS, SPI, MCP2517_INT);
+#else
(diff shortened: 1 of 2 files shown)


Software/Software.ino
```diff
@@ -53,6 +53,8 @@ static ACAN2515_Buffer16 gBuffer;
 #ifdef CAN_FD
 #include "src/lib/pierremolinaro-ACAN2517FD/ACAN2517FD.h"
 ACAN2517FD canfd(MCP2517_CS, SPI, MCP2517_INT);
+#else
+typedef char CANFDMessage;
```
puffnfresh, line 57: Just any random type just to make the code compile. We can't have any void parameters so I just went with `char`.

puffnfresh:
Fixes #333.

dalathegreat (approved):
LGTM! Many people have asked for the PlatformIO support, and it is very nice to keep the oh so important Arduino IDE support intact, since that is what 99% of people use when taking the Battery-Emulator into use!

PR #364, opened 2024-06-30
Improvement: Improve performance on LED driver

dalathegreat:
### What
This PR reduces CPU load, by optmizing the LED driver

### Why
To improve CPU overhead on the Battery-Emulator, less time spent on LED modes, more time spent on actual battery-inverter synergy

### How
Stolen from @Cabooman 's CAN buffer branch 🕴️ 🙊

### DIFF
--- a/Software/src/lib/adafruit-Adafruit_NeoPixel/esp.c
+++ b/Software/src/lib/adafruit-Adafruit_NeoPixel/esp.c
@@ -87,89 +87,84 @@ static void IRAM_ATTR ws2812_rmt_adapter(const void* src, rmt_item32_t* dest, si
   *item_num = num;
 }
 
+static bool rmt_initialized = false;
+static bool rmt_adapter_initialized = false;
+
 void espShow(uint8_t pin, uint8_t* pixels, uint32_t numBytes, boolean is800KHz) {
-  // Reserve channel
-  rmt_channel_t channel = ADAFRUIT_RMT_CHANNEL_MAX;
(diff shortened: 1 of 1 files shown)


Software/src/lib/adafruit-Adafruit_NeoPixel/esp.c
```diff
@@ -87,89 +87,84 @@ static void IRAM_ATTR ws2812_rmt_adapter(const void* src, rmt_item32_t* dest, si
...
+  // rmt_wait_tx_done(config.channel, pdMS_TO_TICKS(100));
 
   // Free channel again
-  rmt_driver_uninstall(config.channel);
-  rmt_reserved_channels[channel] = false;
+  // rmt_driver_uninstall(config.channel);
+  // rmt_reserved_channels[channel] = false;
 
-  gpio_set_direction(pin, GPIO_MODE_OUTPUT);
+  // gpio_set_direction(pin, GPIO_MODE_OUTPUT);
```
Cabooman, line 167: I kept these as nice-to-have in case of issues, but since I haven't noticed any they could be removed

Cabooman (approved):
Looks good, and I've been running it on my end for a while without any apparent issues

Software/src/lib/adafruit-Adafruit_NeoPixel/esp.c
dalathegreat: Good point, removed in 47aabca !

PR #366, opened 2024-07-02
Minor CPU load improvement on CAN driver

dalathegreat:
### What
This PR improves CPU load, less cycles performed when sending a CAN message

### Why
Increase performance!

### How
Instead of clearing event on every CAN send, we only clear events incase we are recovering from a CAN send failure. This saves some CPU cycles.

### DIFF
--- a/Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
+++ b/Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
@@ -11,18 +11,17 @@ int ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
   if (tx_ok) {
     result = CAN_write_frame(p_frame);
     tx_ok = (result == 0) ? true : false;
-    if (tx_ok == false) {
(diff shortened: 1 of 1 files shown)


dalathegreat:
Alternatively, we could save some memory also, need to profile this,

`int ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
  static unsigned long start_time;
  if (tx_ok) { // Last transmit was OK, lets try another one!
    tx_ok = (CAN_write_frame(p_frame) == 0);
    if (!tx_ok) { // We failed to transmit
      #ifdef DEBUG_VIA_USB
      Serial.println("CAN failure! Check wires");
      #endif
      set_event(EVENT_CAN_TX_FAILURE, 0);
      start_time = millis();
    }
  } else { // Retry if we have waited over 20ms
    if ((millis() - start_time) >= 20) {
      tx_ok = true;
      clear_event(EVENT_CAN_TX_FAILURE);
    }
  } 
  return tx_ok ? 0 : -1;
}`

dalathegreat:
Closing, too little improvement to be considered

PR #376, opened 2024-07-09
Feature: Add SSID/PWD writing from webserver!

dalathegreat:
### What
This PR adds Wifi SSID/Password settings to the Webserver, so user can edit them there. Once entered, they get stored to permanent storage and loaded on boot. This PR also enables the LOAD_SAVED_SETTINGS_ON_BOOT by default.

image

### Why
This allows the user to update the board over the air with a software that is compiled without specific WiFi SSID and password. Flashing the binary that is compiled without specific WiFi SSID and password would still allow the board to connect to the WiFi network it was previously connected to. This helps the user to quickly update to the latest version of the software, without having to manually update the WiFi SSID and password in the source code.

### How
This allows the user to flash code without specific WiFi SSID and password to the board, connect to the WiFi access point of the board, and then add the WiFi SSID and password of the network they want the board to connect to.  Ofcourse also setting the password via the .cpp file as before works!

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -320,9 +320,6 @@ void init_stored_settings() {
   settings.clear();  // If this clear function is executed, no settings will be read from storage
 #endif
 
-  Serial.println("SSID before: ");
-  Serial.println(String(ssid.c_str()));
-
   char tempSSIDstring[63];  // Allocate buffer with sufficient size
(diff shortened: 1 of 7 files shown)


dalathegreat:
Merging this for now to get more broad feedback on it in .dev version!

PR #378, opened 2024-07-13
BYD Atto 3: Add contactor closing allowed section

dalathegreat:
### What
This PR adds setting the `battery_allows_contactor_closing` flag to allow contactor control to function

### Why
Some BYD Atto 3 batteries are locked if they are crashed too hard. This fix enables use of contactor control via GPIO pins, to manually force the contactors on.

### How
`battery_allows_contactor_closing` set to true if we are not in faulted mode


juancruz1953 (approved):
Works well - pre-charge and contactors close in the required sequence.

PR #386, opened 2024-07-17
Update webserver.cpp

GubbJVL:
Currently, soc min/max and charge/discharge amp does not accept decimals, due to the use of integers before multiplying. Changed toInt to toFloat, problem solved.

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -123,7 +123,7 @@ void init_webserver() {
   server.on("/updateSocMax", HTTP_GET, [](AsyncWebServerRequest* request) {
     if (request->hasParam("value")) {
       String value = request->getParam("value")->value();
-      datalayer.battery.settings.max_percentage = value.toFloat() * 100;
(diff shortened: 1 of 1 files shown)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -123,7 +123,7 @@ void init_webserver() {
   server.on("/updateSocMax", HTTP_GET, [](AsyncWebServerRequest* request) {
     if (request->hasParam("value")) {
       String value = request->getParam("value")->value();
-      datalayer.battery.settings.max_percentage = value.toInt() * 100;
+      datalayer.battery.settings.max_percentage = value.toFloat() * 100;
```
dalathegreat, line 126: Nice! @GubbJVL , could you explicitly cast the result to uint16_t to make the code more readable and clear about the intended behavior?

`datalayer.battery.settings.max_percentage = static_cast<uint16_t>(value.toFloat() * 100);`

Just as a precaution!

dalathegreat (changes requested):

dalathegreat (approved):

PR #388, opened 2024-07-20
BYD Atto 3: Add stuck SOC% value detection

dalathegreat:
### What
This PR adds improvements to SOC% reading for BYD Atto 3 batteries
- Reading of SOC% via periodically transmitted CAN message
- Detection of stuck SOC%
   - Event for info incase stuck SOC% is noticed

### Why
Previously we have been using voltage based SOC% estimate for the BYD Atto 3 all the time. This has been necessary, since the BMS goes into a semi-locked state if the vehicle is crashed too hard. However, if you use a pristine battery, estimated SOC% is not super accurate, and it would be better to use the real SOC% value sent by the BMS

### How
We start out with normal estimated SOC%, but now read the periodically transmitted SOC%. Incase the read value stays stuck at the same value for 1 hour after startup, we consider the value unusable. If the value moves around 3 times, we consider it usable, and ditch the estimation.

:warning: Incorrect switching from estimate to a locked value could result in damage to the battery. This algorithm needs testing on both locked and unlocked packs before we can merge it. BYD Atto 3 testers wanted!

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -106,7 +106,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
   datalayer.battery.status.voltage_dV = BMS_voltage * 10;
 
   if (BMS_state == UNLOCKED) {  // This value can be frozen if battery is crashed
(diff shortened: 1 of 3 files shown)


dalathegreat:
Initial testing showed that the constantly transmitted CAN message can jump from 55% -> 174%
image

This can not be merged in this state. We need to fix this first.

PR #393, opened 2024-07-27
Update Status-bits PYLON-CAN.cpp

LevinSwe:
- Add Sleep status to open contactors in inverter at Fault
- Change status-value from 0x 11, 12, 13 to 01, 02, 03 to not call for extra chargning.


dalathegreat:
Tested on the Discord with FerroAmp. Thanks for figuring this out @LevinSwe ! 🙌

dalathegreat (approved):

PR #400, opened 2024-08-05
Refactor: CAN abstraction layer

dalathegreat:
### What
This PR refactors all CAN message handling by adding our own CAN struct. The project no longer relies on the miwagner struct for CAN, we have our own definition in place now!

### Why
The amount of variations is starting to grow out of hand. There are so many battery/inverter specific hardcoded combinations in the code. This PR makes it free for the end user to specify any combination of hardware, and on which CAN interface the parts are connected. 

#### Bonus
- CAN-FD add-on chip can now be used for normal CAN also! Just enable `USE_CANFD_INTERFACE_AS_CLASSIC_CAN `in USER_SETTINGS
- This also makes it easier to integrate new hardware boards with the Battery-Emulator
- Compile time is faster, since less includes

### How
The user now configures on which CAN interface each component is connected to. This simplifies all code logic massively.

In the USER_SETTINGS file, you configure how everything is connected
image

The Webserver also shows what the current configuration is:
image

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -535,7 +535,9 @@ void receive_can_native() {  // This section checks if we have a complete CAN me
   if (xQueueReceive(CAN_cfg.rx_queue, &rx_frame, 0) == pdTRUE) {
 
     if (can_config.battery == CAN_NATIVE) {
+#ifndef SERIAL_LINK_RECEIVER
       receive_can_battery(rx_frame);
+#endif  // SERIAL_LINK_RECEIVER
     }
     if (can_config.inverter == CAN_NATIVE) {
 #ifdef CAN_INVERTER_SELECTED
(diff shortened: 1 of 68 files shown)


dalathegreat:
Thanks, @amarofarinha , I will refactor even further, just pushed some breaking changes, but let's see in a few days where this is at :sweat_smile:

dalathegreat:
This is ready for testing! :battery:

Alizainharoon:
Man you on fire 🔥

dalathegreat:
Merging this now! Running great on Double Battery setup (2x LEAF + Modbus BYD), and I have also bench tested the following combinations:
- CAN native
- MCP2515 add-on
- MCP2518 add-on CAN-FD
- MCP2518 add-on CAN (classic mode)

PR #409, opened 2024-08-14
BYD Atto 3: Change CAN id and content for contactor

dalathegreat:
### What
The contactor closing is not working on the BYD Atto 3

### Why
Most likely a typo in the CAN ID (411 / 441)

### How
Fixed ID, and changed content to from an AC charging session


dalathegreat:
Tested on a BYD Atto 3 pack, verified that contactor closing now works! :partying_face: Merging this!

PR #381, opened 2024-07-15
Feature: BMW i3 cellvoltages + small fixes

dalathegreat:
### What
This PR adds cellvoltage monitoring for the BMW i3. It also fixes some minor issues that Havrla noticed while testing the code, like using wrong SOC% value and wrong capacity value.

### Why
Safer operation of the battery. SOC% is now using the display value, since the raw value can sometimes get "stuck" during recalibration. kWh remaining is also rewritten.

### How
Reverse engineering how the cellvalue reading worked, in cooperation with Havrla :raised_hands:

PR #417, opened 2024-08-20
Bugfix: BMW i3 temperature and SOC improvement

dalathegreat:
### What
This PR fixes the min/max temperatures being swapped on the i3, and also changes so we use the DisplaySOC% instead of the raw SOC%.

### Why
The DisplaySOC% is better to use, since it can reach 100%. The raw SOC% is not always reaching the extremes (0 / 100%), so it is safer to use the Display value.

### How
This PR takes the smaller fixes from PR #381 , since the cellvoltages part is not done yet


dalathegreat:
Merging this change, tested OK on the Discord!

Issue #418, opened 2024-08-20
BMW i3: Some packs have other CRC check


dalathegreat:
### What
Some BMW i3 batteries do not pass the CRC check on incoming messages. This causes yellow light to be permanently on, and warning event.

Could be some versions of the BMW i3 SME that dont calculate CRC in the same way as others. We should add a check, if the first message is broken, skip even trying to use CRC

ababians:
This situation appears in my case and regarding a 60Ah battery from November 2014

dalathegreat:
Closing this, solved with #419 , next release included in is 7.1.0

PR #419, opened 2024-08-21
BMW i3: Add skip for CRC incase it fails initially

dalathegreat:
### What
This PR fixes the bug in #418 

### Why
Some i3 packs have constant yellow light on due to failing CRC

### How
We skip CRC entirely incase it fails on the first try


ababians:
After lilygo board was compiled with his code yellow warning has disappeared.  Green light is on. No can warning messages in event webserver page. So iti is solved.

dalathegreat:
Merging this, tested OK!

PR #424, opened 2024-08-25
Update CAN pins from hard coded to using HAL as intended

StarkJohan:
The CAN_TX_PIN and CAN_RX_PIN configuration in the HAL is not currently used. They are both hard coded in Software.ino and this currently overrides the HAL.
This PR fixes this issue.


dalathegreat (approved):
Nice, forgot this when i refactored the CAN handling!

Issue #432, opened 2024-09-03
Tesla batteries: Possible to enter dangerous current values


dalathegreat:
### What
Discord member installed a new Tesla battery, and configured it to work with a 20kW inverter. They raised the MAXCHARGEPOWERALLOWED from 15kW to 20kW in the Tesla.h file. When charging up to 100%, the battery blew the pyrofuse.

Why? Since we do not get the allowed charge value from the Tesla battery (it is 0 all the time), we estimate the allowed charge based on the MAXCHARGEPOWERALLOWED value. When the battery is between 90-99%, we ramp down from the value towards 0. This becomes problematic, if a user sets 50kW allowed, the battery will take extreme amounts of charge at high SOC%, leading to the pyrofuse blowing up.

We need to prevent the battery from being charged with high power when close to 100%.

PR #433, opened 2024-09-03
Tesla S3XY: Change how power ramping is performed

dalathegreat:
### What
This PR fixes the issue noticed in #432 

### Why
For safer battery operation

### How
We now do two things differently.
- Power ramping is no longer tied to normal charge speed define. This avoids battery possibly being charged with 50kW while nearing 100% SOC.
- The Tesla.h file now has a clear distinction which defines should be edited by users
image

Note, for future implementations, we should:

- Try sending the entire vehicle CAN bus towards the battery
- This should at the same time fix that the battery sends 0A allowed Charge/Discharge, so we can stop using estimated value
- Once allowed A is available, use this to perform balancing / graceful stopping of charge at 100% soc


dalathegreat:
Testing completed on one battery, merging this!

PR #434, opened 2024-09-03
Update JAGUAR-IPACE-BATTERY.cpp

gb993:
Adding known Network Management (NM) Frame details to the keep alive list

### DIFF
--- a/Software/src/battery/JAGUAR-IPACE-BATTERY.cpp
+++ b/Software/src/battery/JAGUAR-IPACE-BATTERY.cpp
@@ -42,11 +42,11 @@ static bool HVILBattIsolationError = false;
 static bool HVIsolationTestStatus = false;
 
 /* TODO: Actually use a proper keepalive message */
-CAN_frame ipace_keep_alive = {.FD = false,
+CAN_frame Simulated GWM NM = {.FD = false,
                               .ext_ID = false,
                               .DLC = 8,
-                              .ID = 0x063,
-                              .data = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}};
+                              .ID = 0x59E,
+                              .data = {0x9E, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}};
 CAN_frame ipace_7e4 = {.FD = false,
                        .ext_ID = false,
                        .DLC = 8,


dalathegreat:
Hi @gb993 , did you also intend to send this message? In the send_can_battery() function you can see where the old ipace_keep_alive message was.

gb993:
Hey so the message I added will keep the battery awake - I renmamed it from keep awake message to make it clear what it was meant to do - this may not be correct within the context of the rest of the code and I now realise will probably be missed by a later function call.

dalathegreat:
Do you have a Jag battery to test with? Since you changed both the ID of the message, and the content, it would be good to test. But Jag battery support is not done yet, so not possible to break anything either way :sweat_smile:

gb993:
I dont have one to test it on I am afraid, however I can say with certainty that it is the correct message and content (subject to the age of the vehicle) In earlier vehicles (pre-21 Model year) there are some subtle changes that would need to be made.

dalathegreat:
Okay! We can merge it, but hope someone can continue testing this soon. The initial attempts on the Discord server has died down, only one user has this battery, and with limited access to it.

Software/src/battery/JAGUAR-IPACE-BATTERY.cpp
```diff
@@ -42,11 +42,11 @@ static bool HVILBattIsolationError = false;
 static bool HVIsolationTestStatus = false;
 
 /* TODO: Actually use a proper keepalive message */
-CAN_frame ipace_keep_alive = {.FD = false,
+CAN_frame Simulated GWM NM = {.FD = false,
```
obbardc, line 45: Change this to:

```suggestion
CAN_frame simulated_GWM_NM = {.FD = false,
```

And also change the use of ipace_keep_alive in later lines.

You can also remove the Todo above.

obbardc:
@gb993 Can you help us with understanding what canbus messages need to be sent to request turning the contactors on? And also their status? I have a DBC file but I don't think it's correct for this model.

obbardc:
@gb993 I tested this locally. Sending the message every 100ms keeps the BECM alive, but after a while stops ACKing CANbus messages, goes into sleep mode (stops drawing current from PSU) then immediately wakes up from sleep mode after the next message.

The BECM I am using, is marked as `JLR BEV BECM LG Chem JLR X590 VB BECM Ver1.0 2018.01.05`. Not sure what FW revision.

Sending 0x7e4 testerpresent every 1s seems to also do the same;

gb993:
Could you try sending every 1000ms and see if the behaviour changes please?

gb993:
> @gb993 Can you help us with understanding what canbus messages need to be sent to request turning the contactors on? And also their status? I have a DBC file but I don't think it's correct for this model.

I shall do my best

obbardc:
> Could you try sending the NM message every 1000ms and see if the behaviour changes please?

Same thing - BECM stays alive for 5s, then appears to go to sleep for a few ms (current draw is 0...) until the next message is sent :(

gb993:
Ok I have a theory, any chance you can send me a log please. I'm interested to see which messages you are seeing in the 0x500 range please

obbardc:
Let's communicate in private via mail for now.

obbardc:
This can be closed in favour of #437 :-)
Thanks for your help @gb993 !

PR #435, opened 2024-09-04
Publish Battery Emulator Events to MQTT

amarofarinha:
### What
This PR implements an initial version that now exposes Battery Emulator's internal events via MQTT

### Why
Like the rest of the information already exposed via MQTT, events can be of great importance when exposed as they will allow, among other actions (automations), the possibility of sending alerts by subscribers

### How
The standards already implemented in the MQTT module were followed and a new sensor with properties was created that will be sent whenever there are new events. Autodiscovery was implemented like other existing topics.
events_2
events_1


dalathegreat (approved):
LGTM! :+1:

PR #438, opened 2024-09-05
Feature: Show CONTACTOR_CONTROL status in the WebUI

obbardc:
When CONTACTOR_CONTROL is used, the battery-emulator controls
its own contactors. There is nowhere on the WebUI to show if
the contactors are engaged or not, only an error event is shown
if the contactors are disengaged.
    
Add an entry in the WebUI to show the CONTACTOR_CONTROL status.

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -645,6 +645,16 @@ String processor(const String& var) {
     else
       content += "<h4 style='color: red;'>Pause status: " + String(get_emulator_pause_status().c_str()) + " </h4>";
 
+#ifdef CONTACTOR_CONTROL
+    content += "<h4>Contactors controlled by Battery-Emulator: ";
+    if (datalayer.system.status.contactor_control_closed) {
+      content += "<span style='color: green;'>ON</span>";
+    } else {
+      content += "<span style='color: red;'>OFF</span>";
+    }
+    content += "</h4>";
+#endif
+
     // Close the block
     content += "</div>";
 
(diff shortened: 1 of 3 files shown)


nagyrobi:
A red x would create a false impression that it's faulty. I'd suggest simply use `ON` and `OFF` for contactor.

obbardc:
> A red x would create a false impression that it's faulty. I'd suggest simply use `ON` and `OFF` for contactor.

Agreed.


Also this PR may need reworking for when the multi-battery setup is enabled.

dalathegreat:
This needs to be set for every single battery type if we add this. Otherwise users will worry when they see X or OFF in the UI. This is a quite massive undertaking to visualize this properly for all our battery types!

obbardc:
> This needs to be set for every single battery type if we add this. Otherwise users will worry when they see X or OFF in the UI. This is a quite massive undertaking to visualize this properly for all our battery types!

This has no relation to battery type, it is just to show the state of the contactors which the battery-emulator is controling :-)
But I see that it should be guarded and contactor state only shown in the webui if that feature is enabled :-)

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -644,6 +644,14 @@ String processor(const String& var) {
       content += "<span style='color: red;'>&#10005;</span></h4>";
     }
 
+    content += "<h4>Contactor status: ";
```
dalathegreat, line 647: Can you wrap this with an #ifdef CONTACTOR_CONTROL?

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -644,6 +644,14 @@ String processor(const String& var) {
...
+    if (datalayer.system.status.contactor_closed) {
+      content += "<span>&#10003;</span>";
```
dalathegreat, line 649: Let's change this to either ON or OFF

dalathegreat (approved):

PR #440, opened 2024-09-06
Ignore built binary files

obbardc:
Ignore .bin files.


dalathegreat (approved):
Approved, but please try to adopt the What/Why/How to make it clear what the PR achieves!

PR #445, opened 2024-09-08
Tesla: Add support for 108S Model X +refactor

dalathegreat:
### What
This PR adds support for the 108s Model X battery.

### Why
In the current code if you tried to use this battery, you would get overvoltage and cellvoltage errors.

### How
The PR fixes this by refactoring how Tesla packs are defined when compiling. You can now specify which Tesla battery you are using, previously it was all Model 3. Now there are specific rows for each battery:

image


dalathegreat:
Tested successfully by Discord member! Merging this :sunglasses:

PR #446, opened 2024-09-08
Web Server Authentication

amarofarinha:
### What
This PR implements an initial version for Web Server access authentication

### Why
Being a critical system where some important emulator settings can be changed, there are advantages in protecting it.

### How
An implementation example using the libraries currently in use in the emulator was followed from https://randomnerdtutorials.com/esp32-esp8266-web-server-http-authentication/. This is not very strong protection but enough to prevent non-malicious use from attempted use on the internal network where it runs. The credentials requirement option is defined in User Settings in WEBSERVER_AUTH_REQUIRED and the default value is disabled, maintaining compatibility with current use without authentication. The authentication used here is Digest MD5 Authentication.


dalathegreat (approved):
That was fast! Feels like we just discussed this as an idea yesterday :sweat_smile: Nice work!

PR #453, opened 2024-09-10
Added additional sensors to the MQTT info topic.

amarofarinha:
### What
Added additional sensors to the MQTT info topic.

### Why
To provide more detailed system information.

### How
Included Total Capacity, Remaining Capacity, Max Discharge Power, Max Charge Power, and BMS Status in the MQTT publications to info Topic.

### DIFF
--- a/Software/src/devboard/utils/types.cpp
+++ b/Software/src/devboard/utils/types.cpp
@@ -2,13 +2,20 @@
 
 // Function to get string representation of bms_status_enum
 std::string getBMSStatus(bms_status_enum status) {
-    switch (status) {
-        case STANDBY: return "STANDBY";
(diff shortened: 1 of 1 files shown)


Software/src/devboard/utils/types.cpp
```diff
@@ -0,0 +1,21 @@
+#include "types.h"
+
+// Function to get string representation of bms_status_enum
+std::string getBMSStatus(bms_status_enum status) {
+  switch (status) {
```
dalathegreat, line 5: Is there any way to get the enums pulled from the .h file? So we dont need to update two places in the future. Approved anyhow! :+1:

dalathegreat (approved):

Issue #355, opened 2024-06-21
Sort order for events in 6.2.1


marknisk:
I noticed that the sort order for events is not as the column header says (last event). I'm running v 6.2.1

![image](https://github.com/dalathegreat/Battery-Emulator/assets/66523164/b6197caf-66f7-4438-8407-d8a52eed5984)

marknisk:
Personally I would also prefer SI style for date and time

PR #460, opened 2024-09-16
display events in reverse chronological order

amarofarinha:
### WHAT
This PR changes the events page on the web server to display events in reverse chronological order, showing the most recent event first instead of the internal definition order.

### WHY
Currently, events are displayed based on their internal definition order, which doesn't always reflect the actual sequence in which they occurred. This makes it difficult for users to quickly find the most recent events. Reordering the events based on when they occurred improves the user experience by allowing easier access to the latest activity.
This was also requested by users on https://github.com/dalathegreat/Battery-Emulator/issues/355

### HOW
The events were reordered by timestamp in descending order prior to rendering on the page.
The implementation has been tested to ensure the correct ordering and proper functionality of the events display.

### DIFF
--- a/Software/src/devboard/webserver/events_html.cpp
+++ b/Software/src/devboard/webserver/events_html.cpp
@@ -1,6 +1,4 @@
 #include "events_html.h"
-#include <Arduino.h>
-#include "../utils/events.h"
 
 const char EVENTS_HTML_START[] = R"=====(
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):

dalathegreat:
LGTM, great change!

In the future, having Up/Down buttons next to each category would be awesome, and user could decide which way to sort the event list.

PR #463, opened 2024-09-17
BYD Atto 3: Add way to use SOC% sent by battery

dalathegreat:
### What
This PR adds a configurable option to use SOC% sent by battery

### Why
When Atto 3 packs are crashed hard, they lock themselves and stop updating SOC%. Our previous implementation instead estimates SOC% via battery voltage. This solution is bad, since putting load on the battery will make the reading swing, this making it impossible to use the whole battery.

### How
We add a configurable option to the `BYD-ATTO-3-BATTERY.h` file. If you are certain that your battery is not crashed hard, you can remove the line `USE_ESTIMATED_SOC`, and the code will then instead use the periodically transmitted SOC value. :information_source: If your battery IS locked and you remove this line, SOC% will stay stuck at the value it had when the battery was crashed and airbags popped.

image

The "More battery info" page also now shows what is used. It will either show Estimated from voltage OR Measured by BMS , based on the configuration

image


dalathegreat:
We have three candidates, someone needs to test which one works best and make the integration accordingly;
- Polled SOC
- Highprecision SOC
- Lowprecision SOC

GraphicObession:
> We have three candidates, someone needs to test which one works best and make the integration accordingly;
> 
> * Polled SOC
> * Highprecision SOC
> * Lowprecision SOC

Hi Dala, I am happy to test this if it is still needed?

dalathegreat:
Those who tested this on the Discord ran into the following issues:

_The problem I now seem to be encountering is that the span of the SOC sent by battery is drifting. It used to be 100% when I had at about 96 scaled but know hitting the same cell voltages at 80% and then the deviation starts growing. The cell information does not work for me in HA so can not trend. I'm going to go back to estimate to try to discharge the battery lower. If I was to try and reconnect the donar car to the battery what is the best way to monitor the canbus. I have a spare lilygo but no special software. Ran the battery down using SOC inverter stopped at 5%. Changed to Estimated and was able to run battery down to 16% estimated approx 3200mv. No warnings unable to go lower. As soon as it started charging the Hi started counting up. It has reset the battery range from here so I was able to get it back up to 100% with  the scale in the software set to 1.3 . The hi got back up to 760. I think there is still room to take the battery lower and the Hi count should be higher at the top.
I still had a little left at the top did not go all the way up to 3400mV_

Basically it is behaving a bit strange when using the real SOC sent by battery. Not sure if we want to merge this or not.

GraphicObession:
I think it would be good to merge version as I would be keen to try it as the current version does not seem to give any real view of the SOC.

dalathegreat:
@GraphicObession I suggest you try this PR instead, and report back if the various SOC% moves: https://github.com/dalathegreat/Battery-Emulator/pull/561

dalathegreat:
This PR is IMO ready for merging, if it works well :)

PR #464, opened 2024-09-17
Bugfix: RJXZS BMS, Add ramping charge power + settings

dalathegreat:
### What
This PR adds configurable settings for how much charge/discharge is allowed to the battery. It also adds smooth ramping at the top SOC% area, and a configurable float charge setting.

### Why
Custom batteries require more tweaking of allowed power levels. The built in BMS does not know what battery you have constructed. You need to define the power allowed.

### How
You can define your battery limits in `RJXZS-BMS.h` file

image

### DIFF
--- a/Software/src/battery/RJXZS-BMS.cpp
+++ b/Software/src/battery/RJXZS-BMS.cpp
@@ -89,6 +89,9 @@ void update_values_battery() {
     clear_event(EVENT_SOC_UNAVAILABLE);
   }
 
+  datalayer.battery.status.remaining_capacity_Wh = static_cast<uint32_t>(
+      (static_cast<double>(datalayer.battery.status.real_soc) / 10000) * datalayer.battery.info.total_capacity_Wh);
+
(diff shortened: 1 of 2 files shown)


Software/src/battery/RJXZS-BMS.cpp
```diff
@@ -98,9 +98,20 @@ void update_values_battery() {
   datalayer.battery.status.active_power_W =  //Power in watts, Negative = charging batt
       ((datalayer.battery.status.voltage_dV * datalayer.battery.status.current_dA) / 100);
 
-  datalayer.battery.status.max_charge_power_W = (protective_current * total_voltage) / 10;
```
puffnfresh, line 101: It looks like `protective_current` is now unused. Should that still be parsed from the BMS if it's not needed?

dalathegreat: We sample all the BMS is sending (In the future we will create individual pages on the webserver with more info). So lets keep it!

dalathegreat:
Merging, successfully tested by Paz on the Discord server!

PR #465, opened 2024-09-17
Feature: Contactor closing for eGMP

dalathegreat:
### What
This PR fixes contactor closing for the EGMP platform (Kia EV6 etc.)

### How
Loop of 63 messages added. Other fixes related to CAN-FD
- CANFD buffer overflow event added, incase message gets lost
- Buffer size increased from 16 -> 20 for CANFD acan library


ilia92 (approved):
As owner of EV6 battery, I tested it and it closes the contactors.

Camelrider79:
Confirmed

Gerwaz3:
Hello, 
I connected the KIA EV6 battery to the Sofar 20KTL inverter, RGB Led is green, no errors in events, and the contactor turns on after starting the program for a few seconds and turns off. The battery is discharged, SOC=2.5%, SOH=100%, U=614V. The battery is connected according to the diagram - jumpers and 12K resistors. The FDCAN INIT ERROR error is shown in the events, but only once after switching on the power supply, after the software restart it is gone. The battery parameters read are consistent with the measurements, the inverter also shows SOC and SOH from the battery, it even turns on charging, but when the contactor releases it stops charging. What could be the reason for the contactor disconnecting? 
Best regards

PR #473, opened 2024-09-20
Improve MQTT Connection Resilience and Add Event Tracking for Diagnos…

amarofarinha:
### WHAT
Enhanced MQTT client reconnection logic to make it more resilient to Wi-Fi disconnections and Home Assistant/Mosquitto broker restarts.
Introduced internal events to register MQTT connect and disconnect states for better diagnostic capabilities.
Reduced processor load by introducing a timer for MQTT checks

### WHY
Users reported issues with MQTT client failing to reconnect after Wi-Fi disconnections or when the MQTT broker (Mosquitto/Home Assistant) restarts.
Adding event tracking for connect and disconnect events helps with diagnosing connection issues more effectively in the future.

### HOW
Implemented improved reconnection logic that checks Wi-Fi status before attempting to reconnect to the MQTT broker, preventing unnecessary retries during Wi-Fi outages.
Adjusted the retry mechanism for reconnections, ensuring the MQTT client properly resets after multiple failed attempts.
Added internal event registration to log connect and disconnect states, aiding future debugging efforts.

### DIFF
--- a/Software/src/devboard/wifi/wifi.cpp
+++ b/Software/src/devboard/wifi/wifi.cpp
@@ -24,6 +24,7 @@ static bool hasConnectedBefore = false;
 static uint16_t reconnectAttempts = 0;  // Counter for reconnect attempts
 static uint16_t current_full_reconnect_interval = INIT_WIFI_FULL_RECONNECT_INTERVAL;
 static uint16_t current_check_interval = WIFI_CHECK_INTERVAL;
+static bool connected_once = false;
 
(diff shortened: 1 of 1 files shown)


bubble07 (approved):
Tested ok on Lili with MQTT and Web server active 
Screenshot_2024-09-20-20-00-40-09_40deb401b9ffe8e1df2f1cc5ba480b12

dalathegreat (approved):

PR #474, opened 2024-09-20
RJXZS: Map errors to Battery Emulator events

puffnfresh:
### What

RJXZS errors mostly have equivalents in Battery Emulator. This PR raises an event for each possible error.

### Why

Logging the error events in the Battery Emulator UI is useful for having one place to identify problems. It could also allow Battery Emulator to react to the errors in some cases.

### How

Raises an event depending on the byte response.


Software/src/battery/RJXZS-BMS.cpp
```diff
@@ -509,6 +509,38 @@ void receive_can_battery(CAN_frame rx_frame) {
         use_capacity_to_automatically_reset = (rx_frame.data.u8[3] << 8) | rx_frame.data.u8[4];
```
puffnfresh, line 509: Any idea on what `use_capacity_to_automatically_reset` could mean?

dalathegreat:
Nice, we could also create new events for the warnings/errors that dont match up with existing events.

The pre-commit check fails;
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.

dalathegreat (approved):

PR #476, opened 2024-09-21
bugfix - pause state update was in wrong place.

amarofarinha:
### What
Pause state update (emulator_pause_state_send_CAN_battery) was in wrong place.

### Why
it was just working if FUNCTION_TIME_MEASUREMENT was defined.

### How
moved the code outside FUNCTION_TIME_MEASUREMENT  region


dalathegreat (approved):
Approved. We really should do a refactor and place the #endif //REASON on every single codeblock :sweat_smile:

PR #480, opened 2024-09-23
Solax: Add inverter missing from CAN event, Solax

dalathegreat:
### What
This PR implements a new event, incase the Solax inverter does not send any CAN data, we raise an event.

### Why
To guide the user towards troubleshooting CAN wiring

### How
This new event is raised if Solax CAN communication is gone for too long:
image


Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -88,6 +89,12 @@ void update_values_can_inverter() {  //This function maps all the values fetched
   if (millis() - LastFrameTime >= SolaxTimeout) {
     datalayer.system.status.inverter_allows_contactor_closing = false;
     STATE = BATTERY_ANNOUNCE;
+    inverter_missing_on_can++;
+    if (inverter_missing_on_can > CAN_STILL_ALIVE) {
+      set_event(EVENT_CAN_INVERTER_MISSING, 0);
```
obbardc, line 94: should this event clear `datalayer.system.status.inverter_allows_contactor_closing` ?

dalathegreat: It is cleared 4 rows before the event is fired. No need to clear it really, if we go into a ERROR situation we open contactors. This event is setup as a WARNING, but if it is stable and not triggered accidentally, we could bump up the severity to ERROR.

obbardc: No worries, was just a thought without any real checking of the code ;-)

dalathegreat: Always appreciate extra pairs of eyes :grinning: :eyes:

Software/src/devboard/utils/events.cpp
```diff
@@ -258,6 +259,8 @@ const char* get_event_message_string(EVENTS_ENUM_TYPE event) {
       return "ERROR: High amount of corrupted CAN messages detected. Check CAN wire shielding!";
     case EVENT_CAN_TX_FAILURE:
       return "ERROR: CAN messages failed to transmit, or no one on the bus to ACK the message!";
+    case EVENT_CAN_INVERTER_MISSING:
```
obbardc, line 262: Not really a question for this PR (perhaps we could open an issue) ; but how are events handled when serial transmitter/receiver is used ? Will this event just fire on the inverter side ?

dalathegreat: @obbardc Best to open issue. Would be good to sync all events between the two lilygos. I did not write the implementation for the double lilyGo support!

obbardc: Sure!

Issue #472, opened 2024-09-20
PWM_CONTACTOR_CONTROL does not seem to work on lilygo


obbardc:
`CONTACTOR_CONTROL` seems to work fine; but when enabling `PWM_CONTACTOR_CONTROL` to save some power, the pins just stay low (even without any driver connected) and do not even go high. Reverting to just `CONTACTOR_CONTROL` enabled works fine though.

I will diagnose this once I get another lilygo, but opening an issue in case anyone else has any thoughts.

obbardc:
I have debugged this and fixed it in #481

PR #481, opened 2024-09-24
Fix: PWM_CONTACTOR_CONTROL now actually engages contactors

obbardc:
### What
This PR fixes the broken PWM_CONTACTOR_CONTROL, as reported in https://github.com/dalathegreat/Battery-Emulator/issues/472 as well some small quality-of-life improvements.

When PWM_CONTACTOR_CONTROL was set the contactor coil pins for positive & negative never actually went high

### Why
It's broken.

### How
Root cause was due to a wrong argument passed to `ledcWrite`: the existing code used the PWM channel; while `ledcWrite` actually expects the PWM pin to be passed.


obbardc:
I tested `CONTACTOR_CONTROL` enabled with, `PWM_CONTACTOR_CONTROL` disabled as well as `PWM_CONTACTOR_CONTROL` enabled separately.

dalathegreat (approved):

PR #487, opened 2024-09-26
Feature: Clear events via webpage

dalathegreat:
### What
This PR implements manual event clearing
Bonus fixes:
- Pause button centered properly
- Pylonbattery text on webserver added

### Why
User requested feature

### How
A button is added to the Events webpage. When clicked it asks the user if they want to clear all errors.

image

image

image

### DIFF
--- a/Software/src/devboard/utils/events.cpp
+++ b/Software/src/devboard/utils/events.cpp
@@ -240,12 +240,15 @@ void reset_all_events() {
   events.nof_logged_events = 0;
   for (uint16_t i = 0; i < EVENT_NOF_EVENTS; i++) {
     events.entries[i].data = 0;
+    events.entries[i].state = EVENT_STATE_INACTIVE;
     events.entries[i].timestamp = 0;
     events.entries[i].millisrolloverCount = 0;
     events.entries[i].occurences = 0;
     events.entries[i].log = true;
     events.entries[i].MQTTpublished = false;  // Not published by default
   }
+  events.level = EVENT_LEVEL_INFO;
+  update_bms_status();
 }
 
 void set_event_MQTTpublished(EVENTS_ENUM_TYPE event) {
--- a/Software/src/devboard/utils/events.cpp
+++ b/Software/src/devboard/utils/events.cpp
@@ -236,6 +236,18 @@ void clear_event(EVENTS_ENUM_TYPE event) {
   }
 }
 
+void reset_all_events() {
+  events.nof_logged_events = 0;
+  for (uint16_t i = 0; i < EVENT_NOF_EVENTS; i++) {
+    events.entries[i].data = 0;
+    events.entries[i].timestamp = 0;
+    events.entries[i].millisrolloverCount = 0;
+    events.entries[i].occurences = 0;
+    events.entries[i].log = true;
+    events.entries[i].MQTTpublished = false;  // Not published by default
+  }
+}
+
 void set_event_MQTTpublished(EVENTS_ENUM_TYPE event) {
   events.entries[event].MQTTpublished = true;
 }
--- a/Software/src/devboard/webserver/events_html.cpp
+++ b/Software/src/devboard/webserver/events_html.cpp
@@ -5,10 +5,9 @@ const char EVENTS_HTML_START[] = R"=====(
 )=====";
 const char EVENTS_HTML_END[] = R"=====(
 </div></div>
-<button onclick='clear()'>Clear all events</button>
-<button onclick='home()'>Back to main page</button>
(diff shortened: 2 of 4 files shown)


Software/src/devboard/utils/events.cpp
```diff
@@ -236,6 +236,21 @@ void clear_event(EVENTS_ENUM_TYPE event) {
   }
 }
 
+void reset_all_events() {
+  events.nof_logged_events = 0;
+  for (uint16_t i = 0; i < EVENT_NOF_EVENTS; i++) {
+    events.entries[i].data = 0;
```
obbardc, line 242: would it not be preferable to call `clear_event` ?

Software/src/devboard/webserver/events_html.cpp
```diff
@@ -5,9 +5,9 @@ const char EVENTS_HTML_START[] = R"=====(
...
 </div></div>
-<button onclick='home()'>Back to main page</button>
-<style>.event:nth-child(even){background-color:#455a64}.event:nth-child(odd){background-color:#394b52}</style>
-<script>function showEvent(){document.querySelectorAll(".event").forEach(function(e){var n=e.querySelector(".sec-ago");n&&(n.innerText=new Date(Date.now()-((+n.innerText.split(';')[0])*4294967296+ +n.innerText.split(';')[1])).toLocaleString());})}function home(){window.location.href="/"}window.onload=function(){showEvent()}</script>
+<button onclick="askClear()">Clear all events</button>
+<button onclick="home()">Back to main page</button>
+<style>.event:nth-child(even){background-color:#455a64}.event:nth-child(odd){background-color:#394b52}</style><script>function showEvent(){document.querySelectorAll(".event").forEach(function(e){var n=e.querySelector(".sec-ago");n&&(n.innerText=new Date(Date.now()-(4294967296*+n.innerText.split(";")[0]+ +n.innerText.split(";")[1])).toLocaleString())})}function askClear(){window.confirm("Are you sure you want to clear all events?")&&(window.location.href="/clearevents")}function home(){window.location.href="/"}window.onload=function(){showEvent()}</script>
```
obbardc, line 10: sidenote: Having this code minimised seems.. wrong (e.g someone could introduce an exploit similar to recent xz exploit or bitcoin mining script. Longshot, but possible.). Also minimising reduces readability. Perhaps it could be minimised as part of the build process ? Perhaps one to open as an issue for future rework.

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -499,6 +511,9 @@ String processor(const String& var) {
 #ifdef NISSAN_LEAF_BATTERY
     content += "Nissan LEAF";
 #endif  // NISSAN_LEAF_BATTERY
+#ifdef PYLON_BATTERY
```
obbardc, line 514: Doesn't seem like it should be part of this PR ?

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -810,7 +825,7 @@ String processor(const String& var) {
       content += "<button onclick='PauseBattery(false)'>Resume Battery</button>";
     else
       content += "<button onclick='PauseBattery(true)'>Pause Battery</button>";
-
+    content += " ";
```
obbardc, line 828: Separate commit required ?

Software/src/devboard/utils/events.cpp
dalathegreat: When running clear event you have to specify which event to clear. Having a global reset all makes it easier!

Software/src/devboard/webserver/events_html.cpp
dalathegreat: Agreed. We started to minimise code when we had performance issues. This was before we switched to multicore use, running webserver on the second core. We could actually revert all minification now, since we dont have the performance issue anymore.

Software/src/devboard/webserver/webserver.cpp
dalathegreat: I was sneaking this in so I wouldn't forget it :sweat_smile:

Software/src/devboard/webserver/webserver.cpp
dalathegreat: This was to get the Pause button centered. Noticed it was touching the nearby button while working on the new button. I can add it to the PR desc

Software/src/devboard/utils/events.cpp
obbardc: Sorry, I thought you duplicated the logic in `clear_event` :-)

obbardc: What I really meant was, it may be preferable to have a separate function (also called by clear_event) which clears the struct directly.

PR #489, opened 2024-09-27
Pylon Battery: Add voltage limits, remove faulty safety

dalathegreat:
### What
This PR implements a configurable min/max voltage for Pylon batteries. It also omits the reasonable SOC check on pylon packs, since it is very unreliable here due to being voltage based.

### Why
Since we do not know the voltage limits of the pylon battery the user intends to use, it is better if they specify this themselves.

### How
This is set at the top of the PYLON-BATTERY.cpp file


dalathegreat:
Tested successfully by a person on Discord :partying_face: 

image

PR #497, opened 2024-09-28
Fix: Disable Chained CAN Sending During OTA Updates

amarofarinha:
### What
This change disables certain CAN sending operations that are chained within the receiving methods during an OTA (Over-the-Air) update. Previously, only some sending operations were blocked, leading to incomplete coverage of CAN communication control during the update process.

### Why
The prior implementation did not fully prevent CAN communication during OTA, particularly in cases where sending operations were triggered as part of receiving routines. This oversight allowed some chained communications to continue, potentially leading to incomplete or inconsistent behavior during OTA updates. This update corrects that behavior by ensuring these specific sending operations are disabled, thereby preventing any unintended communication during the OTA process.

### How
Logic has been added to disable all CAN sending operations in the transmit_can method. The system now ensures that any sending chained to receiving actions is fully paused until the update process completes.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -588,8 +588,9 @@ void receive_can_native() {  // This section checks if we have a complete CAN me
 }
 
 void send_can() {
-  if (!can_send_CAN)
+  if (!allowed_to_send_CAN) {
     return;
+  }
 
   send_can_battery();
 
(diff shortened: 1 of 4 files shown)


dalathegreat (approved):
LGTM! But could we rename `can_send_CAN` to `allowed_to_send_CAN`? And put the return inside {} to match the rest of the codestyle?

PR #503, opened 2024-09-29
Bugfix iMIEV: Make cellvolt check work with double

dalathegreat:
### What
This PR fixes a bug that caused cellvoltage min/max to not appear in webserver for iMIEV

### Why
The check was assuming integer values, but we had double datatype

### How
Fixed the if check


dalathegreat:
Tested OK on Discord server

PR #504, opened 2024-09-29
Bugfix iMIEV: Set amount of cellvoltages

dalathegreat:
### What
This PR fixes a bug on imiev, all cellvoltages were missing on webserver cellmonitor page

### Why
Better monitoring of battery

### How
We set the cell amount, this was missing from the code


dalathegreat:
Tested successfully, merging!

PR #506, opened 2024-10-02
Add Support for External Button and Web-Based Equipment Stop Functionality

amarofarinha:
### What
This PR implements both an external equipment stop button and a web-based emergency stop, allowing for full system shutdown in critical situations. Both options halt charge and discharge operations, open battery contactors (if controlled by the battery), and stop all CAN communications.

### Why
The idea for this enhancement originated from @StarkJohan, who provided valuable feedback and insights throughout our discussions. The primary motivation for adding these features is to introduce additional safety mechanisms to the emulator, ensuring that emergency shutdowns can be triggered swiftly via either a physical button or the web interface. These options enhance the safety and responsiveness of the system in case of a critical failure or emergency.

This PR strengthens the emulator's emergency handling capabilities by introducing redundant, user-friendly methods to stop operations in emergency scenarios.

### How
Added support for an external equipment stop button connected to a GPIO pin (EQUIPMENT_STOP_PIN).
Integrated web-based emergency stop control to trigger the same shutdown functionality.
Introduced two modes for button functionality:

- TOGGLE_SWITCH: The button directly reflects its physical state (state ON/OFF). 
- PERSISTENT_ACTIVATION_SWITCH: Implements a short/long press mechanism for activation and reset. The state is persisted between reboots.

Activation/deactivation via the web interface is possible allowing the system to resume.

when in EQUIPMENT STOP:

- Halts all charging and discharging operations.
- Stops all CAN communications.
- Opens battery contactors (if controlled by the emulator).

On boot, if an emergency button is configured, its state will take precedence for the emulator. This functionality can be used with both physical buttons and the web server. If no physical button is defined, the state set via the web server will persist across reboots. I believe this approach covers most of your suggestions, although it’s not possible to accommodate all opinions. I consider this a solid solution, which can always be improved or modified in the future.

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -870,14 +870,14 @@ String processor(const String& var) {
           " onclick=\""
           "if(confirm('This action will open contactors on the battery and stop all CAN communications. Are you "
           "sure?')) { estop(true); }\""
-          ">EQUIPMENT STOP</button><br/>";
+          ">Open Contactors</button><br/>";
     else
       content +=
           "<br/><br/><button style=\"background:green;color:white;cursor:pointer;\""
           "20px;font-size:16px;font-weight:bold;cursor:pointer;border-radius:5px; margin:10px;"
           " onclick=\""
           "if(confirm('This action will restore the battery state. Are you sure?')) { estop(false); }\""
-          ">RESUME FROM EQUIPMENT STOP</button><br/>";
+          ">Close Contactors</button><br/>";
 
     content += "<script>";
     content += "function OTA() { window.location.href = '/update'; }";


nagyrobi:
Long press is not necessarily a good idea.

For this task, an emergency stop button is appropriate, which, when pressed, stays in pressed-in state until being released by a different movement (like rotating it).

With an emergency stop button, you are in stopped/emergency state while it's being pressed, and can only resume to normal when released.

Software/src/devboard/hal/hw_lilygo.h
```diff
@@ -78,6 +81,18 @@
 #endif
 #endif
 
+#ifdef EQUIPMENT_STOP_BUTTON
+#ifdef DUAL_CAN
+#error EQUIPMENT_STOP_BUTTON and DUAL_CAN cannot coexist due to overlapping GPIO pin usage
```
nagyrobi, line 86: No more free pins left on LilyGo?

Software/Software.ino
```diff
@@ -334,10 +349,24 @@ void init_serial() {
 }
 
 void init_stored_settings() {
+  static uint32_t temp = 0;
   settings.begin("batterySettings", false);
 
+  //allways get the emergency stop status
```
obbardc, line 355:
```suggestion
  // Always get the emergency stop status
```

Software/src/devboard/hal/hw_lilygo.h
StarkJohan: No, the lilygo has a limited amount of exposed GPIO.

Software/Software.ino
```diff
@@ -334,10 +349,24 @@ void init_serial() {
...
+  temp = settings.getBool("EMERGENGY_STOP", false);
+  datalayer.system.settings.equipment_stop_active = temp;
```
obbardc, line 357: why not just

```
  datalayer.system.settings.equipment_stop_active = settings.getBool("EMERGENGY_STOP", false);
```
then we can also remove the variable temp ?

Software/Software.ino
```diff
@@ -334,10 +349,24 @@ void init_serial() {
...
+  if (datalayer.system.settings.equipment_stop_active) {
+    set_event(EVENT_EMERGENCY_STOP, 1);
+  }
+
 #ifndef LOAD_SAVED_SETTINGS_ON_BOOT
   settings.clear();  // If this clear function is executed, no settings will be read from storage
+
+  //always save the emergency stop status
+  if (datalayer.system.settings.equipment_stop_active) {
+    settings.putBool("EMERGENGY_STOP", datalayer.system.settings.equipment_stop_active);
```
obbardc, line 367: prefer an enum/define for this to avoid mistyping.

also it is spelt wrong `EMERGENCY`

Software/Software.ino
```diff
@@ -334,10 +349,24 @@ void init_serial() {
 }
 
 void init_stored_settings() {
+  static uint32_t temp = 0;
   settings.begin("batterySettings", false);
 
+  //allways get the emergency stop status
+  temp = settings.getBool("EMERGENGY_STOP", false);
```
obbardc, line 356: spell check `EMERGENCY`

Software/USER_SETTINGS.h
```diff
@@ -65,6 +65,7 @@
 #define MDNSRESPONDER  //Enable this line to enable MDNS, allows battery monitor te be found by .local address. Requires WEBSERVER to be enabled.
 #define LOAD_SAVED_SETTINGS_ON_BOOT  //Enable this line to read settings stored via the webserver on boot (overrides Wifi/battery settings set below)
 //#define FUNCTION_TIME_MEASUREMENT  // Enable this to record execution times and present them in the web UI (WARNING, raises CPU load, do not use for production)
+//#define EQUIPMENT_STOP_BUTTON      // Enable this to have a emergengy/equipment stop button connected to board
```
obbardc, line 68:
```suggestion
//#define EQUIPMENT_STOP_BUTTON      // Enable this to allow an emergency/equipment stop button connected to the Battery-Emulator to disengage the battery
```

Software/src/devboard/hal/hw_stark.h
```diff
@@ -58,6 +58,9 @@ GPIOs on extra header
 #define LED_PIN 4
 #define LED_MAX_BRIGHTNESS 40
 
+// Equipment stop pin
```
obbardc, line 61: Can you call it EMERGENCY throughout ?

Software/src/devboard/safety/safety.cpp
```diff
@@ -196,7 +196,18 @@ void update_machineryprotection() {
 }
 
 //battery pause status begin
-void setBatteryPause(bool pause_battery, bool pause_CAN) {
+void setBatteryPause(bool pause_battery, bool pause_CAN, bool emergency_stop) {
+
+  //fist handle emergency stop / resume
```
obbardc, line 201:
```suggestion
  // First handle emergency stop / resume
```

Software/src/devboard/safety/safety.cpp
```diff
@@ -196,7 +196,18 @@ void update_machineryprotection() {
 }
 
 //battery pause status begin
-void setBatteryPause(bool pause_battery, bool pause_CAN) {
+void setBatteryPause(bool pause_battery, bool pause_CAN, bool emergency_stop) {
+
```
obbardc, line 200: additional line which can go away

Software/src/devboard/safety/safety.cpp
```diff
@@ -245,6 +257,14 @@ void emulator_pause_state_send_CAN_battery() {
   }
 
   allowed_to_send_CAN = (!emulator_pause_CAN_send_ON || emulator_pause_status == NORMAL);
+
```
obbardc, line 260: additional line

Software/src/devboard/utils/events.cpp
```diff
@@ -411,6 +412,8 @@ const char* get_event_message_string(EVENTS_ENUM_TYPE event) {
       return "Info: MQTT connected.";
     case EVENT_MQTT_DISCONNECT:
       return "Info: MQTT disconnected.";
+    case EVENT_EMERGENCY_STOP:
+      return "ERROR: EMERGENCY STOP ACTIVATED!!!";
```
obbardc, line 416: Doesn't follow other errors

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -166,6 +166,23 @@ void init_webserver() {
     }
   });
 
+  // Route for emergency stop/resume
+  server.on("/estop", HTTP_GET, [](AsyncWebServerRequest* request) {
```
obbardc, line 170: can we call the endpoint `/emergencyStop` to match other endpoints ?

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -844,6 +861,21 @@ String processor(const String& var) {
     content += "<button onclick='askReboot()'>Reboot Emulator</button>";
     if (WEBSERVER_AUTH_REQUIRED)
       content += "<button onclick='logout()'>Logout</button>";
+    if (!datalayer.system.settings.equipment_stop_active)
+      content +=
+          "<br/><br/><button style=\"background:red;color:white;cursor:pointer;\""
+          " onclick=\""
+          "if(confirm('This action will open contactors on the battery and stop all CAN communications. Are you "
+          "sure?')) { estop(true); }\""
+          ">EQUIPMENT STOP</button><br/>";
```
obbardc, line 870: use `EMERGENCY`

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -844,6 +861,21 @@ String processor(const String& var) {
...
+    else
+      content +=
+          "<br/><br/><button style=\"background:green;color:white;cursor:pointer;\""
+          "20px;font-size:16px;font-weight:bold;cursor:pointer;border-radius:5px; margin:10px;"
+          " onclick=\""
+          "if(confirm('This action will restore the battery state. Are you sure?')) { estop(false); }\""
+          ">RESUME FROM EQUIPMENT STOP</button><br/>";
```
obbardc, line 877: EMERGENCY

obbardc (changes requested):

Software/src/devboard/webserver/webserver.cpp
StarkJohan: EQUIPMENT STOP! @dalathegreat  :-)

nagyrobi:
> what about having configurable behaviors?

Good!

dalathegreat:
I agree with @nagyrobi , most big red e-stop switches latch once you press them. This is good behaviour, which at the same time prevents the system from accidentally starting operating after a reboot (since the request for e-stop is still active).

I have encountered quite a few e-stop designs over the years, and the majority work this way.

Having us only look at the GPIO state for the E-stop (which **needs** to be a NC circuit to prevent wire break from hindering operation of the switch) also takes out the need to store the e-stop to persistent memory. Since rebooting the emulator with the e-stop engaged will result in it going straight into stopped state.

nagyrobi:
I'm also worried about the [hard exclusion of the other features](https://github.com/dalathegreat/Battery-Emulator/pull/506#discussion_r1783916416) on LilyGo, just because the limited amount of exposed GPIO. Is the SD card used? There are 4 more GPIOs in the SD card socket. I agree they can't be easily accessed, but one could still manage to use one of them together with the other features. The GPIOs should be freely configurable without having to alter the rest of the code too heavily...

nagyrobi:
> Alternatively, we could consider giving the physical button priority over the web interface, meaning that if the button is activated, it wouldn’t be possible to deactivate it via the web interface. What would be the best approach in the schenario of a latched button?

Could be completely two different safety features. One is software stop, and the other just hardware stop. Shouldn't be related to each other. The hardware stop's status should just be shown in the web interface. One should not interfere with each other. Eg. releasing the button shouldn't release the software stop (someone physically at the device may not be aware that somebody else is remotely working and shouldn't mess the state remotely set).

BEMSman:
Shouldn’t the web interface be the final reset that would only provide reset after E-Stop have been reset locally.

StarkJohan:
> I'm also worried about the [hard exclusion of the other features](https://github.com/dalathegreat/Battery-Emulator/pull/506#discussion_r1783916416) on LilyGo, just because the limited amount of exposed GPIO. Is the SD card used? There are 4 more GPIOs in the SD card socket. I agree they can't be easily accessed, but one could still manage to use one of them for this, together with the other features. The GPIOs should be freely configurable without having to alter the rest of the code too heavily...

Most user will not max out the GPIO on the lilygo. Changing the configuration of the pins is very easy in the HAL. 
For those who need more GPIOs there are multiple options. I don't see the limited GPIOs as a reason to not implement excellent safety feartures.

Software/USER_SETTINGS.cpp
```diff
@@ -48,6 +48,13 @@ const char* mqtt_password = "REDACTED";  // Set NULL for no password
 #endif                                   // USE_MQTT
 #endif                                   // WIFI
 
+#ifdef EQUIPMENT_STOP_BUTTON
+// Equipment stop button behavior. Use NC button for safety reasons.
+//TOGGLE_SWITCH                - activated while pressed, deactivated when released, button state is reflected in the emulator state
```
dalathegreat, line 53: We should use simple understandable terms here, my suggestion

//LATCHING_SWITCH       - Normally closed (NC), latching switch. When pressed it activates e-stop
//MOMENTARY_SWITCH - Short press to activate e-stop, long 5s press to deactivate. E-stop is persistent between reboots

However the more I think about it, I really would recommend dropping support for momentary_switch...

StarkJohan: I don't mind the momentary switch support, it's the reset function that need to be very safe and obvious. I don't love the long press but if it's kept, it needs to be rather long.

dalathegreat:
Sorry for joining this review so late, but I'd like to challenge the need to trigger e-stop via the webserver.

To me it seems like we are using the equipment stop terminology wrong in the Webserver. From the Webserver, it makes sense to either "Pause", or "Open Contactors". Can we rename this in the Webserver?

dalathegreat:
I am still also not sure about the benefit of handling a recovery from an "Open contactor" press. To me it makes more sense to latch this, and require a reboot to get out of it.

dalathegreat:
Same goes with the storing to persistent memory. What are we trying to prevent here? Someone opens contactors via webinterface, and then accidentally restarting, and it proceeds to operate? Feels like a very niche scenario that we dont need to consider?

StarkJohan:
> Same goes with the storing to persistent memory. What are we trying to prevent here? Someone opens contactors via webinterface, and then accidentally restarting, and it proceeds to operate? Feels like a very niche scenario that we dont need to consider?

To me the scenario of someone pausing/stopping and counting on this being persistent is a good reason to keep the persistence after a power outage or other accidental reboot.

dalathegreat (approved):

PR #508, opened 2024-10-03
Safety: Refactor cellvoltage safeties

dalathegreat:
### What
This PR improves overall safety for many battery types, by introducing global min/max cellvoltage events. Previously this feature was only available for some battery types (LEAF,i3,imiev,egmp,kia64,kangoo,zoe,tesla,volvo).

### Why
Instead of having the check inside each battery handler, it is moved to the Safety.cpp file.  This makes it inherently safer, it is no longer possible to add a new battery type without a cellvoltage check.

### How
Each battery type initializes the following on startup, or later on in the code when we detect which battery is actually used (for instance 90S and 98S battery in same battery file)

datalayer.battery.info.max_design_voltage_dV = MAX_PACK_VOLTAGE_DV;
datalayer.battery.info.min_design_voltage_dV = MIN_PACK_VOLTAGE_DV;
datalayer.battery.info.max_cell_voltage_mV = MAX_CELL_VOLTAGE_MV;
datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;

From each battery implementation, we just set the min/max allowed cellvoltages and deviation value, in the battery.h file.

This PR also cleans up:
- Unused recalibration&soc check event for tesla battery2

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -395,8 +395,37 @@ void update_values_battery2() {  //This function maps all the values fetched via
 
   datalayer.battery2.status.temperature_max_dC = battery2_temperature_max * 10;  // Add a decimal
 
-  datalayer.battery2.status.cell_min_voltage_mV = datalayer.battery2.status.cell_voltages_mV[0];
-  datalayer.battery2.status.cell_max_voltage_mV = datalayer.battery2.status.cell_voltages_mV[1];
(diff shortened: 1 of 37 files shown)


dalathegreat:
Thanks! It is impossible to test every single battery, so if issues arise with false positive triggering we will have to bugfix them later on. Worst thing that can happen is that it is too safe 😅

Merging!

PR #518, opened 2024-10-06
Feature: Double Atto 3 battery support

dalathegreat:
### What
This PR adds support for running two Atto 3 batteries at the same time. Testing needed

### Why
User requested feature

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -19,13 +19,12 @@ static uint8_t counter_100ms = 0;
 static uint8_t frame6_counter = 0xB;
 static uint8_t frame7_counter = 0x5;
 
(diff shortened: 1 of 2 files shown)


juancruz1953:
The double BYD Atto3 battery works well with the emulator software
BYDatto3-doubleBat
. This was trialed with a double Lilygo, fitted with the dual CAN MCP2515 add-on located on the battery Lilygo. Prior to doubling up, both batteries were individually charged to the same voltage (420V), before paralleling with the requisite fuses in place. Both batteries were simultaneously switched on automatically via the software. With this setup, an interesting observation is that the double battery does charge up if single battery operation is used, but the second battery statistics are obviously not visible. Subsequent checking of the second battery showed that balancing was working in this state.

dalathegreat:
Thanks for testing! Merging this to main, will be included in release 7.5.0

PR #550, opened 2024-10-20
CHAdeMO Vehicles, Battery

NJbubo:
### What
I added the CHAdeMO GPIO configuration.

### Why
The inputs are not working.

### How
It is necessary to configure the GPIOs. In particular, setting the initial state of the outputs and enabling the inputs.

The inputs are not working with this modification!!!

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -174,11 +174,17 @@ static int16_t battery2_temp_polled_max = 0;
 static int16_t battery2_temp_polled_min = 0;
 #endif  // DOUBLE_BATTERY
 
-void print_with_units(char* header, int value, char* units) {
-  Serial.print(header);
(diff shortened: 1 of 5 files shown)


NJbubo:
I update CHADEMO-BATTERY.cpp. After this modification the INPUTs are worging good.

I think this place is not a correct place the GPIO config.

In my test, I use the CHADEMO_PIN_7 input as a start button.
I can't imagine how you intended to use "plug_inserted" as a signal according to the program.
I think the CHAdeMO connector pin 7 is connected directly to GND/0V when viewed from the side of the charger.

dalathegreat:
@smaresca is the expert when it comes to the Chademo testing, got any opinions on this?

smaresca:
@dalathegreat  noted, will review

smaresca:
I am curious @NJbubo -- what model connector are you using?

NJbubo:
Hi. I am using a SUMITUMO SEVD-11U connector.

Issue #479, opened 2024-09-22
Remaining capacity not scaled


bubble07:
### Expected Behavior
Remaining battery capacity sent to inverter is accurate to the remaining capacity when taking consideration of any max /min scaling applied to SoC

### Actual Behavior

Remaining capacity is calculated from real SoC not scaled SoC. Screenshot attached below.
### Steps to Reproduce the Problem

1. Input real battery capacity in user settings (e.g.64000)
2. Input max and min battery capacity and enable scaling (e.g. min 20%, max 80%)
3. Observe BMS remaining capacity is equal to real capacity x real SoC, rather than usable capacity (real capacity x (max-min)) X scaled SoC. Same value also shown in webserver.

### Settings

- Software version: `7.3`
- Battery used: `kia 64kwh`
- Inverter communication protocol: `solax`
- Hardware used for Battery-Emulator: `HW_LILYGO`
- CONTACTOR_CONTROL: `yes
![Screenshot_2024-09-22-20-21-08-52_40deb401b9ffe8e1df2f1cc5ba480b12](https://github.com/user-attachments/assets/d5c93c10-91f6-4e61-afb7-d84073258d24)
`
- DUAL_CAN: `yes`
- WEBSERVER: `yes`
- MQTT: `yes`

PR #557, opened 2024-10-23
Feature/scaled remaining capacity

amarofarinha:
### What

Implemented a new calculation for the reported remaining capacity in watt-hours (reported_remaining_capacity_Wh), accounting for the minimum SOC percentage. The new logic subtracts the capacity reserved by the minimum SOC from the total remaining capacity.

### Why

This change resolves an issue where the remaining capacity did not properly account for the minimum SOC margin, causing inaccurate reporting of available battery capacity. The fix addresses bug #479, ensuring that the battery emulator accurately reflects the usable capacity, excluding the portion that cannot be discharged due to the defined minimum SOC margin.

### How

- Added reported_remaining_capacity_Wh to datalayer and a calculation that removes the unused capacity based on the min_percentage setting (expressed in basis points).
- Updated reported_remaining_capacity_Wh to reflect the adjusted remaining capacity after the minimum SOC margin is applied.
- Changed inverter files to send the newly created datalayer.battery.status.reported_remaining_capacity_Wh instead of the previous remaining_capacity_Wh.


dalathegreat (approved):
Nice change! Thanks for fixing this 👍

PR #561, opened 2024-10-26
Feature: Add more info page for Atto3

dalathegreat:
### What
This PR implements an extra battery info page for the BYD Atto 3

### Why
More insight for users if battery is locked or not, you can see if SOC% moves

### How
The more battery info webpage now contains the following info:

image

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -1,6 +1,7 @@
 #include "../include.h"
 #ifdef BYD_ATTO_3_BATTERY
 #include "../datalayer/datalayer.h"
+#include "../datalayer/datalayer_extended.h"
 #include "../devboard/utils/events.h"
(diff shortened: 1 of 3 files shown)


GraphicObession:
[BYD ATTO3 extended.xlsx](https://github.com/user-attachments/files/17532691/BYD.ATTO3.extended.xlsx)

Thanks Dala!

Unfortunately my BMS does not seem update the SOC at all!  Or am I doing something wrong?   I have attached my observations in an excel sheet.

BTW what is the best way of creating a copying the updated files down?  I copied the text which was pretty cumbersome.

dalathegreat:
@GraphicObession If your values are stuck, it means the battery pack came from a crashed vehicle. You will have to use estimated SOC for the foreseeable future. Currently we have no way to "unlock" a crashed battery. See the Atto3 wiki for more info.

dalathegreat:
Oh, and thanks for testing this. I am merging this!

GraphicObession:
Thanks Dala!

Issue #207, opened 2024-02-26
MQTT sensor names in 5.3.0 changed


MarkoBolt:
The non cell MQTT sensor values have been changed in 5.3.0 resulting in new sensors via MQTT. 
In the new sensor names the lily board ID is present:
```
ESP32_1234567
```

kharnt0x:
@MarkoBolt the issue will be for those with multiple lilygos there will be a namespace conflict I think... the intent was to ensure we keep the sensor names unique

Cabooman:
At least for me and Mbolt (and probably others) the issue is that we'll get duplicate entities for a single setup in e.g. HA when using several lily's for testing purposes. Let's figure out a compile-time switch for differentiating (or not) when we're in dev mode again :)

StarkJohan:
We should take the step to MQTT discovery as this will make these kinds of problems easier to handle. As I see it we need to keep multiple scenarios in mind.

Most users will have one setup and might do some testing with different hardware. This scenario is best handled with generic sensor names and MQTT discovery. If the hardware changes, the sensor names stay the same.

Some users will use multiple setups with the same MQTT broker. In this case the sensors need to be unique on a system basis. I.e. some kind of system identifier is needed and I don't think the default esp host name is a viable solution.

I suggest adding a "hostname" variable in the user settings and swapping to MQTT discovery. This should solve both scenarios, including development.

Cabooman:
@StarkJohan Discovery is currently used for all MQTT data, but with said host name and the library that you’re not too happy with ;)

I was writing basically your post a few hours ago but didn’t manage to finish, so I agree with the scenario or use case perspective. Most will have one lily and one setup, some will then have two or more lilys with one setup and yet fewer will have some form of many lilys and many setups. If this was at work, we would solve it in the cabling, using one or more pins for system identification, but that doesn’t feel very feasible for a number of reasons.

In any case, “nice” entity names should be default. All ideas welcome!

PR #566, opened 2024-10-28
Enhancement/mqtt topic naming

amarofarinha:
### What

- Added customization options for MQTT topic name, object ID prefix, and device name in USER_SETTINGS.cpp to allow consistent naming across different boards.
- Retained previous behavior that calculates names based on the hardware ESP32 ID for users who prefer this method.

### Why

- Addresses the need for easier migration and configuration sharing in Home Assistant, ensuring users can maintain consistent naming without reconfiguration.
- Enables users to preserve their entire Home Assistant configuration even when changing boards.
- This change responds to open issue [#207](https://github.com/dalathegreat/Battery-Emulator/issues/207) regarding naming changes in previous versions.

### How

- Introduced the MQTT_MANUAL_TOPIC_OBJECT_NAME define, enabling users to specify custom naming.
- The default naming format is now based on the manual configuration. This represents a breaking change for MQTT naming; users who wish to maintain the previous behavior must comment out the MQTT_MANUAL_TOPIC_OBJECT_NAME define.
- Default naming format when MQTT_MANUAL_TOPIC_OBJECT_NAME is not defined will revert to 'battery-emulator_esp32-XXXXXX', based on the hardware ID.

### DIFF
--- a/Software/USER_SETTINGS.h
+++ b/Software/USER_SETTINGS.h
@@ -75,7 +75,13 @@
 // #define MQTT  // Enable this line to enable MQTT
 #define MQTT_SERVER "192.168.xxx.yyy"
 #define MQTT_PORT 1883
-#define MQTT_MANUAL_TOPIC_OBJECT_NAME  // Enable this to use custom MQTT topic, object ID prefix, and device name. If not defined, the default naming format 'battery-emulator_esp32-XXXXXX' (based on hardware ID) will be used. See USER_SETTINGS.cpp for configuration options.
+#define MQTT_MANUAL_TOPIC_OBJECT_NAME  // Enable this to use custom MQTT topic, object ID prefix, and device name.    \
+                                       // WARNING: If this is not defined, the previous default naming format         \
+                                       // 'battery-emulator_esp32-XXXXXX' (based on hardware ID) will be used.        \
+                                       // This naming convention was in place until version 7.5.0.                    \
+                                       // Users should check the version from which they are updating, as this change \
+                                       // may break compatibility with previous versions of MQTT naming.              \
+                                       // Please refer to USER_SETTINGS.cpp for configuration options.
 
 /* Home Assistant options */
 #define HA_AUTODISCOVERY  // Enable this line to send Home Assistant autodiscovery messages. If not enabled manual configuration of Home Assitant is required


clowrey:
I like it but think the comments in the user settings cpp file are a little confusing as "default" is without the define, not what is generated by default in the new code.

clowrey:
Oh and do you have a screenshot of what an entity name in HA results from these defaults?

Software/USER_SETTINGS.h
```diff
@@ -75,6 +75,7 @@
 // #define MQTT  // Enable this line to enable MQTT
 #define MQTT_SERVER "192.168.xxx.yyy"
 #define MQTT_PORT 1883
+#define MQTT_MANUAL_TOPIC_OBJECT_NAME  // Enable this to use custom MQTT topic, object ID prefix, and device name. If not defined, the default naming format 'battery-emulator_esp32-XXXXXX' (based on hardware ID) will be used. See USER_SETTINGS.cpp for configuration options.
```
Cabooman, line 78: For the merge, it's probably good to retain previous default behavior, i.e. comment this out

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -68,17 +71,16 @@ SensorConfig sensorConfigs[] = {
 
 };
 
-static String generateCommonInfoAutoConfigTopic(const char* object_id, const char* hostname) {
-  return String("homeassistant/sensor/battery-emulator_") + String(hostname) + "/" + String(object_id) + "/config";
+static String generateCommonInfoAutoConfigTopic(const char* object_id) {
+  return String("homeassistant/sensor/") + String(topic_name) + "/" + String(object_id) + "/config";
```
Cabooman, line 75: Is the String-cast really necessary when the original variable type is String?

Applies to all other places as well

Cabooman (changes requested):
Looks good overall!

clowrey:
So ```esp32_5bcf78_temperature_min``` will become ```esp32_temperature_min``` ? 
Personally I think the HA ID name here should be ```BE_temperature_min``` since ESP32 is not really necessary here and does not describe the device well, I have tons of esp32 devices besides the battery emulator

clowrey:
Also the names of each entity dont need "Battery Emulator" appended to the front of each one, since it is part of the device itself. 
image
Should just say "Temperature Min"

Software/USER_SETTINGS.h
clowrey: Invite for new default being the same.

clowrey:
> > Also the names of each entity dont need "Battery Emulator" appended to the front of each one, since it is part of the device itself. image Should just say "Temperature Min"
> 
> I agree that simplifying entity names could be beneficial. However, i would prefer to address this change in a separate PR to carefully assess any impacts on existing Home Assistant configurations. This way, we ensure a smooth transition for current users who may rely on the current naming format.

I agree it should be separate from this PR. It would make the entities in line with how ESPHome does it now as well, right now trying to find the right entity to use for a dashboard manually is a mess because the "Battery Emulator" overall name is duplicated like this:
image

clowrey:
One last change, could we have the 
```const char* mqtt_device_name = "BatteryEmulator_esp32";  ``` 
    
Instead
    
```const char* mqtt_device_name = "Battery Emulator";  ```

clowrey:
> > One last change, could we have the `const char* mqtt_device_name = "BatteryEmulator_esp32"; `
> > Instead
> > `const char* mqtt_device_name = "Battery Emulator"; `
> 
> i think we have that in this PR image

I mean to make the default ```const char* mqtt_device_name = "Battery Emulator"; ```

PR #568, opened 2024-10-29
Bugfix: Add capacity writing to BYD CAN

dalathegreat:
### What
This PR adds capacity handling for BYD CAN. Previously we incorrectly wrote charge limits to the field that should contain capacity.

image

### Why
We fix this in order to satisfy the inverter, and make the display on inverters show correct info

### How
Content in 0x150 is now formatted correctly


dalathegreat:
Confirmed working on the Discord:

image

PR #570, opened 2024-10-29
Bugfix: Tesla S/3/X/Y , add contactor opening incase of FAULT

dalathegreat:
### What
This PR implements contactor opening incase we encounter a fault

### Why
To increase safety when using a Tesla battery

### How
If we encounter FAULT state, we drop sending the 221_2 message, and after a few seconds we stop sending also the 221_1 message


josiahhiggs (approved):
Much better contactor control under fault conditions now, Tested with TM3 LFP and Solax CAN.

PR #573, opened 2024-10-30
Feature/can fd crystal frequency setting

kukumagi:
### What
This PR implements CAN_FD crystal frequency setting in USER_SETTINGS.

### Why
Since there are now MCP2518FD development boards that have 20MHZ crystals.

### How
In USER_SETTINGS.h if CAN_FD is defined it defines CAN_FD_CRYSTAL_FREQUENCY_MHZ which is by default OSC_40MHz. For the 20Mhz board it should be changed to OSC_20MHz.


dalathegreat (approved):
Nice improvement. Maybe if you have time, could you update the Wiki page for CAN FD add-on?

Issue #569, opened 2024-10-29
Need help with custom battery, CAN  list available


MathieuWuyts:
Hello,

I need help implementing a custom battery. I have a complete CAN list available. Precharge and contactors all programmed within the battery itself, so only needs a closed contact and 12V source to turn on.
I checked the existing .h and .cpp files of a few batteries, but I'm not sure how I can adapt my battery to the existing ones.
Anyone willing to help?

thanks

dalathegreat:
@MathieuWuyts 
What battery manufacturer? What Model? Datasheets? CAN list?

Please attach more information here

MathieuWuyts:
Hi Dala, it is a custom battery by Intercel in the Netherlands. Can list in attachment.
[CAN_Messages_Cellpower_CLPL_Battery.pdf](https://github.com/user-attachments/files/17557457/CAN_Messages_Cellpower_CLPL_Battery.pdf)

dalathegreat:
PR is now open! https://github.com/dalathegreat/Battery-Emulator/pull/575

PR #575, opened 2024-10-30
New battery: Cellpower BMS

dalathegreat:
### What
This PR implements support for the 250kbps Cellpower BMS

### Why
User requested feature

### How
When `CELLPOWER_BMS` is selected in the battery list, we interpret the incoming CAN messages according to the Cellpower protocol supplied in https://github.com/dalathegreat/Battery-Emulator/issues/569


dalathegreat:
Tested OK with battery only, merging this! :partying_face: 
image
image

PR #577, opened 2024-10-30
Speed up value updating

dalathegreat:
### What
This PR speeds up the main loop value update rate from 5seconds -> 1seconds

### Why
It is an industry norm to update values every second. With an 1s update rate we can faster respond to any changing values, and evaluate safety checks more often.

### How
`intervalUpdateValues = INTERVAL_1_S; `


dalathegreat:
Thanks, merging this :raised_hands:

PR #582, opened 2024-11-03
Feature: Add degradation data reset for 30kWH LEAF batteries

dalathegreat:
### What
This PR implements diagnostic commands used to reset degradation data, also known as SOH%

### Why
Some 30kWh packs have a defective BMS firmware, that causes the degradation data to be incorrectly calculated, and SOH% drops too fast. Since we dont have any way to update the software on the battery yet, the next best solution is to reset the SOH% to factory new every once in a while.

### How
There is a button added under the "More battery info" page, which can be pressed

image

The prompt asks the user if they are sure. If OK is pressed, we proceed with the SOH% reset

image


nagyrobi:
Will this reset to 100%?

dalathegreat:
Feature tested OK, successfully restored 2x 30kWh batteries into a state where more energy could be extracted from them ([Faulty BMS firmware revision 4A](https://flipthefleet.org/2018/30-kwh-nissan-leaf-firmware-update-to-correct-capacity-reporting/))

Merging! :raised_hands:

PR #591, opened 2024-11-04
Add Support for Double-Battery in MQTT Integration

amarofarinha:
### What

This PR introduces support for a dual-battery setup in the MQTT integration. When double battery configuration is detected, the MQTT code now publishes information for the second battery as well. It also adds double battery support for TEST-FAKE-BATTERY for testing.

### Why

A user requested this feature to facilitate the use of MQTT with a two-battery system, as this integration was previously missing

### How

Modifications were made to the MQTT code to send information for both batteries.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -860,6 +860,8 @@ void update_scaled_values() {
      * Before we use real_soc, we must make sure that it's within the range of min_percentage and max_percentage.
     */
     uint32_t calc_soc;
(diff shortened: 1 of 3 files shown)


dalathegreat:
Nice! Should we add double-battery support to Test-Fake-Battery, so it becomes easier to test this?

dalathegreat (approved):

dalathegreat (approved):
LGTM!

PR #594, opened 2024-11-05
Bugfix: Add remaining capacity writing for Pylon

dalathegreat:
### What
This PR fixes a bug, when using `PYLON_BATTERY` the remaining capacity would be 0 all the time

image

### Why
Fixing this to correct behaviour

### How
We set capacity left according to SOC%, same as other implementations


dalathegreat:
Tested successfully by a user on the Discord, merging this!

PR #603, opened 2024-11-08
Improvement: Add charge/discharge current to datalayer

dalathegreat:
### What
This PR adds allowed charge/discharge to datalayer. This simplifies the code, since each inverter protocol no longer needs to calculate its own current value.

### Why
All EV batteries send allowed charge/discharge in Watts. This is problematic, since the Inverters often need an amperage limit.

### How
We remove 300 lines of code, and add 100 lines :)

The values are also shown in the Webserver
image


josiahhiggs (approved):
Working great with TM3 LFP and Solax CAN.

PR #617, opened 2024-11-14
Improvements: Webserver / integration simplification

dalathegreat:
### What
This PR implements simplified webserver & battery/inverter integrations

### Why
To simplify further development and make scaling the software up easier

### How
- Instead of large #ifdef blocks inside Webserver.cpp to signal what battery/inverter is used, we now write a text string to the datalayer on startup. The webpage displays both battery_protocol and inverter_protocol. This simplifies development, no longer needed to add inverter/battery specific text outside of the battery's .cpp file.
- init_inverter() now calls setup_inverter() in each inverter protocol. Code required specifically for inverter protocol is now inside that inverter protocols .cpp file. This simplifies development, with less risk of forgetting inv specific stuff
- All #ifdef intervalUpdateValues removed, since code now updates at 1s rate instead of 5s
- Startup DEBUG_VIA_USB texts removed
- BONUS: Some inverter/batteries were missing from the Github workflow. Added them to build system

### DIFF
--- a/.github/workflows/compile-all-batteries.yml
+++ b/.github/workflows/compile-all-batteries.yml
@@ -34,19 +34,25 @@ jobs:
         # These are the batteries for which the code will be compiled.
         battery:
           - BMW_I3_BATTERY
+          - BMW_IX_BATTERY
           - BYD_ATTO_3_BATTERY
           - CELLPOWER_BMS
           - CHADEMO_BATTERY
           - IMIEV_CZERO_ION_BATTERY
           - JAGUAR_IPACE_BATTERY
           - KIA_HYUNDAI_64_BATTERY
+          - KIA_E_GMP_BATTERY
           - KIA_HYUNDAI_HYBRID_BATTERY
+          - MG_5_BATTERY
           - NISSAN_LEAF_BATTERY
           - PYLON_BATTERY
           - RJXZS_BMS
+          - RANGE_ROVER_PHEV_BATTERY
           - RENAULT_KANGOO_BATTERY
+          - RENAULT_TWIZY_BATTERY
           - RENAULT_ZOE_GEN1_BATTERY
           - RENAULT_ZOE_GEN2_BATTERY
+          - SANTA_FE_PHEV_BATTERY
           - TESLA_MODEL_3Y_BATTERY
           - VOLVO_SPA_BATTERY
           - TEST_FAKE_BATTERY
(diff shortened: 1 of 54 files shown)


nagyrobi:
Although it would probably require a serious amount of work at first, maybe it's worth considering implementing all these as ESPHome components.

ESPHome is a very well designed, extremely modular environment. It natively supports CAN and Modbus as frameworks, adding discrete devices on top of these is pretty straightforward. Has webserver support, MQTT fully customizable.
Easier to maintain on long term.

dalathegreat:
> Although it would probably require a serious amount of work at first, maybe it's worth considering implementing all these as ESPHome components.
> 
> ESPHome is a very well designed, extremely modular environment. It natively supports CAN and Modbus as frameworks, adding discrete devices on top of these is pretty straightforward. Has webserver support, MQTT fully customizable. Easier to maintain on long term.

Feel free to fork it there! I personally dont use MQTT or Webserver, I focus mainly on getting a stable and reliable way to re-use the batteries.

clowrey:
> Although it would probably require a serious amount of work at first, maybe it's worth considering implementing all these as ESPHome components.
> 
> ESPHome is a very well designed, extremely modular environment. It natively supports CAN and Modbus as frameworks, adding discrete devices on top of these is pretty straightforward. Has webserver support, MQTT fully customizable. Easier to maintain on long term.

I want to do a small test with ESPhome soon - I already have some of the Tesla M3 battery protocol decoded via ESPhome, mainly just need to emulate the BYD HV protocol - any interest in helping with that @nagyrobi ? I don't think it will be particularly hard and could be done all in YAML, or via an external component. We probably should start another thread on the discord or somehow discuss this elsewhere though.

clowrey:
For this commit I fully support removing as many #ifdef blocks as possible from the code base :). 

Somehow getting the dual batteries to not require doubling the Tesla.cpp file would be a huge improvement too eventually. Probably by making everything classes etc.

dalathegreat:
> Somehow getting the dual batteries to not require doubling the Tesla.cpp file would be a huge improvement too eventually. Probably by making everything classes etc.

This is on my TODO list, I really want to refactor the battery handling as a class (which would get rid of duplicate code). But a bit too much for this PR :)

nagyrobi:
> any interest in helping with that @nagyrobi

Sure!

Issue #614, opened 2024-11-13
More info found in 0x91 frame (BYD-CAN.cpp)


jsphuebner:
https://github.com/dalathegreat/Battery-Emulator/blob/0b966655479860f87d70e3e210dc7240639d7824/Software/src/inverter/BYD-CAN.cpp#L180

While running a charge session I found 2 more items in the message that matched the values polled from modbus:

(byte[2] << 8 + byte[3])/10 is battery current
(byte[4] << 8 + byte[5])/10 is inverter temperature

dalathegreat:
Massive thanks @jsphuebner , added this in PR https://github.com/dalathegreat/Battery-Emulator/pull/618

PR #618, opened 2024-11-14
Feature: Add more BYD inverter mappings

dalathegreat:
### What
This PR adds more inverter content discovered in https://github.com/dalathegreat/Battery-Emulator/issues/614

### Why
Inverter values can be used for future development

### How
More data is extracted from 0x091 message


dalathegreat:
Merging this, no risk of anything since the value is not used anywhere.

PR #639, opened 2024-11-25
Feature: Add initial version of CAN logger

dalathegreat:
### What
This PR implements CAN logging, directly in the Webserver

### Why
To help reverse engineering, make it easier for users to take CAN logs

### How
There is a new page added to the Webserver, "CAN logger". When opened, the system starts logging CAN messages. This is disabled by default to not disturb the system.

image

image

Bonus: This PR also changes the style of the webserver buttons to make them easier to navigate on mobile devices

image


dalathegreat:
@nagyrobi 

Save as .txt button has been added

image

For now the system stores 15k chars, will be revised for future versions.

nagyrobi:
God bless you. If all my wishes could come true like this...

Issue #643, opened 2024-11-27
BYD CAN protocol should send a target charge voltage instead of "MAXVOLTAGE" which it sends now. 


clowrey:
Currently the BYD-CAN.cpp code is like this:

```cpp
  //Map values to CAN messages
  //Maxvoltage (eg 400.0V = 4000 , 16bits long)
  BYD_110.data.u8[0] = (datalayer.battery.info.max_design_voltage_dV >> 8);
  BYD_110.data.u8[1] = (datalayer.battery.info.max_design_voltage_dV & 0x00FF);
  //Minvoltage (eg 300.0V = 3000 , 16bits long)
  BYD_110.data.u8[2] = (datalayer.battery.info.min_design_voltage_dV >> 8);
  BYD_110.data.u8[3] = (datalayer.battery.info.min_design_voltage_dV & 0x00FF);
  ```
This should instead be a set point voltage - IMO a configurable value from the web interface would be ideal so you can use this to limit max charge voltage and periodically peak charge it, without re-compiling.

dalathegreat:
Fix for this is live: https://github.com/dalathegreat/Battery-Emulator/pull/647

PR #647, opened 2024-11-28
Feature: Add target voltage BYD setting, configurable setting!

dalathegreat:
### What
This PR achieves two things
- We no longer send the battery design voltage value to BYD protocol (it expects a target voltage)
   - If user has not defined a value, we send the design value -+ offset in order to not trigger events/faults
   - If user has defined a value, we use that as target
- We add target voltages as persistent setting
   - Bonus, added event to signal incase storing usersettings fail

### Why
Fixes https://github.com/dalathegreat/Battery-Emulator/issues/643

### How
The configurable settings are in the Webserver (and in USER_SETTINGS.h)

image

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -429,9 +429,6 @@ void init_stored_settings() {
     datalayer.battery.settings.max_user_set_discharge_dA = temp;
   }
   datalayer.battery.settings.soc_scaling_active = settings.getBool("USE_SCALED_SOC", false);
-  settings.end();
-
-  settings.begin("batteryExtra", false);
   temp = settings.getUInt("TARGETCHVOLT", false);
   if (temp != 0) {
     datalayer.battery.settings.max_user_set_charge_voltage_dV = temp;
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -375,6 +375,7 @@ void init_serial() {
 
 void init_stored_settings() {
   static uint32_t temp = 0;
+  //  ATTENTION ! The maximum length for settings keys is 15 characters
   settings.begin("batterySettings", false);
 
   // Always get the equipment stop status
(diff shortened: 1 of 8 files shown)


clowrey:
I have tested it and it does work as far as I can tell so far :)

I think the greyed out initial value of 500V maybe could be set to the default from the battery.h file? 
Screenshot_20241203-163214-EDIT (Medium)
Screenshot_20241203-163605 (Medium)
PXL_20241204_023610629 (Medium)

dalathegreat:
@clowrey It looks like the values are not saving right. Are you getting any Event about saving settings failed?

clowrey:
> @clowrey It looks like the values are not saving right. Are you getting any Event about saving settings failed?

It does seem to save - you can see the 360V in one image - just the initial value when first enabled it shows 500V greyed out.

clowrey:
> @clowrey It looks like the values are not saving right. Are you getting any Event about saving settings failed?

You're right I just rebooted and it was at 386V with the custom voltage disabled. 

There is an event - failed to save namespace full?

dalathegreat:
Okay to finalize this feature I am going to need some help. I have tried everything to be able to save these new parameters into the namespace, but it keeps failing. Help! :sweat_smile:

lenvm:
To ensure the relevant bug issue gets closed automatically, I'm adding this comment:

Fixes #643

mvgalen:
> Okay to finalize this feature I am going to need some help. I have tried everything to be able to save these new parameters into the namespace, but it keeps failing. Help! 😅

The maximum length for the key is 15 characters. USE_VOLTAGE_LIMITS exceeds that.

dalathegreat:
Massive thanks @mvgalen , I pushed a small fix, tested working great! Also @clowrey has verified the functionality, so now we can merge this :partying_face:

PR #649, opened 2024-12-01
docs: update README.md

eltociear:
### What
compability -> compatibility

### Why
Why does it do it?

### How
How does it do it?


dalathegreat (approved):
Thanks!

PR #654, opened 2024-12-02
Improvement: General MEB improvements

mvgalen:
### What
This PR improves the MEB support. Removing spurious events and increasing responsiveness of battery voltage and SOC. 

### Why
Issues not yet solved in the late MEB branch.

### How
- Only use values form CAN message if BMS is not in init state.
- Quickly determine number of cell of the pack
- Use broadcast battery_voltage and battery_SOC
- Set default SOC to 0.05% to prevent battery empty event on startup


dalathegreat (approved):
LGTM!

PR #655, opened 2024-12-03
Change web based CAN log file format to CANdump

mvgalen:
### What
Change web based CAN log format to CANdump format.
Improve performance by removing unnecessary strncat and strlens.

### Why
Make logfiles directly compatible with SavvyCAN.

### How
Slightly changed format. 
Tested downloaded file with SavvyCAN


Software/Software.ino
```diff
@@ -642,44 +642,37 @@ void print_can_frame(CAN_frame frame, frameDirection msgDir) {
...
 
   if (datalayer.system.info.can_logging_active) {  // If user clicked on CAN Logging page in webserver, start recording
-
-    char message_string[128];  // Buffer to hold the message string
-    int offset = 0;            // Keeps track of the current position in the buffer
-
+    char *message_string = datalayer.system.info.logged_can_messages;
+    int offset = datalayer.system.info.logged_can_messages_offset;  // Keeps track of the current position in the buffer
+    
+    if (offset + 128 > sizeof(datalayer.system.info.logged_can_messages)) {
```
obbardc, line 648: style nit: move size variable before the length check

Software/Software.ino
```diff
@@ -642,44 +642,37 @@ void print_can_frame(CAN_frame frame, frameDirection msgDir) {
...
+      // Not enough space, reset and start from the beginning
+      offset = 0;
+    }
+    size_t message_string_size = sizeof(datalayer.system.info.logged_can_messages);
```
obbardc, line 652: style nit: declare where other variables are declared, at the top of the block

Software/Software.ino
```diff
@@ -642,44 +642,37 @@ void print_can_frame(CAN_frame frame, frameDirection msgDir) {
 #endif  //#DEBUG_CAN_DATA
 
   if (datalayer.system.info.can_logging_active) {  // If user clicked on CAN Logging page in webserver, start recording
-
-    char message_string[128];  // Buffer to hold the message string
-    int offset = 0;            // Keeps track of the current position in the buffer
-
+    char *message_string = datalayer.system.info.logged_can_messages;
```
obbardc: would it be better to copy into the string e.g sprintf ?

Software/Software.ino
```diff
@@ -642,44 +642,37 @@ void print_can_frame(CAN_frame frame, frameDirection msgDir) {
...
-        strnlen(datalayer.system.info.logged_can_messages, sizeof(datalayer.system.info.logged_can_messages));
-    size_t available_space =
-        sizeof(datalayer.system.info.logged_can_messages) - current_len - 1;  // Space left for new data
-
-    if (available_space < strlen(message_string) + 1) {
-      // Not enough space, reset and start from the beginning
-      current_len = 0;
-      datalayer.system.info.logged_can_messages[0] = '\0';
-    }
+    offset += snprintf(message_string + offset, message_string_size - offset, "\n");
```
obbardc, line 674: perhaps `\r\n` for windows users compatibility in notepad ?

obbardc (changes requested):

Software/Software.ino
mvgalen: Sorry, I am afraid I don't understand your suggestion. Can you show the changed code?

obbardc:
```suggestion
    if (sizeof(datalayer.system.info.logged_can_messages) > offset + 128) {
```
looks cleaner to me.

Software/Software.ino
mvgalen: Ok.

Software/Software.ino
mvgalen: What do you mean? Nothing is copied here? snprintf's are below this block.

Software/Software.ino
mvgalen: Do you really think this is necessary? The newer windows wordpad also natively supports only \n.

Software/Software.ino
obbardc: Sorry. misread.

Software/Software.ino
obbardc: No, that's fine :-) Just mentioned it as something minor to think about.

Software/Software.ino
mvgalen: Hmm, I get your point, but at the same time it seems less intuitive, since you place the comparator in the wrong direction ( > vs. < ). I definitively find it easier to read the offset+value is greater than the size.

obbardc: that's fine. :-)

obbardc (approved):

mvgalen:
@obbardc is anything still blocking review approval?

dalathegreat (approved):
LGTM! Now we just need to update the can-logging wiki page to explain the format

mvgalen:
I will update the wiki later this morning.

Op di 3 dec. 2024 22:35 schreef Daniel Öster ***@***.***>:

PR #662, opened 2024-12-08
Improvement: Moving secrets from USER_SETTINGS.cpp to platformio config

No-Signal:
### What
This PR moves user secrets from USER_SETTINGS.cpp & USER_SETTINGS.h into a platformio config file USER_SECRETS.ini

### Why
This change makes it less likely that secrets are accidentally exposed by users committing them and makes it easier to maintain these secrets during development

### How
- USER_SECRETS.ini is included as additional platformio configuration in platformio.ini
- A template file USER_SECRETS_TEMPLATE.ini is copied to USER_SECRETS.ini if it does not exist and acts as a base configuration file
- USER_SECRETS.ini has been added to .gitignore to ensure it is not accidentally tracked

### DIFF
--- a/Software/USER_SETTINGS.h
+++ b/Software/USER_SETTINGS.h
@@ -82,8 +82,6 @@
 #define WIFI
 //#define WIFICONFIG  //Enable this line to set a static IP address / gateway /subnet mask for the device. see USER_SETTINGS.cpp for the settings
 #define WEBSERVER  //Enable this line to enable WiFi, and to run the webserver. See USER_SETTINGS.cpp for the Wifi settings.
-#define WEBSERVER_AUTH_REQUIRED \
-  false  //Set this line to true to activate webserver authentication (this line must not be commented). Refer to USER_SETTINGS.cpp for setting the credentials.
 #define WIFIAP  //Disable this line to permanently disable WIFI AP mode (make sure to hardcode ssid and password of you home wifi network). When enabled WIFI AP can still be disabled by a setting in the future.
 #define MDNSRESPONDER  //Enable this line to enable MDNS, allows battery monitor te be found by .local address. Requires WEBSERVER to be enabled.
 #define LOAD_SAVED_SETTINGS_ON_BOOT  //Enable this line to read settings stored via the webserver on boot (overrides Wifi/battery settings set below)
@@ -92,7 +90,6 @@
 
 /* MQTT options */
 // #define MQTT  // Enable this line to enable MQTT
-#define MQTT_PORT 1883
 #define MQTT_MANUAL_TOPIC_OBJECT_NAME  // Enable this to use custom MQTT topic, object ID prefix, and device name.    \
                                        // WARNING: If this is not defined, the previous default naming format         \
                                        // 'battery-emulator_esp32-XXXXXX' (based on hardware ID) will be used.        \
--- a/Software/USER_SETTINGS.h
+++ b/Software/USER_SETTINGS.h
@@ -92,7 +92,6 @@
 
 /* MQTT options */
 // #define MQTT  // Enable this line to enable MQTT
-#define MQTT_SERVER "192.168.xxx.yyy"
 #define MQTT_PORT 1883
 #define MQTT_MANUAL_TOPIC_OBJECT_NAME  // Enable this to use custom MQTT topic, object ID prefix, and device name.    \
                                        // WARNING: If this is not defined, the previous default naming format         \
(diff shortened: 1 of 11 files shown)


dalathegreat:
Nice improvement! Does it work good for Arduino IDE users also?

beadon:
I also like this approach.  Testing ..

No-Signal:
A different approach would be required to support Arduino IDE users which would require a one time manual action which I avoided with Platformio.

The best option I can see to support both Arduino IDE and Platformio is to create a USER_SECRETS_TEMPLATE.h file that users would need to copy to USER_SECRETS.h once manually. USER_SECRETS.h would be added to .gitignore to avoid accidental commits.

dalathegreat:
Yeah well 99% of users are Arduino IDE users, so I guess we would need to cater to those :sweat_smile:

No-Signal:
I have updated the method to support both Arduino & VSCode users, it does require the manual step of copying the template file which I have added to the readme

dalathegreat:
Doing this change will be a kind of "breaking" change that will require a new Youtube video to be put up. Otherwise my email will explode :sweat_smile: 

But I like the change!

No-Signal:
For such a small code change the impact is significant. If there are other settings to move now is the time to do it

lenvm:
@No-Signal, you should be able to add and commit the file `USER_SECRETS.h` in one commit, and only add this file to `.gitignore` in the next commit. That removes the need for all users to do the copy from `USER_SECRETS.template.h` to `USER_SECRETS.h`.

No-Signal:
@lenvm my understanding is that once a file is being tracked in git adding it to .gitignore won't stop it tracking future changes

lenvm:
@No-Signal, this can be solved.

Please do the following on your branch:

- comment `USER_SECRETS.h` in `.gitignore`
- `git add .gitignore; git commit -m "comment USER_SECRETS.h in .gitignore"`
- `git mv Software/USER_SECRETS.TEMPLATE.h Software/USER_SECRETS.h`
- `git commit -m "rename USER_SECRETS.TEMPLATE.h to USER_SECRETS.h`
- uncomment `USER_SECRETS.h` in `.gitignore`
- `git add .gitignore; git commit -m "uncomment USER_SECRETS.h in .gitignore"`

No-Signal:
@lenvm if I follow your steps then all subsequent changes to USER_SECRETS.h are tracked by git which is what I'm attempting to avoid with this PR

lenvm:
Git shouldn’t, be tracking those changes after the last commit, where you add it to .gitignore. Try it out for yourself to confirm!

No-Signal:
@lenvm my understanding is that everyone that pulls the code would need to run the git rm command for this approach to work

lenvm:
Let’s try it out and see!

Even if that would be the case, it would be less intrusive than a breaking change for every user of the code, especially those that just download and compile and never plan to do any coding.

No-Signal:
I have pushed my test feature branch up so you can test https://github.com/No-Signal/Battery-Emulator/tree/feature/user_secrets2

This solution is certainly easier for people to get up and running and was the initial method I tried before testing platformio then the template approach

The downside is it makes it much easier for people to accidentally commit secrets which is why most projects I have seen use a template file instead

lenvm:
I have checked cloned your repository, and you are right.

Conclusions:
- Directly after checking out the repository, the change made to `USER_SECRETS.h` are indeed tracked by git.
- The changes are not tracked once you run the command `git rm --cached Software/USER_SECRETS.h`. Running this command leaves you with a `changes to be committed: deleted:` .
- As per GitHub documentation, and "example" file (i.e. template) is the way to go, like you mentioned. See [Git: Ignore already checked in file](https://gist.github.com/tsrivishnu/a2f3adbbca9fcad5f3597af301ad1abb). If we go this route, we should add the copy step to the GitHub Actions as well (`cp Software/USER_SECRETS.template.h Software/USER_SECRETS.h`, to make sure the workflows don't fail.

Software/USER_SETTINGS.h
```diff
@@ -92,7 +92,6 @@
 
 /* MQTT options */
 // #define MQTT  // Enable this line to enable MQTT
-#define MQTT_SERVER "192.168.xxx.yyy"
 #define MQTT_PORT 1883
```
lenvm, line 95: Perhaps also include this setting in `USER_SECRETS.h`

lenvm:
@dalathegreat, what do you say? Shall we merge this?

The change is a real improvement, but does still impact your YouTube video :)

dalathegreat (approved):
LGTM :+1: Let's merge this since we are indeed releasing a major update :)

PR #665, opened 2024-12-10
Update BMW-I3-BATTERY.cpp

beadon:
Saw this in the issue tracker, seems easy enough to fix. Looks like a copy + paste error which just results in over-writing some cell data accidentally.

### What
extracting data from a CAN message correctly.

### Why
Why does it do it?

Previous implementation seems to accidentally overwrite an array position.

### How
How does it do it?

Magic bytes.


dalathegreat (approved):

dalathegreat:
Thanks @beadon ! Merging this!

PR #666, opened 2024-12-12
Atto 3: Update scaling on real SOC

### LINKED ISSUE #590: BYD Atto 3 Battery issue
BYD Atto 3 Battery issue

### Actual Behavior

we use your emulator on BYD Atto 3 Battery we found that the real SOS is stuck on 94.90 or 77.50 and keep change from this tow SOC value even if i keep discharge the battery the value keep the same.
we test that on other battery (BYD atto 3 ) and we have the same value 
so if you have a solution for that to use it

thanks,

### Steps to Reproduce the Problem

1.
1.
1.

### Settings

- Software version:7.5.0 ``
- Battery used:BYD Atto 3 ``
- Inverter communication protocol: PYLON_CAN and the inverter is sofar hyd 20k ``
- Hardware used for Battery-Emulator: `HW_LILYGO`
- CONTACTOR_CONTROL: `no`
- DUAL_CAN: `yes/no`
- WEBSERVE
Screenshot_20241101_150829_Chrome
Screenshot_20241101_150836_Chrome
R: `yes/no`
- MQTT: `yes/no`

dalathegreat:
### What
There was one decimal too much on the real Atto 3 SOC% value sent by battery

### Why
To fix being able to use real soc

### How
*10 instead of *100

Fixes #590

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -118,7 +118,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
   datalayer.battery.status.real_soc = estimateSOC(datalayer.battery.status.voltage_dV);
   SOC_method = ESTIMATED;
 #else  // Pack is not crashed, we can use periodically transmitted SOC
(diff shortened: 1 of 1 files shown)


lenvm:
Fixes #590

lenvm (approved):

PR #668, opened 2024-12-12
Bugfix: PWM causes precharge to break

dalathegreat:
### What
Using PWM contactor control breaks the precharge pin. This PR fixes that

### Why
During development of PWM control, precharge was never tested

### How
This PR fixes the bug, and implements additional improvements to GPIO contactor control

- Precharge pin now works when using PWM :bug: 
- Precharge duration now configurable in USER_SETTINGS
- Precharge logic improved significantly, the new logic is:
   - Turn on negative contactor
   - Wait a few ms, turn on precharge resistor
   - Wait the user specified time (Set in USER_SETTINGS). Then turn on positive contactor
   - Wait a few ms, turn off precharge resistor (and economize contactors if PWM is on)
- Webserver now shows precharge time instead of :x: to not confuse users:
image
- Webserver now work with PWM and shows off/economized state:
image

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -132,7 +132,7 @@ State contactorStatus = DISCONNECTED;
 #define NEGATIVE_CONTACTOR_TIME_MS \
   500  // Time after negative contactor is turned on, to start precharge (not actual precharge time!)
 #define PRECHARGE_COMPLETED_TIME_MS \
-  1000  // Time after precharge completed, that we should turn off the precharge resistor (and economize if PWM is on)
+  1000  // After successful precharge, resistor is turned off after this delay (and contactors are economized if PWM enabled)
 #define PWM_Freq 20000  // 20 kHz frequency, beyond audible range
 #define PWM_Res 10      // 10 Bit resolution 0 to 1023, maps 'nicely' to 0% 100%
 #define PWM_HOLD_DUTY 250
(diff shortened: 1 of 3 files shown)


Software/Software.ino
```diff
@@ -126,30 +126,29 @@ State contactorStatus = DISCONNECTED;
...
-/* NOTE: modify the precharge time constant below to account for the resistance and capacitance of the target system.
- *	t=3RC at minimum, t=5RC ideally 
- */
-#define PRECHARGE_TIME_MS 160
-#define NEGATIVE_CONTACTOR_TIME_MS 1000
-#define POSITIVE_CONTACTOR_TIME_MS 2000
+#define NEGATIVE_CONTACTOR_TIME_MS \
+  500  // Time after negative contactor is turned on, to start precharge (not actual precharge time!)
+#define PRECHARGE_COMPLETED_TIME_MS \
+  1000  // Time after precharge completed, that we should turn off the precharge resistor (and economize if PWM is on)
```
lenvm, line 135: this comment is not quite clear
`that we should turn off the precharge resistor`

dalathegreat: Fixed in https://github.com/dalathegreat/Battery-Emulator/pull/668/commits/602fc6dc921968f0630d40d8a946694f8aba4711

dalathegreat:
Merging this, successfully tested the PWM fix by users on the Discord server :raised_hands:

PR #673, opened 2024-12-14
Bugfix: Add startup delay for contactors, fix PWM sequence

dalathegreat:
### What
This PR fixes a bug, where if PWM was used and the board was reset via powercut (Not reproducible via Webserver reboot), the next time it booted it could occasionally engage the startup sequence incorrectly

image

### Why
We fix this, since an incorrect startup sequence will skip precharge, and potentially damage components. Phenomenon still unclear why it happens on PWM setups, but this PR fixes it. 

### How
We fix the issue with a very simple fix. Instead of hammering the statemachine 10ms after CPU poweron, we wait 10 seconds until we start the precharge sequence. This has multiple benefits apart from fixing the glitched PWM startup, now incase a critical FAULT is encountered before the system has started (for instance battery sends something via CAN), we can now act on this and not turn on the battery at all, increasing safety.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -827,6 +827,13 @@ void handle_contactors() {
   }
 
   unsigned long currentTime = millis();
+
+  if (currentTime < INTERVAL_10_S) {
+    // Skip running the state machine before system has started up.
+    // Gives the system some time to detect any faults from battery before blindly just engaging the contactors
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


mvgalen (commented):
Are we sure we want to delay 10 seconds? Are there really batteries that take this long to startup? Is something like 4 seconds not long enough? If doing repeated power up tests, thsi 10s will start to get annoying very quick i think.

mvgalen (approved):
Looks good to me.

PR #675, opened 2024-12-15
BYD Atto 3: Make startup less prone to throw Events

dalathegreat:
### What
This PR makes the Atto 3 less likely to throw Events the first seconds on startup

image

### Why
Feedback from flying tools video on the Atto 3, which reports CELL DEVIATION and UNDERVOLTAGE as soon as the battery starts.

### How
The issue was split into two parts
- Cell max/min values where updated async during OBD2 polling, causing the min to actually be larger than max for a few seconds. By adding ABS subtraction to compare the diff, we catch this edge case
- Pack undervoltage on startup. Due to the Datalayer having 370.0V as startup value, which is 10V lower than the Atto 3 minimum voltage (380.0V), we now init the value to 380.1V to avoid false positive events.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -419,6 +419,9 @@ void setup_battery(void) {  // Performs one time setup at startup
   datalayer.battery.info.max_cell_voltage_mV = MAX_CELL_VOLTAGE_MV;
   datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
   datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
+  datalayer.battery.status.voltage_dV =
+      MIN_PACK_VOLTAGE_DV +
+      1;  // Init values in datalayer too low, causes safeties to trigger before values are sampled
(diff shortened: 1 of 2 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -419,6 +419,9 @@ void setup_battery(void) {  // Performs one time setup at startup
   datalayer.battery.info.max_cell_voltage_mV = MAX_CELL_VOLTAGE_MV;
   datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
   datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
+  datalayer.battery.status.voltage_dV =
+      MIN_PACK_VOLTAGE_DV +
+      1;  // Init values in datalayer too low, causes safeties to trigger before values are sampled
```
lenvm, line 424: Please add the values provided in the PR description directly to the code.

dalathegreat: Good point, fixed in [22f3fdd](https://github.com/dalathegreat/Battery-Emulator/pull/675/commits/22f3fddd3042d6f4161f1eb2b9bc1117451f6712)

lenvm (approved):

Issue #680, opened 2024-12-19
Lilygo constantly rebooting


tjjq44:
After flashing the board, Wifi AP isn't activated and I see this constantly looping on serial monitor...

ELF file SHA256: b5fd3e836

Rebooting...
ets Jul 29 2019 12:21:46

rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:4916
load:0x40078000,len:16436
load:0x40080400,len:4
ho 8 tail 4 room 4
load:0x40080404,len:3524
entry 0x400805b8

assert failed: tcp_alloc /IDF/components/lwip/lwip/src/core/tcp.c:1851 (Required to lock TCPIP core functionality!)

Backtrace: 0x400827d1:0x3ffb3010 0x4008e651:0x3ffb3030 0x40094a46:0x3ffb3050 0x40103903:0x3ffb3180 0x40103a7d:0x3ffb31a0 0x400dd6d4:0x3ffb31c0 0x400e3446:0x3ffb3210 0x400d933d:0x3ffb3230 0x400d326e:0x3ffb3280 0x4008f39a:0x3ffb32a0

### Settings

- Software version: `7.92`
- Battery used: `None (for now)`
- Inverter communication protocol: `CAN (Sunny Boy Storage 3.7)`
- Hardware used for Battery-Emulator: `HW_LILYGO`
- CONTACTOR_CONTROL: `yes`
- CAN_ADDON: `yes`
- WEBSERVER: `yes`
- MQTT: `no`

dalathegreat:
HI @tjjq44 , could you attach your configuration files (USER_SETTINGS.h and .cpp ), but be sure to remove Wifi login info from them!

dalathegreat:
Issue seems repeatable after updating ESP32 to v 3.1.0 

![image](https://github.com/user-attachments/assets/e04c3bda-a303-4cdc-af70-4e99fd18d67d)

@tjjq44 you might be able to continue by downgrading ESP32 to v3.0.3 or similar older version.

tjjq44:
Thank you! This time it boots ok but... in both mode (AP and normal wifi) I can't log in the webserver. 

[...]

PR #681, opened 2024-12-19
Bugfix: Board bootloops and insta crashes

dalathegreat:
### What
This PR fixes the bootlooping bug encountered in #680 

### Why
This was caused by the latest latest ESP32 3.1.0 release breaking the AsyncTCP library

### How
The me-no-dev/AsyncTCP library has not been maintained for over 5 months. We switch to the forked mathieucarbou-AsyncTCP branch instead now that has the fix required to work with the latest ESP32 release


wjcloudy (approved):
LGTM!

PR #687, opened 2024-12-20
Bugfix: avoid that GitHub Action workflow strategy expansion exceeds 256 results

lenvm:
### What
This PR updates the GitHub Action workflow that compiles all battery and inverter combinations.

### Why
Currently the "Compile all combinations" workflow fails, with the error `Strategy expansion exceeded 256 results for job 'build-matrix'`. See [Compile All Combinations | v7.9.2 #159](https://github.com/dalathegreat/Battery-Emulator/actions/runs/12339753775) as an example.

This PR avoids that GitHub Action workflow strategy expansion exceeds 256 results for job build-matrix

### How
It does this by splitting the workflow into two parts:
- Part 1: batteries starting with A-N
- Part 2: batteries starting with M-Z


dalathegreat (approved):
LGTM, we are taking lots of resources running all these tests :sweat_smile:

PR #690, opened 2024-12-21
Change Serial logging to flexible logging

mvgalen:
### What
This PR reroutes Serial output to a Logging class, which for now only support compile time switching between serial output and web based log output, but this could be made more flexible in the future.
Example output below:
image

### Why
To be able to see debug output through the web interface, for installations where serial logging is not very convenient. Also to be able to better see the order in which events etc. are happening. 

### How
The log output is placed in the same buffer as the web based can_logging, to save on memory and because both are not needed at the same time. If can logging is enabled, the web based logging is disabled. Also placing the output in memory seems to be much less cpu intensive than the serial-logging, where each byte is written separately to the hardware serial.
The display mechanism is the same as for the web based can-logging. Which means that load from the web transfer happens at the moment the user reloads the log page, which can be timed so it happens after the issue that someone may be trying to log and therefore has no influence on the issue.


Software/USER_SETTINGS.h
```diff
@@ -66,6 +66,11 @@
 
 /* Other options */
 //#define DEBUG_VIA_USB          //Enable this line to have the USB port output serial diagnostic data while program runs (WARNING, raises CPU load, do not use for production)
+//#define DEBUG_VIA_WEB          //Enable this line to log diagnostic data while program runs, which can be viewed via webpage (WARNING, slightly raises CPU load, do not use for production)
```
dalathegreat, line 69: For future? implementation once the DEBUG_VIA_WEB works good, we could consider removing it as a compile option altogether. We could then do the same as CAN logging via webserver, and toggle it on/off whenever the webpage is active. This would:

- Simplify the USER_SETTINGS, remove 5 rows
- Make it easier to right away debug an issue when needed, without the need to recompile

Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
```diff
@@ -132,7 +132,7 @@ void update_values_battery() {  //This function maps all the values fetched via
 
   datalayer.battery.status.cell_min_voltage_mV = min_cell_mv_value;
   datalayer.battery.status.cell_max_voltage_mV = max_cell_mv_value;
-  datalayer.battery.status.voltage_dV = static_cast<uint32_t>((calculated_total_pack_voltage_mV / 100));  // mV to dV
+  datalayer.battery.status.voltage_dV = static_cast<uint32_t>((calculated_total_pack_voltage_mV / 100));  // mV to dV}
```
dalathegreat, line 135: Unintended change?

Software/src/battery/MG-5-BATTERY.cpp
```diff
@@ -43,7 +43,7 @@ void update_values_battery() {  //This function maps all the values fetched via
 
   datalayer.battery.status.temperature_max_dC;
 
-#ifdef DEBUG_VIA_USB
+#ifdef DEBUG_LOG
```
dalathegreat, line 46: This block could be removed, unused

Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.cpp
```diff
@@ -217,7 +217,7 @@ void update_values_battery() {  //This function maps all the values fetched via
   datalayer_extended.zoePH2.battery_soc_min = battery_soc_min;
   datalayer_extended.zoePH2.battery_soc_max = battery_soc_max;
 
-#ifdef DEBUG_VIA_USB
```
dalathegreat, line 220: This block could be removed, unused

mvgalen: Ah, leftovers from before the rebase. I will remove it.

Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
mvgalen: Yes, indeed.

Software/src/battery/MG-5-BATTERY.cpp
mvgalen: Ah, leftovers from before the rebase. I will remove it.

Software/USER_SETTINGS.h
mvgalen: Yes, ofcourse. However, first let's keep this as an option. Though we could change the default to enabled. Because I think the overhead from the logging is very low. However if someone thinks he runs into an issue with it, he can easily temporarily remove the logging. Shall I change it to default enabled?

dalathegreat: I think it would be nice to:

- remove DEBUG_VIA_WEB option from USER_SETTINGS.h to make it easier for users.
   - Automatically enable logging when user opens the Debug webpage (See how CAN logging does it)
   - Automatically DISABLE logging when user returns to main page to save system resources

Make the calls to logging.print exit early incase logging is not turned on (Webpage not open)

mvgalen: Hmm, if we do that, in my experience you always miss exactly the thing why you are looking in the log (because something strange has occurred).

dalathegreat: Good point. Ok, let's leave it as-is!

dalathegreat (approved):

PR #693, opened 2024-12-22
Improvement/tesla advanced battery

josiahhiggs:
### What
This PR implements Tesla advanced battery info to the webserver, reorganize of data and labelling for reference.

### Why
Why does it do it? There is so much information that is able to be read from the battery, but was not coded to read it or display the data.

### How
How does it do it? All data frames in the TESLA-BATTERY.cpp were updated, adding information to the datalayer_extended and adding to the webserver under the advanced_battery.html. Factoring was moved from the Tesla-BATTERY.cpp to the advanced_battery.html using float, also added static const char to correctly display data in the webserver. Also separated the 0x352 with mux and labeled in the webserver. This data has not been completely verified as accurate.

Tesla more battery info 1
Tesla More Battery Info 2
Tesla More Battery Info 3
Tesla More Battery Info 4
Tesla More Battery Info 5


Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -1227,14 +2654,14 @@ the first, for a few cycles, then stop all  messages which causes the contactor
 #endif  //defined(TESLA_MODEL_SX_BATTERY) || defined(EXP_TESLA_BMS_DIGITAL_HVIL)
 
   //Send 30ms message
-  if (currentMillis - previousMillis30 >= INTERVAL_30_MS) {
```
dalathegreat, line 1230: Is it really OK to change the 30ms sending to 50ms? Should we test on more batteries before merging?

Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -44,43 +45,214 @@ static uint16_t battery_nominal_energy_remaining = 0;       // kWh
...
 static uint8_t battery_packCtrsClosingAllowed = 0;
 static uint8_t battery_pyroTestInProgress = 0;
-static uint8_t battery_battTempPct = 0;
-static uint32_t battery_packMass = 0;
-static uint32_t battery_platformMaxBusVoltage = 0;
-static uint32_t battery_packConfigMultiplexer = 0;
-static uint32_t battery_moduleType = 0;
-static uint32_t battery_reservedConfig = 0;
-//Fault codes
+static uint8_t battery_packCtrsOpenNowRequested = 0;
```
dalathegreat, line 103: To increase readability and make the logic easier, all values that hold either a true/false could instead of being defined as: 

`static uint8_t battery_packCtrsOpenNowRequested = 0;`

Instead be defined as:

`static bool battery_packCtrsOpenNowRequested = false;`

Minor improvement for easier maintainability!

Software/src/battery/TESLA-BATTERY.cpp
josiahhiggs: I forgot that I made that correction in the code. I have tested it at 50ms for the past month with no problems. All the DBC and CAN dumps I have looked at indicate the 50ms is correct. For the context of this PR, I will return it back to 30ms and introduce the 50ms in another PR that aligns with it.

Software/src/battery/TESLA-BATTERY.cpp
josiahhiggs: That makes sense, I can make the changes for that. What would be better, false/True or no/Yes, displayed on the webserver?

dalathegreat: From a user perspective, I think

Error: Not active / Active! 

makes sense. But anything is fine there! I just worry about maintainability :)

dalathegreat (approved):
LGTM! Nice improvement for people wanting more data out of their Tesla batteries!

PR #697, opened 2024-12-23
Crashfix: Restore Async lib, implement Mutex locking

dalathegreat:
### What
This PR restore the library change performed in PR #681

### Why
Changing that library resulted in so many changes, that we could not really troubleshoot an issue on the Stark board related to connectivity. We rollback this change

### How
To get the ESP32 v3.1.0 working, we implement TCP_MUTEX_LOCK(); and TCP_MUTEX_UNLOCK(); functions inside the me-no-dev/AsyncTCP library.


dalathegreat:
Merging this to prepare for 8.0.0 release!

PR #699, opened 2024-12-24
Reword functions names used for CAN communication

lenvm:
### What
This PR updates the function names used for CAN communication, as it was confusing what each function was used for. For instance the difference between `send_can` and `transmit_can` was not clear.

It also adds a new function `receive_can`, to simplify the code used in `Software.ino`.

### Why
This change should make the code easier to read and easier to maintain.

### How
It changes the function names used for CAN communication, for instance:
- `receive_can` becomes `map_can_frame_to_variable`
- `receive_can_addon` becomes `receive_frame_can_addon`
- `transmit_can` becomes `transmit_can_frame`
- `send_can` becomes `transmit_can`

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -214,13 +214,7 @@ void core_loop(void* task_time_us) {
 #endif
 
     // Input, Runs as fast as possible
-    receive_can_native();  // Receive CAN messages from native CAN port
-#ifdef CANFD_ADDON
-    receive_canfd_addon();  // Receive CAN-FD messages.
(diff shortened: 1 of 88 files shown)


dalathegreat (approved):
LGTM! Only pet peeve would be the wording on "map_can_frame_to_variable_battery", we dont necessarily use all incoming frames, and on some we dont even use the content, we just perform some other function once the message is incoming. For instance "handle_incoming_can_frame_battery" would be closer to the truth.

PR #700, opened 2024-12-24
Make MCP2515 and MCP2518FD (MCP2517) work simultaneously

lenvm:
### What
This PR makes the CAN add-on and CANFD add-on boards work simultaneously.

### Why
It does this because for some hardware setups it is useful to be able to use both an MCP2515 and MCP2518FD.

### How
It does this by making the code for MCP2515 and MCP2517 unique, such that they code does not conflict.


dalathegreat (approved):
Good improvement, thanks!

Issue #565, opened 2024-10-28
Tesla battery contactor close control before safety is checked


frasersimpson83:
### Expected Behavior
Just when discussing this with other users in the Tesla thread, we realised that the emulator code will just send the required CAN commands to close contactors regardless of whether or not the safety boundries in the software are met.

eg cell deviation too high or low, cell voltage too high or low etc.

We thought that maybe it would make more sense for the Emulator to sit in a 'checking' state to wait for cell voltages (I know this can take a minute) and once it's checked all that to make sure it's within bounds then it can start sending CAN packets to close contactors.

That seems to make more sense.    As long as it shows on the web interface it's just waiting to check out data before starting battery contactor close that seems to be safer in my mind and I assume not too hard to implement as it's all there just flipping it around to check before sending CAN instead of after.

josiahhiggs:
I'm happy to have a look into this. @dalathegreat, do you have any examples of where this code is best to modify?

dalathegreat:
@josiahhiggs , easy fix, just add a delay on the start of the send_can function, so that the function returns right away if not enough time has passed 

Somewhere here: https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/src/battery/TESLA-BATTERY.cpp#L1199

josiahhiggs:
I have attempted to make the recommended changes and PR 702 is now up for review.

PR #702, opened 2024-12-26
Enhancement/tesla safety check

josiahhiggs:
### What
This PR changes the init cell voltage values from 3700mV to 3300mV to stop false under/over cell voltage events after reboot of BE. See issue #565 Tesla battery contactor close control before safety is checked.

### Why
Why does it do it? Tesla users have noted cell voltage over and/or under voltage events that log after reboot. This is set due to Tesla allowing contactors to close but the BE has not read cell voltages yet and safety checks set events.

### How
How does it do it? In the TESLA-BATTERY.cpp file the default value for cell voltages was originally set to 3700mV. That is the value that the Emulator uses before actual mapping of CAN data. This has now been adjusted to 3300mV to be inline with the LFP battery voltages.

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -68,7 +68,7 @@ CAN_frame SOLAX_187E = {.FD = false,  //Needed for Ultra
                         .ext_ID = true,
                         .DLC = 8,
                         .ID = 0x187E,
-                        .data = {0x0, 0x2D, 0x0, 0x0, 0x0, 0x5F, 0x0, 0x0}};
+                        .data = {0x60, 0xEA, 0x0, 0x0, 0x64, 0x0, 0x0, 0x0}};
 CAN_frame SOLAX_187D = {.FD = false,  //Needed for Ultra
                         .ext_ID = true,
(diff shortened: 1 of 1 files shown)


josiahhiggs:
@dalathegreat just letting you know that this is not working like I thought it would, seems to cause the program to stop completely for the delay. I'm outside of my knowledge at this point and can't find any examples to reference. Any pointers would be appreciated.

josiahhiggs:
Made some progress on this one by adjusting the init cell voltage! Now we just need to stop the contactors from cycling on reboot.

dalathegreat:
@josiahhiggs 
You can block the startup contactor closing right away, by adding an if check at startup right when the send_can_battery(); function begins.
```
if (currentMillis < INTERVAL_3_S) {
 return;
}
```

However this has a drawback, the millis will overflow after 49days and cause a 3s hitch in the CAN sending routine.

A more robust solution would be to signal in the 0x332 message that all cellvoltages have been read, for instance setting a flag (cellvoltagesRead = true), and then in the send can function check if this is set

```
if (!cellvoltagesRead ) {
 return; //All cellvoltages not read yet, do not proceed with contactor closing
}
```

PR #703, opened 2024-12-26
Indicate that SMA inverter allows contactor closing via LED

lenvm:
### What
This PR implements turning on/off an LED pin to indicate that SMA inverter allows/does not allow contactor closing, respectively.

To make use of the implemented code, a GPIO pin must be defined in the hardware file used. For instance
`#define INVERTER_CONTACTOR_ENABLE_LED_PIN GPIO_NUM_2` can be defined in `hw_liligo.h` when using one of the SD card pins for the LED.

### Why
This PR is implemented to get visual information if the SMA inverter request to close the contactors is sent and recognized up by the battery emulator.

### How
The code checks whether the SMA inverter allows contactor closing, and turns on an LED when the inverter allows contactor closing, and turns off the LED when the inverter does not allow contactor closing.

### DIFF
--- a/Software/src/inverter/BYD-SMA.cpp
+++ b/Software/src/inverter/BYD-SMA.cpp
@@ -130,10 +130,18 @@ void update_values_can_inverter() {  //This function maps all the values fetched
   }
 
   //Error bits
-  if (!datalayer.system.status.inverter_allows_contactor_closing) {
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):
Looking forward to seeing the HAL for STM32 DEV board!

Issue #86, opened 2023-11-11
Leaf 62kWh, SOC and GIDS freezes


StarkJohan:
On my Leaf 62kWh battery, SOC and GIDS shows very strange values after a few days of running grid support services through Checkwatt. SOC/GIDS is suddenly **not affected by slight charge/discharge** (1A) while battery and cell voltage remains correct and increases with low rate charge/discharge as well as larger loads. SOC and GIDS only changes if charge/discharge rate is significant.

In my setup this leads to SOC/GIDS of around 65% while the true charge of the battery is at the full 96%. It also means the true SOC can be pushed outside the software set limits. Bad news on more than one level.

This issue likely occurs because of the way Checkwatt controls the battery SOC with Solax inverters. They always aim for 65% which means the battery is constantly changing between charge/discharge at a very low level to stay at 65% SOC (at the inverter). The leaf BMS does not like this and bugs out after a few days. A reboot of the BMS resets the SOC/GIDS values but might put the true SOC outside the configured limits which is not good.

We need to improve software checks to make this, and similar scenarios impossible. Some basic checks are in place. This will avoid bad battery failures but does not truly solve the problem as the battery will be put in fault mode until manually rebooted. 
When installed in a car the BMS power is cut for two minutes every 24 hours. 

A simple timer or otherwise controlled 12V input could reboot the BMS on a certain interval, e.g. break 12V power 1 minute every 24h. The software needs to handle this scenario correctly to make sure everything continues to operate after a reboot. I have not yet properly tested this approach but it should work without any large code additions.
I would like an integrated solution much like the contactor control. If using contactor control and dual can the lilygo is out of available pins. If dual can is not needed, I suggest "borrowing" on of those pins. 
I'm working on an all-in-one hardware solution to replace the lilygo and I intend to make room for another 12V "channel" in addition to the three needed for contactor control.

_The fault mode could be delayed with a software mod to make the 64-65-66% percent interval to be virtually larger but this is a very specific addition and most likely not useful in any other setups._

_Values from battery
Real SOC%: 53% , GIDS: 355 (x77Wh) , Battery gen: 2 , Max cell voltage: 4152mV , Min cell voltage: 4119mV , Cell deviation: 33mV Values going to inverter
SOH%: 94% , SOC% scaled: 42% , Voltage: 396V , Max discharge power: 30000W , Max charge power: 11000W , Max temp: 22°C , Min temp: 21°C 
[...]

PR #706, opened 2024-12-29
Feature: Add periodic BMS power reset

dalathegreat:
### What
This PR implements a periodic BMS reset. It achieves this by toggling a GPIO pin every 24h. This 3.3V pin can be used to control the 12V power going into the BMS via an SSR or other type of relay.

### Why
Some automotive BMSes like the Nissan LEAF BMS does not work well when powered on continously. According to OEM specs, a powercycle is required every 24hours to keep the BMS working as intended. Failure to reset the BMS can cause SOC% to drift overtime, as reported in #86 

### How
We add a new user configurable feature in the USER_SETTINGS, called `PERIODIC_BMS_RESET`. When this is enabled, a GPIO pin (See the HAL file for your hardware to see which pin), will be toggled ON upon start, and after 24 hours it will turn off for 30s. When the power is cut to this pin, the emulator is put into Paused state, meaning no CAN messages will be sent towards the battery, and the charge/discharge limits will go to 0W. After the 30 seconds have passed, power is turned back on to the pin, and the emulator is unpaused.

### DIFF
--- a/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
+++ b/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
@@ -39,6 +39,11 @@ unsigned long negativeStartTime = 0;
 unsigned long prechargeCompletedTime = 0;
 unsigned long timeSpentInFaultedMode = 0;
 #endif
+unsigned long currentTime = 0;
+unsigned long lastPowerRemovalTime = 0;
+const unsigned long powerRemovalInterval = 24 * 60 * 60 * 1000;  // 24 hours in milliseconds
+const unsigned long powerRemovalDuration = 30000;                // 30 seconds in milliseconds
+bool isBMSResetActive = false;
 
 void set(uint8_t pin, bool direction, uint32_t pwm_freq = 0xFFFF) {
 #ifdef PWM_CONTACTOR_CONTROL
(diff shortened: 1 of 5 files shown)


No-Signal:
@dalathegreat it would be great to have an option to trigger this remotely via MQTT or an HTTP post as well as a timer.

In my case I'd like to restart the BMS during the overnight window of cheap electric where I normally charge the batteries and expect others would like to do similar.

Let me know what you think and if you want to get the current version out first as happy to submit a PR.

dalathegreat:
Sounds good @No-Signal , we can do it in a PR after this is in. At the same time, forcing a balance charge will also be something that will be nice to have controllable via MQTT, as implemented in PR #718

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -202,3 +213,39 @@ void handle_contactors_battery2() {
...
+  currentTime = millis();
+
+  // Check if 24 hours have passed since the last power removal
+  if (currentTime - lastPowerRemovalTime >= powerRemovalInterval && !isBMSResetActive) {
+    lastPowerRemovalTime = currentTime;  // Record the time when BMS reset was started
+
+    // Set emulator state to paused (Max Charge/Discharge = 0 & CAN = stop)
+    // TODO: We try to keep contactors engaged during this pause, and just ramp power down to 0.
+    // If this turns out to not work properly, set also the third option to true to open contactors
+    setBatteryPause(true, true, false, false);
```
mvgalen, line 233: This might not be ideal for batteries with internal contactors like MEB. They need to be opened before shutting down the BMS. I should look into what the current behavior is, equipment stop may already gracefully open the contactors.

dalathegreat: Is there any indication that MEB needs periodic reset? Mainly this whole PR is only based on LEAF batteries needing this

Software/src/devboard/hal/hw_lilygo.h
```diff
@@ -51,6 +51,7 @@
 #define POSITIVE_CONTACTOR_PIN 32
 #define NEGATIVE_CONTACTOR_PIN 33
 #define PRECHARGE_PIN 25
+#define BMS_POWER 18  // Note, this pin collides with CAN add-ons and Chademo
```
mvgalen, line 54: Wouldn't it be nice to change it depending on if CAN_ADDON or CONTACTOR_CONTROL are defined? That would mean it would not always be the same pin, but would make is available in more scenario's.

mvgalen (changes requested):
Added two comments, the rest looks good to me.

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
mvgalen: I don't know yet. Lets keep this section as it is for now.

Software/src/devboard/hal/hw_lilygo.h
dalathegreat: Yes, BUT, since the GPIO structure is getting quite complicated on the LilyGo, I personally think it is better that the end user assumes some planning responsibility here, and manages it to another pin if they have some special combination. For the LEAF (which this feature is for), the pins dont collide with Automatic contactor control, which is the most important factor.

mvgalen: Ok, makes sense. I do think it would be best if we dictate (via ifdefs) the possible options, but let's wait until somebody asks for it.

mvgalen (approved):
LGTM.

ruffieuxh:
Cool! Many thanks!
Got a question: How do I see, if the feature works without standing by the battery when it happens? Is there an event logged?

PR #707, opened 2024-12-29
Adding ability to log CAN frames to SD card

No-Signal:
### What
This PR implements logging of CAN frames to an SD card

### Why
To enable logging of CAN frames over time to debug issues

### How
Implements a ring buffer that CAN frames are pushed onto which are then written to the SD card using the second core to take load off the primary core.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -18,6 +18,7 @@
 #include "src/communication/rs485/comm_rs485.h"
 #include "src/communication/seriallink/comm_seriallink.h"
 #include "src/datalayer/datalayer.h"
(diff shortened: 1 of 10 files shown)


dalathegreat:
Nice work! 

Minor, might be good to add a compile error incase someone tried to use this with any other hardware than Lilygo? Maybe add these lines to \Software\src\include.h

```
#ifdef LOG_CAN_TO_SD
#if !defined(HW_LILYGO)
#error The SD card logging feature is only available on LilyGo hardware
#endif
#endif
```

dalathegreat (approved):

PR #708, opened 2024-12-29
Adding feature to log to SD Card

No-Signal:
### What
This PR implements logging messages to an SD Card where available

### Why
This will help to diagnose issues when an emulator reboots as the log is persisted

### How
This is an extension to the sdcard methods to add a second buffer for logs which are sent from the logger and written to the SD Card on the second CPU core


dalathegreat (approved):
LGTM! 👍

No-Signal (commented):
Changes look great thanks!

PR #713, opened 2024-12-29
Make BMS_POWER hardware agnostic

lenvm:
### What
This PR ensures that `BMS_POWER` can be used by all hardware variants.

While we are at it, this PR also updates a mistake in a comment from `HW_STARK` to `HW_3LB`.

### Why
To ensure that functions are available for all types of hardware.

### How
By updating a define from `#ifdef HW_STARK` to `#ifdef BMS_POWER`.


dalathegreat:
Good change, BUT, this collides a bit with PR #706 
Let's change it there instead?

lenvm:
@dalathegreat, fine by me! Please implement it in #706!

PR #714, opened 2024-12-29
Bugfix: ensure code compiles for hardware that does not have SD card pins

lenvm:
### What
This PR fixes the issue that the code does not compile anymore for hardware that does not have SD card pins defined, since the SD functions were introduced in #707.

### Why
To ensure the code compiles for all hardware types.

### How
By adding `#if defined()` statements around the code that requires the hardware pins to be defined.

### DIFF
--- a/Software/src/devboard/sdcard/sdcard.cpp
+++ b/Software/src/devboard/sdcard/sdcard.cpp
@@ -1,6 +1,9 @@
 #include "sdcard.h"
 #include "freertos/ringbuf.h"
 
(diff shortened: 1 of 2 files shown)


No-Signal (approved):
Looks good

dalathegreat (approved):

dalathegreat:
@lenvm with this nice change, maybe we can remove the #error compilation set in Battery-Emulator\Software\src\include.h?

**#ifdef LOG_CAN_TO_SD**

PR #715, opened 2024-12-29
Add hardware to GitHub Actions workflows

lenvm:
### What
This PR implements automatic builds for the different supported hardware platforms, using GitHub Actions.

### Why
It does this, to ensure that the code compiles for all different types of hardware. This prevents the need for bug fixes such as #714.

### How
A GitHub Action workflow is added to compile the code for different hardware setups. This check is done on each push and each pull request.

### DIFF
--- a/.github/workflows/compile-all-batteries.yml
+++ b/.github/workflows/compile-all-batteries.yml
@@ -79,6 +79,10 @@ jobs:
         # These are the emulated inverter communication protocols for which the code will be compiled.
         inverter:
           - BYD_CAN
+        # These are the supported hardware platforms for which the code will be compiled.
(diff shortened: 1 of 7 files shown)


dalathegreat:
@lenvm 
Now that by default the LilyGo HW is not selected, we could make the compilation #error a bit easier to understand. I suggest changing the Battery-Emulator/Software/src/include.h

`#error You must select a HW to run on! `

into

`#error You must select a target hardware in the USER_SETTINGS.h file! `

lenvm:
Good idea! Implemented in 58c15c5

dalathegreat (approved):

PR #716, opened 2024-12-30
Enhancement/tesla safety check

josiahhiggs:
### What
This PR changes the init cell voltage values from 3700mV to 3300mV to stop false under/over cell voltage events after reboot of BE. See issue https://github.com/dalathegreat/Battery-Emulator/issues/565 Tesla battery contactor close control before safety is checked.

### Why
Why does it do it? Tesla users have noted cell voltage over and/or under voltage events that log after reboot. This is set due to Tesla allowing contactors to close but the BE has not read cell voltages yet and safety checks set events.

### How
How does it do it? In the TESLA-BATTERY.cpp file the default value for cell voltages was originally set to 3700mV. That is the value that the Emulator uses before actual mapping of CAN data. This has now been adjusted to 3300mV to be inline with the LFP battery voltages. Also added code recommended by @dalathegreat in old PR 702, cellvoltagesRead before sending CAN to close contactors. Updated the send cycle time to 50 MS as DBC files indicate. Old PR 702 was closed due to merge conflicts I couldn't figure out how to handle.


Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -2629,6 +2633,10 @@ the first, for a few cycles, then stop all  messages which causes the contactor
 
   unsigned long currentMillis = millis();
 
+  if (!cellvoltagesRead) {
```
dalathegreat, line 2636: One thing, do we have any Tesla batteries that DONT send the 0x332 message? Reason I ask, is if there is, we end up in a situation where contactors never close, and without any indication whatsoever why. 

If this is a real worry, we could add an event incase cellvoltages haven't been read for over a minute?

dalathegreat (approved):

PR #718, opened 2024-12-30
Feature: Add forced Tesla LFP balancing functionality

dalathegreat:
### What
This PR implements forced balancing functionality for Tesla LFP batteries (Model 3/Y)

### Why
To help keep LFP batteries balanced outside of a Tesla vehicle. This mimics more how the batteries would be commanded in a vehicle.

### How
The forced balancing mode can be enabled via the Webserver. It is recommended to run this once per month on Tesla LFP batteries. **Before enabling, make sure battery is close to full, and you have enough sun or grid power to continue charging and forcing the balance!**

When enabling the `Manual LFP balancing` , the balancing runs for the specified amount of time, set in `Balancing max time`. The balancing parameters, Balancing float power , Max battery voltage , Max cell voltage and Max cell voltage deviation can also be adjusted from here.

image

### DIFF
--- a/Software/src/battery/TESLA-BATTERY.cpp
+++ b/Software/src/battery/TESLA-BATTERY.cpp
@@ -904,6 +904,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   // During forced balancing request via webserver, we allow the battery to exceed normal safety parameters
   if (datalayer.battery.settings.user_requests_balancing) {
+    datalayer.battery.status.real_soc = 9900;  //Force battery to show up as 99% when balancing
     datalayer.battery.info.max_design_voltage_dV = datalayer.battery.settings.balancing_max_pack_voltage_dV;
     datalayer.battery.info.max_cell_voltage_mV = datalayer.battery.settings.balancing_max_cell_voltage_mV;
     datalayer.battery.info.max_cell_voltage_deviation_mV =
--- a/Software/src/battery/TESLA-BATTERY.cpp
+++ b/Software/src/battery/TESLA-BATTERY.cpp
@@ -901,6 +901,15 @@ void update_values_battery() {  //This function maps all the values fetched via 
     datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_NCA_NCM;
     datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_NCA_NCM;
   }
+
+  // During forced balancing request via webserver, we allow the battery to exceed normal safety parameters
+  if (datalayer.battery.settings.user_requests_balancing) {
+    datalayer.battery.info.max_design_voltage_dV = datalayer.battery.settings.balancing_max_pack_voltage_dV;
+    datalayer.battery.info.max_cell_voltage_mV = datalayer.battery.settings.balancing_max_cell_voltage_mV;
+    datalayer.battery.info.max_cell_voltage_deviation_mV =
+        datalayer.battery.settings.balancing_max_deviation_cell_voltage_mV;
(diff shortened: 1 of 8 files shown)


Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -901,6 +901,16 @@ void update_values_battery() {  //This function maps all the values fetched via
     datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_NCA_NCM;
     datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_NCA_NCM;
   }
+
+  // During forced balancing request via webserver, we allow the battery to exceed normal safety parameters
+  if (datalayer.battery.settings.user_requests_balancing) {
+    datalayer.battery.status.real_soc = 9900;  //Force battery to show up as 99% when balancing
```
mvgalen, line 907: Why is the soc forced to 99%? As far as I can see, it is not used programmatically?

Software/src/devboard/safety/safety.cpp
```diff
@@ -238,6 +238,22 @@ void update_machineryprotection() {
...
+    if (datalayer.battery.settings.balancing_start_time_ms == 0) {
+      datalayer.battery.settings.balancing_start_time_ms = millis();
+      //TODO, raise info event? Forced balancing starting!
+    }
+
+    // Check if the elapsed time exceeds the balancing time
+    if (millis() - datalayer.battery.settings.balancing_start_time_ms >= datalayer.battery.settings.balancing_time_ms) {
+      datalayer.battery.settings.user_requests_balancing = false;
+      datalayer.battery.settings.balancing_start_time_ms = 0;  // Reset the start time
+      //TODO, raise info event? Balancing time elapsed. Turning off...
```
mvgalen, line 254: perhaps a logging.println("TESLA: Balancing time finished"); ?

Software/src/battery/TESLA-BATTERY.cpp
dalathegreat: Incase the battery shows 100%, charging would stop in the Safety.cpp file. So we need to ensure it does not.

Could also be solved with an if == 100, then 99 check

mvgalen: Wouldn't it make more sense to have the safety check take the user_requests_balancing into account? And have safety then also look at he max charge power for instance? Only if the combination is satisfied may charging continue.

Software/src/devboard/safety/safety.cpp
dalathegreat: Think Event will make more sense, due to this might being triggered via MQTT and will be hard to see when it started/completed otherwise. Tesla code has so much on the logging already!

But let's leave this for future development, seems like we need to merge this to .dev to get more feedback on actual behaviour

Software/src/battery/TESLA-BATTERY.cpp
dalathegreat: Let's leave this for future development, seems like we need to merge this to .dev to get more feedback on actual behaviour

josiahhiggs:
The MAX_XELL_DEVIATION_LFP needs to be raised to 400mV to stop cell deviation errors from being set after the Manual session has ended. Also the MAX_CELL_VOLTAGE_LFP needs to be raised to at least 3650mv to stop cell overvoltage errors.

Software/src/devboard/safety/safety.cpp
dalathegreat: I added event in [a60a1c3](https://github.com/dalathegreat/Battery-Emulator/pull/718/commits/a60a1c3b5c8c7d107c9ae61ad4147b699d4462c9)

PR #722, opened 2024-12-31
Improvement: Remove Tesla watchdog spamming 

dalathegreat:
### What
This PR makes the Tesla debug messages a bit more manageable

### Why
During normal operation of the battery, the Tesla code spams the following error codes every second:

```
       1.000 ERROR: The processor has experienced a reset due to watchdog reset
       1.000 ERROR: BMS_a035_SW_Isolation
       1.000 ERROR: BMS_a036_SW_HvpHvilFault
       1.000 ERROR: BMS_a088_SW_VcFront_MIA_InDrive
       1.000 ERROR: BMS_a090_SW_Gateway_MIA
       1.000 ERROR: BMS_a091_SW_ChargePort_MIA
       1.000 ERROR: BMS_a092_SW_ChargePort_Mia_On_Hv
       1.000 ERROR: BMS_a094_SW_Drive_Inverter_MIA
       1.000 ERROR: BMS_a170_SW_Limp_Mode
```

The MIA messages are good info for further development, hinting towards CAN messages from these units are missing. We should improve the software by sending more CAN towards the battery.

However, the first message about the Watchdog is quite unnecessary, and causes users to incorrectly report this as a severe issue. The Tesla battery operates perfectly OK, so spamming about this watchdog every seconds really does not help the debug log at all.

### How
We uncomment the watchdog message for now. Further development should focus on sending more messages towards the battery to make it happier.

### DIFF
--- a/Software/src/battery/TESLA-BATTERY.cpp
+++ b/Software/src/battery/TESLA-BATTERY.cpp
@@ -2732,7 +2732,7 @@ void printFaultCodesIfActive() {
   }
   // Check each symbol and print debug information if its value is 1
   // 0X3AA: 938 HVP_alertMatrix1
-  printDebugIfActive(battery_WatchdogReset, "ERROR: The processor has experienced a reset due to watchdog reset");
+  //printDebugIfActive(battery_WatchdogReset, "ERROR: The processor has experienced a reset due to watchdog reset"); //Uncommented due to not affecting usage
   printDebugIfActive(battery_PowerLossReset, "ERROR: The processor has experienced a reset due to power loss");
   printDebugIfActive(battery_SwAssertion, "ERROR: An internal software assertion has failed");
   printDebugIfActive(battery_CrashEvent, "ERROR: crash signal is detected by HVP");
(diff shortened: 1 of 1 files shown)


Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -2732,7 +2732,7 @@ void printFaultCodesIfActive() {
   }
   // Check each symbol and print debug information if its value is 1
   // 0X3AA: 938 HVP_alertMatrix1
-  printDebugIfActive(battery_WatchdogReset, "ERROR: The processor has experienced a reset due to watchdog reset");
+  //printDebugIfActive(battery_WatchdogReset, "ERROR: The processor has experienced a reset due to watchdog reset"); //Uncommented due to not affecting usage
```
levibutler, line 2735: should the comment read `Commented due to not affecting usage`?
Is it worth keeping this line commented vs removing? or replacing with something like `battery_WatchdogReset debug log removed pending further development to send additional messages to the battery. PR #722`

levibutler (approved):
approved with comment

Issue #705, opened 2024-12-28
Cellmonitor not showing data for LEAF ZE0 with version 8.dev


mahoekst:
### Expected Behavior

Expect cellmonitor to propagate data after a few minutes

### Actual Behavior

Webpage stays empty.

### Steps to Reproduce the Problem

Connect battery only, configure settings for LEAF battery only (no inverter). Startup and wait a few minutes, screen stays unpopulatad.
Switched back to 7.9.2 and the data appeared after a few minutes.

### Settings

- Software version: `8.dev downloaded 28-12-2024'
- Battery used: `LEAF ZE0`
- Inverter communication protocol: ``
- Hardware used for Battery-Emulator: `HW_LILYGO`
- CONTACTOR_CONTROL: `no`
- CAN_ADDON: `no`
- WEBSERVER: `yes`
- MQTT: `no`

dalathegreat:
Attempted fix for this is in the PR #731

PR #731, opened 2025-01-02
Regression: LEAF, Speedup polling, fix requests for cellvoltage

dalathegreat:
### What
This PR fixes the issue reported in #705 

### Why
So we can read cellvoltages again

### How
At the same time, we speedup the handling. Instead of waiting 100seconds before starting to poll, we now wait 20 seconds.


No-Signal:
Tested working on a dual LEAF battery setup

Issue #723, opened 2024-12-31
8.0.dev: Web Logging causes Panic crash


dalathegreat:
### Expected Behavior

DEBUG_VIA_WEB displays debug messages

### Actual Behavior

DEBUG_VIA_WEB causes board to panic reset with the Event RESET_PANIC stored

### Steps to Reproduce the Problem

Enable the following configuration options

1. DEBUG_VIA_WEB
2. TESLA_MODEL_3Y_BATTERY
3. FOXESS_CAN

Flash board and browse to Log page.

Refresh page, it corrupts and crashes the board

![image](https://github.com/user-attachments/assets/d6fb7f9d-771a-4463-a521-d581650b8931)

### Settings

- Software version: `8.0.dev`
- Battery used: `TESLA_MODEL_3Y_BATTERY`
- Inverter communication protocol: `FOXESS_CAN`
- Hardware used for Battery-Emulator: `HW_STARK`
- CONTACTOR_CONTROL: `no`
- CAN_ADDON: `no`
- WEBSERVER: `yes`
- MQTT: `no`

PR #733, opened 2025-01-02
Enhancement/templating fix

mvgalen:
### What
This PR fixes issue https://github.com/dalathegreat/Battery-Emulator/issues/723

### Why
% characters were crashing the templating engine in the async web server component.

### How
For the log pages we now bypass the templating engine. Which removes the need for a template character, and thus does not limit the characters that can appear in the log file.
This also reverts PR https://github.com/dalathegreat/Battery-Emulator/pull/724 since those changes are no longer needed.


dalathegreat (approved):
LGTM :+1:

PR #734, opened 2025-01-02
Slow down polling of ParameterPolling KIA / Hyundai 

LevinSwe:
### What
Slow down ParameterPolling of Kia / Hyundai battery

### Why
Sometimes the CellVoltages are corrupted due to high CAN-load / CPU load

### How
Polling of Parameter ID's polled every other time instead of each 100ms cycle.

### DIFF
--- a/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
+++ b/Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
@@ -264,13 +264,13 @@ void handle_incoming_can_frame_battery(CAN_frame rx_frame) {
       startedUp = true;
 
       //PID data is polled after last message sent from battery every other time:
-      if (poll_data_pid >= 10) {  //polling one of ten PIDs at 100ms*2, resolution = 2s
-        poll_data_pid = 0;
-      }
       if (holdPidCounter == true) {
         holdPidCounter = false;
       } else {
         holdPidCounter = true;
+        if (poll_data_pid >= 6) {  //polling one of six PIDs at 100ms*2, resolution = 1200ms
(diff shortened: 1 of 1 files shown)


Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
```diff
@@ -262,27 +263,32 @@ void handle_incoming_can_frame_battery(CAN_frame rx_frame) {
...
+          transmit_can_frame(&KIA64_7E4_id2, can_config.battery);
+        } else if (poll_data_pid == 3) {
+          transmit_can_frame(&KIA64_7E4_id3, can_config.battery);
+        } else if (poll_data_pid == 4) {
+          transmit_can_frame(&KIA64_7E4_id4, can_config.battery);
+        } else if (poll_data_pid == 5) {
+          transmit_can_frame(&KIA64_7E4_id5, can_config.battery);
+        } else if (poll_data_pid == 6) {
+          transmit_can_frame(&KIA64_7E4_id6, can_config.battery);
+        } else if (poll_data_pid == 7) {
```
dalathegreat, line 287: Since this slows down reading slightly, we should remove the unused 7-10 pid increments. Can you rewrite this to only handle 6 pids?

dalathegreat (changes requested):

Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
mvgalen: Change look good to me.

mvgalen (approved):
LGTM.

dalathegreat (approved):

PR #739, opened 2025-01-03
Bugfix: Improve SOLAX can performance

dalathegreat:
### What
This PR makes the Solax protocol more gentle on the CAN bus load

### Why
To avoid overflowing the send buffer

### How
Instead of batch sending 9x messages, we send each message with a 2ms delay inbetween.

At the same time this PR does some general cleanup of the Solax implementation


dalathegreat:
Closing this one, bad idea

PR #741, opened 2025-01-03
3LB: Add support for BMS2 power

dalathegreat:
### What
This PR implements support for running 2x batteries, with dedicated BMS2 power pin. The 3LB hardware is the first board that has this option defined

### Why
To improve support for using 2x LEAF batteries at the same time

### How
If the `BMS_2_POWER` pin is defined, we also use it


mvgalen (approved):
LGTM.

PR #742, opened 2025-01-03
Fix bugs in printf logging code & add version to log

mvgalen:
### What
Fix bugs in printf logging code & add Battery Emulator version and build date/time to log

### Why
Fix the bug and being able to see the exact build version that is running

### How
Fixed the code. Added printf containing version and date/time.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -97,6 +97,10 @@ Logging logging;
 void setup() {
   init_serial();
 
+  // We print this after setting up serial, such that is also printed to serial with DEBUG_VIA_USB set.
+  logging.printf("Battery emulator %s build " __DATE__
+                 " " __TIME__ "\n", version_number);
+
   init_stored_settings();
 
 #ifdef WIFI
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -98,8 +98,7 @@ void setup() {
   init_serial();
 
   // We print this after setting up serial, such that is also printed to serial with DEBUG_VIA_USB set.
-  logging.printf("Battery emulator %s build " __DATE__
-                 " " __TIME__ "\n", version_number);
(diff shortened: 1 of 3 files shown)


Software/src/devboard/utils/logging.h
```diff
@@ -4,14 +4,21 @@
...
+
 class Logging : public Print {
   void add_timestamp(size_t size);
 
  public:
   virtual size_t write(const uint8_t* buffer, size_t size);
   virtual size_t write(uint8_t) { return 0; }
   void printf(const char* fmt, ...);
-  Logging() {}
+  Logging() {
```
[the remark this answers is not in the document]
mvgalen: I'll add a space tomorrow and see if anything can be done to make sure the build date/time also ends up in the serial channel.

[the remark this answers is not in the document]
mvgalen: Done.

Software/Software.ino
```diff
@@ -97,6 +97,9 @@ Logging logging;
 void setup() {
   init_serial();
 
+  // We print this after setting up serial, such that is also printed to serial with DEBUG_VIA_USB set.
```
No-Signal, line 100: Good improvement for diagnostics, if we were to move the following 2 lines from the logging loop it could also log to sd card where enabled. This feature isn't likely widely used right now and only works on the lilygo so may not be important.

  init_logging_buffers();
  init_sdcard();

No-Signal (approved):
Approved on the assumption that SD logging of the version isn't needed

Software/Software.ino
mvgalen: Yeah, might bu usefull, but maybe in another PR?

No-Signal: Agreed, no need to delay this PR

dalathegreat (approved):

PR #745, opened 2025-01-03
Kia64: Add sanity checking for last cellvoltages

dalathegreat:
### What
This PR implements a sanity check on the last cellvoltages for Kia64/40kWh packs

### Why
Some 90S packs incorrectly report some garbage on the unpopulated cells. Attached is a screenshot from a Kia 40kWh 375A0 - K4403
image

### How
We check the cells above 85. If the value seems sus, we zero it


mvgalen (approved):
LGTM.

PR #746, opened 2025-01-03
Adding ability to remotely trigger a BMS reset via MQTT

No-Signal:
### What
This PR implements the ability to trigger a BMS reset remotely via MQTT. It also puts the groundwork in place for further MQTT messages in the future.

### Why
This allows you to decide when to run the BMS reset for example ensuring it runs over night during cheap electric rates vs during peak rates.

### How
The BMS reset logic is expanded to support both a timed and a triggered version. If remote is enabled an MQTT message can trigger the reset. If HA auto discovery is enabled it adds a button to trigger the reset.


dalathegreat (approved):
Nice work! :raised_hands:

PR #749, opened 2025-01-04
Bugfix: Make Volvo/Polestar use BMS allowed charge/discharge values

dalathegreat:
### What
This PR allows for safer operation of Volvo/Polestar SPA batteries

### Why
To not damage the battery while charging. Before this PR, allowed charge/discharge was locked at 30kW, even at 99%!

### How
We now use the Charge/Discharge power limit values that the BMS sends. Example of values at 95.53%SOC

image


dalathegreat:
Confirmed working by @GoSmartEV over on the Discord :partying_face: 

image

PR #751, opened 2025-01-04
Improvement: Move BYD startup messages to 10s block

dalathegreat:
### What
This PR adds a delay on the initial data sending we do on the BYD CAN protocol

### Why
Instead of sending the messages the first 1ms after CPU starts up, we wait 10 seconds until we fire this batch send.

### How
In theory this should help some startup race conditions observed by users

### DIFF
--- a/Software/src/inverter/BYD-CAN.cpp
+++ b/Software/src/inverter/BYD-CAN.cpp
@@ -83,7 +83,7 @@ static int16_t inverter_temperature = 0;
 static uint16_t remaining_capacity_ah = 0;
 static uint16_t fully_charged_capacity_ah = 0;
 static long inverter_timestamp = 0;
-static bool initialDataSent = 0;
+static bool initialDataSent = false;
 
 void update_values_can_inverter() {  //This function maps all the values fetched from battery CAN to the correct CAN messages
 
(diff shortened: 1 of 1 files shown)


GoSmartEV:
10 seconds is too short, as the battery in a Volvo, for example, only starts after 10 seconds. Then the inverter takes about 10-20 seconds to begin sending data over the CAN. Please increase this time to 60 seconds.

dalathegreat:
@GoSmartEV , changed to 60 seconds in commit [b81e0f5](https://github.com/dalathegreat/Battery-Emulator/pull/751/commits/b81e0f5647874651605c0610c5e15d10dcdcfa16)

GoSmartEV:
> @GoSmartEV , changed to 60 seconds in commit [b81e0f5](https://github.com/dalathegreat/Battery-Emulator/pull/751/commits/b81e0f5647874651605c0610c5e15d10dcdcfa16)


It doesn’t recognize it in the right way the battery and invertor 


IMG_5791
IMG_5792

dalathegreat:
Hmm, not really sure how to solve this. If this PR did not help anything, I suggest we close it as wont do for now, and investigate your startup issue more in detail

GoSmartEV:
> Hmm, not really sure how to solve this. If this PR did not help anything, I suggest we close it as wont do for now, and investigate your startup issue more in detail

the program does not correctly determine where the battery and where the inverter is

But the CAN message are at the correct

GoSmartEV:
IMG_5912
IMG_5913

dalathegreat:
So the problem is most likely related to OFFGRID use, incase the Inverter is not powered on, the messages sent towards it will result in the board going into CAN_TX_FAILURE.

We need to add an OFFGRID toggle, where the code ignores total CAN failures and properly allows the system to start

GoSmartEV:
> So the problem is most likely related to OFFGRID use, incase the Inverter is not powered on, the messages sent towards it will result in the board going into CAN_TX_FAILURE.
> 
> We need to add an OFFGRID toggle, where the code ignores total CAN failures and properly allows the system to start

Yes, this helps, and yes, it’s working for off grid 👌🏻

PR #753, opened 2025-01-05
Improvement: Update CANFD library with 2517 compat toggle

dalathegreat:
### What
This PR adds the suggested changes to the pierremolinaro/acan2517FD library as suggested [in this PR](https://github.com/pierremolinaro/acan2517FD/pull/34)

### Why
There is a performance gain, disabling 2517FD compat mode, it no longer disables interrupts during SPI transactions which slightly increases performance and avoids delaying interrupts. There should be no downside to enabling this, since [our Wiki](https://github.com/dalathegreat/Battery-Emulator/wiki/CAN%E2%80%90FD-add%E2%80%90on-(MCP2518FD)#hardware) suggests using the 251**8**FD chip instead.

### How
Disabling the MCP2517 compatibility mode can be done in ACAN2517FD.h by uncommenting the `#define DISABLEMCP2517FDCOMPAT` line. By default we keep this disabled until someone has had time to test it properly on a heavy bus load CAN-FD setup.


mvgalen (approved):
LGTM.
Will test it on my setup and report results.

PR #754, opened 2025-01-06
New battery: Add support for Sono Motors Battery

dalathegreat:
### What
This PR implements support for the elusive 54kWh LFP battery, found in the Sono Motors Sion pre-production vehicles

### Why
Feature requested by Patreon supporter!

### How
Initial support for the Sion added in this PR. Fault codes can be mapped later on for easier debugging.


dalathegreat:
Merging this. Note that this will NOT be added to build system, due to the rarity of these batteries.

PR #756, opened 2025-01-06
Improvement: Remove MCP2517 support for code speedup

dalathegreat:
### What
This PR enabled the optimization developed in PR https://github.com/dalathegreat/Battery-Emulator/pull/753

### Why
We drop MCP2517 support in favour of the newer MCP2518. By dropping the support we get improved CAN performance

### How
See PR https://github.com/dalathegreat/Battery-Emulator/pull/753 for more info. This PR just enabled the speedup


dalathegreat:
@mvgalen We could do that, but since I checked the Wiki and all hardware we use, there are no boards with MCP2517 floating around, so maybe better to just add it to the Wiki that if you really want to have a bad time with the MCP2517, you can go into this .h file and enable the hardware workaround?

mvgalen (approved):
If problems arise, we move the define to USER_SETTINGS.

PR #764, opened 2025-01-07
Feature: Tesla, add isolation clearing function

dalathegreat:
### What
This PR implements isolation fault clearing via the More Battery Info page for Tesla batteries

### Why
Good to have diagnostic feature

### How
Pressing the "Clear Isolation Fault" under the "More Battery Info" page will send diagnostic messages towards the battery

image

CAN messages sent:

image

NOTE: For future implementation, we should instead reply when the UDS messages arrive from battery. This should still work fine, but no feedback from battery is taken into account.


mvgalen (approved):
LGTM.

PR #770, opened 2025-01-08
Improvement: Reduce sketch size

dalathegreat:
### What
This PR implements a few optimizations to reduce sketch and RAM usage

### Why
We are soon running out of sketch space

### How
This PR performs the following optimizations, for a total of -0.2% reduction of sketch usage

- Unused LED state (rainbow) removed from code
- Datatypes optimized in uptime library
- Unused code removed from Events
- Unnecessary text removed from all events
- Unnecessary files removed from the codebase


No-Signal (approved):
Tested locally and all seems to be working

PR #777, opened 2025-01-10
Tesla: Add overdischarge prevention

dalathegreat:
### What
This PR implements an attempt at overdischarge prevention for Tesla

### Why
One user reported an overdischarge event which led to contactor opening. This should not happen, battery should just stop discharge

### How
We now check if we are approaching cellvoltage limit, and if we do, we zero out the allowed discharge power.


dalathegreat:
This PR is now closed, in favour of #782

PR #779, opened 2025-01-11
Removing duplicate start script

No-Signal:
### What
Removes duplicate start script block that breaks the SOH reset feature


dalathegreat (approved):
Appreciate it!

PR #781, opened 2025-01-11
Update VOLVO-SPA-BATTERY.cpp

GoSmartEV:
Correct command for opening and closing contactors

0x140 Byte 8 Bit 0 is DcDcActivationRequest
0x140 Byte 8 Bit 1 is DcDcActivationRequest_Valid (should be 1 to have DcDcActivationRequest bit valid)

So 0x83 on that bytes means that the ECM requests to activate the DC-DC.

But we don’t want to do it , so we should send 0x82


dalathegreat (approved):
Thanks!

PR #782, opened 2025-01-11
Feature: Add CRITICAL cell voltage safeties

dalathegreat:
### What
This PR implements a new type of cellvoltage safeties, CRITICAL cell events.

### Why
Many users have reported back that their packs touched the extreme ends of the allowed voltages, either high or low side. When this happens, we previously entered FAULT state right away, stopped using battery, and opened contactors. This is not good, since undercharging can lead to cellvoltages dropping further, causing hard recoveries.

### How
Now we switch things up a bit. Normal cell under/over voltage events will now stop either charging/discharging, but keep the emulator running in Warning state. This allows the battery to normally recover from this event. As an added safety, if the cells continue to drift towards the dangerzone, at 100mV past the safety, we enter a CRITICAL cellvoltage error state that stops the emulator (FAULT), which opens contactors.

So to summarize, this change makes it easier for people to recover from low/high cells, and allows the emulator to take more corrective action.


wjcloudy (approved):
LGTM

PR #786, opened 2025-01-12
Bugfix: Move Battery2 power calculation outside SOC scaling function

dalathegreat:
### What
This PR fixes a bug related to using 2x batteries without SOC scaling

### Why
Old software only calculated power for battery2 when "SOC scaling" option was turned on

### How
Massive thanks to @pmajor021 for finding and reporting this bug!


dalathegreat:
Merging, tested successfully by pmajor021!

Issue #774, opened 2025-01-09
BMW I3, BMS starting slowly,  Lilygo starting fast


havrla:
In some cases, the battery BMS starts up slower than Lilygo. The BMS does not send valid data about the battery voltage and lilygo is already working with it. Lilygo ends with an error. The emergency solution is to restart Lilygo after the BMS/battery is already working. However, this is not optimal. Is it possible to wait 1-2s when starting up until the BMS in the battery starts up?
Or diferent source problem :-)

Thanks

![Screenshot_20250109_165859_com android chrome](https://github.com/user-attachments/assets/f251aa87-b581-4e1f-8772-4000030df0aa)

dalathegreat:
Thanks for the report, fix is now in this PR: https://github.com/dalathegreat/Battery-Emulator/pull/789

PR #789, opened 2025-01-12
Bugfix: BMW i3, Add sanity check to minmax reading

dalathegreat:
### What
This PR implements a sanity check on the min/max cellvoltage readings

### Why
The BMW i3 BMS can give strange values the first few seconds on a start, causing events to be incorrectly triggered. This has been reported in #774 

### How
Incorrect readings range between 6000-8000mV. We now do a quick sanity check to see if value is below 4500. If it is, we use the value.


wjcloudy (approved):
LGTM

Issue #792, opened 2025-01-14
PAUSE_END phase trigger BATTERY_VALUE, BATTERY_OVERVOLTAGE and SOC_PLAUSABILITY_ERROR


luipir:
### Expected Behavior

Normal work of the emulator wthout triggering any alarm or error

### Actual Behavior

when PAUSE_END is reached, a set of error ar triggered as in the image
![Screenshot_from_2025-01-13_21-56-09](https://github.com/user-attachments/assets/ad3ad027-a776-4b55-9659-56f05831b343)

The error is triggered every day as 18:06 

### Settings

detailed configuration:
https://github.com/luipir/Battery-Emulator/commit/71e3fdefa1db1c0ab3d30dd18d3fff8fa14c62d2

in resume the conf is:
```
#define NISSAN_LEAF_BATTERY

/* Select inverter communication protocol. See Wiki for which to use with your inverter: https://github.com/dalathegreat/BYD-#define BYD_CAN  //Enable this line to emulate a "BYD Battery-Box Premium HVS" over CAN Bus

/* Select hardware used for Battery-Emulator */
#define HW_STARK

/* Contactor settings. If you have a battery that does not activate contactors via CAN, configure this section */
#define PRECHARGE_TIME_MS 500  //Precharge time in milliseconds. Modify to suit your inverter (See wiki for more info)
#define CONTACTOR_CONTROL  //Enable this line to have the emulator handle automatic precharge/contactor+/contactor- closing sequence (See wiki for pins)
#define PWM_CONTACTOR_CONTROL  //Enable this line to use PWM for CONTACTOR_CONTROL, which lowers power consumption and heat generation. CONTACTOR_CONTROL must be enabled.
#define PERIODIC_BMS_RESET  //Enable to have the emulator powercycle the connected battery every 24hours via GPIO. Useful for some batteries like Nissan LEAF

#define CRYSTAL_FREQUENCY_MHZ 8  //CAN_ADDON option, what is your MCP2515 add-on boards crystal frequency?
#define CANFD_ADDON_CRYSTAL_FREQUENCY_MHZ \
  ACAN2517FDSettings::OSC_40MHz  //CANFD_ADDON option, what is your MCP2518 add-on boards crystal frequency?
#define WIFI
#define WIFICONFIG  //Enable this line to set a static IP address / gateway /subnet mask for the device. see USER_SETTINGS.cpp for the settings
[...]

PR #793, opened 2025-01-14
Bugfix: Make PERIODIC_BMS_RESET not cause events after reset

dalathegreat:
### What
This PR attempts to fix the bug reported in #792

### Why
PERIODIC_BMS_RESET was a newly introduced feature in 8.x.x , it periodically reboots the battery BMS to prevent memory leaks. However, the LEAF battery sends some garbage data upon poweron, which we incorrectly interpreted as sane values. This caused many events to trigger once we resumed using the battery.

### How
We now sanity check voltage measurements, if they are 0x3FF we skip using it. This is the default value when you boot


No-Signal (approved):
Good fix

PR #801, opened 2025-01-16
Bugfix: Pylon CAN, add status bit writing to 0x4250

dalathegreat:
### What
This PR fixes a small bug, we have only been writing status bits to the message 0x4251 , but not to the 0x4250 message

### Why
We fix this, to be able to better relay status when using the SEND_0 option

### How
Thanks to Stefan for noticing this on the Discord :raised_hands:


mvgalen (approved):
LGTM.

PR #803, opened 2025-01-19
Add iX balancing requests

wjcloudy:
### What
This PR implements regular requests for the BMS to start balancing and sets some valid values at startup.

### Why
Improves support for iX battery

### How
Additional of two UDS routine requests sent regularly


dalathegreat (approved):
LGTM 👍

Issue #805, opened 2025-01-19
Tesla DEFINE LFP move to 'user_settings.h'


TheOfficialBatteryMan:
Selecting LFP when using an LFP Tesla battery is the safest method, as it means the software knows what voltages it is working with. This can get forgotten because this DEFINE setting is hidden in the battery/Tesla.h file

We need to move this to User_settings.h so it is not forgotten about for new people and in multiple updates for more experienced people.

This improvement will aid safety while also providing a better user experience.

dalathegreat:
Implemented in PR https://github.com/dalathegreat/Battery-Emulator/pull/806 , thanks for the improvement suggestion!

PR #806, opened 2025-01-19
Improvement: Move LFP option, rearrange USER_SETTINGS

dalathegreat:
### What
This PR cleans up the arrangement in USER_SETTINGS file. It has recently grown quite rapidly, so some new features were a bit all over the place. It also moves the Tesla LFP definition from the tesla.h file, into USER_SETTINGS

### Why
It was too easy for users to forget to set the LFP define, as reported in #805 

### How
We move `LFP_CHEMISTRY` definition to USER_SETTINGS


mvgalen (approved):
LGTM.

PR #808, opened 2025-01-20
Improvement: Tesla blown pyro event

dalathegreat:
### What
This PR introduces an event for detecting blown pyro fuse on Tesla batteries

### Why
To aid troubleshooting Tesla batteries

### How
We now fire an event about fuse being blown if the voltage is 1.0V

Bonus, removed some unnecessary header include files for quicker compilation time


mvgalen (approved):
LGTM.

PR #810, opened 2025-01-20
Cleanup: Move global charger variables to datalayer

dalathegreat:
### What
This PR moves the global charger variables from Software.ino into the datalayer

### Why
Better coding standards. Less includes and repetitions

### How
Same way as other components, batteries, shunts, etc.


mvgalen (approved):
LGTM. Nice improvement.

PR #811, opened 2025-01-20
Bugfix: Nissan LEAF, move battery-allows-closing writing

dalathegreat:
### What
This PR moves the `battery_allows_contactor_closing` writing for Nissan LEAF

### Why
Periodic BMS reset would always cause this red X on the webserver, confusing users

image

It also confused users, some packs never set this bit as stated in the LEAF wiki

_NOTE: with some Leaf BMS (so far only noticed on the 30kWh 2016) the line batteryAllowsContactorClosing = true is never set: https://github.com/dalathegreat/Battery-Emulator/blob/d759946cc41882576b6c27be095fffe0f47ba191/Software/src/battery/NISSAN-LEAF-BATTERY.cpp#L745 . This prevents the automatic contactor sequence. A workaround is to force the value to always true. So if you are having problems with the contactor sequence, change the above line to datalayer.system.status.battery_allows_contactor_closing = true;_

### How
Instead of dynamically writing it based on CAN, always allow contactor closing on boot

This also makes the workaround no longer required

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -766,11 +766,6 @@ void handle_incoming_can_frame_battery(CAN_frame rx_frame) {
       battery_Relay_Cut_Request = ((rx_frame.data.u8[1] & 0x18) >> 3);
       battery_Failsafe_Status = (rx_frame.data.u8[1] & 0x07);
       battery_MainRelayOn_flag = (bool)((rx_frame.data.u8[3] & 0x20) >> 5);
-      if (battery_MainRelayOn_flag) {
-        datalayer.system.status.battery_allows_contactor_closing = true;
-      } else {
-        datalayer.system.status.battery_allows_contactor_closing = false;
-      }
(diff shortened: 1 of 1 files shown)


mvgalen (commented):
I understand that the old situation is not ideal, but the always allowing the contactors to close (even when BMS is being reset) seems a bit strange to me.
Shouldn't at least some sanity checks be done before enabling battery_allows_contactor_closing? Like, BMS powered and no faults present?

dalathegreat:
Very good point @mvgalen , I changed this in commit [850922b](https://github.com/dalathegreat/Battery-Emulator/pull/811/commits/850922b0f0cfd82d49e9f56cde823cc22d248468) , so now we need to get CAN from battery before we allow closing to happen!

mvgalen:
Yes, that is better. But are there no sanity checks we could do, like checking soc?

dalathegreat:
@mvgalen It is better to try and use the battery to prevent it from under/overcharging. Incase it is impossible to use it, we set fault and open contactors. I think this implementation is much better than it was before :)

mvgalen (approved):
LGTM.

Issue #679, opened 2024-12-18
Webs server display of "Contactors controlled by emulator, state:" not working correctly


davidhgriffiths:
### Expected Behavior

When the contactors open after an inverter request, the web server should display the OFF state.

### Actual Behavior

When the contactors open after an inverter request the web server continues to display the  ON state

I think the bug is caused by the absence of the following line at about line 128 of webserver.cpp

```
datalayer.system.status.contactors_engaged = false;
```

Currently `datalayer.system.status.contactors_engaged` only gets set to false if contractor opening is requested from the web page, not if it is requested by the inverter or estop

mvgalen:
Just so that I understand the issue correct: You mean that currently if an inverter request opening of contactors it does not show on the webpage? For which inverter have you seen this?

davidhgriffiths:
Yes your understanding is correct - if the inverter requests opening of the contractors the contractors open but the web page continues to show them closed.  The inverter is a Solax X3 G4. With the change I suggested it works correctly.

dalathegreat:
Thanks for the report @davidhgriffiths , I have issued a fix for this in PR https://github.com/dalathegreat/Battery-Emulator/pull/812

dalathegreat:
Closing as completed, now in main, target release 8.3.0 🥳  Thanks again for the report!

PR #812, opened 2025-01-20
Bugfix: Make Webserver show correct contactor status when Solax opens them

dalathegreat:
### What
This PR fixes the bug reported in #679 

### Why
To make it easier for users to see actual contactor state

### How
Currently datalayer.system.status.contactors_engaged only gets set to false if contractor opening is requested from the web page, not if it is requested by the inverter or estop.


mvgalen (approved):
LGTM.

PR #814, opened 2025-01-22
Bugfix: Halt only CAN sending towards battery while BMS resetting

dalathegreat:
### What
This PR makes the BMS Reset only halt CAN message sending towards battery

### Why
The current implementation pauses CAN sending globally. That means that if you use a CAN based inverter + Period BMS reset, the inverter will also get starved of CAN messages while the reset is active. This is not desired, since it risks using power from battery while it is off. This new fix makes it pause only sending CAN towards battery while it is down.

Another issue was also that the original implementation only stopped ESP32.CAN , which was the first CAN interface. So it was impossible to use the feature right on setups where battery was not connected to fist CAN port

### How
Moved the `isBMSResetActive` into the datalayer `datalayer.system.status.BMS_reset_in_progress`. Also simplified the code a bit.

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -1064,9 +1064,19 @@ void handle_incoming_can_frame_battery(CAN_frame rx_frame) {
   }
 }
 void transmit_can_battery() {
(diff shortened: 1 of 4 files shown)


mvgalen (approved):
LGTM.

PR #816, opened 2025-01-23
Feature: Add support for Double Kia-Hyundai 64/40kWh batteries

dalathegreat:
### What
This PR implements support for running 2x Kia-Hyundai  64/40kWh batteries at the same time :battery: :battery: 

### Why
Feature requested by Discord member

### How
Same way as other DoubleBattery implementations


mvgalen (approved):
- I did not check the code, in general it seems ok.
- We need to ASAP discuss on how to handle multiple batteries better. E.g. putting variables in structs so they can be an argument to battery functions.

PR #826, opened 2025-01-27
Improvement: Add more CAN findings for Bolt-Ampera integration

dalathegreat:
### What
This PR adds more CAN information about the IDs used in this integration

### Why
To assist with finalizing the integration

### How
We now use correct terminology on the CAN messages in the system. This will aid reverse engineering efforts


mvgalen (approved):
LGTM.

greg8787cpu:
I am working on a Bolt 2017 battery. Sending a query 7E4h there is no response 7ECh. I need to get SOC. Sending 07E4 8 03 22 43 AF 00 00 00 00 there is no response. Asking the battery itself via the diagnostic interface about SOC (7E4h 43AF) there is also no 7ECh. When I view the communications in the car there are responses 7ECh but I do not know which module. Attached is a PCANview file with the car. Can anyone help?

[BOLT ignition ON in the car question about SOC.csv](https://github.com/user-attachments/files/19565700/BOLT.ignition.ON.in.the.car.question.about.SOC.csv)

PR #827, opened 2025-01-27
Tesla: Send more CAN, remove double battery

dalathegreat:
### What
This PR adds more CAN message sending towards the battery. It also removes Double Battery support for Tesla packs.

### Why
Integration is not considered 100%. Easier to develop for now

### How
We now send 0x241, 0x242, 0x243 and 0x129 towards the battery

Delete many lines related to Double Battery support


mvgalen (approved):
LGTM.

Issue #824, opened 2025-01-27
RESET PANIC and Board Resetting -> MQTT?


obsoolete:
Running `8.0.0`, I experienced 3 "RESET PANIC" events over the course of about 1 week. Dala suggested MQTT was causing this problem, so I updated to `8.3-dev` (`a46c6f4a`), and after flashing the board became un-usable. I estimate every 30-60s the board would crash and reset. 

I enabled `DEBUG_VIA_USB` and was able to capture the log attached in **MQTTCrashLog.txt**. I have also uploaded my **USER_SETTINGS** files. Of note in the **USER_SECRETS** file:

```cpp
#define WEBSERVER_AUTH_REQUIRED true
#define MQTT_USER ""
#define MQTT_PASSWORD ""
```

Am currently running 8.2.0 which has been solid for the last 30 hours.

N.B: I have home assistant connected to the MQTT server (with the MQTT integration), and previously there was nice parsing of all the messages from the battery emulator (i.e. recognised BE as a device and created entities for all the data). Since moving to 8.2.0, this has all disappeared. Messages are still being sent to the server (i.e. checked using MQTTX). Might have to make a separate issue for this.

[MQTTCrashLog.txt](https://github.com/user-attachments/files/18557676/MQTTCrashLog.txt)
[USER_SETTINGS.h](https://github.com/user-attachments/files/18557687/USER_SETTINGS.h.txt)
[USER_SETTINGS.cpp](https://github.com/user-attachments/files/18557743/USER_SETTINGS.cpp.txt)

dalathegreat:
@obsoolete , the Wiki page for MQTT says;
"The username and password for MQTT can be set in the USER_SECRETS file. Incase you want to use no login info, set fields to NULL"

Does the crash happen if you set
```
#define MQTT_USER NULL
#define MQTT_PASSWORD NULL
```

dalathegreat:
Nevermind, I managed to get a crash reproduced with the settings file included. I dont have an MQTT server, so my crash might be different;

```
E (11085) esp-tls: [sock=49] select() timeout
E (11085) transport_base: Failed to open a new connection: 32774
E (11085) mqtt_client: Error transport connect
E (11249) outbox: outbox_enqueue(53): Memory exhausted
E (11251) outbox: outbox_enqueue(53): Memory exhausted
E (11253) outbox: outbox_enqueue(53): Memory exhausted
E (11255) outbox: outbox_enqueue(53): Me
[...]

PR #836, opened 2025-01-30
Bugfix for mqtt crash & platformio update

No-Signal:
This PR adds a new setting MQTT_QOS with a default value of 0 to stop crashes when the mqtt server is unavailable but still allow higher QoS when desired

It also updates the platformio config to use ESP 3.1.1 as it was using an out of date version

Fix for #824


dalathegreat:
Thanks for fixing the crashing issue :raised_hands: Shall we add this info to the Wiki, that the QoS can be increased (at the risk of crashing?)

Issue #831, opened 2025-01-30
MQTT build errors when autodiscovery is disabled and mqtt is enabled


amarofarinha:
Hi.
when building with autodiscovering off, i have build erros.
Version 8.4.dev  (is is probably also on 8.3.0)

![Image](https://github.com/user-attachments/assets/69ec106b-c888-4a74-9a4e-c5be135e431c)

![Image](https://github.com/user-attachments/assets/757d35fd-f7f6-4c1d-b1ee-9152201f06cb)

dalathegreat:
Closing as completed, fix included in release v8.4.0 🥳

PR #837, opened 2025-01-31
Fix for compilation errors when HA_AUTODISCOVERY is commented out

greenoem:
What

This PR fixes MQTT build errors when autodiscovery is disabled and mqtt is enabled as per issue #831 - pre-commit now run/checked :-)

Why

Build fails when #define HA_AUTODISCOVERY is commented out (disabled) in USER_SETTINGS.h

How

ifdefs changed to incorporate missing function definitions

### DIFF
--- a/Software/src/devboard/mqtt/mqtt.cpp
+++ b/Software/src/devboard/mqtt/mqtt.cpp
@@ -38,7 +38,6 @@ static void publish_values(void) {
 }
 
 #ifdef HA_AUTODISCOVERY
-
 static bool ha_common_info_published = false;
 static bool ha_cell_voltages_published = false;
 static bool ha_events_published = false;
@@ -132,6 +131,7 @@ void set_common_discovery_attributes(JsonDocument& doc) {
   doc["payload_not_available"] = "offline";
   doc["enabled_by_default"] = true;
 }
+#endif  // HA_AUTODISCOVERY
 
 void set_battery_attributes(JsonDocument& doc, const DATALAYER_BATTERY_TYPE& battery, const String& suffix) {
   doc["SOC" + suffix] = ((float)battery.status.reported_soc) / 100.0;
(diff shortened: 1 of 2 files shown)


Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -538,6 +540,6 @@ void mqtt_loop(void) {
 }
 
 bool mqtt_publish(const char* topic, const char* mqtt_msg, bool retain) {
-  int msg_id = esp_mqtt_client_publish(client, topic, mqtt_msg, strlen(mqtt_msg), MQTT_QOS, retain);
+  int msg_id = esp_mqtt_client_publish(client, topic, mqtt_msg, strlen(mqtt_msg), 1, retain);
```
No-Signal, line 543: Looks like the QoS fix has regressed

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -425,6 +425,7 @@ void mqtt_message_received(char* topic, int topic_len, char* data, int data_len)
   }
 #endif  // REMOTE_BMS_RESET
 
+#ifdef HA_AUTODISCOVERY
```
No-Signal, line 428: These are commands that can be executed and are not part of the HA Autodiscovery process so shouldn't be excluded

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -132,6 +131,7 @@ void set_common_discovery_attributes(JsonDocument& doc) {
   doc["payload_not_available"] = "offline";
   doc["enabled_by_default"] = true;
 }
+#endif  // HA_AUTODISCOVERY
 
 void set_battery_attributes(JsonDocument& doc, const DATALAYER_BATTERY_TYPE& battery, const String& suffix) {
```
No-Signal, line 136: Would it be better to move this function down to keep all Autodiscovery methods together?

No-Signal (changes requested):
I think there are a couple of bugs to be fixed

Software/src/devboard/mqtt/mqtt.cpp
greenoem: Fixed, will amend commit

Software/src/devboard/mqtt/mqtt.cpp
greenoem: These commands reference generateButtonTopic which is part of the main HA Autodiscovery block (line 40+)?

Software/src/devboard/mqtt/mqtt.cpp
greenoem: Agreed

Software/src/devboard/mqtt/mqtt.cpp
No-Signal: I should have moved generateButtonTopic outside of the HA Autodiscovery block as it is needed to check received commands

No-Signal (approved):
Looks great

PR #840, opened 2025-01-31
BYD Atto 3: Change temperature reading + code cleanup

dalathegreat:
### What
This PR changes the source for BYD Atto 3 temperature readings

### Why
Some users reported that values can glitch out when rebooting

### How
Instead of reading all daughterboard temperatures and doing mathematical operations to figure out min/max, we now just use the temperature values sent via PID polls-

Bonus, this PR cleans up the code


mvgalen (approved):
LGTM.

PR #843, opened 2025-02-02
Add warning event if SD card fails to initialize

No-Signal:
### What
This PR adds a warning event on SD initialization failure to make it more visible when a failure occurs. It includes the messages that a power reset is required if you end up in this state.


dalathegreat (approved):
LGTM 👍

PR #847, opened 2025-02-03
Periodic BMS reset now at a specific time with the help of NTP

laughingguffly:
### What
This PR implements the ability to reset the BMS at a specific time. 
Also fixes the bug where the bms power is restored during a bms reset event but the bms is not given time to startup and CAN overrun messages appear in logs.

### Why
To add more flexiblity to periodic reset. Currently it resets every 24 hours from when the emulator is turned on.

### How
How does it do it?

As the Emulator board has no concept of time. The emulator can now sync with an NTP server (Using espressif Time library)  at startup and work out the number of ms between now and a time in the future. That time difference is used as an offset when performing the BMS Reset.

Timezone, NTP servers and BMS Reset time are all configurable parameters.

---

Screenshot 2025-02-03 at 14 29 23
1-e0f5-4039-99c7-1970f2797d81" />


Software/USER_SETTINGS.h
```diff
@@ -46,7 +46,7 @@
 //#define AFORE_CAN        //Enable this line to emulate an "Afore battery" over CAN bus
```
dalathegreat, line 46: Try to restore changes to USER_SETTINGS so we can merge this PR. Only change the new row you added, PERIODIC_BMS_RESET_AT

Software/src/devboard/utils/ntp_time.cpp
```diff
@@ -0,0 +1,64 @@
...
+  configTzTime(time_zone, ntpServer1, ntpServer2);
+  struct tm timeinfo;
+
+  // Wait for time to be set
+  for (int i = 0; i < 5; i++) {
+    if (getLocalTime(&timeinfo)) {
+      logging.println("Time retrieved from NTP Server");
+      break;
+    }
+    delay(1000);
```
dalathegreat, line 17: Delay function cannot be used in embedded code. This will block CAN sending, Modbus handler, etc.

Software/USER_SETTINGS.h
laughingguffly: done

Software/src/devboard/utils/ntp_time.cpp
laughingguffly: I've removed it. It should be  fine without.

Software/Software.ino
```diff
@@ -44,6 +44,10 @@
 #include "src/devboard/mqtt/mqtt.h"
 #endif  // MQTT
 #endif  // WIFI
+#ifdef PERIODIC_BMS_RESET_AT
+#include "src/devboard/utils/ntp_time.h"
```
No-Signal, line 48: I could see ntp being a useful optional add on in general, it would allot events and debug logging to show the date and time of the event instead of just the uptime figure.

Software/Software.ino
```diff
@@ -124,6 +127,9 @@ void setup() {
 
   xTaskCreatePinnedToCore((TaskFunction_t)&core_loop, "core_loop", 4096, &core_task_time_us, TASK_CORE_PRIO,
                           &main_loop_task, CORE_FUNCTION_CORE);
+#ifdef PERIODIC_BMS_RESET_AT
```
No-Signal, line 130: If wifi is slow to connect could this fail to set the time? It would be good to periodically sync time to avoid drift and moving this code to a loop would avoid the wifi connection issue at startup

Software/src/devboard/utils/ntp_time.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+const unsigned long millisInADay = 24 * 60 * 60 * 1000;  // 24 hours in milliseconds
+
+unsigned long long getNtpTimeInMillis() {
+  configTzTime(time_zone, ntpServer1, ntpServer2);
+  struct tm timeinfo;
+
+  // Wait for time to be set
+  for (int i = 0; i < 5; i++) {
+    if (getLocalTime(&timeinfo)) {
```
No-Signal, line 13: Should this set a warning event on failure so people are aware it has failed and the restart wont happen when expected?

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -249,7 +249,7 @@ void handle_BMSpower() {
 
 #ifdef PERIODIC_BMS_RESET
   // Check if 24 hours have passed since the last power removal
-  if (currentTime - lastPowerRemovalTime >= powerRemovalInterval) {
+  if ((currentTime + bmsResetTimeOffset) - lastPowerRemovalTime >= powerRemovalInterval) {
```
No-Signal, line 252: Will this be impacted by daylight savings?

laughingguffly: It only does the time sync once on start up. If the clocks change its going to be out.

Software/src/devboard/utils/ntp_time.cpp
laughingguffly: Yes I'll add a new event type.

Software/Software.ino
laughingguffly: What was thinking behind not doing NTP sync  on the emulator  and instead always counting from 0 millis. Did you not want a board that was reliant on internet connection ? According to the espressif docs there are robust enough timers in the esp32 for keeping time although the recommendation is to sync hourly. Also you would have to think about the other hardware that is supported and implications for that.  My thinking was to keep it simple for this feature :)

Software/Software.ino
laughingguffly: As I'm calculating an offset I wanted to do it as early as possible. If I had it in the main loop and constantly retrying I might end calculating an unaccurate off-set.  e.g Maybe the internet doesn't work for first 30 mins then I calculate it.. A v2 of this would just have the main timer synced to NTP and the reset would be done on reaching a set time but that would involve refactoring lots of other things I don't know much about. I'm time poor :)

Software/USER_SETTINGS.h
```diff
@@ -139,8 +141,6 @@
 #define BATTERY_MAXTEMPERATURE 500
 // -250 = -25.0 °C , Min temperature (Will produce a battery frozen event if below)
 #define BATTERY_MINTEMPERATURE -250
-// 150 = 15.0 °C , Max difference between min and max temperature (Will produce a battery temperature deviation event if greater)
```
dalathegreat, line 142: You have accidentally rebased on an older version, that removed some functionality from this file!

dalathegreat (changes requested):

Software/USER_SETTINGS.h
laughingguffly: Will take a look

laughingguffly:
@dalathegreat fixed user_settings.h ready to review now.

dalathegreat (approved):

dalathegreat:
And final hurdle, the [code formatting](https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#code-formatting)! Try running the `pre-commit run --all-files`

PR #848, opened 2025-02-03
MEB: improvements

mvgalen:
### What
- Fixes temperature conversion
- Adds battery temp-point temperatures
- Adds cell temperatures.
- Makes automatic precharge more reliable/easier to debug
- Adds events to mark begin/end of balancing
- Add event for high temperature (also triggered for high difference between cells)
- Uses cell voltage min/max from CAN message instead of calculating from cell voltages
- Add battery temperature value
- Adds all new values to advanced battery page.
- Updates TODOs
### Why
To fix incorrect values in the webpage.
To have more info available in advanced battery page. 
To fix difficult precharge behaviour with the 82kWh battery model.

### How
Use PID value conversions as found on the web.
Couple precharge state machine to CAN messages generated, instead of hope-for-the-best asynchronous behaviour.

Tested with 62kWh and 82kWh batteries.


Software/src/battery/MEB-BATTERY.cpp
```diff
@@ -561,33 +552,23 @@ void update_values_battery() {  //This function maps all the values fetched via
...
-      min_cell_mv_value = std::min(min_cell_mv_value, voltage_mV);
-      max_cell_mv_value = std::max(max_cell_mv_value, voltage_mV);
-    }
-  }
-  // If all array values are 0, reset min/max to 3700
-  if (min_cell_mv_value == std::numeric_limits<uint16_t>::max()) {
-    min_cell_mv_value = 3700;
-    max_cell_mv_value = 3700;
+  if (datalayer.battery.status.temperature_max_dC > 400) {
+    set_event_latched(EVENT_BATTERY_OVERHEAT, datalayer.battery.status.temperature_max_dC);
```
dalathegreat, line 562: Battery overheat already set in Safety.cpp (at 50C) , so this one would trigger before at 40C. Seems like overkill to also check it here?

Software/src/battery/MEB-BATTERY.cpp
```diff
@@ -561,33 +552,23 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+  } else {
+    if (labs(datalayer.battery.status.temperature_max_dC - datalayer.battery.status.temperature_min_dC) > 100)
```
dalathegreat, line 564: I like this check! High deviation between high and low. But wouldn't it be better if it was placed in Safety.cpp, and with another Event name? EVENT_TEMPERATURE_DEVIATION ? That way other integrations would also get this nice check!

Software/src/battery/MEB-BATTERY.cpp
mvgalen: Ok, didn't know, I'll remove it here.

Software/src/battery/MEB-BATTERY.cpp
mvgalen: Yes, good idea. We would need to see how it pans out. 10degr is just some arbitrary value. I will remove it here and do a PR to add it to safety.

dalathegreat (approved):
LGTM :+1: Nice work on these improvements!

PR #849, opened 2025-02-03
BYD Atto3: Add 10 temperature sensors to datalayer

dalathegreat:
### What
This PR makes it so you can track all 10 temperature sensors inside the battery

It also adds a method to skip using a specific temperature sensor, if you know it is broken.

Example, Sensor number 4 broken

image

image

### Why
Better way to monitor the battery.

Also adds a way to use a battery without needing to replace a daughterboard with broken temp sensor (Note, requires deep knowledge of actual battery status)

### How
All 10x measurements added to extended datalayer

Configurable option in .h file, SKIP_TEMPERATURE_SENSOR_NUMBER that can be tuned between 1-10 if needed

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -135,6 +135,7 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   datalayer.battery.status.cell_min_voltage_mV = BMS_lowest_cell_voltage_mV;
 
+#ifdef SKIP_TEMPERATURE_SENSOR_NUMBER
   // Initialize min and max variables for temperature calculation
   battery_calc_min_temperature = battery_daughterboard_temperatures[0];
   battery_calc_max_temperature = battery_daughterboard_temperatures[0];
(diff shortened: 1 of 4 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -135,9 +135,35 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+  battery_calc_min_temperature = battery_daughterboard_temperatures[0];
+  battery_calc_max_temperature = battery_daughterboard_temperatures[0];
+
+  // Loop through the array of 10x daughterboard temps to find the smallest and largest values
+  // Note, it is possible for user to skip using a faulty sensor in the .h file
+  if (SKIP_TEMPERATURE_SENSOR_NUMBER == 1) {  //If sensor 1 is skipped, init minmax to sensor 2
+    battery_calc_min_temperature = battery_daughterboard_temperatures[1];
+    battery_calc_max_temperature = battery_daughterboard_temperatures[1];
+  }
+  for (int i = 1; i < 10; i++) {
```
No-Signal, line 149: Should this start at sensor 0?

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -135,9 +135,35 @@ void update_values_battery() {  //This function maps all the values fetched via
...
+    if (i == (SKIP_TEMPERATURE_SENSOR_NUMBER - 1)) {
```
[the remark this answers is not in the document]
dalathegreat: For now it is so rare that a sensor would be broken. Best to keep it safe, and only allow bypassing a single faulted sensor. We also dont know if the BMS does something when sensors fail, so let's not risk it :sweat:

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
dalathegreat: Not needed, since we init the value to [0] element at the start

dalathegreat:
Change has been successfully tested by Discord member!

No-Signal (approved):

PR #850, opened 2025-02-03
MEB: Add CRC checking on incoming CAN messages

dalathegreat:
### What
This PR implements CRC checking on incoming messages from BMS

### Why
So we are able to detect corrupted CAN messages, and alert user to shield CAN wiring. Safer operation of MEB battery

### How
We now check the incoming CRC on key messages. If the actual CRC does not match the calculated CRC, we skip reading the rest of the message and instead increment the datalayer.battery.status.CAN_error_counter++;

If the CAN_error_counter grows too large, safety function reacts and raises an event


mvgalen (commented):
I think we can do the checks with less duplicated code. I have an idea how to do that.

dalathegreat:
Good improvements @mvgalen , have you tested if the code works towards a battery?

mvgalen (approved):
LGTM.

PR #851, opened 2025-02-04
Add event for large temperature difference in battery

mvgalen:
### What
Event when temperature difference between min and max temperature is larger than 15 degrees C.

### Why
When ambient temperature is low, a large temperature difference may exist in the battery before the max temp reaches 50 degrees C. However large temperature difference are a sign something is wrong in the battery.

### How
Compare min and max temp difference to a threshold.


Software/src/devboard/safety/safety.cpp
```diff
@@ -41,6 +41,14 @@ void update_machineryprotection() {
     clear_event(EVENT_BATTERY_FROZEN);
   }
 
+  if (datalayer.battery.status.temperature_max_dC - datalayer.battery.status.temperature_min_dC >
+      BATTERY_MAX_TEMPERATURE_DEVIATION) {
+    set_event_latched(EVENT_BATTERY_TEMP_DEVIATION_HIGH,
```
dalathegreat, line 46: Any reason you call set_event_latched instead of just set_event ?

Software/src/devboard/utils/events.cpp
```diff
@@ -218,6 +218,7 @@ void init_events(void) {
   events.entries[EVENT_MQTT_DISCONNECT].level = EVENT_LEVEL_INFO;
   events.entries[EVENT_EQUIPMENT_STOP].level = EVENT_LEVEL_ERROR;
   events.entries[EVENT_SD_INIT_FAILED].level = EVENT_LEVEL_WARNING;
+  events.entries[EVENT_BATTERY_TEMP_DEVIATION_HIGH].level = EVENT_LEVEL_ERROR;
```
dalathegreat, line 221: Deviation should IMO be a warning!

mvgalen: Yeah, sorry, I meant to change that, but forgot.

Software/src/devboard/safety/safety.cpp
mvgalen: I would think it serious enough to not let a user clear it. If temperature difference is reduced it will clear itself.

Software/src/devboard/safety/safety.cpp
```diff
@@ -41,6 +41,14 @@ void update_machineryprotection() {
     clear_event(EVENT_BATTERY_FROZEN);
   }
 
+  if (datalayer.battery.status.temperature_max_dC - datalayer.battery.status.temperature_min_dC >
```
dalathegreat, line 44: I think we should we perform ABS subtraction. There are risks that some integrations update min/max values at different rates, and this can lead to odd behaviour if we dont do ABS subraction, same as on cellvoltage diff

dalathegreat (approved):
LGTM now! :+1:  Very nice safety improvement!

PR #854, opened 2025-02-04
Tesla S/3/X/Y: Update rampdown power

dalathegreat:
### What
This PR updates the default rampdown powervalue for Tesla batteries

### Why
We had an user that overcharged their Tesla LFP pack, since the current values still allow for almost 7kW charging at 97%SOC.

image
image

### How
Instead of starting the rampdown from 15kW at 90%, we now start at 10kW. This change will hopefully allow for the cellvoltage safeties to have more time to react, and write allowed power to 0W incase it starts to get out of hand


No-Signal (approved):
Looks good

Gangster8214:
Hello.
I always changed this parameter to 5 kW before.
I think it is better.

https://github.com/user-attachments/assets/48058947-f1ed-4ad3-b318-2bb5f7099925

PR #856, opened 2025-02-04
iX Battery Type - Send BMS Reset at startup

wjcloudy:
### What
A fix to reset battery at BatEmu startup

### Why
Previously if BMS was already in error state at battery emulator it will fail to auto reset.

### How
Sends a BMS reset can frame at boot. Also extends stale timeout as it's possible to trigger it in solar/use equilibrium.


dalathegreat (approved):
LGTM!

PR #537, opened 2024-10-14
Add new battery implementation "Renault Twizy" (LV)

M4GNV5:
### What
This PR implements the CAN bus protocol of Renault Twizy Batteries.
~~Not yet tested with real hardware. Also includes some `TODOs` where I was unsure about the correct way to implement something. A review would be much appreciated.~~
Tested and reviewed

### Why
See: #536 

### How
This is just a straightforward decoder for CAN messages sent by a Renault Twizy BMS. The protocol is [documented by the Twizy community here](https://docs.google.com/spreadsheets/d/1gOrG9rnGR9YuMGakAbl4s97a6irHF6UNFV1TS5Ll7MY/edit?gid=0#gid=0).
For now we do not wake up the battery. For a still built-in battery one can simply turn on the vehicle. For a plain battery (without vehice) this is a TODO.

PR #857, opened 2025-02-05
New Battery: Daly SmartBMS Support

M4GNV5:
### What
This PR implements the Daly SmartBMS RS485 protocol, allowing to talk to and obtain battery data from these systems.

### Why
This is yet another LV component. Many DIY 48V batteries use Daly components, since they are cheap.

There are non-smart Daly versions which do not have communication and "smart" versions which come with CAN and RS485. While implementing the CAN communication is simple(r), I opted for the RS485 to keep the lilygo CAN port unused allowing to use it for the very common pylon-lv CAN protocol for the inverter side.
The payloads for both CAN and RS485 are the same, so reusing some of this code for also implementing the CAN communication should be easy if desired.

### How
This is the first RS485 battery implementation, so I had to make some additions to the core.

### When
While I was planning to use [Twizy batteries as storage](https://github.com/dalathegreat/Battery-Emulator/pull/537), in the end we obtained only cells from Twizy packs (no BMS included) (at only 10€/kWh). I had to change plans and bought a Daly BMS for one of the 4 packs.
I will convert this to a regular PR once everything is tested and running with my Deye LV inverter (pylon-lv protocol).

### Wiki
Not sure if I can edit the wiki directly?
usual daly bms voltage range: 4S-24S ~= 14V-100V


dalathegreat:
Nice work! I created this Wiki page which can be used for adding info about the Daly SmartBMS https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Daly-SmartBMS

M4GNV5:
I made some successfull tests. The Deye inverter got the daly data from battery emulator and displayed it on the screen. I did not connect power yet to make sure everything works. I actually found some problems with the pylon LV implementation as well and fixed those.
Let me know if you think a separate PR is required for this.

Below some images. The cabling is still quite experimental :laughing:. I will create better pictures for the wiki once everything is set up.
I will probably do a full test with power on ~wednesday.

photo_2025-02-10_11-11-28

photo_2025-02-10_11-24-47

Alizainharoon:
What version of deye are you using? 
I am using the SUN-8K-SG01HP2-US-AM2 version of Deye and it seems that it is not supported yet and I do not know what data is required for it to be supported

Software/src/inverter/PYLON-LV-CAN.cpp
```diff
@@ -55,12 +58,16 @@ void update_values_can_inverter() {
   PYLON_355.data.u8[2] = (datalayer.battery.status.soh_pptt / 10) & 0xff;
   PYLON_355.data.u8[3] = (datalayer.battery.status.soh_pptt / 10) >> 8;
 
-  PYLON_356.data.u8[0] = datalayer.battery.status.voltage_dV & 0xff;
-  PYLON_356.data.u8[1] = datalayer.battery.status.voltage_dV >> 8;
+  int16_t voltage_cV = datalayer.battery.status.voltage_dV * 10;
+  int16_t temperature = datalayer.battery.status.temperature_min_dC;
```
dalathegreat, line 62: What does the temperature variable achieve? Would it be better to put in average_temperature instead, based on min+max/2? Some other protocols are doing this

Software/src/inverter/PYLON-LV-CAN.cpp
```diff
@@ -42,9 +42,12 @@ CAN_frame PYLON_35E = {.FD = false,
 void update_values_can_inverter() {
   // This function maps all the values fetched from battery CAN to the correct CAN messages
 
-  // TODO: officially this value is "battery charge voltage". Do we need to add something here to the actual voltage?
-  PYLON_351.data.u8[0] = datalayer.battery.status.voltage_dV & 0xff;
-  PYLON_351.data.u8[1] = datalayer.battery.status.voltage_dV >> 8;
+  // XXX: this value is "battery charge voltage". We add 1V here to the current voltage to achieve charging
+  int16_t charge_voltage_dV = datalayer.battery.status.voltage_dV + 1;
+  if (charge_voltage_dV > datalayer.battery.info.max_design_voltage_dV)
```
dalathegreat, line 47: We have a method for user to manually specify target charge voltage, consider implementing also this if you have time!

Software/src/inverter/PYLON-LV-CAN.cpp
M4GNV5: I first tought about using min_temperature when it is <10, max temperature when it is >20 and interpolate in the middle. It added a lot of complexity though.
I guess we can just go with (min+max)/2 instead

PR #858, opened 2025-02-05
kostal inverter crc and byte scrambling rewritten

gwk1:
### What
Kostal RS485 protocol CRC calculation and byte scrambling improvements

### Why
Byte scrambling has been resolved, allowing the removal of 2-byte float tweaks from the code. Additionally, CRC calculation can now be performed correctly.

### DIFF
--- a/Software/src/inverter/KOSTAL-RS485.cpp
+++ b/Software/src/inverter/KOSTAL-RS485.cpp
@@ -126,7 +126,9 @@ static void dbg_message(const char* msg) {
 #endif
 }
 
-void scramble_null_bytes(byte* lfc, int len) {
+/* https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing#Encoding_examples */
+
+void null_stuffer(byte* lfc, int len) {
   int last_null_byte = 0;
   for (int i = 0; i < len; i++) {
     if (lfc[i] == '\0') {
@@ -144,7 +146,7 @@ static void send_kostal(byte* frame, int len) {
 byte calculate_kostal_crc(byte* lfc, int len) {
   unsigned int sum = 0;
   if (lfc[0] != 0) {
-    printf("WARNING: first byte should be 0, but is 0x%02x\n", lfc[0]);
+    logging.printf("WARNING: first byte should be 0, but is 0x%02x\n", lfc[0]);
   }
   for (int i = 1; i < len; i++) {
     sum += lfc[i];
@@ -226,8 +228,10 @@ void update_RS485_registers_inverter() {
     CyclicData[59] = 0x00;
   }
 
-  float2frame(CyclicData, (float)(datalayer.battery.info.total_capacity_Wh / nominal_voltage_dV * 10),
-              30);  // BAttery capacity Ah
+  if (nominal_voltage_dV > 0) {
+    float2frame(CyclicData, (float)(datalayer.battery.info.total_capacity_Wh / nominal_voltage_dV * 10),
+                30);  // BAttery capacity Ah
+  }
   float2frame(CyclicData, (float)datalayer.battery.status.temperature_max_dC / 10, 38);
   float2frame(CyclicData, (float)datalayer.battery.status.temperature_min_dC / 10, 42);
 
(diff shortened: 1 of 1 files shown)


Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -191,60 +182,63 @@ void update_RS485_registers_inverter() {
...
-    frame2[57] = 0x03;
-    frame2[59] = 0x02;
+    CyclicData[59] = 0x02;
+  } else {
+    CyclicData[59] = 0x00;
   }
 
-  float2frame(frame2, (float)datalayer.battery.status.temperature_max_dC / 10, 38);
-  float2frame(frame2, (float)datalayer.battery.status.temperature_min_dC / 10, 42);
+  float2frame(CyclicData, (float)(datalayer.battery.info.total_capacity_Wh / nominal_voltage_dV * 10),
```
dalathegreat, line 229: I don't see how it could happen, but if for some odd reason `nominal_voltage_dV` would be 0, this would be a division by 0 and crash the board. If you have time, consider adding a check to only update this value if `nominal_voltage_dV` > 0

Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -148,17 +126,30 @@ static void dbg_message(const char* msg) {
...
   dbg_frame(frame, len, "TX");
   Serial2.write(frame, len);
 }
 
-byte calculate_longframe_crc(byte* lfc, int lastbyte) {
+byte calculate_kostal_crc(byte* lfc, int len) {
   unsigned int sum = 0;
-  for (int i = 0; i < lastbyte; ++i) {
+  if (lfc[0] != 0) {
+    printf("WARNING: first byte should be 0, but is 0x%02x\n", lfc[0]);
```
dalathegreat, line 147: replace `printf` with `logging` used by other integrations, that way user can see this even in the Webserver Log, and not only in the USB debug

Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -148,17 +126,30 @@ static void dbg_message(const char* msg) {
 #endif
 }
 
+void scramble_null_bytes(byte* lfc, int len) {
```
lewurm, line 129: "scramble" is something I came up with when reversing the BYD firmware, but later found that this is called "byte stuffing"

maybe rename it and put a reference to wikipedia?
https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing#Encoding_examples

Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -191,60 +182,63 @@ void update_RS485_registers_inverter() {
...
   //  Some current values causes communication error, must be resolved, why.
-  //  float2frameMSB(frame2, (float)datalayer.battery.status.current_dA / 10, 20);  // Peak discharge? current (2 byte float)
-  //  float2frameMSB(frame2, (float)datalayer.battery.status.current_dA / 10, 24);
-
-  float2frameMSB(frame2, (float)datalayer.battery.status.max_discharge_current_dA / 10, 28);  // BAttery capacity Ah
+  //  float2frame(CyclicData, (float)datalayer.battery.status.current_dA / 10, 18);  // Peak discharge? current (2 byte float)
+  //  float2frame(CyclicData, (float)datalayer.battery.status.current_dA / 10, 22);
 
-  float2frameMSB(frame2, (float)datalayer.battery.status.max_discharge_current_dA / 10, 32);
+  float2frame(CyclicData, (float)datalayer.battery.status.max_discharge_current_dA / 10, 26);
```
lewurm, line 205: I guess the goal is to keep the semantic changes at a minimum.  Thus shouldn't we keep 
```
float2frame(CyclicData, (float)datalayer.battery.status.max_discharge_current_dA / 10, 30);
```
too?

then again it would be kind of wrong anyways, so maybe leave it as-is and hope it still works 😅

lewurm (approved):

Software/src/inverter/KOSTAL-RS485.cpp
gwk1: That's the battery capacity (Ah) at bytes 30-33, which is populated below.  Previously, the byte pointers were mixed up.

Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -328,7 +322,7 @@ void receive_RS485()  // Runs as fast as possible to handle the serial stream
             }
 
             if (headerA && (RS485_RXFRAME[6] == 0x4A) && (RS485_RXFRAME[7] == 0x08)) {  // "frame 1"
-              send_kostal(frame1, 40);
+              send_kostal(BATTERY_INFO, 40);
```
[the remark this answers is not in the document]
gwk1: Good point, if float(nominal voltage) has null bytes, things go wrong.

PR #860, opened 2025-02-07
Pylon: Fix compilation error when using SET_30K_OFFSET

dalathegreat:
### What
This PR fixes a compilation error, when using SET_30K_OFFSET

image

### Why
We fix is to users can use Pylon CAN with all inverters

### How
Fix the compilation errors


mvgalen (approved):
LGTM.

Issue #825, opened 2025-01-27
24kWh leaf wrong capacity, reset degradation is not working


Kontich:
Hello, thank you for great project <3 

I have question regarding capacity on webserver: it shows 21024Wh in settings, after changing this value to 24000 and refreshing page it goes back to 21024Wh. 
Battery Emulator 8.2.0
I am not sure if it is expected because my SOH is 57%

also I did try to reset degradation data but there is an error:

![Image](https://github.com/user-attachments/assets/e77daa00-58b2-4e5c-91c7-26c5715e7ad4)

![Image](https://github.com/user-attachments/assets/c128b464-8738-4ec1-becb-6d35d07dbe3d)

dalathegreat:
On the Nissan LEAF, the code uses automatic detection of kWh available. However, this only works for 30-40-62kWh batteries, we just assume that 24kWh are at 100% and have 21kWh net capacity (273GIDS). This could be improved by multiplying this value by SOH (*0.57). I will mark this issue as a bug, and we can work on it for the next release!

The Reset Degradation Data was mostly developed for 30kWh packs. There is no gain to reset capacity on 24kWh, but if you really want to do it, you can use the dedicated app developed by safetyuggs, it is available on the MyNissanLEAF forum!

dalathegreat:
This is now fixed with PR https://github.com/dalathegreat/Battery-Emulator/pull/862

PR #862, opened 2025-02-08
Nissan LEAF: Make Wh max value more realistic based on SOH

dalathegreat:
### What
This PR fixes the Wh bug reported in https://github.com/dalathegreat/Battery-Emulator/issues/825#issuecomment-2615887506

### Why
Make the Wh value more realistic on 24kWh packs. 30/4/62kWh packs automatically fill in the available Wh based on maxGids, which is not available on 24kWh packs

### How
We now scale the 24kWh value based on SOH%

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -202,7 +202,7 @@ void update_values_battery() { /* This function maps all the values fetched via 
   datalayer.battery.status.current_dA =
       (battery_Current2 * 5);  //0.5A/bit, multiply by 5 to get Amp+1decimal (5,5A = 11)
 
-  if (LEAF_battery_Type == ZE0_BATTERY) {  //battery_Max_GIDS is stuck at 273 on ZE0
+  if (battery_Max_GIDS == 273) {  //battery_Max_GIDS is stuck at 273 on ZE0
     datalayer.battery.info.total_capacity_Wh = ((battery_Max_GIDS * WH_PER_GID) * (battery_StateOfHealth / 100));
   } else {  //battery_Max_GIDS updates on newer generations, making for a total_capacity_Wh value that makes sense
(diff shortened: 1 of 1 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -202,7 +202,11 @@ void update_values_battery() { /* This function maps all the values fetched via
   datalayer.battery.status.current_dA =
       (battery_Current2 * 5);  //0.5A/bit, multiply by 5 to get Amp+1decimal (5,5A = 11)
 
-  datalayer.battery.info.total_capacity_Wh = (battery_Max_GIDS * WH_PER_GID);
+  if (battery_Max_GIDS == 273) {  //battery_Max_GIDS is stuck at 273 on ZE0
+    datalayer.battery.info.total_capacity_Wh = ((battery_Max_GIDS * WH_PER_GID) * (battery_StateOfHealth / 100));
```
mvgalen, line 206: I think we have a loss of precision due to int conversion here. battery_StateOfHealth is an int type, which means that battery_StateOfHealth/100 always returns 0. Either we need to do this calculation as float or use: ((battery_Max_GIDS * WH_PER_GID * battery_StateOfHealth) / 100)

mvgalen (changes requested):
See comment in the code.

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
dalathegreat: Thanks! What was I thinking last night :sweat_smile:  Fixed in https://github.com/dalathegreat/Battery-Emulator/pull/862/commits/1e20b171df7ddf5f11b80334b2b4fa58e11b6773

mvgalen: That's why reviews matter, even for trivial stuff.

mvgalen (approved):
LHTM now ;-)

PR #864, opened 2025-02-09
Improvement: Tigthen cellvoltage warnings on most batteries

dalathegreat:
### What
This PR lowers the threshold needed to trigger a cellvoltage warning on most battery implementations

### Why
Many batteries were left at the default 500mV diff value, before raising an event. The idea was that feedback should be gathered from each integration, and these limits refined over time. They have now been set to the same value for a long time, so this change is long overdue. 

### How
We now tighten the limits significantly for those batteries left with essentially un-set limits


Tinyman7613 (approved):
Looks good

PR #866, opened 2025-02-09
SOLAX-CAN: message 187E should contain total_capacity.

mvgalen:
### What
On Solax CAN message 187E should contain the total capacity of the battery.
### Why
To make solax modbus report the bms capacity value correct.

### How
Report the correct value.


dalathegreat (approved):
:+1:

PR #874, opened 2025-02-10
Feature: Add limiting factor to Webserver

dalathegreat:
### What
This PR implements an user requested feature, Webserver to show what is limiting power in/out of the battery.

### Why
Feature requested by Discord users

### How
We implement a few new variables to the datalayer

- user_settings_limit_charge
- user_settings_limit_discharge

Based on these, we show in the Webserver when the charge/discharge amperage values are limited by the battery itself (BMS), or by the user configurable settings (Manual)

image

We also show if the Inverter or the Battery is the limiting factor, at the Battery Charging/Discharging text, using the following variables

- inverter_limits_discharge
- inverter_limits_charge

image

image

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -885,8 +885,18 @@ String processor(const String& var) {
     } else {
       content += formatPowerValue("Max discharge power", datalayer.battery.status.max_discharge_power_W, "", 1);
       content += formatPowerValue("Max charge power", datalayer.battery.status.max_charge_power_W, "", 1);
-      content += "<h4 style='color: white;'>Max discharge current: " + String(maxCurrentDischargeFloat, 1) + " A</h4>";
-      content += "<h4 style='color: white;'>Max charge current: " + String(maxCurrentChargeFloat, 1) + " A</h4>";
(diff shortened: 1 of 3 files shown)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -947,9 +957,22 @@ String processor(const String& var) {
...
     } else if (datalayer.battery.status.current_dA < 0) {
-      content += "<h4>Battery discharging!</h4>";
-    } else {  // > 0
-      content += "<h4>Battery charging!</h4>";
+      content += "<h4>Battery discharging!";
+      if (datalayer.battery.settings.inverter_limits_discharge) {
+        content += " (Inverter limiting)</h4>";
+      } else {
+        //Note, this can also be settings limiting
+        content += " (Battery limiting)</h4>";
```
mvgalen, line 965: Would an if/else here for settings_limiting not be a good idea?

dalathegreat: Yes good idea. Added this check in https://github.com/dalathegreat/Battery-Emulator/pull/874/commits/e4ac6d789902d6d4007161b53b94a3628c66641b

mvgalen (approved):
Nice.

Issue #873, opened 2025-02-10
Deye in offgrid does not stop discharge if SOC% becomes unreasonable


dalathegreat:
Noticed by user who accidentally managed to get a non-PCS Tesla pack to overdischarge. SOC% became stuck at 100%, and Deye in offgrid mode is only respecting SOC%, nothing else. So all safeties that triggered that wrote allowed power to 0W became useless, the Deye only waited on SOC% to drop.

The user was extremely unlucky here. Suggestion would be to write SOC% to either 0% or 100% incase allowed charge/discharge values go to 0W. This would have saved the situation

PR #875, opened 2025-02-10
BYD CAN: Safety, add overriding SOC incase of 0W allowed

dalathegreat:
### What
This PR implements a safety check to the BYD CAN protocol. Incase the battery no longer wants to charge, we send 100% SOC, and incase it no longer wants to discharge, we send 0% SOC

### Why
To avoid offgrid Deye inverters overdischarging batteries, as reported in https://github.com/dalathegreat/Battery-Emulator/issues/873

### How
If checks


mvgalen (approved):
LGTM.

PR #879, opened 2025-02-11
Init event handling before any use of events

mvgalen:
### What
Move init of events to earlier in initialisation, to prevent and event being generated before the event system is initialised.

### Why
Prevent crashes due to uninitialized values/pointers.

### How
Move the init_events function call to earlier in the setup routine.


dalathegreat (approved):
Nice catch! :+1:

PR #888, opened 2025-02-13
Ferroamp: Add linear interpolation to fake LFP voltage range

### LINKED ISSUE #887: Ferroamp high cell warning
To work around Ferroamp's monitoring system, which only supports LFP chemistry, we need to modify or fake the reported cell voltage for other chemistries to mimic an LFP signature.

dalathegreat:
### What
This PR implements rescaling of non LFP packs, into LFP cellvoltage ranges.

### Why
To fix #887 , this avoids raising suspicion of LFP overvoltage when using NCM batteries. 

### How
Linear interpolation to remap the value from the range [2500-4200] to [2500-3400]

### DIFF
--- a/Software/src/inverter/FERROAMP-CAN.cpp
+++ b/Software/src/inverter/FERROAMP-CAN.cpp
@@ -134,9 +134,24 @@ CAN_frame PYLON_4291 = {.FD = false,
                         .ID = 0x4291,
                         .data = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}};
 
+static uint16_t cell_tweaked_max_voltage_mV = 3300;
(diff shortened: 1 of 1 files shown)


LevinSwe:
The code at row 410..413, is after END_IF, looks like a copy from row 373..376. But with wrong CAN-ID in last two 412&413? 
Or is it the code for BMS temperature that should send CellMaxTemp?

dalathegreat:
@LevinSwe This is kind of intentional. We had to revert Ferroamp to this older implementation of Pylon from 8.1, which has ifdef bugs. So we on purpose are using the odd implementation, since it works with Ferroamp.

Tinyman7613:
it works :)

Tinyman7613 (approved):
It works!

Issue #890, opened 2025-02-14
EPS32 3.1.2 is crashing OTA page


dalathegreat:
After updating a board to [ESP32 3.1.2 ](https://github.com/espressif/arduino-esp32/releases/tag/3.1.2), navigating to the OTA page no longer works. Going here will cause the Hardware ID and Firmware Version fields to read (error), and crash the board

![Image](https://github.com/user-attachments/assets/9783efcf-eacc-4c8e-8213-3d9867521221)

![Image](https://github.com/user-attachments/assets/e52c980f-e7e7-4d31-a739-96bfa19dbe6a)

Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.

Workaround, use ESP32 3.1.1 until a fix can be deployed

PR #891, opened 2025-02-14
Library: Update ElegantOTA from 3.1.5 -> 3.1.6

dalathegreat:
### What
This PR , along with the accidental force push I did to main https://github.com/dalathegreat/Battery-Emulator/commit/87050791ad049c3deafddbf2849ec316e785a997 , fixes the crashing issue

### Why
After ESP32 v3.1.2 was taken into use, navigating to the OTA update page would result in a crash, as reported in https://github.com/dalathegreat/Battery-Emulator/issues/890

### How
We update ElegantOTA to the latest version (3.1.5 -> 3.1.6), along with switching to the AsyncTCPsock lib with official ESP32Async-AsyncTCP library, which is tested for compatibility with ESP32 releases


dalathegreat:
@amarofarinha , this PR effectively reverts the https://github.com/dalathegreat/Battery-Emulator/pull/461/ , in order to get the board to stop crashing

Issue #877, opened 2025-02-11
Re-enable the task watchdog


Dibblah1:
Currently, the task watchdog - which protects against a single task spinning and locking a core - seems to be disabled relatively early in boot, since this refactoring commit:

https://github.com/dalathegreat/Battery-Emulator/commit/01f1ac24f75c1269714df773b2d5c37bfc7b70f6#diff-f842a223a55460366f4d94fb1a5eb44048f82179a59a5f8ae812cad360abaa96R168

I have recently (probably due to the recent MQTT bug) had 3 lockups, where there is no external activity from the emulator - the LED is green but not pulsing, there is no modbus traffic to the inverter. 

The task watchdog should be enabled and a panic action should be considered if a critical task is not being allowed to be scheduled (from my understanding, this is not default)

PR #892, opened 2025-02-14
Feature: Enable task watchdogs

dalathegreat:
### What
This PR enables task watchdogs for both the core loop and the connectivity loop. The task watchdog prevents a process from spinning indefinitely and locking up a core.

### Why
During the refactoring in 5.x.x, task watchdogs were accidentally disabled! This could lead to unresponsive cores and system instability. See issue [#877](https://github.com/dalathegreat/Battery-Emulator/issues/877).

### How
We now trigger a panic event in case of a lockup, ensuring the board can recover from fault scenarios more reliably.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -73,6 +73,8 @@ TaskHandle_t logging_loop_task;
 
 Logging logging;
 
+#define WDT_TIMEOUT_SECONDS 5  // If code hangs for longer than this, it will be rebooted by the watchdog
+
 // Initialization
 void setup() {
   init_serial();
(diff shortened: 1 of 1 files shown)


Software/Software.ino
```diff
@@ -118,10 +120,19 @@ void setup() {
...
   pinMode(0, INPUT_PULLUP);
 
-  esp_task_wdt_deinit();  // Disable watchdog
-
   check_reset_reason();
 
+  // Initialize Task Watchdog for subscribed tasks
+  esp_task_wdt_config_t wdt_config = {
+      .timeout_ms = WDT_TIMEOUT_SECONDS * 1000,                        // Convert seconds to milliseconds
+      .idle_core_mask = (1 << CORE_FUNCTION_CORE) | (1 << WIFI_CORE),  // Watch both cores
```
Cabooman, line 128: The wifi core SHOULD always be different from the "core core", but who knows what the future brings :) There should be a define from somewhere in the board layer that states the number of cores, so that we can hardcode the mask to 1, 3, 7, etc

Cabooman (approved):

PR #893, opened 2025-02-14
(Hopfully) Final DALY BMS & Pylon LV fixes

M4GNV5:
### What
I did some more testing with by BMS & Inverter and managed to successfully charge and discharge my battery.
I did however find some remaining incorrect value conversions which are fixed with this PR

### Why
Make DALY and PylonLV work.

Side note I also added more text and pictures to the [Daly DIY LV Battery Wiki page](https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Daly-SmartBMS).


dalathegreat (approved):
Awesome work @M4GNV5 :raised_hands:

Issue #886, opened 2025-02-12
Improvement: RJXZS, guide user towards clearing faults in App


dalathegreat:
Currently the RJXZS can throw many different faults:

![Image](https://github.com/user-attachments/assets/aecb2c0f-268b-48c3-bc69-a89c05daf0b3)

However, this confuses users greatly, with no indication that it is the RJXZS that is causing this. We need to set another event at the same time as these, to guide the user to clear the historical log in the RJXZS smartphone app

Example, overvoltage event causing confused user:

![Image](https://github.com/user-attachments/assets/99e4269a-5ff2-4140-adfc-39f2982ed46f)

PR #897, opened 2025-02-15
Feature: Add event for RJXZS BMS error and guide user how to clear

dalathegreat:
### What
This PR adds a generic event for when the RJXZS BMS has errors active

### Why
Easier for end users to understand the system. As reported in https://github.com/dalathegreat/Battery-Emulator/issues/886

### How
Incase the fault code bits are high, we fire an info event that tells the user:

"Error code active in RJXZS BMS. Clear via their smartphone app!"


dalathegreat:
Confirmed working on the Discord by tester!

image

Issue #882, opened 2025-02-12
BE with RJXZS allways shows  "Battery charging"


pazdzierzz:
BE displays allways "Battery charging", we forgot to correct that bug during integration :)

![Image](https://github.com/user-attachments/assets/8f36b1c8-bb13-4901-89e7-cf5c97a06fcb)

dalathegreat:
PR with attempted fix is up here https://github.com/dalathegreat/Battery-Emulator/pull/898

PR #898, opened 2025-02-16
Bugfix: Add current direction for RJXZS

dalathegreat:
### What
This PR fixes a bug that caused all current directions to appear as charging

### Why
Forgotten during development: https://github.com/dalathegreat/Battery-Emulator/issues/882

### How
We now check message 0x51 and take the direction from this message. Based on this info we add sign to value


dalathegreat:
Tested successfully on the Discord :partying_face: 

image

Issue #901, opened 2025-02-18
[BUG] OTA page lost return to battery-emulator buttons in 8.6.0


marknisk:
After 8.6.0 OTA page does not have buttons that can be used to return to Battery-Emulator main view.

**To Reproduce**
a) hit "OTA Update" in the main page
b) after 8.6.0 is flashed the user sees OTA page without buttons

**Expected behavior**
Buttons that make it easy to navigate back to Battery-Emulator main page

**Screenshots**

![Image](https://github.com/user-attachments/assets/1cce473d-8933-496a-85fe-f51187a04b7f)

**Version and settings (please complete the following information):**
- Software version: 8.6.0
- Battery used: NISSAN_LEAF
- Inverter communication protocol: BYD_MODBUS
- Hardware used for Battery-Emulator: HW_LILYGO
- CONTACTOR_CONTROL: yes
- MQTT: 

![Image](https://github.com/user-attachments/assets/729fc811-f913-406f-b0d1-fd0cc1dc00a0)

yes

dalathegreat:
Yes this customization is what prompted us to release 8.6.0, the ESP32 3.1.2 update caused this page to crash. By reverting our customization we got it going again. We need to look at how to re-implement it in a way that does not cause crashes. @amarofarinha might be able to solve this once they have time over!

dalathegreat:
ESP32 v3.1.3 is now releaesed, which fixes the ArduinoJSON string copy crash. I have reverted the removal of the OTA buttons in this PR https://github.com/dalathegreat/Battery-Emulator/pull/902

dalathegreat:
Fix now in main. Version 8.7.0 will restore the OTA page functionality!

PR #902, opened 2025-02-19
Bugfix: Restore OTA back buttons

dalathegreat:
### What
This PR reverts the removal of the "Back to Battery-Emulator" buttons on the OTA page, to avoid crashing with ESP32 3.1.2

image

### Why
ESP32 3.1.3 was recently released as a hotfix for the ArduinoJSON string copy crash. We can now take into use our own buttons again :partying_face: , Many reported on this, and issues were made, https://github.com/dalathegreat/Battery-Emulator/issues/901

### How
Revert the changes in https://github.com/dalathegreat/Battery-Emulator/pull/891


mvgalen (approved):
LGTM.

Issue #903, opened 2025-02-19
[BUG] safety.cpp setting charge_power_W can be overridden by battery updating allowed charge power


Dibblah1:
On batteries that update their own max or min charge powers (for example, Kia Niro or Renault Zoe), the values set by safety.cpp (not sure if this actually is set for a battery comms loss) will be overridden on  the next run through update_values_battery(). 

**To Reproduce**
On a running battery which updates charge powers, disconnect the data CAN from the battery
Wait for a poll cycle of the battery
Power gets set back to the previously read actual battery requested power.

**Expected behavior**
Emulator probably should set an overriding value in safety critical situations. This includes cell overvoltage, undervoltage (BMS should already do this), loss of battery comms, etc. This is in addition to opening the contactors if appropriate, in case they're welded or otherwise unable to function.

Code:
[RENAULT-ZOE-GEN1-BATTERY.cpp#L90](https://github.com/dalathegreat/Battery-Emulator/blob/18878e040a92364a3de76e1fff3e005bdf219530/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp#L90)

[RENAULT-ZOE-GEN1-BATTERY.cpp#L156](https://github.com/dalathegreat/Battery-Emulator/blob/18878e040a92364a3de76e1fff3e005bdf219530/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp#L156)

**Screenshots**

![Image](https://github.com/user-attachments/assets/f3fc6c0f-50fb-4c8f-aa88-436a5467eb34)

![Image](https://github.com/user-attachments/assets/d11eaaf9-ee65-4d3b-8b8e-9f7d4e6833da)

![Image](https://github.com/user-attachments/assets/3ba529c4-4590-40fd-9b23-cc0544c97b7a)

**Version and settings (please complete the following information):**
- Software version: [8.3.0]
- Battery used: [Renault Zoe Gen 1]
- Inverter communication protocol: [BYD_MODBUS]
- Hardware used for Battery-Emulator: [HW_LILYGO]
- CONTACTOR_CONTROL: [yes]
- MQTT: [yes/no]

dalathegreat:
Fix up in https://github.com/dalathegreat/Battery-Emulator/pull/904

PR #904, opened 2025-02-20
Bugfix: Add writing of values to 0 incase of critical fault

dalathegreat:
### What
This PR fixes the bug reported in #903 

### Why
Safer operation of battery

### How
Incase `datalayer.battery.status.bms_status == FAULT` we now force allowed charge/discharge values to 0 always in SAFETY.cpp


mvgalen (approved):
LGTM.

Dibblah1:
Can confirm functionality. Battery emulator, on entering FAULT sets charge / discharge power and therefore current to 0 after around 10 seconds after the fault condition. When a non-latching fault condition (in this case lack of communication) is cleared, overridden charge power is repopulated with the battery values.

PR #912, opened 2025-02-22
Kostal communication improvements

gwk1:
##What
Kostal RX/TX handling has been rewritten.

##Why
The first implementation was quite fragile and contained many incorrect assumptions.

##How
The processing of received data frames has been clarified.
Null stuffing and CRC calculations have been added.

### DIFF
--- a/Software/src/inverter/KOSTAL-RS485.cpp
+++ b/Software/src/inverter/KOSTAL-RS485.cpp
@@ -214,23 +214,35 @@ void update_RS485_registers_inverter() {
        datalayer.battery.info.min_design_voltage_dV);
   float2frame(BATTERY_INFO, (float)nominal_voltage_dV / 10, 6);
 
-
   float2frame(CyclicData, (float)datalayer.battery.info.max_design_voltage_dV / 10, 10);
 
   float2frame(CyclicData, (float)average_temperature_dC / 10, 14);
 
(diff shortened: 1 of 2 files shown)


Software/src/inverter/KOSTAL-RS485.cpp
```diff
@@ -265,105 +304,95 @@ void receive_RS485()  // Runs as fast as possible to handle the serial stream
...
-        } else {
-          dbg_frame(RS485_RXFRAME, 10, "RX (dropped)");
+          rx_index = 0;
         }
         rx_index = 0;
       }
     }
-    if (rx_index >= 10) {
-      dbg_frame(RS485_RXFRAME, 10, "RX (!RX_allow)");
+    if (rx_index >= 299) {
```
dalathegreat, line 386: rx_index is casted as uint8_t, max value it can reach is 255. This check needs changing

dalathegreat (changes requested):

Software/src/inverter/KOSTAL-RS485.cpp
gwk1: https://github.com/dalathegreat/Battery-Emulator/pull/912/commits/1aeb8b9c467be9e2ee8ca61ad35a6a68ae93d28a

dalathegreat (approved):

PR #916, opened 2025-02-25
Microfeature: Make LED modes customizable in USER_SETTINGS

dalathegreat:
### What
This PR makes the LED pattern customizable in USER_SETTINGS.h

### Why
Previously you had to edit Software/src/system_settings.h to get the LED modes to change. Not a lot of users even knew about this feature

### How
Now we move it to USER_SETTINGS, for easy tuning


wjcloudy (approved):
LGTM!

Issue #920, opened 2025-02-25
[BUG]BMW i3, no cell voltages information via MQTT


pmajor021:
**Describe the bug**
cell voltages are not displayed on HomeAssistant (MQTT)

**To Reproduce**
check it on HA

**Expected behavior**
see voltages for 96 (or 2x96 for double battery)

**Screenshots**

**Version and settings (please complete the following information):**
- Please attach your USER_SETTINGS files for easier debugging
- Software version: 8.6.0
- Battery used: [BMW_i3]
- Inverter communication protocol: [BYD_MODBUS]
- Hardware used for Battery-Emulator: [HW_LILYGO]
- CONTACTOR_CONTROL: [no]
- MQTT: [yes]

PR #921, opened 2025-02-25
fix for issue #920

pmajor021:
### What
fixing the issue #920 with i3 batteries not sending cell voltage information via MQTT

### Why
Why does it do it?

### How
adding the number of cells info

### DIFF
--- a/Software/src/inverter/KOSTAL-RS485.cpp
+++ b/Software/src/inverter/KOSTAL-RS485.cpp
@@ -20,6 +20,7 @@ static unsigned long currentMillis;
 static unsigned long startupMillis = 0;
 static unsigned long contactorMillis = 0;
 
+static uint16_t rx_index = 0;
 static boolean RX_allow = false;
 
 union f32b {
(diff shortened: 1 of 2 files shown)


Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -273,6 +273,7 @@ static uint8_t battery_status_diagnosis_powertrain_maximum_multiplexer = 0;
 static uint8_t battery_status_diagnosis_powertrain_immediate_multiplexer = 0;
 static uint8_t battery_ID2 = 0;
 static uint8_t battery_soh = 99;
+static uint8_t detected_number_of_cells = 96;
```
dalathegreat, line 276: Better to use a define since the value never changes

#define NUMBER_OF_CELLS 96

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -407,6 +408,8 @@ void update_values_battery2() {  //This function maps all the values fetched via
     }
   }
 
+  datalayer.battery2.info.number_of_cells = detected_number_of_cells;
```
dalathegreat, line 411: Since this is only written in startup, it is better to move these into the `void setup_battery(` function.

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -1060,9 +1065,9 @@ void transmit_can_battery() {
         case CELL_VOLTAGE_CELLNO:
           current_cell_polled++;
           if (current_cell_polled > 96) {
-            datalayer.battery.info.number_of_cells = 97;
+            datalayer.battery.info.number_of_cells = 96;
```
dalathegreat, line 1068: datalayer.battery.info.number_of_cells = NUMBER_OF_CELLS

Software/src/battery/BMW-I3-BATTERY.cpp
pmajor021: I looked at how it is with the iX battery and adapted this to it. It works for me ;)

Software/src/battery/BMW-I3-BATTERY.cpp
pmajor021: better to delete these lines, since already defined in void setup_battery(void) function, as mentioned above:
 ```
             datalayer.battery.info.number_of_cells = 97;
#ifdef DOUBLE_BATTERY
            datalayer.battery2.info.number_of_cells = 97;
#endif
```
```

dalathegreat (approved):
Approved! Code formatting fails, best to run the pre-commit hook, explained here how: https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#code-formatting

PR #923, opened 2025-02-26
Feature: Kia eGMP, Add SOC estimation from cellvoltage min-max

dalathegreat:
### What
This PR implements a bandaid solution, SOC% estimation based on cellvoltage min/max

### Why
The SOC% is not usable on eGMP packs. To use the batteries, SOC% is mandatory. So we now estimate this value

### How
We calculate a SOC% value from the min and max cellvoltages. We then use either one that is the lowest.


mvgalen (approved):
LGTM.

PR #927, opened 2025-03-01
Feature: CAN log playback via Webserver

dalathegreat:
### What
This PR implements a new feature, CAN log replaying via Webserver

### Why
Awesome reverse engineering tool

### How
Added a new page called "CAN Replay", where you can upload .log files and play them back towards an interface. Both normal CAN and CAN-FD is supported. You can also tick the "Loop" checkbox, to have the replay start over automatically when reaching the end.

image


nagyrobi:
What happens when it reaches to the end? Maybe a number input box where one could set a number of repeat loops, where 0 would mean repeat infinitely?

dalathegreat:
@nagyrobi Good idea, I added the ability to loop the log 

image

wjcloudy (approved):
Looks like this is useful in current state!

1ni3shi:
Would this allow for CAN instructions to be sent to the battery BMS?

PR #930, opened 2025-03-03
Minor (final) Daly BMS  changes

M4GNV5:
Some minor leftover changes I made to the Daly BMS code. As of today battery is in daily use and works fine.

The changes include:
- Adjust webservice to display RS485 for battery connection when `RS485_BATTERY_SELECTED` is set
- Add a low temperature limit to Daly BMS (checking the various battery datasheets this could be way more complicated including charge/discharge differences, temperature to power curves, 30s peak power etc. For now this very simple approach should keep the battery safe when having the maximum charge/discharge low enough)
- Adjust default Daly BMS settings to the most common LV battery voltage range

Final installation Pictures:

photo_2025-03-03_09-42-25

signal-2025-03-01-163024

### DIFF
--- a/Software/src/battery/DALY-BMS.cpp
+++ b/Software/src/battery/DALY-BMS.cpp
@@ -37,6 +37,9 @@ void update_values_battery() {
   else if (SOC > 8000)
     adaptive_power_limit = ((10000 - (uint32_t)SOC) * POWER_PER_PERCENT) / 100;
 
+  if (temperature_min_dC < LOW_TEMP_POWER_LIMIT_START && adaptive_power_limit > LOW_TEMP_POWER_LIMIT)
+    adaptive_power_limit = LOW_TEMP_POWER_LIMIT;
+
   if (adaptive_power_limit < datalayer.battery.status.max_charge_power_W)
     datalayer.battery.status.max_charge_power_W = adaptive_power_limit;
   if (SOC < 2000 && adaptive_power_limit < datalayer.battery.status.max_discharge_power_W)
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
LGTM :+1: Nice work!

M4GNV5:
I was thinking about changing the low temperature adjustment once more. Maybe using a configurable linear function or something. Maybe wait until tomorrow before merging this.

image

dalathegreat:
Consider using the defines in USER_SETTINGS.h ?

// 500 = 50.0 °C , Max temperature (Will produce a battery overheat event if above)
#define BATTERY_MAXTEMPERATURE 500
// -250 = -25.0 °C , Min temperature (Will produce a battery frozen event if below)
#define BATTERY_MINTEMPERATURE -250

Software/src/battery/DALY-BMS.cpp
```diff
@@ -1,11 +1,9 @@
 #include "DALY-BMS.h"
 #include <cstdint>
-#include "../include.h"
-#include "RJXZS-BMS.h"
-#ifdef DALY_BMS
 #include "../datalayer/datalayer.h"
 #include "../devboard/utils/events.h"
-#include "RENAULT-TWIZY.h"
+#include "../include.h"
+#ifdef DALY_BMS
```
dalathegreat, line 6: The line #ifdef DALY_BMS needs to be the second row. Otherwise you start to include stuff, even though DALY is not used

dalathegreat (changes requested):

M4GNV5:
I fixed your include order comment, rebased your latest commits and updated this PR.

PR #933, opened 2025-03-03
Improvements for Volvo hybrid batteries

nmainil:
### What
This PR implements improvements and fixes for the Volvo hybrid batteries, it should be possible to close contactors automatically but it's left commented out, please check the updated wiki on wiring to remove some DTC and check the specific thread on Discord.

### Why
Read-only support was implemented first, time for a step further.

### How
Mostly by sending the same control messages from the supported VOLVO full EV batteries. There are still DTCs that keep coming but they doesn't seem to affect contactor closing/opening. Please keep in mind this was tested on only one battery and that your experience may vary. And also keep in mind this was only tested with a very light load consisting of a DC/DC converter. Working with an inverter and drawing "real" current is still untested.


dalathegreat (approved):
LGTM :+1:

PR #934, opened 2025-03-03
New battery: Add support for 69kWh Volvo XC40 battery

dalathegreat:
### What
This PR implements initial support for the 96S Volvo XC40 69kWh battery

### Why
The current 78kWh SPA battery implementation has some hardcoded stuff for 108S and 78kWh. This caused issues when running with a 69kWh battery

### How
We now initialize to safe values, and once we determine how many cells are present, we switch to the correct value handling


dalathegreat:
Tested OK on the Discord server! :partying_face:

PR #946, opened 2025-03-06
Improvement: Optimize CAN-Native driver and event handling

dalathegreat:
### What
This PR improves low level performance of the CAN driver. It also changes how CAN TX fail events are handled

### Why
Performance improvements are nice :sunglasses: 

### How
By moving out serial printing / event triggering from inside the CAN driver, we increase performance significantly in the event that it would fail to send. The SAFETY.cpp loop now keeps track of every CAN interface, and checks every 1000ms if something has occurred. 

The PR also changes a native CAN TX fail event from Error -> Warning. This was an old assumption that a single TX fail on this CAN channel would be tied to battery, but with the new can_config option the battery is necessarily not connected to the CAN native port. We keep the 60s without CAN from battery as an error event.

### DIFF
--- a/Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
+++ b/Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
@@ -5,20 +5,14 @@
 int ESP32CAN::CANInit() {
   return CAN_init();
 }
-int ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
+bool ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
   static unsigned long start_time;
-  int result = -1;
+  bool result = false;
   if (tx_ok) {
     result = CAN_write_frame(p_frame);
-    tx_ok = (result == 0) ? true : false;
-    if (tx_ok == false) {
-      #ifdef DEBUG_VIA_USB
-      Serial.println("CAN failure! Check wires");
-      #endif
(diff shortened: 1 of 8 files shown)


Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
```diff
@@ -5,20 +5,14 @@
 int ESP32CAN::CANInit() {
   return CAN_init();
 }
-int ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
+bool ESP32CAN::CANWriteFrame(const CAN_frame_t* p_frame) {
   static unsigned long start_time;
-  int result = -1;
+  bool result = false;
   if (tx_ok) {
     result = CAN_write_frame(p_frame);
```
Cabooman, line 12: Since CAN_write_frame returns an int and not a bool, it'd be proper to change the return type to bool for CAN_write_frame the same way it was done for ESP32CAN::CANWriteFrame

Cabooman (changes requested):

Software/src/lib/miwagner-ESP32-Arduino-CAN/ESP32CAN.cpp
dalathegreat: Good catch, fixed in https://github.com/dalathegreat/Battery-Emulator/pull/946/commits/476c54711fc1a0e305d67559082ce936dc108091

Cabooman (approved):

PR #949, opened 2025-03-06
Improvement: Event loop optimization

dalathegreat:
### What
This PR implements optimizes the main loop

### Why
More performance

### How
We perform the following optimizations:
- Clear the void loop
   - Remove the run_event_handling() calls from void loop. Not needed, called at other times in the Event handler
   - Removed the 10s profiling task, since it performs nothing
   - Void loop no longer checks millis() to see if millisOverflowCounter should be incremented. Moved to update_calculated_values() section, which takes less system resources


Cabooman (approved):
LGTM :)

PR #953, opened 2025-03-07
Bugfix: Slow webserver 

dalathegreat:
### What
This PR fixes the webpage slowness causing CAN_OVERRUNs to happen frequently when using the Webserver

### Why
Regression caused between 8.5.0 -> 8.6.0

### How
We revert the AsyncTCP library change with AsyncTCPsock. This library change was a bad idea :see_no_evil:


lenvm (approved):

Cabooman (approved):
LGTM

PR #959, opened 2025-03-10
Remove EE event log

Cabooman:
### What
This PR removes the EEPROM event log

### Why
It is not used and too difficult to maintain

### How
Delete, delete, delete


dalathegreat (approved):
Approving, moving into testing stage!

Issue #960, opened 2025-03-10
[BUG]   Can buffer full event with mcp2515 addon board


ErikssonPer:
Fresh download of version 8.8.0
Lilygo with can addon mcp2515 board
Compiled for Growatt hv inverter and pylon battery

![Image](https://github.com/user-attachments/assets/3e73f1da-1aa9-4da1-a914-d17573047f86)

Have CAN buffer full event but the system is working  like normal when I check can traffic, it both sends and recieves data from inverter and bms and it's currently charging.

This the first time to use addon board for me, had 2x lilygo's before, which worked without events.

dalathegreat:
Hi @ErikssonPer , do you have the Growatt HV inverter, or the Pylon Battery attached to the MCP2515 board?

ErikssonPer:
> Hi [@ErikssonPer](https://github.com/ErikssonPer) , do you have the Growatt HV inverter, or the Pylon Battery attached to the MCP2515 board?

Inverter on MCP2515

dalathegreat:
Good catch, I have submitted a pull request to fix this. We are sending too many CAN messages at once which makes the buffer overflow.

You can test this PR: https://github.com/dalathegreat/Battery-Emulator/pull/980

Direct download link: https://github.com/dalathegreat/Battery-Emulator/archive/refs/heads/bugfix/growatt-CAN-load.zip

Let me know if it works better! :)

ErikssonPer:
> Let me know if it works better! :)

Tested and can confirm it is fixed.

Outstanding piece of software!

dalathegreat:
Thanks for testing, the PR is now merged to main!

PR #980, opened 2025-03-16
Bugfix: Growatt HV: Refactor CAN sending to be performed in batches

dalathegreat:
### What
This PR implements batch sending of the periodic CAN data for Growatt HV protocol

### Why
By trying to send 20 messages at the same time, we risk easily overflowing the CAN send buffer, as seen in https://github.com/dalathegreat/Battery-Emulator/issues/960#issue-2907055552

### How
We now instead send the messages in 5 batches (5x4 messages), with 10ms inbetween each batch send. The same approach has been used on other CAN heavy inverters like Foxess


dalathegreat:
Tested OK by @ErikssonPer :raised_hands:

PR #982, opened 2025-03-17
Update GROWATT-HV-CAN.cpp

ErikssonPer:
Added line,
GROWATT_3110.data.u8[7] = 0x00;  // Clear all bits 

Had problem with the inverter never enabled charging or discharging but dsipalyed values from the bms.
Found that the status bit wasn't correct, tried this and it works good now.


dalathegreat (approved):
Awesome, thanks for the contribution! :raised_hands:

PR #983, opened 2025-03-17
MEB: Improve MEB status output.

mvgalen:
### What
Adds:
BMS_fault_performance
BMS_fault_emergency_shutdown_crash
BMS_error_shutdown_request
BMS_error_shutdown
SOH calculation.

Fixes:
(hopefully) the capacity calculation from Ah to kWh.

### Why
To help with debugging battery packs where the pyro went off, or with precharge issues.
Resolves issue where kWh reported was larger than design capacity.

### How
Adds the four fields to the extended datalayer and to the webpage code.

### DIFF
--- a/Software/src/battery/MEB-BATTERY.cpp
+++ b/Software/src/battery/MEB-BATTERY.cpp
@@ -540,14 +540,20 @@ void update_values_battery() {  //This function maps all the values fetched via 
 
   datalayer.battery.status.real_soc = battery_SOC * 5;  //*0.05*100
 
-  datalayer.battery.status.soh_pptt;
-
   datalayer.battery.status.voltage_dV = BMS_voltage * 2.5;  // *0.25*10
 
   datalayer.battery.status.current_dA = (BMS_current - 16300);  // 0.1 * 10
 
   datalayer.battery.info.total_capacity_Wh =
-      ((float)datalayer.battery.info.number_of_cells) * 3.6458 * ((float)BMS_capacity_ah) * 0.2 * 1.13;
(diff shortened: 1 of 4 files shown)


Software/src/battery/MEB-BATTERY.cpp
```diff
@@ -540,14 +540,20 @@ void update_values_battery() {  //This function maps all the values fetched via
...
-      ((float)datalayer.battery.info.number_of_cells) * 3.6458 * ((float)BMS_capacity_ah) * 0.2 * 1.13;
+      ((float)datalayer.battery.info.number_of_cells) * 3.67 * ((float)BMS_capacity_ah) * 0.2 * 1.02564;
+
+  int Wh_max = 61832 * 0.935 / 1.02564;  // 108 cells
+  if (datalayer.battery.info.number_of_cells <= 84)
+    Wh_max = 48091 * 0.9025 / 1.02564;
+  else if (datalayer.battery.info.number_of_cells <= 96)
+    Wh_max = 82442 * 0.9025 / 1.02564;
+  if (BMS_capacity_ah > 0)
+    datalayer.battery.status.soh_pptt = 10000 * datalayer.battery.info.total_capacity_Wh / Wh_max;
```
dalathegreat, line 556: Maybe good to add a check if the SOH% calculation goes over 100, it should cap at 100. Can happen with crashed packs

dalathegreat (changes requested):

Software/src/battery/MEB-BATTERY.cpp
mvgalen: Ah, I now realise this SOH calculation can go wrong if number of cells is not yet known...

PR #986, opened 2025-03-19
Inverter allows

gwk1:
### What
inverter_allows_contactor_closing is set to true when the inverter requests battery info frame

### Why
For an unknown reason, the inverter doesn't always send the battery control message after startup.


lewurm:
Does that work with your S-Box setup?

gwk1:
> Does that work with your S-Box setup?

Yes, and Mibha verified that this works with the BMW i3 battery.

lewurm (approved):

PR #1002, opened 2025-03-24
Add reported_total_capacity_Wh and use it for SOLAX.

mvgalen:
### What
- Add reported_total_capacity_Wh to datalayer and webpage.
- Use reported_total_capacity_Wh and SOH in SOLAX messages.

### Why
To make sure values (remaining kwh/chargable kwh) Solax inverter computes from total_capacity make sense when SOC scaling is active. 

### How
Calculate reported total from total_capacity by taking soc scaling into account.


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -1061,6 +1061,8 @@ String processor(const String& var) {
     content += "<h4 style='color: white;'>Current: " + String(currentFloat, 1) + " A</h4>";
     content += formatPowerValue("Power", powerFloat, "", 1);
     content += formatPowerValue("Total capacity", datalayer.battery.info.total_capacity_Wh, "h", 1);
+    content += formatPowerValue("Scaled total capacity", datalayer.battery.info.reported_total_capacity_Wh, "h", 1);
```
dalathegreat, line 1064: Since this is only used for one single inverter, I suggest we dont show it on the Webserver. It will confuse everyone that is not using Solax

dalathegreat: I like the option to only show the Scaled values if scaling is turned on. Care to add this to the PR? :)

mvgalen: Sure thing.

Software/src/devboard/webserver/webserver.h
```diff
@@ -103,6 +103,9 @@ void onOTAEnd(bool success);
 template <typename T>
 String formatPowerValue(String label, T value, String unit, int precision, String color = "white");
 
+template <typename T>  // This function makes power values appear as W when under 1000, and kW when over
+String formatPowerValue(T value, String unit, int precision);
```
dalathegreat, line 107: Question, is there a reason we need to have two formatPowerValue functions?

mvgalen: One also has the label and color as arguments. I could make the top one use the bottom one.
For the combined lines I needed  a formatPowerValue function without a header and h4 start and end tags to combine the two values on one line.

dalathegreat: If possible, please try combine the functions, the less amounts of rows in the webserver, the better!

mvgalen: I have already combined the lines in the webserver output (see the screenshot on the discord dev thread). Do you also mean to improve the code?

dalathegreat: Totally optional, I am approving this :+1:

dalathegreat (approved):

PR #1009, opened 2025-03-26
Improvement: Code cleanup :broom: 

dalathegreat:
### What
This PR cleans up the main loop, and makes it faster to execute

### Why
I like efficient code

### How
We perform the following optimizations:

- Remove unnecessary WDT init
   - Fixes "task_wdt: esp_task_wdt_init(515): TWDT already initialized" on bootup in serial log 
- Remove unnecessary variable for timekeeping (saves memory)
- Move time measurement inside 10/1000ms (reduces CPU load slightly)
- Simplify bms_status logging (saves memory)
- Remove ESP32 reset comments
- Simplify pause_state function call, add it as 1s task instead of own dedicated 2s task
- Simplify watchdog setup (saves memory)
- Further simplify profiling 10s task timers (saves both memory and CPU load)

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -56,17 +56,11 @@ unsigned long previousMillis10ms = 0;
 unsigned long previousMillisUpdateVal = 0;
 unsigned long lastMillisOverflowCheck = 0;
 // Task time measurement for debugging and for setting CPU load events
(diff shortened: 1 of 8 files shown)


mvgalen (approved):
LGTM.

PR #1012, opened 2025-03-26
Combine GitHub Action to compile all combinations into one file and reduce triggers

lenvm:
### What
This PR:
- Combines GitHub Actions to compile all combinations into one file.
- Adds missing inverters.
- Splits the GitHub Action into 3 parts, now that additional battery and inverter combinations are tested, to prevent the GitHub error that shows when the matrix expansion exceeds 255.
- Reduces the number of times the GitHub Action is triggered.

### Why
- To keep checking that all combinations compile correctly.
- To reduce the resources needed to perform this check.

### How
- By combining the two existing files into 1 file.
- By adding missing inverters.
- By updating the `on: release: types:` used, to just 1 (`published`), rather than 5 (`created, edited, prereleased, released, published`). To check any commit, the workflow dispatch can be used to trigger the GitHub Action manually.


dalathegreat (approved):
Nice work!

PR #1015, opened 2025-03-27
Bugfix: Tesla, extend pyrofuse event check

dalathegreat:
### What
This PR extends the voltage range check for determining if the Pyrofuse is blown on a tesla battery

### Why
The original check did not catch all cases of blown pyro, since it just checked if voltage is 1.0VDC

### How
The new check verifies if the voltage is between 0.5V-5.0V, and if it is, we fire the event!


mvgalen (approved):
LGTM

PR #1031, opened 2025-04-05
Improvement: Performance and cleanup

dalathegreat:
### What
This PR aims to improve overall performance of the code

### Why
Less chance of timing issues

### How
The following optimizations are done:

- Remove unused int64_t task_times_us for reduced memory usage
- All time measurement places are now wrapped in ifdef FUNCTION_TIME_MEASUREMENT blocks, to prevent unnecessary function calls to micros, reducing CPU usage
- ESP32 platformio file updated to v3.2.0

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -78,18 +78,18 @@ void setup() {
   init_stored_settings();
 
 #ifdef WIFI
-  xTaskCreatePinnedToCore((TaskFunction_t)&connectivity_loop, "connectivity_loop", 4096, NULL,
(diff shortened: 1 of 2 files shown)


mvgalen (changes requested):
Perhaps it would be good to remove the "task_time_us" as parameter name from the loop functions as well?

dalathegreat:
> Perhaps it would be good to remove the "task_time_us" as parameter name from the loop functions as well?

Fixed in [d92d9d1](https://github.com/dalathegreat/Battery-Emulator/pull/1031/commits/d92d9d1fb6b4161ff585894d002b6c30b5378a7e)

mvgalen (approved):
LGTM now.

Issue #972, opened 2025-03-14
[BUG] "The board was reset due to an exception or panic. Inform developers!" upon startup


uitgeslapen:
**Describe the bug**
Was running 8.5.0, upgraded to 8.8.0 via OTA. Noticed the reset after upload took longer as expected, the Lilygo led was yellow. After that MEB battery didn't close contacters. And event log reported "The board was reset due to an exception or panic. Inform developers!"

**To Reproduce**
Upgrade to 8.8.0 using my setup file and hopefully it reproduces...

**Expected behavior**
No "The board was reset due to an exception or panic. Inform developers!" error

**Screenshots**

**Version and settings (please complete the following information):**

[USER_SETTINGS .h.txt](https://github.com/user-attachments/files/19254825/USER_SETTINGS.h.txt)

- Software version: 8.8.0
- Battery used: MEB
- Inverter communication protocol: BYD_CAN
- Hardware used for Battery-Emulator: LILYGO [HW_LILYGO/HW_STARK/HW_DEVKIT]
- CONTACTOR_CONTROL: NO, but am using precharge
- MQTT: No

dalathegreat:
Got this info on Discord, but I have not been able to reproduce the crash yet
```

       0.063 Battery emulator 8.9.dev build Mar 17 2025 19:27:35
       0.067 Event: The board was reset from a power-on event. Normal operation
E (356) task_wdt: esp_task_wdt_init(515): TWDT already initialized
       0.076 Battery BMS state changed to: DISCONNECTED
Guru Meditation Error: Core  1 panic'ed (InstrFetchProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x0000001d  PS      : 0x00060130  A0      : 0x800eaf30  A1      : 0x3ffbd270  
A2      : 0x3ffcadb8  A3      : 0x3ffdb450  A4      : 0x3ffbd278  A5      : 0x00000001  
A6      : 0x3ffcb4a0  A7      : 0x3ffdb430  A8      : 0x800eaeca  A9      : 0x00000001  
A10     : 0x0000006e  A11     : 0x3ffdbda0  A12     : 0x0000001d  A13     : 0x00000000  
A14     : 0x3ffb5a7c  A15     : 0x3ffcad1c  SAR     : 0x00000017  EXCCAUSE: 0x00000014  
[...]

PR #1037, opened 2025-04-06
Crash fix: Remove faulty DEBUG_LOG section

dalathegreat:
### What
This PR fixes a crash (Core  1 panic'ed (LoadProhibited). Exception was unhandled.) that would occur when the DEBUG_LOG option was enabled in the code

### Why
Crashes are bad. Hopefully fixes https://github.com/dalathegreat/Battery-Emulator/issues/972

### How
Quickfix, remove the faulty code section

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -290,10 +290,7 @@ void core_loop(void* task_time_us) {
       datalayer.system.status.wifi_task_10s_max_us = 0;
       datalayer.system.status.mqtt_task_10s_max_us = 0;
     }
-#endif  // FUNCTION_TIME_MEASUREMENT
-#ifdef DEBUG_LOG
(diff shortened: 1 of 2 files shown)


mvgalen (changes requested):
I think the issue is this: logging.printf("Battery%s BMS state changed to: OK\n");
The %s has no matching parameter. It think it would be better to remote the %s.

For MEB the real bms state is very usefull log/debug info.

dalathegreat:
@mvgalen , let's re-implement this in the MEB then, since no other batteries are using the real bms state

It also takes performance to iterate this function at lightning fast speed each core_loop. Would make more sense to just update it in the MEB once it changes.

mvgalen (approved):
We need to move this code to MEB.

PR #1038, opened 2025-04-06
Improvement: Zoe Gen2, feedback fixes

dalathegreat:
### What
This PR improves the ZoeGen2 handling

### Why
User feedback on integration

### How
- We now send ID 0x373 every 100ms instead of 200ms, and with the correct still alive formatting
- We now poll the current every second, instead of every 8.2seconds. The current measurement is very important for the integration
- We now use the 12V measurement, and trigger an event if it goes lower than 11V


mvgalen (approved):
LGTM.

PR #1040, opened 2025-04-06
Feature: Add negative SOC scaling 📉

dalathegreat:
### What
This PR implements the possibility to do negative SOC scaling

### Why
Some inverters restrict the possibility to use the entire battery. For instance with Fronius inverters, it is not possible to go below 5% battery SOC%. This can now be circumvented, by applying negatively scaled SOC.

User requested feature, fixes 1039

### How
The settings page now accepts -10.0% as the new minimum

image

Example, real battery SOC is at 1.2%, but we tell inverter that we are still at 10.2%. This would be perfect for an inverter brand that never allows discharging past 10%.

image


dalathegreat:
Merging this, testing OK :raised_hands:

Issue #1048, opened 2025-04-10
(Feature request) Geely Geometry C 2022


Abedashkr:
Hello, I need your help in creating a programming file for the Arduino application to operate the Geely Geometry C battery from the simulator and operate it for charging and consumption and see all the data.

* This the text document CANBUS file - Geely Geometry C *

[ch1_20250410_130932.txt](https://github.com/user-attachments/files/19686590/ch1_20250410_130932.txt)

![Image](https://github.com/user-attachments/assets/0bf2a4c6-ff9c-4d9e-a8ac-8c618bea07f0)
![Image](https://github.com/user-attachments/assets/7a2941c0-d600-4e3a-9e72-0c956f5bb3db)
![Image](https://github.com/user-attachments/assets/9b879c7c-fd29-492b-af10-3cb3016a9d5e)
![Image](https://github.com/user-attachments/assets/d24739aa-c19f-494a-99ad-5e739c5073fd)
![Image](https://github.com/user-attachments/assets/ce1d5ae6-9ecd-480e-8243-3f9345edd572)
![Image](https://github.com/user-attachments/assets/49edf096-b991-4511-acb5-46df898d4beb)
![Image](https://github.com/user-attachments/assets/ab68c0cf-584d-432e-a143-75d816fff3da)
![Image](https://github.com/user-attachments/assets/8a4d54ba-e51d-4005-b32c-8a65f32658a8)
![Image](https://github.com/user-attachments/assets/ee1b4de1-ffea-4c72-b7cd-0e60d027626f)
![Image](https://github.com/user-attachments/assets/a86a14ed-f24e-42a4-8147-3040628670c9)

dalathegreat:
Hi,
This is great start! From the CAN log you took, I can search for some data in the log, but I have some questions:

- This is the battery only in the log? 
- Can you say what total pack voltage is at the moment? 3XX V?
- Can you say what one cell voltage is? 3###mV?
- How hot is the battery when the log was taken? 10*C? 20*C?

I am guessing the contactors inside the battery does not engage, and we need to send CAN towards the battery. For this, a CAN log from a full vehicle is needed.

PR #1049, opened 2025-04-10
Feature: Add starting point for Geely Geometry C battery :battery: 

dalathegreat:
### What
This PR implements support for [Geely Geometry C ](https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Geely-Geometry-C) battery

### Why
User requested feature. Fixes https://github.com/dalathegreat/Battery-Emulator/issues/1048

### How
WIP integration, further refinement needed

/* TODO
- Contactor closing: CAN log needed from complete H-CAN of Geely Geometry C vehicle. We are not sure what needs to be sent towards the battery yet to get contactor closing working. DTC readout complains that a "Power CAN BUS Data Missing" message is still missing
- Unsure if the current CAN sending routine is enough to keep BMS alive 24/7. Testing needed
- There are a few UNKNOWN PID polls, these need to be decoded
- Some critical values are still missing:
   - Current sensor (Mandatory!)
   - Max charge power (can be estimated)
   - Max discharge power (can be estimated)
   - Cell voltage min/max (not mandatory, but very nice to have)
   - All cell voltages (not mandatory, but very nice to have)
/*


Software/src/battery/GEELY-GEOMETRY-C-BATTERY.cpp
mohammad2510, line 1: how to add these files into my file in my pc ?

mohammad2510: > how to run this code 
i tryed to copy the volvo spa info and try  compilling 
the can worked with volvo cx30 

483505421_1221448709587937_1381445818207545458_n
489116650_1181094000148775_3796616922535086321_n

Software/src/battery/GEELY-GEOMETRY-C-BATTERY.h
```diff
@@ -0,0 +1,220 @@
...
+#define GEELY_GEOMETRY_C_BATTERY_H
+#include "../datalayer/datalayer.h"
+#include "../datalayer/datalayer_extended.h"
+#include "../include.h"
+#include "CanBattery.h"
+
+#define BATTERY_SELECTED
+#define SELECTED_BATTERY_CLASS GeelyGeometryCBattery
+
+#define MAX_PACK_VOLTAGE_70_DV 4420  //70kWh
```
kyberias, line 11: If these are basically private constants used by this battery implementation alone, I recommend moving them in the class's private-block and changing them to "static const int" for example. That way they don't leak accidentally to other code when this header is included if they happen to have the same name.

kyberias (approved):

Software/src/battery/GEELY-GEOMETRY-C-BATTERY.h
dalathegreat: Good suggestion, fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1049/commits/e153be1c77282dc9022c4ab17271a5fa2d406dca

PR #1052, opened 2025-04-13
Bugfix: Add rescaling of CMFA-EV raw SOC

dalathegreat:
### What
This PR implements scaling on the raw SOC value reported by the CMFA EV battery

### Why
Raw SOC value will show 90% when full, so we cannot use it as-is

### How
The software adds two new defines

```
#define MAXSOC 9000  //90.00 Raw SOC displays this value when battery is at 100%
#define MINSOC 500  //5.00 Raw SOC displays this value when battery is at 0%
```

We use these to re-scale the raw SOC to a usable value, using a static version of a stochastic oscillator (https://en.wikipedia.org/wiki/Stochastic_oscillator). This makes it possible to get a nice usable display value out of the raw value.

### DIFF
--- a/Software/src/battery/CMFA-EV-BATTERY.cpp
+++ b/Software/src/battery/CMFA-EV-BATTERY.cpp
@@ -5,12 +5,6 @@
 #include "../devboard/utils/events.h"
 #include "CMFA-EV-BATTERY.h"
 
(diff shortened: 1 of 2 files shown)


danpadure:
looks good

PR #1057, opened 2025-04-13
Atto3: Add crashed BMS unlocking :lock: 

dalathegreat:
### What
This PR makes it possible to unlock crashed BYD Atto3 batteries :closed_lock_with_key: 

### Why
When Atto 3 EVs are crashed hard enough to trigger airbags, the BMS will lock itself and prevent contactor closing. The SOC% will also stop updating. This is problematic, and users of these batteries have suffered. Previous workarounds have been to bypass contactors, and estimate SOC% from voltage.

This is no longer a problem, since we now have the OEM commands to unlock crashed batteries :raised_hands: 

### How
This feature adds a button in the "More Battery Info" page in the webserver, that when pressed will unlock the crashed battery.

image

 After battery is unlocked, you can remove the "USE_ESTIMATED_SOC" from the BYD-ATTO-3-BATTERY.h file. Enjoy the easy contactor closing that this feature brings :raised_hands: 

Massive thanks to EVL on the Discord for providing the log files from a BYD VDS2100 Diagnostic Tester Scan Tool

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -7,15 +7,17 @@
 
 /* Notes
   - SOC% by default is now ESTIMATED.
-  - If you have a non-crashed pack, enable using real SOC. See Wiki for info.
-  - TODO: In the future, we might be able to unlock crashed batteries and get SOC going always
+  - If you have a crash-locked pack, run the "Unlock crashed BMS" from the "More Battery Info" page
+  - After battery is unlocked, you can remove the "USE_ESTIMATED_SOC" from the BYD-ATTO-3-BATTERY.h file
 */
 
 /* Do not change code below unless you are sure what you are doing */
 #define NOT_DETERMINED_YET 0
 #define STANDARD_RANGE 1
 #define EXTENDED_RANGE 2
+#define NOT_RUNNING 0xFF
 static uint8_t battery_type = NOT_DETERMINED_YET;
+static uint8_t stateMachineClearCrash = NOT_RUNNING;
 static unsigned long previousMillis50 = 0;   // will store last time a 50ms CAN Message was send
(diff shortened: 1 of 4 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -455,6 +476,27 @@ void transmit_can_battery() {
 #ifdef DOUBLE_BATTERY
     transmit_can_frame(&ATTO_3_441, can_config.battery_double);
 #endif  //DOUBLE_BATTERY
+    switch (stateMachineClearCrash) {
+      case 0:
```
lenvm, line 480: To make the code easier to read and maintain, perhaps use an enum for the different cases.

dalathegreat: Good idea, fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1057/commits/94c813f3adebbaf6dd302b32fe975a907fc12b20

mvgalen (commented):
Does the pack report somewhere that it is locked? If so, it would make sense to make estimated soc depend on that value, instead of requiring a recompile after unlocking the battery.

lenvm (approved):

mvgalen (approved):
Does the pack report somewhere that it is locked? If so, it would make sense to make estimated soc depend on that value, instead of requiring a recompile after unlocking the battery.
Since we do not have this info, we have to keep the current setup.

dalathegreat:
I have updated the Wiki with all the info we have so far on the Battery Unlock. To avoid this PR hanging forever, and get more testers on it, I will merge this PR now.

Wiki section: https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-BYD-Atto-3-%E2%80%90-Yuan-Plus#how-do-i-unlock-a-crashed-battery

PR #1060, opened 2025-04-14
Improved SoC perc-voltage matrix, limits, added internal resistance c…

ilia92:
Improved SoC perc-voltage matrix, limits, added internal resistance compensation feature

### What
1. The percentage-voltage matrix limit is changed down to 3V, because people with E-GMP batteries reported dangerous imbalance at low SoC (incl. me) . Real results: Under 3.3V >150mV deviation, triggering a event. 
Also, we must avoid hitting MIN_CELL_VOLTAGE_MV 2950 due to normal operation. 
custom_battery_curve

2. This PR implements compensation of estimating SoC, based on the battery internal resistance.
e.g if we pull energy and the voltage drop is for instance 10V and 10V means 2% difference, with this, the reported SoC will be exactly same at idle and at 10A (until of course the real energy drops). Similar for charging

### Why
Improvement

### How
N/A


Software/src/battery/KIA-E-GMP-BATTERY.cpp
```diff
@@ -66,23 +66,24 @@ static uint8_t EGMP_3XF_counter = 0;
...
+                            3350, 3358, 3367, 3375, 3384, 3392, 3401, 3410, 3418, 3427, 3435, 3444, 3452, 3461, 3470,
+                            3478, 3487, 3495, 3504, 3512, 3521, 3530, 3538, 3547, 3555, 3564, 3572, 3581, 3590, 3598,
+                            3607, 3615, 3624, 3632, 3641, 3650, 3658, 3667, 3675, 3684, 3692, 3701, 3710, 3718, 3727,
+                            3735, 3744, 3752, 3761, 3770, 3778, 3787, 3795, 3804, 3812, 3821, 3830, 3838, 3847, 3855,
+                            3864, 3872, 3881, 3890, 3898, 3907, 3915, 3924, 3932, 3941, 3950, 3953, 3959, 3969, 3980,
+                            3993, 4007, 4023, 4041, 4060, 4080, 4102, 4124, 4148, 4173, 4200};
+
+// Function to estimate SOC based on cell voltage
+uint16_t estimateSOCFromCell(uint16_t cellVoltage) {
   if (cellVoltage >= voltage[0]) {
```
dalathegreat, line 87: You have re-arranged the voltage table from low-high. This check originally checked if cellvoltage is over 4200mV, and if so, sets soc to 100%. Now it incorrectly checks if cellvoltage is over 3000mV, and sets soc to 0% if the voltage is over 3000mV

dalathegreat (approved):

PR #1066, opened 2025-04-16
New web features - API endpoints and /metrics Prometheus

ilia92:
What
This PR implements simple API READ endpoints
and /metrics for Prometheus - https://prometheus.io/docs/introduction/overview/

Why
Why does it do it?
Extending the functionality of BE

How
How does it do it?
#define API_ENDPOINTS //When enabled, json API endpoints are available
and
#define PROMETHEUS_METRICS //When enabled, /metrics will be available for Prometheus scraping

How to use it:
/api/status - Status of the BE
Example:
{
"system": {
"version": "8.11.api",
"uptime": "1d 8h 32m 49s",
"inverter_protocol": "Pylontech battery over CAN bus",
"inverter_brand": "",
"battery_protocol": "Kia/Hyundai EGMP platform",
"temperature": 38.89,
"emulator_status": "RUNNING",
"equipment_stop": false,
"led_status": "GREEN"
},
"battery": {
"soc_real": 67.66,
"soc_reported": 67.66,
"soc_scaling_active": false,
"soh": 100,
"voltage": 729,
"current": -4.6,
"power": -3312,
"capacity_total_wh": 77100,
"capacity_reported_wh": 77000,
"capacity_remaining_wh": 52165,
"capacity_reported_remaining_wh": 52165,
"max_discharge_power": 20000,
"max_charge_power": 20000,
"max_discharge_current": 25,
"max_charge_current": 25,
"user_settings_limit_discharge": true,
"user_settings_limit_charge": true,
"inverter_limits_discharge": true,
"inverter_limits_charge": true,
"cell_min_voltage": 3760,
"cell_max_voltage": 3800,
"cell_delta": 40,
"cell_delta_limit": 150,
"temp_min": 10,
"temp_max": 12,
"bms_status": 3,
"bms_status_text": "ACTIVE",
"contactor_status": true,
"battery_allows_contactor_closing": true,
"inverter_allows_contactor_closing": true
}
}

/api/cells - Only cells voltage info (for each cell)
{
"cells": [
{
"index": 1,
"voltage": 3780
},
{
"index": 2,
"voltage": 3780
},
{
"index": 3,
"voltage": 3780
},
{
...

/api/events - Events
Example:
{
"events": [
{
"id": 67,
"description": "The board was reset via software, webserver or OTA. Normal operation",
"timestamp": 98,
"level": "INFO",
"occurences": 1,
"active": true
},
{
"id": 83,
"description": "Wifi connected.",
"timestamp": 194,
"level": "INFO",
[...]

PR #1072, opened 2025-04-18
New web features - API endpoints and /metrics Prometheus

ilia92:
I hope this time checks will succeed :)

Improved code after this: https://github.com/dalathegreat/Battery-Emulator/pull/1066

What
This PR implements simple API READ endpoints
and /metrics for Prometheus - https://prometheus.io/docs/introduction/overview/

Why
Why does it do it?
Extending the functionality of BE

How
How does it do it?
#define API_ENDPOINTS //When enabled, json API endpoints are available
and
#define PROMETHEUS_METRICS //When enabled, /metrics will be available for Prometheus scraping

How to use it:
/api/status - Status of the BE
Example:
{
"system": {
"version": "8.11.api",
"uptime": "1d 8h 32m 49s",
"inverter_protocol": "Pylontech battery over CAN bus",
"inverter_brand": "",
"battery_protocol": "Kia/Hyundai EGMP platform",
"temperature": 38.89,
"emulator_status": "RUNNING",
"equipment_stop": false,
"led_status": "GREEN"
},
"battery": {
"soc_real": 67.66,
"soc_reported": 67.66,
"soc_scaling_active": false,
"soh": 100,
"voltage": 729,
"current": -4.6,
"power": -3312,
"capacity_total_wh": 77100,
"capacity_reported_wh": 77000,
"capacity_remaining_wh": 52165,
"capacity_reported_remaining_wh": 52165,
"max_discharge_power": 20000,
"max_charge_power": 20000,
"max_discharge_current": 25,
"max_charge_current": 25,
"user_settings_limit_discharge": true,
"user_settings_limit_charge": true,
"inverter_limits_discharge": true,
"inverter_limits_charge": true,
"cell_min_voltage": 3760,
"cell_max_voltage": 3800,
"cell_delta": 40,
"cell_delta_limit": 150,
"temp_min": 10,
"temp_max": 12,
"bms_status": 3,
"bms_status_text": "ACTIVE",
"contactor_status": true,
"battery_allows_contactor_closing": true,
"inverter_allows_contactor_closing": true
}
}

/api/cells - Only cells voltage info (for each cell)
{
"cells": [
{
"index": 1,
"voltage": 3780
},
{
"index": 2,
"voltage": 3780
},
{
"index": 3,
"voltage": 3780
},
{
...

/api/events - Events
Example:
{
"events": [
{
"id": 67,
"description": "The board was reset via software, webserver or OTA. Normal operation",
"timestamp": 98,
"level": "INFO",
"occurences": 1,
"active": true
},
{
"id": 83,
"description": "Wifi connected.",
"timestamp": 194,
"level": "INFO",
"occurences": 1,
"active": false
}
]
}

/api/full - Full info, combining above

/metrics - Prometheus-ready endpoint fof scraping the data

Tested!

Additional info:
Cost:
with API: Sketch uses 1202168 bytes (61%)
without: Sketch uses 1193116 bytes (60%)

Once merged, i will write a wiki documentation


ilia92:
Unblvbl, unit tests passed :) I'm also currently testing it (main code since few days) and works pretty good.

Working exporter / http://battery-emulator/metrics on every minute / + some improvised graphs:
image

Software/src/devboard/webserver/api/api_helpers.h
```diff
@@ -0,0 +1,38 @@
+#ifndef API_HELPERS_H
+#define API_HELPERS_H
+
+#include <Arduino.h>
+
+// Helper function to add a field to the JSON string
+inline void addJsonField(String& json, const String& key, const String& value, bool addComma = true) {
```
kyberias, line 7: I think ArduinoJSON is already included in battery emulator and probably should be used instead of implementing custom JSON manipulation functions.

Software/src/devboard/webserver/api/api_status_html.cpp
```diff
@@ -0,0 +1,207 @@
+#include "api_status_html.h"
```
kyberias, line 1: If this API provides JSON content, perhaps the name of the file should not be "HTML".

kyberias:
My 0.02€... 

Adding JSON API would make more sense if the web UI also would use it instead of rendering directly to HTML. Adding JSON API will effectively add identical code that renders the same thing as the HTML API but instead in JSON format. 

Prometheus metrics API adds value only for Prometheus users. I'm not sure it makes much sense in adding it to a project like battery emulator. Even though it can be conditionally compiled, it would still add maintenance burden. Aren't there some external  solutions that bridge JSON API to Prometheus?

dalathegreat:
Yes, I think we can close this for the time being, and instead work on making the Webserver better in general (Not having to F5 the whole page to get values to update would be an amazing start) 

But still, very nice work @ilia92 :raised_hands:

PR #1073, opened 2025-04-19
Common standard image build

kyberias:
### What
This proof-of-concept PR includes the changes necessary to build all battery and inverter protocol implementations into one binary and allows battery and inverter to be selected at run-time in emulator settings. It still also supports selecting a single battery or inverter at build-time as before.

Great care was taken to retain the existing structure of the code: no unnecessary files were added or removed and changes were targeted in existing places when possible.

I recommend this POC PR to be used as a basis for the final implementation.

Notable missing features:

- Some battery / inverter combinations have GPIO conflicts that should be handled at run-time. The current build-time checks prevent compilation of all batteries.
- Some BMS not included in the build
- Error handling in out-of-memory situations
- Support for other HW than Lilygo
- Complete charger selection

### Why
A common image would make it much easier to take battery emulator in use since it ideally would not require users to build the image at all, simply transfer it to their HW of choice. This would be a much more robust and reliable experience for the users. There are also obvious benefits to code maintainability and reuse. This also provides new opportunities to refactor other parts of the implementation to make much more settings at run-time.

### How
Existing implementation already has clear separation of battery/inverter implementation into C-style modules but since the interface was with global functions (e.g. transmit_can_battery), simply building all the modules into binary would introduce linker conflicts. This PR introduces some **abstract base classes** for batteries and inverters that the battery and inverter modules can then implement in a concrete class.

User can select a battery and the identifier of the battery type is stored in settings. When the emulator starts it constructs a concrete battery class based on this selection. The main loop may then use the abstract interface to call battery dependent functions as before. Same mechanism is used for inverter protocols.

Earlier code had a lot static data in battery implementations. Compiling all this in would overflow RAM budget, so some of that is moved to battery class's private members.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -9,6 +9,8 @@
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
 
+#include "src/battery/Battery.h"
+
 #include "src/communication/can/comm_can.h"
 #include "src/communication/contactorcontrol/comm_contactorcontrol.h"
 #include "src/communication/equipmentstopbutton/comm_equipmentstopbutton.h"
@@ -66,6 +68,9 @@ TaskHandle_t mqtt_loop_task;
 
 Logging logging;
 
+InverterProtocol* inverter;
+BatteryBase* battery;
+
 // Initialization
 void setup() {
   init_serial();
@@ -92,20 +97,43 @@ void setup() {
                           WIFI_CORE);
 #endif
 
-  init_CAN();
+// When a single battery/inverter is configured, we don't care what user may have selected earlier,
+// we must use the one and only type.
+#ifndef BUILD_EM_ALL
+  userSelectedBatteryType = *(supported_battery_types().begin());
+  userSelectedInverter = *(supported_inverter_protocols().begin());
+#endif
+
+  inverter = init_inverter_protocol(userSelectedInverter);
+  if (inverter == nullptr) {
+    // TODO: Handle error, this could be out-of-memory
+  }
+
+  battery = init_battery(userSelectedBatteryType);
+  if (battery == nullptr) {
+    // TODO: Handle error, this could be out-of-memory
+  }
+
+  if (inverter->usesCAN() || battery->usesCAN()) {
+    init_CAN();
+  }
 
   init_contactors();
 
 #ifdef PRECHARGE_CONTROL
-  init_precharge_control();
+
+  if (battery->requiresPrechargeControl()) {
+    init_precharge_control();
+  }
 #endif  // PRECHARGE_CONTROL
 
-  init_rs485();
+  if (inverter->usesRS485() || battery->usesRS485()) {
+    init_rs485();
+  }
 
-#if defined(CAN_INVERTER_SELECTED) || defined(MODBUS_INVERTER_SELECTED) || defined(RS485_INVERTER_SELECTED)
-  setup_inverter();
-#endif
-  setup_battery();
+  inverter->setup();
+  battery->setup();
+
 #ifdef EQUIPMENT_STOP_BUTTON
   init_equipment_stop_button();
 #endif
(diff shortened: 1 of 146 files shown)


kyberias:
I've figured out how to handle double battery with this design so that each battery has its own object.

Software/Software.ino
```diff
@@ -511,15 +574,17 @@ void update_calculated_values() {
...
-  update_values_can_inverter();
-#endif  // CAN_INVERTER_SELECTED
-#ifdef MODBUS_INVERTER_SELECTED
-  update_modbus_registers_inverter();
-#endif  // CAN_INVERTER_SELECTED
-#ifdef RS485_INVERTER_SELECTED
-  update_RS485_registers_inverter();
-#endif  // CAN_INVERTER_SELECTED
+  if (inverter->usesCAN()) {
+    inverter->update_values_can_inverter();
```
kyberias, line 578: Looks like we could just have inverter implement a single update_values() method that the main loop here calls and the inverter implementation then decides what to do. The main loop doesn't need to know here whether it's CAN or RS485 or something else.

Software/src/battery/BATTERIES.h
```diff
@@ -147,17 +188,6 @@ void setup_can_shunt();
 #include "VOLVO-SPA-HYBRID-BATTERY.h"
 #endif
 
-void setup_battery(void);
```
kyberias, line 150: These no longer exist since we're using the Battery and InverterProtocol interface to call.

Software/src/battery/BOLT-AMPERA-BATTERY.cpp
```diff
@@ -19,114 +21,6 @@ TODOs left for this implementation
 - SOH% (low prio))
 */
 
-/*TODO, messages we might need to send towards the battery to keep it happy and close contactors
```
kyberias, line 22: Here the static variables are moved to the respective battery's class header file to minimize static data. There are many others but not for all battery implementations.

Software/src/battery/BOLT-AMPERA-BATTERY.h
```diff
@@ -176,7 +176,125 @@
 #define POLL_7E7_CELL_119 0x4257
 #define POLL_7E7_CELL_120 0x4258
 
-void setup_battery(void);
-void transmit_can_frame(CAN_frame* tx_frame, int interface);
+class BoltAmperaBattery : public CanBattery {
+ public:
+  BoltAmperaBattery() : CanBattery(BoltAmpera) {}
+  virtual const char* name() { return Name; };
```
kyberias, line 182: The name is provided directly by the battery object. This eliminates the redundant copying of strings to datalayer.

(BTW there are many other such data that could be provided by the battery object directly and not copied.)

Software/src/battery/BOLT-AMPERA-BATTERY.cpp
```diff
@@ -813,9 +707,7 @@ void transmit_can_battery() {
   }
 }
 
-void setup_battery(void) {  // Performs one time setup at startup
-  strncpy(datalayer.system.info.battery_protocol, "Chevrolet Bolt EV/Opel Ampera-e", 63);
```
kyberias, line 817: Copying of strings eliminated. See below.

No-Signal:
> I've figured out how to handle double battery with this design so that each battery has its own object.

If we can move from a single battery to a battery array then it could easily scale to additional batteries rather than a single option of double battery

kyberias:
> > I've figured out how to handle double battery with this design so that each battery has its own object.
> 
> If we can move from a single battery to a battery array then it could easily scale to additional batteries rather than a single option of double battery

Since the rest of the code is currently hard-code with exactly two batteries, I probably won't implement an array, but that is of course possible once these changes have been implemented.

Software/USER_SETTINGS.h
```diff
@@ -8,6 +8,14 @@
 /* There are also some options for battery limits and extra functionality */
 /* To edit battery specific limits, see also the USER_SETTINGS.cpp file*/
 
+// Defining BUILD_EM_ALL builds in support for all batteries and inverter protocols
+// and allows selecting them at run-time.
+#define BUILD_EM_ALL
+
+#ifdef BUILD_EM_ALL
+#define HW_LILYGO
```
lenvm, line 16: Would it be possible to use the BUILD_EM_ALL flag with other hardware as well? I would suggest to make the `BUILD_EM_ALL` macro compatible with all supported hardware, and not hardcode that it must be used with Lilygo hardware.

dalathegreat (approved):
Extremely nice work! Let's get the branch in shape and start testing! :raised_hands:

PR #1084, opened 2025-04-22
Moving battery min max values into base class

No-Signal:
### What
This PR moves battery min max values into the base class. This removes many warnings for redefining values and will allow future simplification of the datalayer

### DIFF
--- a/Software/src/battery/Battery.h
+++ b/Software/src/battery/Battery.h
@@ -66,8 +66,12 @@ class BatteryBase {
   virtual void receive_RS485() = 0;
 
   //5000 = 500.0V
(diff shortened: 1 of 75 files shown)


Software/src/battery/Battery.h
```diff
@@ -65,6 +65,18 @@ class BatteryBase {
   virtual void transmit_RS485() = 0;
   virtual void receive_RS485() = 0;
 
+  //5000 = 500.0V
+  virtual uint16_t max_cell_voltage_mv() {
```
kyberias, line 69: Would it actually make sense to make these abstract, that is not define default zero and force the battery implementations to make a decision?

PR #1085, opened 2025-04-23
BYD Modbus: Update FW and Serial number

dalathegreat:
### What
This PR updates the BYD reported FW/Serial data

### Why
To increase compatibility and not appear as an old battery from 2021

This PR also fixes the issue that Fronius Software "ROW 1.36.5-1" released on April 23rd breaks compatibility with the Battery-Emulator project.

### How
We update the FW version from 3.16 -> 3.26. We also add a default serial number to the data field.

image

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -30,9 +30,9 @@ void handle_static_data_modbus_byd() {
   static uint16_t byd_data[] = {16985, 17408, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
   static uint16_t battery_data[] = {16985, 17440, 16993, 29812, 25970, 31021, 17007, 30752,
                                     20594, 25965, 26997, 27936, 18518, 0,     0,     0};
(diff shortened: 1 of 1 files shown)


pmajor021:
Date is invalid. 
I deciphered the data: P030 T020 Z200800 348834, where YY:20, MM:08, DD:00. Also the last part shouldn't be a 4 digit value?

dalathegreat:
@pmajor021 I have updated the PR, new serial is:

"P030T020Z1911180001"

SW "3.26"

One thing I wonder still about is the extra 00000 padding in serial_data

image

pmajor021:
Latest FW for BMU 3.26, and for BMS is 3.31.

```
#!/usr/bin/env python3

import pprint
import requests

pp = pprint.PrettyPrinter(depth=6)

r = requests.post('https://deapp.byd.com/api/V1/Http2/GetFileNewLink', json={"SoftVersion": "V1.0"})
print(r.status_code)
pp.pprint(r.json())
```

dalathegreat:
Like discussed on Discord, this fix is now working perfectly and allowing the Emulator to function with the latest Fronius software. Let's merge this as-is, and continue in another PR to get the SN to show up via API! :raised_hands:

PR #1090, opened 2025-04-26
Refactoring: Simplify can sending routine and performance check

dalathegreat:
### What
This PR simplifies CAN sending routine in each integration. It also renames the CAN_OVERRUN event to TASK_OVERRUN with a more reassuring info text

### Why
To increase performance
To reduce complexity
To reduce support load regarding the CAN_OVERRUN events

### How
Performance is increased by reducing amount of millis() calls
TASK_OVERRUN is checked in main loop instead of each CAN sending routine


lenvm (approved):

mvgalen (approved):
LGTM.

PR #1094, opened 2025-04-28
SMA: Add Event for missing enable line

dalathegreat:
### What
This PR adds an event, to detect if Enable line has been missing for a long time

### Why
To aid troubleshooting

### How
If we don't detect any activity on the enable line for 30 minutes, we raise an Info event;

`INFO - Inverter has not toggled Enable signal for a long time. Check Wiring!`


Software/src/devboard/utils/events.cpp
```diff
@@ -269,6 +270,8 @@ const char* get_event_message_string(EVENTS_ENUM_TYPE event) {
              "Check other error code for reason!";
     case EVENT_MODBUS_INVERTER_MISSING:
       return "Modbus inverter has not sent any data. Inspect communication wiring!";
+    case EVENT_NO_ENABLE_DETECTED:
+      return "Inverter has not toggled Enable signal for a long time. Check Wiring!";
```
lenvm, line 274: I think “enabled” is a better word than “toggled”, as toggled seems to indicate 0 —> 1 —> 0

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -119,6 +120,17 @@ static void dbg_contactors(const char* state) {
 void handle_contactors() {
 #if defined(SMA_BYD_H_CAN) || defined(SMA_BYD_HVS_CAN) || defined(SMA_TRIPOWER_CAN)
   datalayer.system.status.inverter_allows_contactor_closing = digitalRead(INVERTER_CONTACTOR_ENABLE_PIN);
+
+  if (!datalayer.system.status.inverter_allows_contactor_closing) {
+    timeSpentWithoutEnableTurnedOn++;
+
+    if (timeSpentWithoutEnableTurnedOn >
+        180000) {  //Increments once every 10ms. 100/second*60seconds*30minutes = 180000
```
lenvm, line 128: I think it would be better to use a slower loop for this counter. What do you recon?

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -119,6 +120,17 @@ static void dbg_contactors(const char* state) {
 void handle_contactors() {
 #if defined(SMA_BYD_H_CAN) || defined(SMA_BYD_HVS_CAN) || defined(SMA_TRIPOWER_CAN)
   datalayer.system.status.inverter_allows_contactor_closing = digitalRead(INVERTER_CONTACTOR_ENABLE_PIN);
+
+  if (!datalayer.system.status.inverter_allows_contactor_closing) {
```
lenvm, line 124: “inverter allows contactor closing” and “enable turned on” is not really consistent. I’d suggest to choose one, and use it consistently.

Probably something like “inverterBatteryEnableOn” could work, starting with the sender, then the content, and finally the state

dalathegreat: No it has to be inverter_allows_contactor_closing, in order for GPIO contactor control to open contactors if the signal goes low

Software/src/devboard/utils/events.cpp
dalathegreat: Improved in https://github.com/dalathegreat/Battery-Emulator/pull/1094/commits/792eac2858e950013704e70710b05ce925cf7598

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
dalathegreat: Improved in https://github.com/dalathegreat/Battery-Emulator/pull/1094/commits/84a5fdf580d561a7d9492a756978a3c9501dc0eb , 10ms -> 1000ms,  this also takes more into consideration the compile all inverter combos, by checking it in each inverter protocol that requires it!

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
lenvm: Do you mean it needs to be that signal, and not another one?

I’m suggesting to align the naming of the signals

dalathegreat: Yes it needs to be the inverter_allows_contactor_closing which the whole code is already based on to determine contactor closing. The Enable line is used to toggle on/off HV. So we write the enable line status to the inverter_allows_contactor_closing to make the contactors open up when it goes low

lenvm: Yes, I’m aware of that!

Perhaps just change `timeSpentWithoutEnableTurnedOn` to something like `timeWithoutInverterAllowsContactorClosing`

dalathegreat: Fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1094/commits/576e8af921b4968b1f2acba5cb0cec204e561a14

lenvm (approved):

PR #1095, opened 2025-04-28
Bugfix: Edit SMA inverter input pin for Stark CMR

dalathegreat:
### What
This PR changes the default pin used for SMA Enable line signal

### Why
Initial implementation picked a free GPIO (pin19). This pin is not optimal, since the Stark CMR has a dedicated SIGNAL IN pin (pin2) that can take the 11V input directly

### How
19 changes to 2


StarkJohan (approved):
LGTM. Now it corresponds to the SIGNAL IN GPIO which is the best way to handle this SMA signal.

PR #1096, opened 2025-04-28
Common base class for batteries

kyberias:
### What
This PR implements a minimal, abstract base class, which is a new type of battery API. When later implemented by all batteries, it allows all battery code to be compiled into common image. This first stage implements a minimal change: one battery implementation (Zoe Gen1) changed to this design.

This PR contains no functionality changes.

### Why
A base class approach is needed to gradually convert the code base to support common image.

### How

CanBattery base class is an abstract base class for batteries (supporting CAN bus). The API is identical to the C-style API but allows in the future multiple run-time implementations of batteries.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -1,0 +1,30 @@
+#include "BATTERIES.h"
+#include "../include.h"
+
+// These functions adapt the old C-style global functions battery-API to the
+// object-oriented battery API.
+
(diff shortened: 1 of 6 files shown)


kyberias:
Looks like a build check fails due to some tool install, and not related to PR content. Could someone with write permissions start a rerun, please? @dalathegreat

dalathegreat (approved):
LGTM :+1: Great first step towards building all at once!

PR #1098, opened 2025-04-30
Common base class for inverter protocols

kyberias:
### What
This PR implements a minimal, abstract base class for **inverter protocol** implementations. When later implemented by all inverter protocols, it allows all inverter code to be compiled into common image. This first stage implements a minimal change: one inverter implementation (BYD Modbus) changed to this design with relevant adaptation code.

This PR contains no functionality changes.

### Why
A base class approach for inverter protocols is needed to gradually convert the code base to support common image.

### How
InverterProtocol and ModbusInverterProtocol abstract classes are the base classes for batteries (first supporting only Modbus). The API is identical to the C-style API but allows in the future multiple run-time implementations of inverter protocols.


dalathegreat (approved):
Nice work! :+1:

PR #1100, opened 2025-05-01
Use common battery base class for Jaguar Ipace battery

kyberias:
Convert Jaguar I-PACE battery implementation to use the common base class. The concrete battery class selected is indicated by SELECTED_BATTERY_CLASS define.


dalathegreat (approved):
Bit by bit! :+1:

PR #1101, opened 2025-05-01
BYD Atto 3: Tweak SOC voltage estimation lookup tables + More PID polls!

dalathegreat:
### What
This PR tweaks the SOC% voltage lookup tables + add new PID polls  + removes estimated charge speed in favor of actual BMS allowed charge speed

### Why
Voltage table: To avoid large cell deviations that occur when targeting the same pack voltage as the car would charge the cells to

PID polls: To gain more info about the BYD Atto 3 batteries

Charge allowed W: To make safer end of charge rampdown, without making the LFP cells imbalanced

### How
Preface, SOC% estimation via voltage is bad, especially bad when it comes to LFP cells, which have an extremely flat voltage curve. If you have a locked Atto3 battery, try to unlock it with the unlock command script to start with.

The new PID polls (14x) are available via the "More Battery Info" page

One of the new PID polls are "Allowed Charge Power". By using this value, we avoid getting large cell deviations at 100% state of charge, since we ramp down the power when we approach full battery.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -91,12 +91,17 @@ CAN_frame ATTO_3_7E7_POLL = {.FD = false,
                              .data = {0x03, 0x22, 0x00, 0x05, 0x00, 0x00, 0x00, 0x00}};
 
 // Define the data points for %SOC depending on pack voltage
-const uint8_t numPoints = 14;
-const uint16_t SOC[numPoints] = {10000, 9970, 9490, 8470, 7750, 6790, 5500, 4900, 3910, 3000, 2280, 1600, 480, 0};
-const uint16_t voltage_extended[numPoints] = {4400, 4230, 4180, 4171, 4169, 4160, 4130,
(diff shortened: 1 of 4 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -161,7 +255,7 @@ void update_values_battery() {  //This function maps all the values fetched via
 
   datalayer.battery.status.max_discharge_power_W = MAXPOWER_DISCHARGE_W;  //TODO: Map from CAN later on
 
-  datalayer.battery.status.max_charge_power_W = MAXPOWER_CHARGE_W;  //TODO: Map from CAN later on
+  datalayer.battery.status.max_charge_power_W = BMS_allowed_charge_power * 10;  //Scaling unknown, *10 best guess
```
lenvm, line 258: Perhaps add a `FIXME:` here

dalathegreat: Fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1101/commits/d5976e36727dac424f73c1bd5c3d937ca404341f

PR #1106, opened 2025-05-03
Microfeature: Add high temperature fault event :thermometer: 

dalathegreat:
### What
This PR adds a high temperature Fault event

### Why
To be able to perform a graceful shutdown incase the Battery-Emulator overheats. It is better to stop gracefully, than wait for CPU to crash.

### How
If the CPU temperature exceeds 110°C, we Fault the emulator. If we are able to cool down below 105°C, we resume operation.


mvgalen (approved):
LGTM.

Issue #926, opened 2025-03-01
[BUG] BMW I3  double,  data not colecting.


havrla:
Situation:
Double battery setup: the first LilyGO communicates with the batteries, while the second one communicates with the inverter.

1. lilygo double battery 50A balancing 25 and 25A , perfect .
![Image](https://github.com/user-attachments/assets/a4d76a0f-4945-4036-89f1-f6ae73b25b00)

2. lilygo  reporting only first battery.   2. lily must reporting colecting data to Invertor .
![Image](https://github.com/user-attachments/assets/d3930566-9e93-4dc9-ab53-079e0febd042)  

Please fix,  2. lilygo must reporting colecting data to invertor.
SOC Management:
    0–50%: The battery with the lower SOC is used.
    50–100%: The battery with the higher SOC is used.

Energy in Wh – simply sum the values.

Temperatures – use the highest and lowest values from the entire system.

Charging and Discharging Power Strategy
This is an interesting strategy. Fortunately, the battery has sufficient power, exceeding that of standard inverters, so we can simplify the approach.
    1.Maximum charging and discharging power will be reported from the battery that reports the lower value.
    2. If a better strategy is proposed, it can be considered, but the minimum strategy is:
        If the reported power value is 0, external charging should be stopped.

Regarding the values of individual cells, compare all cells (i.e., 2 × 96), and report the maximum and minimum values.

Note: Some BMS/batteries from around 2013 do not immediately send valid data after startup. Instead, they start transmitting accurate values after approximately 1 minute to 1 hour.
For example, in the first screen, the second battery reports 3.7V per cell, but this is not the real value—it’s just a default value for the emulator until it reads the correct data from CAN. Similarly, an SOH of 99% is also incorrect but eventually corrects itself over time.

thanks thanks

havrla:
SOC Management:
0–50%: The battery with the lower SOC is used.
50–100%: The battery with the higher SOC is used.
Maybe propably functional:

Left battery present SOC
![Image](https://github.com/user-attachments/assets/e2a4f070-e985-4e40-b728-fe3f8a509761)

Right battery present SOC
![Image](https://github.com/user-attachments/assets/ecf2c7aa-994e-4d0e-8865-8a12ed98a374)

dalathegreat:
Closing as obsolete, much has happened since FW 8.7!

PR #1110, opened 2025-05-05
Bugfix: Scaled capacity overhaul

### LINKED ISSUE #396: Double Battery - Capacity handling
The total kWh number should be dynamic, according to how many batteries are joining in. Current implementation only uses battery 1 kWh numbers and sends this to the inverter

### LINKED ISSUE #1104: Scaled remaining capacity: 0 Wh when 95 and -5% is used
**Describe the bug**
Scaling to max 95% and min -5% SOC shows scaled remaining capacity as 0 Wh.

**To Reproduce**
Set scaling to max 95% and min -5%

**Expected behavior**
Showing the correct scaled remaining capacity.

**Screenshots**

Image
Image

- Please attach your USER_SETTINGS files for easier debugging
- Software version: 8.12.dev (also the same with 8.11)
- Battery used: BYD Atto 3
- Inverter communication protocol: Solax
- Hardware used for Battery-Emulator: HW_STARK
- CONTACTOR_CONTROL: no
- MQTT: yes

dalathegreat:
### What
This PR overhauls the scaled value calculations

### Why
The calculation has remained untouched for a long time. It has many bugs and missing functionality
Fixes #1104 
Also fixes #396 
Partially makes #926 better

### How
We fix several issues:
- When negative scaling is used, it no longer results in the capacity remaining jumping to 0 for the whole range, as reported in 1104
- When double batteries are used, we now report the total capacity for bat1+bat2 towards the inverter
- We fix a place in the code where division by 0 could occur (if user sets min/max SOC setting to same values), making the board crash :warning: 
- We increase system performance, by removing millis() calls in the update_calculated_values() function

On Double Battery setups, the capacity is now summed together (Visible in Battery1 Reported fields)
image

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -496,11 +496,11 @@ void update_calculated_values() {
     datalayer.battery.status.reported_remaining_capacity_Wh = datalayer.battery2.status.remaining_capacity_Wh;
   }
 #endif  // DOUBLE_BATTERY
-  // Check if millis() has overflowed. Used in events to keep better track of time
-  if (millis() < lastMillisOverflowCheck) {  // Overflow detected
+  // Check if millis has overflowed. Used in events to keep better track of time
+  if (currentMillis < lastMillisOverflowCheck) {  // Overflow detected
     datalayer.system.status.millisrolloverCount++;
   }
-  lastMillisOverflowCheck = millis();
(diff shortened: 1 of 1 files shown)


dalathegreat:
Merging this, tested OK with 2x LEAF batteries :heavy_check_mark:

PR #1112, opened 2025-05-06
Bugfix: Add battery closing allowed flag on startup

dalathegreat:
### What
This PR sets the battery_allows_contactor_closing = true; on some integrations that were missing it

### Why
So users can enjoy automated GPIO contactor closing on these batteries. One user reported that their Kangoo pack was not working with GPIO, this PR fixes that issue.

### How
We set the datalayer.system.status.battery_allows_contactor_closing = true; inside the setup_battery function on the integrations that were missing it.


mvgalen (approved):
LGTM.

PR #1114, opened 2025-05-06
Nissan Leaf battery class with double battery support

kyberias:
### What
This PR ports Nissan Leaf battery to use the class interface. In addition this introduces a way to implement double battery support with the new class interface.

### Why
This is part of a continuous effort to bring all batteries to class based interface.

### How
Nissan Leaf battery specific code is implemented as C++ class. Two instances of the same class is used when double battery support is activated. This design eliminates the need of duplicate code. Each instance uses separate CAN bus and writes to a different part of the "datalayer" similarly as the earlier double battery implementation. Static variables are moved to class members to allow two identical copies to exist in memory at the same time. The two battery instances operate identically with the exception of the first battery only updating the "extended datalayer".

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -9,220 +9,48 @@
 #include "../datalayer/datalayer_extended.h"  //For "More battery info" webpage
 #include "../devboard/utils/events.h"
 
-/* Do not change code below unless you are sure what you are doing */
-static unsigned long previousMillis10 = 0;   // will store last time a 10ms CAN Message was send
-static unsigned long previousMillis100 = 0;  // will store last time a 100ms CAN Message was send
-static unsigned long previousMillis10s = 0;  // will store last time a 1s CAN Message was send
-static uint8_t mprun10r = 0;                 //counter 0-20 for 0x1F2 message
-static uint8_t mprun10 = 0;                  //counter 0-3
-static uint8_t mprun100 = 0;                 //counter 0-3
-
-// These CAN messages need to be sent towards the battery to keep it alive
-CAN_frame LEAF_1F2 = {.FD = false,
-                      .ext_ID = false,
-                      .DLC = 8,
-                      .ID = 0x1F2,
-                      .data = {0x10, 0x64, 0x00, 0xB0, 0x00, 0x1E, 0x00, 0x8F}};
-CAN_frame LEAF_50B = {.FD = false,
-                      .ext_ID = false,
-                      .DLC = 7,
-                      .ID = 0x50B,
-                      .data = {0x00, 0x00, 0x06, 0xC0, 0x00, 0x00, 0x00}};
-CAN_frame LEAF_50C = {.FD = false,
-                      .ext_ID = false,
-                      .DLC = 6,
-                      .ID = 0x50C,
-                      .data = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00}};
-CAN_frame LEAF_1D4 = {.FD = false,
-                      .ext_ID = false,
-                      .DLC = 8,
-                      .ID = 0x1D4,
-                      .data = {0x6E, 0x6E, 0x00, 0x04, 0x07, 0x46, 0xE0, 0x44}};
-// Active polling messages
-uint8_t PIDgroups[] = {0x01, 0x02, 0x04, 0x83, 0x84, 0x90};
-uint8_t PIDindex = 0;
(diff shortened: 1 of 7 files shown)


kyberias:
Careful review is needed to make sure my assumption is correct that two instances can be used to implement double battery. The code for the second battery looked identical to the first one with the difference that it was using a different CAN bus and datalayer object. This is handled by the code in the PR.

I recommend that we recruit someone with double battery to test this PR before merging.

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -1505,24 +936,15 @@ void decodeChallengeData(unsigned int incomingChallenge, unsigned char* solvedCh
...
   strncpy(datalayer.system.info.battery_protocol, "Nissan LEAF battery", 63);
   datalayer.system.info.battery_protocol[63] = '\0';
-  datalayer.battery.info.number_of_cells = 96;
-  datalayer.battery.info.max_design_voltage_dV = MAX_PACK_VOLTAGE_DV;
-  datalayer.battery.info.min_design_voltage_dV = MIN_PACK_VOLTAGE_DV;
-  datalayer.battery.info.max_cell_voltage_mV = MAX_CELL_VOLTAGE_MV;
-  datalayer.battery.info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
-  datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
-
-#ifdef DOUBLE_BATTERY
```
kyberias, line 1518: Both batteries are initialized with identical values, so it is sufficient to do this once for a single object.

dalathegreat:
Initial benchtesting looks good!

image

I only have 1 CAN channel Kvaser, so had to switch between CAN0 and CAN1. But it populates the data correctly :raised_hands: 

I will move towards real battery testing within a few days!

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
dalathegreat: One issue with this. If both batteries are initialized with identical values, the check_interconnect_available() function in Software.ino will allow the secondary battery to join the DC link straight on boot. We need the secondary battery voltage to be initialized to 0V, like in the previous implementation.

Alternatively, we could remake the initial sanity check in check_interconnect_available to abort if voltages are 3700dV (the initial value)

kyberias: Perhaps I can't read the original code correctly... but seems to me these particular values are initialized to identical values in original code for both batteries right? Battery(1) values are intialized with constants and then the battery2 fields are copied from battery(1).

kyberias: Ok got it, this needs to happen for the second battery in init which is different for the first:

battery2_Total_Voltage2 = 0;

kyberias: Fixed by initializing the second battery's voltage to 0 in constructor. This should match the original implementation.

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -338,430 +168,45 @@ void update_values_battery() { /* This function maps all the values fetched via
...
-      //ZE1 2018-2023 battery detected!
-      LEAF_battery2_Type = ZE1_BATTERY;
-      break;
-    case 0x79B:
-      stop_battery_query = true;             //Someone is trying to read data with Leafspy, stop our own polling!
-      hold_off_with_polling_10seconds = 10;  //Polling is paused for 100s
-      break;
-    case 0x7BB:
-
-      if (stop_battery_query) {  //Leafspy/Service request is active, stop our own polling
```
kyberias, line 624: This case (7BB)  has a bit more code in the "first battery" code copy. See line #310. Is it ok to run that code for the second battery as well? @dalathegreat

kyberias: There is a comment "// This section checks if we are doing a SOH reset towards BMS. If we do, all 7BB handling is halted"

dalathegreat: If you mean stateMachineClearSOH , then yeah, it does not matter that you try to reset both batteries. It's actually quite OK!

dalathegreat (approved):
Confirmed working on my 2x30kWh LEAF install! Nice work @kyberias :raised_hands:

kyberias:
Ok I guess I'll merge it then. :)

PR #1117, opened 2025-05-07
BYD CAN inverter protocol class

kyberias:
Adds a base class for CAN inverters and implements a class for the BYD CAN inverter.


dalathegreat (approved):
LGTM :+1:

PR #1123, opened 2025-05-09
Updating Test Fake Battery to use the base image

No-Signal:
This PR updates the test fake battery to use the new base image to allow easy local testing


kyberias:
I suggest we first merge PR #1114 . Hopefully today.

Software/src/battery/CanBattery.h
```diff
@@ -1,17 +1,41 @@
...
 // No support for double battery yet.
 class CanBattery {
  public:
+  CanBattery(DATALAYER_BATTERY_TYPE* datalayer_ptr, bool* allows_contactor_closing_ptr,
+             DATALAYER_INFO_NISSAN_LEAF* extended, int targetCan) {
+    datalayer_battery = datalayer_ptr;
+    allows_contactor_closing = allows_contactor_closing_ptr;
+    datalayer_nissan = extended;
+    can_interface = targetCan;
+    Serial.println("Constructor1 called");
```
kyberias, line 19: I don't think we should be printing to serial port here.

Software/src/battery/CanBattery.h
```diff
@@ -1,17 +1,41 @@
...
+    Serial.println("Constructor2 called");
+  }
   virtual void setup(void) = 0;
   virtual void handle_incoming_can_frame(CAN_frame rx_frame) = 0;
   virtual void update_values() = 0;
   virtual void transmit_can(unsigned long currentMillis) = 0;
+
+ protected:
+  DATALAYER_BATTERY_TYPE* datalayer_battery;
+  DATALAYER_INFO_NISSAN_LEAF* datalayer_nissan;
```
kyberias, line 36: Let's not put battery specific stuff in a base class. Base class should define the abstract interface that is common to all batteries.

kyberias:
May I suggest that only the Fake battery would be converted to using the base class in this PR? See my BMW I3 PR as an example.

My plan was to convert all the double batteries to use the base class (in separate PRs) and then refactor the base class(es) if it looks like there are opportunities to have a specialized based class for batteries supporting double operation.

Software/src/battery/CanBattery.h
No-Signal: I accidentally left this in from testing, will remove

Software/src/battery/CanBattery.h
No-Signal: Agree no battery specifics should be in the base class, with the current implementation this feels like a good interim approach until the datalayer is merged into the batteries at a later stage as fewer changes will be required.

No-Signal:
> May I suggest that only the Fake battery would be converted to using the base class in this PR? See my BMW I3 PR as an example.
> 
> My plan was to convert all the double batteries to use the base class (in separate PRs) and then refactor the base class(es) if it looks like there are opportunities to have a specialized based class for batteries supporting double operation.

My thoughts were that this implementation balances separation of concerns with the current state of the migration effort.

It also enables all batteries to benefit from double battery support once converted even if they didn't previously.

Software/src/battery/CanBattery.h
No-Signal: Actually I can see a better approach, let me update

kyberias:
I wouldn't yet allow all the batteries to take in the double battery feature. I believe DOUBLE_BATTERY has had a specific impact on a limited number of batteries and IMHO we should keep it that way during this conversion.

In fact, currently, if you enable DOUBLE_BATTERY for a battery type that doesn't support it, the code doesn't even compile. That should also be the case after this conversion.

IMHO this initial conversion to using the base class should not add/remove any functionality. Just laying out the ground work stp-by-step for the upcoming common image.

Perhaps @dalathegreat has some insight into this topic.

Software/src/battery/CanBattery.h
kyberias: My reason for adding the datalayer pointer to Nissan class was merely the fact that the two instances of the class for each of the double batteries would need to update different parts of the datalayer.

For batteries not supporting double battery setup this is not required and they can update the one and only datalayer directly.

When converting battery code to use the base class I want to retain the original code as much as possible so that it's easy to review the PR and see the mistakes. That's why I didn't want to use the datalayer pointer in other batteries (where it is not absolutely needed).

Hope this makes my approach more clear.

No-Signal:
I've refactored the base constructor to push the battery specific logic to their constructors to keep everything for a battery in a single location.

The idea was to minimise the scale of future updates, for example changing the constructor in the future could require all batteries to be updated in a single PR so getting it in place early avoids that.

The suggestion isn't to allow DOUBLE_BATTERY for batteries where it isn't tested, this implementation makes it easier to implement in the future. The current logic of failing a build for batteries that don't support DOUBLE_BATTERY can and probably should for now remain in place.

kyberias:
It's a reasonable idea but I still haven't integrated the RS485 inverters and batteries which bring additional requirements to the base class. I was kinda hoping to get this effort a bit farther so that we can see the big picture. This shouldn't take too many days or weeks. I'm sure we can't avoid having to refactor the base class in the future any way. For example, getting rid of the datalayer brings additional functionality to the classes and again many files will change.

The current plan is now to make straight-forward conversion of the battery and inverter implementation to the simple class design so that we can be relatively sure we maintain functionality. I would like to focus on this first.

dalathegreat:
Shall we rebase this PR, or close it?

kyberias:
Since the fake battery conversion has now been merged to main, I suggest we close this.

PR #1125, opened 2025-05-10
Bugfix for mqtt memory leak

No-Signal:
Fix for memory leak in mqtt message receive function found by @kyberias

### DIFF
--- a/Software/src/devboard/mqtt/mqtt.cpp
+++ b/Software/src/devboard/mqtt/mqtt.cpp
@@ -484,6 +484,7 @@ void mqtt_message_received(char* topic_raw, int topic_len, char* data, int data_
   if (strcmp(topic, generateButtonTopic("STOP").c_str()) == 0) {
     setBatteryPause(true, false, true);
   }
+  free(topic);
 }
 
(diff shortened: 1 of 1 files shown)


Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -484,6 +484,7 @@ void mqtt_message_received(char* topic_raw, int topic_len, char* data, int data_
   if (strcmp(topic, generateButtonTopic("STOP").c_str()) == 0) {
     setBatteryPause(true, false, true);
   }
+  free(topic);
```
kyberias, line 487: Now that we're at it, we could also add a null check for the strndup. This would prevent crashes in any low memory situations. The command couldn't be handled but at least we wouldn't crash with a null pointer.

kyberias: BTW, since we're C++ and are already using Strings here, this whole function could be refactored to use a String parameter instead of pointers. Caller could create the String and we wouldn't need to allocate anything. strcmps would become string comparisons etc.

I'm not saying you should do it in this PR, this is just something that occurred to me right now.

dalathegreat (approved):
Nice work you two! :raised_hands:

PR #1126, opened 2025-05-11
BMW I3 battery implemented as a class with double battery support

kyberias:
### What
BMW I3 battery implementation converted to use the base class approach with double battery support.

### Why
This is yet another step towards being able to build an image with all batteries.

### How
The same approach was used as with Nissan Leaf with the additional mechanism to handle different wakeup GPIO for the two batteries.


dalathegreat (approved):
Nice work, feels good to get rid of so much duplicate code on these double-battery integrations!

PR #1127, opened 2025-05-11
Ferroamp CAN inverter protocol code converter to use the base class

kyberias:
This PR is yet another straight-forward inverter protocol implementation conversion to use the base class.


dalathegreat (approved):
:+1:

PR #1128, opened 2025-05-11
Improvement: millis() reduction

dalathegreat:
### What
This PR reduces amounts of millis() function calls

### Why
millis() calls take performance. The less you do this, the better

### How
We instead pass the time from the main loop onto the functions

### DIFF
--- a/Software/src/battery/KIA-E-GMP-BATTERY.cpp
+++ b/Software/src/battery/KIA-E-GMP-BATTERY.cpp
@@ -161,7 +161,7 @@ uint16_t selectSOC(uint16_t SOC_low, uint16_t SOC_high) {
 }
 
 /* These messages are needed for contactor closing */
-unsigned long startMillis;
+unsigned long startMillis = 0;
 uint8_t messageIndex = 0;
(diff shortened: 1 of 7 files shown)


Software/src/battery/KIA-E-GMP-BATTERY.cpp
```diff
@@ -1131,9 +1131,6 @@ void transmit_can_battery(unsigned long currentMillis) {
 void setup_battery(void) {  // Performs one time setup at startup
   strncpy(datalayer.system.info.battery_protocol, "Kia/Hyundai EGMP platform", 63);
   datalayer.system.info.battery_protocol[63] = '\0';
-
-  startMillis = millis();  // Record the starting time
```
kyberias, line 1135: If the starting time is not recorded as before, it remains zero and the first check comparing that to actual millis will probably always pass and messages start sending (in transmit_can_battery) sooner than before. I don't know whether this is an issue, probably not.

kyberias (approved):

Software/src/battery/KIA-E-GMP-BATTERY.cpp
dalathegreat: Not an issue in this case, but nicely spotted :+1:

PR #1129, opened 2025-05-11
SMA-BYD-HVS: update battery error handling, and CAN inverter still alive

lenvm:
### What
This PR:
- changes `SMA_158` content depending on `battery_allows_contactor_closing`, rather than `inverter_allows_contactor_closing`
- adds `CAN_inverter_still_alive` for all CAN messages sent by SMA

### Why
- To ensure the inverter does not receive an error bit, whenever `inverter_allows_contactor_closing` is `false`. This error bit should be sent by the battery emulator and received by the inverter when the battery does not allow contactor closing.
- To ensure the `CAN_inverter_still_alive` is set to `true` by all messages that are sent by the inverter

### DIFF
--- a/Software/src/inverter/SMA-BYD-HVS-CAN.cpp
+++ b/Software/src/inverter/SMA-BYD-HVS-CAN.cpp
@@ -138,17 +138,22 @@ void update_values_can_inverter() {  //This function maps all the values fetched
   }
 
   //Error bits
-  if (datalayer.system.status.inverter_allows_contactor_closing) {
+  if (datalayer.system.status.battery_allows_contactor_closing) {
     SMA_158.data.u8[2] = 0xAA;
+  } else {
+    SMA_158.data.u8[2] = 0x6A;
+  }
+
+  // Inverter allows contactor closing
+  if (datalayer.system.status.inverter_allows_contactor_closing) {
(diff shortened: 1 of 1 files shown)


Software/src/inverter/SMA-BYD-HVS-CAN.cpp
```diff
@@ -138,17 +138,22 @@ void update_values_can_inverter() {  //This function maps all the values fetched
...
-  if (datalayer.system.status.inverter_allows_contactor_closing) {
+  if (datalayer.system.status.battery_allows_contactor_closing) {
     SMA_158.data.u8[2] = 0xAA;
+  } else {
+    SMA_158.data.u8[2] = 0x6A;
+  }
+
+  // Inverter allows contactor closing
+  if (datalayer.system.status.inverter_allows_contactor_closing) {
 #ifdef INVERTER_CONTACTOR_ENABLE_LED_PIN
```
dalathegreat, line 149: Maybe wrap the #ifdef around the whole function check?

lenvm: Done ✅

dalathegreat (approved):
Nice work! :+1:

PR #1140, opened 2025-05-13
BYD Atto 3 battery converted to use the base class

kyberias:
### What and How
This PR converts the BYD Atto 3 battery implementation to use the base class with duplicate double battery code eliminated.

### Why
One step closer to being able to build in all the batteries for common image.


kyberias:
Please review in detail as I may have missed out something important with double battery stuff although it looked quite straight-forward.

dalathegreat (approved):
Looks good, this will have the same effect as on t he LEAF double battery, running any diagnostic command will run it towards both batteries. It is not optimal, but it does not matter for this PR.

I will look into splitting the More Battery Info view in 2 if Double Battery is used, that way we can move towards independed diagnostics for double setups.

PR #1143, opened 2025-05-14
Convert the remaining inverter protocols to use the common base class and build them all

kyberias:
### What and how
This PR converts the remaining inverter protocols to use the common base class and also configures the build to build them all. The activated inverter protocol is still selected based on build-time define.

### Why
This brings the project closer to the goal of having all inverters and batteries in a single image.


kyberias:
Still need to check that all the preprocessor defines don't collide.

dalathegreat (approved):
Massive amount of work, extremely nicely done!

PR #1152, opened 2025-05-18
Contactor closing variable fix and more double batteries converter to use the base class 

kyberias:
- Rename `battery2_allows_contactor_closing` to `battery2_allowed_contactor_closing` to clarify that it means that the (second) battery itself doesn't dictate contactor closing but rather the variable gives battery code that permission and the decision is made elsewhere.
- Fix usage of "allows contactor closing" and "allowed contactor closing"
- Convert rest of the batteries supporting double battery to the new class model


dalathegreat (approved):
Very nice reduction with -800 lines :raised_hands:

PR #1153, opened 2025-05-18
Microfeature: Zoe, Refactor and add More Battery Info page

dalathegreat:
### What
This PR refactors the Renault Zoe implementation

### Why
This implementation is used by many different Renault vehicles, (Zoe 22/41 , Kangoo 33, Fluence ZE 22/36) , and it is a bit unclear what is sent by what generation of vehicle. This PR aims to make further improvements to the integration easier

### How

- We remove the calculated min/max cellvoltage lookups, and instead use the values sent constantly via CAN
- We add comments on those CAN messages that we are sure is sent by a specific model
- More Battery Info page added for the Zoe integration, so we can look at diagnostic data in more detail


kyberias (approved):
LGTM AFAIK

PR #1157, opened 2025-05-21
Pylon Battery: Add startup values for SOH and SOC

dalathegreat:
### What
This PR makes the startup of Pylon Battery more smooth

### Why
Prevents two incorrect events from firing at startup before we have values
image

### How
SOH is initialized to 100% instead of 0%
SOC is initialized to 50% instead of 0%


dalathegreat:
Merging, tested OK on the Discord!

PR #1162, opened 2025-05-21
The remaining 21 batteries and BMS converted to use the base class

kyberias:
This is a straight-forward conversion of the remaining batteries and BMS code converted to use the CanBattery base class. This allows all the batteries to be built into the image.


dalathegreat (approved):
Go Kyberias! :raised_hands:

PR #1169, opened 2025-05-23
SMA_BYD_H inverter: add content previously added to SMA_BYD_HVS

lenvm:
### What
This PR implements the changes done for `SMA_BYD_HVS` in #1129, for `SMA_BYD_H`

### Why
To ensure upgrades are done for all protocols

### How
By adding the content from #1129 to the `SMA_BYD_H` inverter protocol.

### DIFF
--- a/Software/src/inverter/SMA-BYD-H-CAN.cpp
+++ b/Software/src/inverter/SMA-BYD-H-CAN.cpp
@@ -63,19 +63,22 @@ void SmaBydHInverter::
   }
 
   //Error bits
-  if (datalayer.system.status.inverter_allows_contactor_closing) {
+  if (datalayer.system.status.battery_allows_contactor_closing) {
     SMA_158.data.u8[2] = 0xAA;
-#ifdef INVERTER_CONTACTOR_ENABLE_LED_PIN
-    digitalWrite(INVERTER_CONTACTOR_ENABLE_LED_PIN,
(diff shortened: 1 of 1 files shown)


Software/src/inverter/SMA-BYD-H-CAN.cpp
```diff
@@ -166,26 +169,69 @@ void SmaBydHInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
     case 0x560:  //Message originating from SMA inverter - Init
       datalayer.system.status.CAN_inverter_still_alive = CAN_STILL_ALIVE;
       break;
+    case 0x561:
```
kyberias, line 172: More of a stylish issue and about the principle of DRY: why not group the identical cases and let them fall back to single statement to set the CAN_inverter_still_alive = CAN_STILL_ALIVE?

Or even: let the default-case return (without setting still alive) and other cases fall back to single line that sets the CAN_STILL_ALIVE everytime some message is recognized.

dalathegreat (approved):

Software/src/inverter/SMA-BYD-H-CAN.cpp
kyberias: I mean they could still be in numeric order. 

```
case 0x560:
case 0x561:
case 0x562:
case 0x563:
      datalayer.system.status.CAN_inverter_still_alive = CAN_STILL_ALIVE;
      break;
```

PR #1173, opened 2025-05-24
Feature: Add separate Deye protocol

dalathegreat:
### What
This PR implements a separate inverter protocol for when using Deye inverters

### Why
Deye inverters are not really compliant with the BYD_CAN protocol, and will start to oscillate when the allowed charge goes below 1A. Deye seems to target a voltage instead, and wants to float charge. To avoid massive modifications to the BYD_CAN protocol, and risk affecting other inverters, we instead branch off a separate DEYE_BYD_CAN protocol.

### How
This implementation introduces DEYE_BYD_CAN, with the following tweaks:

- If charge allowed goes below 1A, we cap it artifically to 1A towards the inverter
- If charge allowed is 0A, we set SOC% to 100% (avoids offgrid Deye overcharging the battery)
- If discharge allowed is 0A, we set SOC% to 0% (avoids offgrid Deye overdischarging the battery)
- TODO: Voltage control needs tweaking

### DIFF
--- a/Software/USER_SETTINGS.h
+++ b/Software/USER_SETTINGS.h
@@ -50,9 +50,9 @@
 /* Select inverter communication protocol. See Wiki for which to use with your inverter: https://github.com/dalathegreat/Battery-Emulator/wiki */
 //#define AFORE_CAN        //Enable this line to emulate an "Afore battery" over CAN bus
 //#define BYD_CAN          //Enable this line to emulate a "BYD Battery-Box Premium HVS" over CAN Bus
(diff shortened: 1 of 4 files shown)


mvgalen (commented):
Looking at this PR, I wonder if this is the right approach. It seems only the update_values fuction differs. Would it not be more appropriate to subclass the BYD_CAN class and only override the update_values function. I think this would make sense with the transition to the single image as well.

dalathegreat:
Closing this for now, branch is stale

PR #1179, opened 2025-05-25
Pylon: Better handling of charge/discharge allowed flags

dalathegreat:
### What
This PR improves charge/discharge allowed flags for Pylon CAN

### Why
Previously we only set charge/discharge to forbidden if the entire Emulator went into a fault state. This is bad.

### How
We now set charge to forbidden, incase the allowed charge power goes to 0A. Same with discharge logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,11 @@ This software enables EV battery packs to be used for stationary storage. It ach
 ![Fronius](https://github.com/dalathegreat/Battery-Emulator/assets/26695010/741c3237-8074-4891-9cd1-f47f0fe45cb5)
 
 
-## Hardware requirements 📜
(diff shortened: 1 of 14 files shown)


mvgalen (changes requested):
The old code also set:
    PYLON_4280.data.u8[2] = 0xAA;
    PYLON_4280.data.u8[3] = 0xAA;
    PYLON_4281.data.u8[2] = 0xAA;
    PYLON_4281.data.u8[3] = 0xAA;

Are those no longer needed?

dalathegreat:
Those bits have no function according to the Pylon HV protocol doc!

mvgalen (approved):
LGTM.

PR #1186, opened 2025-05-28
Feature: Add balancing status (LEAF & ZoeGen2)

dalathegreat:
### What
This PR implements balancing status, initially for Nissan LEAF and ZoeGen2 batteries, and also sets the foundation on how to implement the same on other battery types

### Why
To be able to see that battery is balancing properly long term

### How
LEAF: We now read Group6 from the Nissan LEAF BMS, which contain the balancing resistor status (aka shunt status). We map this to the datalayer, into the newly created cell_balancing_status[]

The cellmonitor view now also shows the balancing status in the graph view:

image

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.h
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.h
@@ -155,8 +155,8 @@ class NissanLeafBattery : public CanBattery {
   uint8_t group_7bb = 0;
   bool stop_battery_query = true;
   uint8_t hold_off_with_polling_10seconds = 2;  //Paused for 20 seconds on startup
-  uint16_t battery_cell_voltages[97];           //array with all the cellvoltages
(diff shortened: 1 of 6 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.h
```diff
@@ -156,6 +156,7 @@ class NissanLeafBattery : public CanBattery {
   bool stop_battery_query = true;
   uint8_t hold_off_with_polling_10seconds = 2;  //Paused for 20 seconds on startup
   uint16_t battery_cell_voltages[97];           //array with all the cellvoltages
+  bool battery_balancing_shunts[97];            //array with all the balancing resistors
```
mvgalen, line 159: Why is the length of the array 97, if we are only using 96 elements?

dalathegreat:
Merging this, as tested OK by several people. Zoe2 is untested, but if there are issues we can fix that in a separate PR!

PR #1189, opened 2025-05-31
Convert two remaining batteries to use the base class

kyberias:
This PR converts the two remaining batteries Foxess and Imiev CZero Ion to use the common battery base class so that we can later build them all into a common image.


dalathegreat (approved):
:+1:

PR #1192, opened 2025-06-01
Update PYLON-CAN.h to allow high capacity >255Ah

Morris3190:
What

This PR fixes a bug that prevents a config for a high capacity battery to compile
Why

As the low-byte of the 732x Pylon frame isn't correctly cast from int -> uint8_t, the compiler won't accept values > 255.
How

Correct masking and casting of low-byte of the AH_Capacity fixes the issues.
Implemented twice for the frame 7320 and 7321


kyberias (approved):
LGTM

dalathegreat:
Thanks for the contribution @Morris3190 :raised_hands:

PR #1193, opened 2025-06-01
Batteries implement HTML renderers for status and commands

kyberias:
Reduce dependencies to specific battery types in web server code by offloading battery specific HTML rendering and commands to the abstract Battery interface and related HTML renderer objects.


Software/src/devboard/webserver/advanced_battery_html.cpp
```diff
@@ -1,9 +1,62 @@
 #include "advanced_battery_html.h"
 #include <Arduino.h>
+#include <vector>
 #include "../../datalayer/datalayer.h"
 #include "../../datalayer/datalayer_extended.h"
 #include "../../include.h"
 
+std::vector<BatteryCommand> battery_commands = {
+    {"clearIsolation", "Clear isolation fault", "clear any active isolation fault?",
```
dalathegreat, line 10: There should be comments on these, to improve maintainability. Which commands belong to which battery type? A simple //BMW i3 , or //Renault Zoe2 comment will suffice!

Software/src/devboard/webserver/cellmonitor_html.cpp
```diff
@@ -16,21 +16,24 @@ String cellmonitor_processor(const String& var) {
...
     content += ".voltage-values { margin-bottom: 10px; }";  // Style for voltage values section
-#ifdef DOUBLE_BATTERY
-    content +=
-        "#graph, #graph2 {display: flex;align-items: flex-end;height: 200px;border: 1px solid #ccc;position: "
-        "relative;}";
-#else
-    content += "#graph {display: flex;align-items: flex-end;height: 200px;border: 1px solid #ccc;position: relative;}";
-#endif
+
+    if (battery2) {
```
dalathegreat, line 20: Not a problem per se, but the PR 1186 collides a bit here, I can refactor mine to match the new battery2

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -572,102 +600,8 @@ void init_webserver() {
...
-    }
-    datalayer_extended.VolvoPolestar.UserRequestBECMecuReset = true;
-    datalayer_extended.VolvoHybrid.UserRequestBECMecuReset = true;
-    request->send(200, "text/plain", "Updated successfully");
-  });
-
-#ifdef TEST_FAKE_BATTERY
   // Route for editing FakeBatteryVoltage
-  server.on("/updateFakeBatteryVoltage", HTTP_GET, [](AsyncWebServerRequest* request) {
+  server.on("/updateBatteryVoltage", HTTP_GET, [](AsyncWebServerRequest* request) {
```
dalathegreat: I liked the old fake voltage name, since now it is a bit confusing what the updateBatteryVoltage means. Setting it to a fake value had a clear testing implication

dalathegreat (changes requested):

dalathegreat:
Oh, and please use the What/Why/How template for PRs if possible :)

Software/src/devboard/webserver/advanced_battery_html.cpp
kyberias: I was thinking these may not (all) be battery specific, but rather generic in a way that any battery could, for example, implement the "clear isolation fault" command.

Software/src/devboard/webserver/webserver.cpp
kyberias: I'll change it back to fake.

Software/src/devboard/webserver/advanced_battery_html.cpp
kyberias: The idea here is that battery may implement supports_clear_isolation() and return true which enables this command for that battery. By default that is false and the command is not enabled. We can see what commands Tesla supports by looking at Tesla file.

Software/src/battery/BMW-IX-HTML.h
```diff
@@ -0,0 +1,54 @@
...
+    content += "<h4>Min Cell Design Voltage: " + String(datalayer.battery.info.min_cell_voltage_mV) + " mV</h4>";
+    content +=
+        "<h4>Min Cell Voltage Data Age: " + String(datalayer_extended.bmwix.min_cell_voltage_data_age) + " ms</h4>";
+    content +=
+        "<h4>Max Cell Voltage Data Age: " + String(datalayer_extended.bmwix.max_cell_voltage_data_age) + " ms</h4>";
+    content += "<h4>Allowed Discharge Power: " + String(datalayer.battery.status.max_discharge_power_W) + " W</h4>";
+    content += "<h4>Allowed Charge Power: " + String(datalayer.battery.status.max_charge_power_W) + " W</h4>";
+    content += "<h4>T30 Terminal Voltage: " + String(datalayer_extended.bmwix.T30_Voltage) + " mV</h4>";
+    content += "<h4>Detected Cell Count: " + String(datalayer.battery.info.number_of_cells) + "</h4>";
+    static const char* balanceText[5] = {"0 No balancing mode active", "1 Voltage-Controlled Balancing Mode",
```
dalathegreat, line 28: Since you are in here, could you make the HTML files safer from out of bounds access? Say the code glitches and writes 255 to the datalayer_extended.bmwix.balancing_status , which would result in some nasty behavior. Got any ideas? :)

kyberias: I think the array indexes just need to be validated either by the code writing to the fields or the renderer code here. But perhaps that could be fixed in a separate PR?

Software/src/devboard/webserver/cellmonitor_html.cpp
dalathegreat: Ugh, I forgot to merge this PR before, and now 1186 was merged, creating a conflict here now :see_no_evil:

kyberias: No worries, I can take a look...

PR #1194, opened 2025-06-01
Bugfix - Update Ferroamp-Can.h to allow large capacity > 255Ah

Morris3190:
### What
This PR implements same change as for Pylon-Can HV

### Why
As both implementations share the same code-base, it's better to align them and fix both implementations

### How
Copy the same code for Frame 7320/7321


dalathegreat (approved):
Massive thanks @Morris3190 :+1:

PR #1196, opened 2025-06-01
Convert Ferroamp/Pylon inverter defines to private constants to avoid conflict

kyberias:
### What
This PR converts the preprocessor define settings related to CAN communication in Ferroamp and Pylon inverters to private static constants to avoid them leaking to change behavior of each other.

### Why
This fixes the bug of sending multiple CAN messages.

### How
The settings are converted into private constants that cannot leak to other classes.

### DIFF
--- a/Software/src/inverter/PYLON-CAN.h
+++ b/Software/src/inverter/PYLON-CAN.h
@@ -20,11 +20,15 @@ class PylonInverter : public CanInverterProtocol {
   void send_system_data();
   void send_setup_info();
 
-#define SEND_0  //If defined, the messages will have ID ending with 0 (useful for some inverters)
-//#define SEND_1 //If defined, the messages will have ID ending with 1 (useful for some inverters)
-#define INVERT_LOW_HIGH_BYTES  //If defined, certain frames will have inverted low/high bytes \
-                                    //useful for some inverters like Sofar that report the voltages incorrect otherwise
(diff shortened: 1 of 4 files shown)


Software/src/inverter/PYLON-CAN.h
```diff
@@ -20,11 +20,15 @@ class PylonInverter : public CanInverterProtocol {
   void send_system_data();
   void send_setup_info();
 
-#define SEND_0  //If defined, the messages will have ID ending with 0 (useful for some inverters)
-//#define SEND_1 //If defined, the messages will have ID ending with 1 (useful for some inverters)
-#define INVERT_LOW_HIGH_BYTES  //If defined, certain frames will have inverted low/high bytes \
-                                    //useful for some inverters like Sofar that report the voltages incorrect otherwise
-  //#define SET_30K_OFFSET  //If defined, current values are sent with a 30k offest (useful for ferroamp)
+  //If true, the messages will have ID ending with 1 (useful for some inverters), otherwise 0
+  static const bool SEND_1 = false;
```
dalathegreat, line 24: Think this requires Wiki updating also before merging

kyberias: Fixed.

kyberias:
Ok looks like if the original Ferroamp code is built with INVERT_LOW_HIGH_BYTES not defined, there are syntax errors. For example, max_charge_current variable is nowhere to be found.

Seems like that code path has not been used for a while?

PR #1199, opened 2025-06-02
Bugfix: Nissan LEAF, make total capacity scale according to SOH%

dalathegreat:
### What
This PR fixes a bug related to Nissan LEAF total battery capacity reporting

### Why
Before this fix, only the 24kWh LEAF pack had its battery size dynamically scaled according to SOH%

### How
We now use the same method on all LEAF battery packs.


dalathegreat:
Merging this, tested OK by people on the Discord :raised_hands:

PR #1201, opened 2025-06-02
fix RS485Battery::transmit_rs485 not being defined/used correctly

M4GNV5:
### What
This PR fixes compiling with battery set to PYLON LV

### Why
The [call to `transmit_rs485` here](https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/Software.ino#L276), supplies a parameter, however [the callee](https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/src/battery/BATTERIES.cpp#L76) is currently defined without the parameter, leading to compilation errors.

### How
Add the missing parameter.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -73,8 +73,8 @@ void handle_incoming_can_frame_battery2(CAN_frame rx_frame) {
 #endif
 
 #ifdef RS485_BATTERY_SELECTED
-void transmit_rs485() {
-  ((RS485Battery*)battery)->transmit_rs485();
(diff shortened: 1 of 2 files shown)


kyberias:
DalyBMS is the only RS485 battery. Perhaps you mean that? Seems like it's missing from compile-all-batteries.yaml and that's why we didn't catch this. Please add it there.

dalathegreat (approved):
Thanks for the contribution @M4GNV5 :raised_hands:

dalathegreat:
Added Daly to build system in https://github.com/dalathegreat/Battery-Emulator/pull/1206

PR #1205, opened 2025-06-03
Update readme to remove confusing diagrams

jymbob:
### What
This PR gives an introduction to what BE is rather than diving straight into implementation
### Why
New users need to be guided to the Wiki rather than making a start. Changes based on clearer understanding of problems and hardware limitations don't belong on the front page.

### How
Removed diagrams, references to Leaf/Fronius/LilyGo, added intro explaining the main concept of BE


dalathegreat (approved):
Massive thanks for this @jymbob :raised_hands:

jymbob:
No problem. I can manage a bit of copy editing. It might be worth considering setting up a github pages site for this project at some point so it looks a bit more polished. Probably after the unified config lands though!

PR #1208, opened 2025-06-04
Generalize comm 'transmit' and introduce common class for shunts

kyberias:
### What
This PR introduces a common base class for shunts and eliminates some compile time defines that differentiate batteries by type.

### Why
Elimination of compile-time selection of functionalities is required for the common image.

### How
A Transmitter interface (abstract base class) is introduced that is implemented by all batteries and other objects that need to "transmit" data by some time interval. The caller (main loop) doesn't need to know what objects are transmitting, it just needs to call transmit(millis) for the registered interfaces. This allows us to eliminate e.g. RS485_BATTERY_SELECTED usage.

Also, MQTT dependency to DOUBLE_BATTERY and other compile-time dependencies are reduced.

### DIFF
--- a/.github/workflows/compile-all-batteries.yml
+++ b/.github/workflows/compile-all-batteries.yml
@@ -114,4 +114,5 @@ jobs:
       # in the build matrix, and using build flags to define the
       # battery and inverter set in the build matrix.
       - name: Compile Sketch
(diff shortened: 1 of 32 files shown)


dalathegreat (approved):
:+1:

PR #1212, opened 2025-06-07
Add feature to changing SOC method in UI on BYD Atto3

freddanastrom:
### What
Makes it possible to change between measured and estimated SOC on BYD Atto3 batteries.

### Why
To easier switch between the modes if the SOC drifts and the battery needs deeper discharge.

### How
Adding a button on the more info page and dynamically switch between the modes.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.h
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.h
@@ -57,6 +57,13 @@ class BydAttoBattery : public CanBattery {
 
   void reset_crash() { datalayer_bydatto->UserRequestCrashReset = true; }
 
+  #ifndef USE_ESTIMATED_SOC
+    // Changing SOC method in UI is only enabled if we initially use measured SOC
+    bool supports_change_SOC_method() { return true; }
+  #endif
+
+  void change_SOC_method() { datalayer_bydatto->ChangeSOCMethod = true; }
+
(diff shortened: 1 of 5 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.h
```diff
@@ -57,6 +57,13 @@ class BydAttoBattery : public CanBattery {
 
   void reset_crash() { datalayer_bydatto->UserRequestCrashReset = true; }
 
+#ifndef USE_ESTIMATED_SOC
+  // Changing SOC method in UI is only enabled if we initially use measured SOC
+  bool supports_change_SOC_method() { return true; }
+#endif
+
+  void change_SOC_method() { datalayer_bydatto->ChangeSOCMethod = true; }
```
kyberias: Seems to me we could directly toggle the SOC_method boolean and we don't need the ChangeSOCMethod variable.

change_SOC_method() could be renamed as toggle_SOC_method() if that's what it does.

freddanastrom: That makes sense now that you mention it.. This is my first contribution to the project and I was looking at the crash unlock button and trying to mimic it, but over complicated it in the process.

dalathegreat (approved):
Nice work, this will save lots of time for BYD rebalancing! :raised_hands:

PR #1214, opened 2025-06-09
Batteries and inverter protocols receive data using run-time registered interfaces

kyberias:
### What
This PR further eliminates remaining compile-time conditions (CAN_INVERTER_SELECTED / RS485_INVERTER_SELECTED) to allow more run-time selection of options.

### Why
Run-time selection of battery and inverter protocol types is required to build a common image.

### How
Software detects the comm interface (CAN / RS485)  used by batteries by allowing the objects to implement either CanReceiver or RS485Receiver interfaces (abstract classes). Batteries and inverter protocols register/subscribe themselves (e.g. register_can_receiver()) to either CAN or RS485 module as receivers and the comm module uses these interfaces to distribute relevant comm frames to the receivers.

### DIFF
--- a/Software/src/inverter/INVERTERS.cpp
+++ b/Software/src/inverter/INVERTERS.cpp
@@ -8,9 +8,11 @@ void setup_inverter() {
     return;
   }
 
+#ifdef SELECTED_INVERTER_CLASS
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Nice work again :raised_hands:

PR #1216, opened 2025-06-09
SMA: Add event for pairing event to aid troubleshooting

dalathegreat:
### What
This PR adds an info event when SMA pairing attempts are made 

### Why
For easier troubleshooting of SMA installs

### How
We now fire an Info event, every time the SMA inverter tries to initiate a pairing event. We send this text to let the user know that now it will be up to the Enable line sensing if this will succeed or not:

"SMA inverter trying to pair, contactors will close and open according to Enable line"


dalathegreat:
Merging this, for quicker feedback loops!

PR #1219, opened 2025-06-10
Bugfix: Make Kia/Hyundai 90S celldetection more reliable

dalathegreat:
### What
This PR makes the determination between 90S and 98S Kia/Hyundai batteries more robust

### Why
One user on the Discord had this happen to them:

image

90S battery incorrectly setup as 98S

### How
We now check if the cell98 is between 2000-4300mV, and if it is, we are on a 98S battery. Otherwise 90S.


dalathegreat:
Fix confirmed working great on Discord! 
image

PR #1221, opened 2025-06-10
Improvement: Make Webserver More Battery Info page more crash proof

### LINKED ISSUE #1202: [BUG] Webserver code may cause crashes if battery reports unexpected values
For example,

```
    static const char* LEAFgen[] = {"ZE0", "AZE0", "ZE1"};
    content += "<h4>LEAF generation: " + String(LEAFgen[datalayer_extended.nissanleaf.LEAF_gen]) + "</h4>";

```

Here if LEAF_gen integer for any reason is out of bounds 0-2, the array is used with an invalid index and likely causes a crash. This needs to be fixed by validating the values before using them as indices.

dalathegreat:
### What
This PR prevents out of bounds array access on the HTML pages

### Why
Fixes #1202 , webserver no longer crashes if an unexpected value occurs

### How
Instead of using const char* array[ ] , we use switch statements with a default, to prevent array overruns. eg.
switch(value){
case 1: content+="first string"; break;
default: content+="??"; break;
}

### DIFF
--- a/Software/src/battery/BMW-IX-HTML.h
+++ b/Software/src/battery/BMW-IX-HTML.h
@@ -28,33 +28,33 @@ class BmwIXHtmlRenderer : public BatteryHtmlRenderer {
     content += "<h4>Balancing: ";
     switch (datalayer_extended.bmwix.balancing_status) {
       case 0:
-        content += String("0 No balancing mode active</h4>");
+        content += "0 No balancing mode active</h4>";
         break;
       case 1:
-        content += String("1 Voltage-Controlled Balancing Mode</h4>");
+        content += "1 Voltage-Controlled Balancing Mode</h4>";
         break;
       case 2:
-        content += String("2 Time-Controlled Balancing Mode with Demand Calculation at End of Charging</h4>");
+        content += "2 Time-Controlled Balancing Mode with Demand Calculation at End of Charging</h4>";
         break;
       case 3:
-        content += String("3 Time-Controlled Balancing Mode with Demand Calculation at Resting Voltage</h4>");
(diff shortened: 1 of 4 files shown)


Software/src/battery/BMW-IX-HTML.h
```diff
@@ -25,13 +25,37 @@ class BmwIXHtmlRenderer : public BatteryHtmlRenderer {
...
-                                         "2 Time-Controlled Balancing Mode with Demand Calculation at End of Charging",
-                                         "3 Time-Controlled Balancing Mode with Demand Calculation at Resting Voltage",
-                                         "4 No balancing mode active, qualifier invalid"};
-    content += "<h4>Balancing: " + String((balanceText[datalayer_extended.bmwix.balancing_status])) + "</h4>";
-    static const char* hvilText[2] = {"Error (Loop Open)", "OK (Loop Closed)"};
-    content += "<h4>HVIL Status: " + String(hvilText[datalayer_extended.bmwix.hvil_status]) + "</h4>";
+    content += "<h4>Balancing: ";
+    switch (datalayer_extended.bmwix.balancing_status) {
+      case 0:
+        content += String("0 No balancing mode active</h4>");
```
kyberias: BTW, I don't think it's necessary to use the String() constructor here, you can just append a string literal, the String class happily concats that as well. See row 28.

dalathegreat: Fixed in [95957a5](https://github.com/dalathegreat/Battery-Emulator/pull/1221/commits/95957a503c8c16bd2c031634d7ba53b3502d980f)

PR #1226, opened 2025-06-11
Renault ZoePH2: Correct location of balancing status

dalathegreat:
### What
This PR fixes the balancing status for Zoe2

### Why
Initial implementation read the wrong bits, misleading users about how battery is balancing

### How
We now use the correct bit location


dalathegreat:
Tested OK on the Discord, merging this!

PR #1234, opened 2025-06-13
Feature: Add support for Double ZoePH2 batteries 🔋 🔋

dalathegreat:
### What
This PR implements support for running 2x Zoe PH2 52kWh packs in parallel

### Why
User requested feature on the Discord!

### How
Same way as other Double battery integrations

### DIFF
--- a/Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.h
+++ b/Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.h
@@ -10,6 +10,23 @@
 
 class RenaultZoeGen2Battery : public CanBattery {
  public:
+  // Use this constructor for the second battery.
+  RenaultZoeGen2Battery(DATALAYER_BATTERY_TYPE* datalayer_ptr, DATALAYER_INFO_ZOE_PH2* extended,
+                        CAN_Interface targetCan)
+      : CanBattery(targetCan) {
+    datalayer_battery = datalayer_ptr;
+    allows_contactor_closing = nullptr;
+    datalayer_zoePH2 = extended;
+
+    battery_pack_voltage = 0;
+  }
+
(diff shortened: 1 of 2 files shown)


Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.h
```diff
@@ -23,6 +40,13 @@ class RenaultZoeGen2Battery : public CanBattery {
 
  private:
   RenaultZoeGen2HtmlRenderer renderer;
+
+  DATALAYER_BATTERY_TYPE* datalayer_battery;
+  DATALAYER_INFO_ZOE_PH2* datalayer_zoePH2;
```
kyberias, line 47: Looks like datalayer_zoePH2 is not used for anything.

dalathegreat: Datalayer for secondary battery seems broken on all double battery integrations. It seems to fetch the values from battery1 and populate it onto battery2. Let's fix that in a separate PR for all double bats?

kyberias: Are you sure though? I don't see that. 

For example Leaf doesn't write to neither data layer directly, but uses a pointer (datalayer_battery) and when the second battery object is constructed, it is given a pointer to the second datalayer correctly:

`        battery2 = new NissanLeafBattery(&datalayer.battery2, nullptr, can_config.battery_double);
`

dalathegreat: I can return to this once I am back from vacation and can inspect the double-bat setup I have running!

dalathegreat:
Merging this to prevent branch from becoming stale, lots of dev going on!

PR #1242, opened 2025-06-16
Fix MQTT-related crash

### LINKED ISSUE #1228: [BUG] Board Reset - Panic / Exception
**Describe the bug**
Flashed v8.14.0 but am receiving a CPU panic/exception causing a reboot.  Noticed straight away afterflashing that the GUI was very laggy and within a few seconds, was given a yellow status.  Captured a log via serial to help debugging which is attached.

Of note, my Atto-3 battery is not crashed - so only other modification is commenting out the estimated SOC in this version so that I could switch modes via the GUI if needed using the new feature (why i was updating in the first place).

The same user settings for version 8.13.0 work fine - nothing was changed as far as user settings between versions.

**To Reproduce**
Upload to board v8.14.0 and within a few seconds of the reboot, the hardware resets.

**Expected behavior**
After flashing the BE should operate as usual.

**Screenshots**

Image

- Please attach your USER_SETTINGS files for easier debugging
- Software version: 8.0.14
- Battery used: BYD-ATTO3
- Inverter communication protocol: BYD_MODBUS
- Hardware used for Battery-Emulator: HW_LILYGO
- CONTACTOR_CONTROL: N/A
- MQTT: Yes

[USER_SETTINGS.h.txt](https://github.com/user-attachments/files/20689852/USER_SETTINGS.h.txt)

[Lilygo_DebugLog.txt](https://github.com/user-attachments/files/20689823/Lilygo_DebugLog.txt)

kyberias:
### What
This PR fixes #1228 in MQTT implementation's array arithmetic that caused a crash.

### Why
Battery emulator must be stable with MQTT.

### How
The actual array arithmetic error was eliminated by not writing to arrays but in a safer list. The list is not accessed with an index but with a safer iterator.

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -120,11 +120,11 @@ void setup() {
       .trigger_panic = true                                            // Enable panic reset on timeout
   };
 
-  init_mqtt();
-
   // Start tasks
 
 #ifdef MQTT
+  init_mqtt();
+
   xTaskCreatePinnedToCore((TaskFunction_t)&mqtt_loop, "mqtt_loop", 4096, NULL, TASK_MQTT_PRIO, &mqtt_loop_task,
                           WIFI_CORE);
 #endif


kyberias:
Tested with Nissan battery + BYD modbus and with fake double battery setups.

dalathegreat (approved):
Massive thanks again @kyberias :raised_hands: So many users were plagued by this!

PR #1246, opened 2025-06-18
Improvement: BYD Atto 3 show correct estimated SOC

freddanastrom:
### What
Correctly shows estimated SOC on the more info page when measured SOC is used.

### Why
Currently the high precision value is showed as estimated SOC when measured SOC is used.
Always showing what the label says it is supposed to show makes more sense and also makes it easier to compare estimated and measured SOC, to for example identify SOC drift. 

### How
By always estimated the SOC from voltage and store the value in "battery_estimated_SOC", to be displayed in the more info page, even when measured SOC is used.

This also makes way for future implementation of automatic switching between SOC methods when large delta between measured and estimated SOC is identified.


freddanastrom:
Hm..
Maybe this "Merge branch" commits that I synced to my fork from main during testing shouldn't have been included in this PR?

dalathegreat:
That can be solved by performing a "Squash Merge" when merging this, no worries!

PR #1247, opened 2025-06-18
Experimental common image

kyberias:
### What
This PR introduces a new (experimental) build that makes all the batteries and inverters available for run-time selection with some other relevant settings.

### Why
This is the first incarnation of the "common image" that minimizes any requirements to make configuration at compile time. This version is intended for further testing.

### How
Code base can be compiled with COMMON_IMAGE define which allows all the critical configuration be performed by the user at run-time in the web UI. At the same time, the existing compile-time configuration can still be used.


.github/workflows/compile-common-image-lilygo.yml
```diff
@@ -0,0 +1,57 @@
+# This is the name of the workflow, visible on GitHub UI.
```
dalathegreat, line 1: Would we be able to remove all other builds when adding this? It seems like this common build would be able to catch issues in one go, without needing to build each component separately?

kyberias: As long as we want to support the USER_SETTINGS.h it would make sense to build something with that mechanism here as well. I'm thinking we want to support both at least for a while.

PR #1249, opened 2025-06-18
Bugfix: Kia-64/40 cell count detection

dalathegreat:
### What
This PR makes the check for determining if we are on 98S or 90S more reliable

### Why
Some 98S owners reported seeing 90S

### How
We fix the broken code :see_no_evil:


dalathegreat:
Tested OK by people on the Discord! Merging this! :+1:

PR #1251, opened 2025-06-19
Feature: BYD Atto 3 mapping unknown polls to support charged/discharged energy

freddanastrom:
### What
This one maps some of the previous unknown polls to total charged and discharged kWh and Ah.

### Why
To see how many cycles your battery has done.

### How
Maps the unknown polls and also enables the feature to have it published to MQTT.


dalathegreat (approved):
Awesome reverse engineering :raised_hands:

PR #1263, opened 2025-06-26
Feature: Add support for 28kWh Kia/Hyundai packs

dalathegreat:
### What
This PR adds support for 28kWh Kia/Hyundai batteries with these part codes:

- 37501 G7200 is Hyundai Ioniq 28kWh
- 37501 G7250 is Hyundai Ioniq 28kWh

### Why
User requested feature

### How
We need to omit the water check, and skip the cellvoltage min/max if we are on a 28kWh battery


dalathegreat:
Closing this. CAN data is too different, will add a new integration instead

PR #1264, opened 2025-06-26
Feature: Add more battery info page for Chademo V2X

dalathegreat:
### What
This PR adds a "More Battery Info" page for the Chademo V2X implementation

### Why
More easily see data, and start/stop session

### How

image

### DIFF
--- a/Software/src/battery/Battery.h
+++ b/Software/src/battery/Battery.h
@@ -1,8 +1,55 @@
 #ifndef BATTERY_H
 #define BATTERY_H
 
-#include "../datalayer/datalayer.h"
+#include <vector>
 #include "src/devboard/webserver/BatteryHtmlRenderer.h"
+
+enum class BatteryType {
+  None = 0,
+  BmwSbox = 1,
+  BmwI3 = 2,
+  BmwIx = 3,
+  BoltAmpera = 4,
+  BydAtto3 = 5,
+  CellPowerBms = 6,
+  Chademo = 7,
+  CmfaEv = 8,
+  Foxess = 9,
+  GeelyGeometryC = 10,
+  OrionBms = 11,
+  Sono = 12,
+  StellantisEcmp = 13,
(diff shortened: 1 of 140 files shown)


Software/src/battery/Battery.h
```diff
@@ -46,6 +48,8 @@ class Battery {
   virtual void request_close_contactors() {}
   virtual void toggle_SOC_method() {}
   virtual void set_factory_mode() {}
+  virtual void chademo_restart() {}
```
kyberias, line 98: Since Battery is the generic base class for all batteries, could we find a more generic term to refer to this functionality instead of "Chademo"? Maybe not, I don't know, I'm not that familiar with Chademo. Can some other battery type ever support the same commands?

dalathegreat: No this is extremely unique to Chademo, so no other integration will ever use the same command

Software/src/battery/CHADEMO-BATTERY.h
```diff
@@ -39,19 +51,17 @@ class ChademoBattery : public CanBattery {
...
-    CHADEMO_CONNECTED,
-    CHADEMO_INIT,  // intermediate state indicating CAN from Vehicle not yet received after connection
-    CHADEMO_NEGOTIATE,
-    CHADEMO_EV_ALLOWED,
-    CHADEMO_EVSE_PREPARE,
-    CHADEMO_EVSE_START,
-    CHADEMO_EVSE_CONTACTORS_ENABLED,
-    CHADEMO_POWERFLOW,
-  };
+#define CHADEMO_FAULT 0
```
kyberias, line 54: I think using the original enum type would be superior here providing at least some level of type safety and protection. The enum could be defined in the public-block here so that it's visible to other code as necessary. 

Preprocessor defines easily leak to other code causing havoc albeit these have the CHADEMO-prefix.

PR #1268, opened 2025-06-27
New Battery: Add support for Ioniq28kWh

dalathegreat:
### What
This PR adds support for the 28kWh Ioniq Electric batteries

### Why
User requested feature on the Discord

### How
Massive thanks to Elco for providing the CAN logs :raised_hands:


Software/src/battery/HYUNDAI-IONIQ-28-BATTERY.cpp
```diff
@@ -0,0 +1,385 @@
+#include "../include.h"
+#ifdef HYUNDAI_IONIQ_28_BATTERY
```
kyberias: Remove this ifdef to fix the build. All battery code is now compiled in regardless of defines. The define only affects which battery object gets instantiated.

Software/src/datalayer/datalayer_extended.h
```diff
@@ -409,6 +409,16 @@ typedef struct {
...
 
+typedef struct {
+  uint8_t total_cell_count = 0;
+  int16_t battery_12V = 0;
+  int8_t temperature_water_inlet = 0;
+  int8_t powerRelayTemperature = 0;
+  uint8_t batteryManagementMode = 0;
+  uint8_t BMS_ign = 0;
+  uint8_t batteryRelay = 0;
+} DATALAYER_INFO_IONIQ28;
```
kyberias: Maybe not add more extended datalayers that then gets removed a bit later? :) See PR #1291 for a model how to allow access to this data for the renderer without adding a struct.

dalathegreat: I have tried to remove the extended datalayer now in later commits, but dang, it keeps insta-crashing on boot

Issue #1266, opened 2025-06-27
[BUG] No data on the More Battery Info Page


MihaitaSF:
**Describe the bug**
The installation went smoothly following the instructions on the wiki and discord server.
Everything works fine, the battery charges and discharges.
The More Battery Info page is empty and the following message appears: "No extra information available for this battery type"
I tried with 8.14.0 and 8.14.1.

**Screenshots**

![Image](https://github.com/user-attachments/assets/5e062ea7-fe1a-4c2a-934b-eecbafd0e58d)

![Image](https://github.com/user-attachments/assets/14c43e27-0825-47e2-8253-15058e589f9e)

![Image](https://github.com/user-attachments/assets/23d4bc27-d142-4208-bf66-f184fe2b7440)

**Version and settings (please complete the following information):**

- Software version: 8.14.0 and 8.14.1
- Battery used: Polestar 2 4WD 78 kW 
- Inverter communication protocol: BYD Battery-Box Premium HVS over CAN Bus Dey
- Hardware used for Battery-Emulator: LILYGO
- CONTACTOR_CONTROL: NO
- MQTT: NO

dalathegreat:
Fixed in main, will be included in the next release (8.15.0), thanks for reporting this @MihaitaSF

PR #1270, opened 2025-06-27
Fix Volvo Spa more battery info page

kyberias:
### What
This fixes bug #1266.

### How
The battery class for Volvo Spa had a method missing to return the HTML renderer for battery specific page.


dalathegreat (approved):
:+1:

PR #1271, opened 2025-06-27
Replace BMW I3 extended data layer with battery class getter methods

kyberias:
### What
This PR eliminates BMW I3 extended data layer.

### Why
Copying data to the extended data layer structure is redundant and wasteful since the same data is already available in the battery object.

### How
We extend the BMW I3 battery class with the same getter methods that the HTML renderer can now use directly only when needed.

### DIFF
--- a/Software/src/battery/BMW-I3-HTML.cpp
+++ b/Software/src/battery/BMW-I3-HTML.cpp
@@ -1,0 +1,89 @@
+#include "BMW-I3-HTML.h"
+#include "../include.h"
+#include "BMW-I3-BATTERY.h"
+
+String BmwI3HtmlRenderer::get_status_html() {
+  String content;
+
+  content += "<h4>SOC raw: " + String(batt.SOC_raw()) + "</h4>";
+  content += "<h4>SOC dash: " + String(batt.SOC_dash()) + "</h4>";
+  content += "<h4>SOC OBD2: " + String(batt.SOC_OBD2()) + "</h4>";
+  static const char* statusText[16] = {
+      "Not evaluated", "OK", "Error!", "Invalid signal", "", "", "", "", "", "", "", "", "", "", "", ""};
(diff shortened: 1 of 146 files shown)


Software/src/battery/BMW-I3-HTML.cpp
```diff
@@ -0,0 +1,89 @@
...
+  content += "<h4>SOC OBD2: " + String(batt.SOC_OBD2()) + "</h4>";
+  static const char* statusText[16] = {
+      "Not evaluated", "OK", "Error!", "Invalid signal", "", "", "", "", "", "", "", "", "", "", "", ""};
+  content += "<h4>Interlock: " + String(statusText[batt.ST_interlock()]) + "</h4>";
+  content += "<h4>Isolation external: " + String(statusText[batt.ST_iso_ext()]) + "</h4>";
+  content += "<h4>Isolation internal: " + String(statusText[batt.ST_iso_int()]) + "</h4>";
+  content += "<h4>Isolation: " + String(statusText[batt.ST_isolation()]) + "</h4>";
+  content += "<h4>Cooling valve: " + String(statusText[batt.ST_valve_cooling()]) + "</h4>";
+  content += "<h4>Emergency: " + String(statusText[batt.ST_EMG()]) + "</h4>";
+  static const char* prechargeText[16] = {"Not evaluated",
```
dalathegreat, line 27: Maybe we could in this same PR switch over the array reading to the more safe format? Like in https://github.com/dalathegreat/Battery-Emulator/pull/1221

Seems like I forgot the BMW i3 in that PR :see_no_evil:

kyberias: Oh sure why not.

dalathegreat (approved):
Nice work. Now we just need to do them all to finally be able to delete datalayer_extended :sweat:

PR #1276, opened 2025-06-30
Use the correct CAN interface when Zoe sends in double battery setup

### LINKED ISSUE #1274: [BUG] Double Battery Renault ZOE gen1 and GoodWe inverter not working
I wrote about it on Discord, but I thought it would be better to write a Github issue. 
Maybe there is someone who has the same problem, or who has a working setup?
I think i tried everything possible :D
In all exampel i does not change any cabels. 
The only thing i change is Switch 1 and Switch 2 Power from the battery (Se picture.)
And offcourse  downloading the BE files.

When defining like in Exampel 1.(One battery)
(Version 8.13, but i also tried with version 8.14.0, 8.14.1)
CAN1 = Inverter  
CAN2 = Battery 2
Switch 1 OFF
Switch 2 ON
Everything works with one battery

When defining like in Exampel 2.(One battery)
(Version 8.13, but i also tried with version 8.14.0, 8.14.1)
CAN 1 = Inverter AND battery 1.
Switch 1 ON
Switch 2 OFF
Everything works with one battery

When defining like in Exampel 3 (Two battery)
(Version 8.14.1)
CAN1 = Inverter and Battery 1. 
CAN2 = Battery 2. 

There is no RX at CAN 2. 

- Software version: 8.14.1
- Battery used: 2x 41kWh Renault ZOE Gen 1
- Inverter communication protocol: CAN
- Hardware used for Battery-Emulator: HW_STARK
- CONTACTOR_CONTROL: yes
- MQTT: yes

[filer.zip](https://github.com/user-attachments/files/20984164/filer.zip)
Image
Image

Maybe i can help somehow? :-)

I know i dont need pre charge for battery 2, i will remove it when the CAN is working

kyberias:
### What
This PR likely fixes #1274 

### How
The current Zoe Gen1 code incorrectly always sent to battery 1 CAN instead of the object's configured CAN interface.

### DIFF
--- a/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
+++ b/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
@@ -166,7 +166,7 @@ void RenaultZoeGen1Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
       switch (frame0) {
         case 0x10:  //PID HEADER, datarow 0
           requested_poll = rx_frame.data.u8[3];
-          transmit_can_frame(&ZOE_ACK_79B, can_config.battery);
+          transmit_can_frame(&ZOE_ACK_79B, can_interface);
 
           if (requested_poll == GROUP1_CELLVOLTAGES_1_POLL) {
             cellvoltages[0] = (rx_frame.data.u8[4] << 8) | rx_frame.data.u8[5];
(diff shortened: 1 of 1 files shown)


kyberias:
I've been thinking of adding a transmit_can method in the Battery (and inverter) base class, so that battery implementations could just send a frame without knowing anything about the interface (the base class could hide that). That would have protected from this kind of bugs.

dalathegreat (approved):
:+1:

PR #1277, opened 2025-06-30
Feature: Sofar CAN protocol :zap: 

dalathegreat:
### What
This PR makes the Sofar CAN protocol work. Massive thanks to Pazdzierz on the Discord for these findings!

### Why
Sofar CAN is a great alternative to Pylon CAN for Sofar inverters, since on Sofar CAN you can parallel 15 batteries on the same CAN bus!

### How
This PR makes SOFAR CAN finally work. 

Bonus, In the SOFAR-CAN.h, configure the line

`  uint8_t battery_index = 0;            // Predefined battery ID (0–15)`

To set it to the battery ID you want to emulate!

Maybe this PR will contain a way to set the ID on the fly via Webserver, if I have time :sweat_smile:

### DIFF
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -289,7 +289,7 @@ String settings_processor(const String& var) {
         "updatePassword?value='+encodeURIComponent(value),true);xhr.send();}}";
     content +=
         "function editSofarID(){var value=prompt('For double battery setups. Which battery ID should this emulator "
-        "send? Enter new value between "
+        "send? Remember to reboot after configuring this! Enter new value between "
         "(0-15):');if(value!==null){if(value>=0&&value<=15){var xhr=new "
         "XMLHttpRequest();xhr.onload=editComplete;xhr.onerror=editError;xhr.open('GET','/"
         "updateSofarID?value='+value,true);xhr.send();}else{alert('Invalid value. Please enter a value between 0 "


pazdzierzz:
Battery ID - works (still tested with fake battery in second terminal).

device info is not updated (not critical for system to work)


IMG_3782
IMG_3781
IMG_3780

pazdzierzz:
Oh, don’t worry about bat2 voltage - it’s measured by inverter, not CAN related

pazdzierzz:
we can add below lines to transmit device name/capacity - it should be visible on screen. I will be able to test it in a week. Vacations time ;-)

 // frame 0x35E – Manufacturer Name ASCII
  memset(SOFAR_35E.data.u8, 0, 8);                                    
  strncpy((char*)SOFAR_35E.data.u8, BatteryType, 8);                 

  // frame 0x35F – Battery Nominal Capacity
  SOFAR_35F.data.u8[4] = BatteryCapacity_Ah & 0x00FF;                 
  SOFAR_35F.data.u8[5] = (BatteryCapacity_Ah >> 8);

dalathegreat:
Good idea @pazdzierzz , added name and capacity reporting in https://github.com/dalathegreat/Battery-Emulator/pull/1277/commits/25c60cb42b2a26b33c82dab554af310d8bab856a

dalathegreat:
@pazdzierzz , The Battery ID is now configurable via Webserver!

image

image

pazdzierzz:
@dalathegreat, one more thing, noticed generic 370V battery voltage in Sofar app. 
image

pazdzierzz:
i think that we can merge it! 

Total Voltage is not updating, checked in logs, looks correct. When I find whats what I will share my findings.

Generally speaking Sofar CAN is operational! (one week of uptime)

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -149,6 +149,9 @@ String settings_processor(const String& var) {
     if (inverter) {
       content += "<h4 style='color: white;'>Inverter interface: <span id='Inverter'>" +
                  String(inverter->interface_name()) + "</span></h4>";
+      content += "<h4 style='color: white;'>Sofar battery ID: " +
```
dalathegreat, line 152: @kyberias , any idea on how to hide this setting, so that it is only visible when Sofar protocol is in use?

kyberias: Good question. Perhaps this could be generalized a bit. The InverterProtocol class could implement a method called bool supports_battery_ID() that only returns true for Sofar. This code here could then use that to decide whether to show it.

dalathegreat: Merging this for now, to get the PR amount down (We'll refactor the settings page anyhow later!)

wreuvers: > Good question. Perhaps this could be generalized a bit. The InverterProtocol class could implement a method called bool supports_battery_ID() that only returns true for Sofar. This code here could then use that to decide whether to show it.

I have the CAN definitions for the 30kW Sandi Inverter (and also the 10kW - I have both). Can you point me in the direction of how to setup a defintion/code this inverter?
I also have the CAN definitions for the 6.6kW (and 3.3kW) Elcon chargers (https://elconchargers.com/?page_id=67). 
I am new to DalaTheGreat (amazing stuff), although I have been using Tesla Model 3 packs for 6 years (using the EVTV controller) on my island in the Bahamas. And I am in the process of tranisitioning to the Lilygo CAN and would love to contribute too.

dalathegreat: @wreuvers Please log a new Github Issue feature request, and attach the CAN definitions there. You are commenting on a closed Pull Request :sweat_smile: 

Here is where to request: https://github.com/dalathegreat/Battery-Emulator/issues

PR #1282, opened 2025-07-01
Proof of concept unit tests

No-Signal:
Draft PR to implement unit tests, this initial version has a set of tests for the safety features but many more can be added

Due to the complexity of dependencies the unit tests are configured to run on an ESP32


platformio.ini
```diff
@@ -18,5 +18,26 @@ monitor_speed = 115200
 monitor_filters = default, time, log2file
 board_build.partitions = min_spiffs.csv
 framework = arduino
-build_flags = -I include
+build_flags = 
+    -I include
+    -DCOMMON_IMAGE
```
No-Signal, line 23: Need to remove these test Includes

No-Signal:
To run the tests on a remote device an auth token secret needs to be added to the action secrets.
Example working test run on my fork https://github.com/No-Signal/Battery-Emulator/actions/runs/16032887620/job/45237573355#step:7:19920

=================================== SUMMARY ===================================
Environment    Test                  Status    Duration
-------------  --------------------  --------  ------------
test           devboard/test_safety  PASSED    00:02:35.233
================= 32 test cases: 32 succeeded in 00:02:35.233 =================

Software/Software.ino
```diff
@@ -68,6 +68,8 @@ TaskHandle_t mqtt_loop_task;
 
 Logging logging;
 
+#ifndef PIO_UNIT_TESTING
```
kyberias, line 71: Would it be possible to build the unit tests without this file entirely or have a separate Software.ino for that so that we wouldn't have these #ifndefs related to unit testing in production code?

Software/Software.ino
```diff
@@ -577,3 +465,119 @@ void check_reset_reason() {
       break;
   }
 }
+
+void core_loop(void*) {
```
kyberias, line 469: Is the core_loop just moved from one place to another or were there some changes in the content? Perhaps the move could be reverted if there were no actual changes?

Software/Software.ino
No-Signal: For the tests to work they need to build the full battery emulator firmware. There are a lot of variables in the Software.ino file that would need to be replicated in a test file and kept in sync so having one #ifndef while not ideal is a lot easier to manage.

Software/Software.ino
No-Signal: No changes were made to the core_loop, it was moved as compilation of the test project fails as the check_interconnect_available and update_calculated_values functions are defined lower in the file.

Software/src/devboard/utils/events.h
```diff
@@ -156,6 +154,7 @@ void set_event(EVENTS_ENUM_TYPE event, uint8_t data);
 void clear_event(EVENTS_ENUM_TYPE event);
 void reset_all_events();
 void set_event_MQTTpublished(EVENTS_ENUM_TYPE event);
+bool is_event_set(EVENTS_ENUM_TYPE event);
```
kyberias: Looks like this is not needed since the existing API can be used like this (from events_html.cpp):

      event_pointer = get_event_pointer((EVENTS_ENUM_TYPE)i);
      if (event_pointer->occurences > 0) {
        ...
      }

It would be preferable that we wouldn't need to add any unit-test specific code.

Software/Software.ino
kyberias: Ok, perhaps at some point we can move those global (?) variables somewhere else where the belong and we don't need to build this file at all with the tests.

Software/Software.ino
kyberias: I wonder why they work with the normal build? Perhaps add a forward declaration?

Software/src/devboard/utils/events.h
No-Signal: Good spot, have updated the test to use the existing functionality

dalathegreat:
Let's close this PR, since https://github.com/dalathegreat/Battery-Emulator/pull/1358 was merged?

PR #1288, opened 2025-07-03
Add feature to set a custom device hostname

### LINKED ISSUE #1170: [Feature request] Add hostname
**Feature request introduction**
It would be nice if the Battery Emulator could be found on a local network via a hostname, rather than its IP address.

**Describe the solution you'd like**
Add a hostname as a user setting, which is used as hostname for the battery emulator.

**Describe alternatives you've considered**
Perhaps a hostname on the access point may also be useful, to not have to type/remember `192.168.4.1`.

**Additional context**
Reference to implement setting a custom hostname: [ESP32 Setting a Custom Hostname](https://randomnerdtutorials.com/esp32-set-custom-hostname-arduino/)

freddanastrom:
### What
This makes it possible to set a custom device hostname, instead of the standard hostname esp32- followed by part of the mac adress.
This PR also makes the hostname visible at the emulator front webpage.

Should close #1170

### Why
To make the emulator accessible by the selected hostname, instead of IP or a hostname that is difficult to remember.
Which especially is beneficial on installations with multiple emulators.

### How
By enabling #define HOSTNAME in USER_SETTINGS.h and optionally specifying a hostname in USER_SETTINGS.cpp

### DIFF
--- a/Software/USER_SETTINGS.cpp
+++ b/Software/USER_SETTINGS.cpp
@@ -21,11 +21,11 @@ volatile CAN_Configuration can_config = {
     .shunt = CAN_NATIVE                   // (OPTIONAL) Which CAN is your shunt connected to?
 };
 
-std::string ssid = WIFI_SSID;               // Set in USER_SECRETS.h
-std::string password = WIFI_PASSWORD;       // Set in USER_SECRETS.h
-const char* ssidAP = "Battery Emulator";    // Maximum of 63 characters, also used for device name on web interface
-const char* passwordAP = AP_PASSWORD;       // Set in USER_SECRETS.h
-const uint8_t wifi_channel = 0;             // Set to 0 for automatic channel selection
+std::string ssid = WIFI_SSID;             // Set in USER_SECRETS.h
+std::string password = WIFI_PASSWORD;     // Set in USER_SECRETS.h
+const char* ssidAP = "Battery Emulator";  // Maximum of 63 characters, also used for device name on web interface
+const char* passwordAP = AP_PASSWORD;     // Set in USER_SECRETS.h
+const uint8_t wifi_channel = 0;           // Set to 0 for automatic channel selection
 
 #ifdef WEBSERVER
 const char* http_username = HTTP_USERNAME;  // Set in USER_SECRETS.h
--- a/Software/USER_SETTINGS.cpp
+++ b/Software/USER_SETTINGS.cpp
@@ -21,11 +21,11 @@ volatile CAN_Configuration can_config = {
     .shunt = CAN_NATIVE                   // (OPTIONAL) Which CAN is your shunt connected to?
 };
 
-std::string ssid = WIFI_SSID;             // Set in USER_SECRETS.h
-std::string password = WIFI_PASSWORD;     // Set in USER_SECRETS.h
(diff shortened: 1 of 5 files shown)


freddanastrom:
I have been debating with myself regarding two things here.

- Should the hostname be visible on the start page or not to keep the start page as clutter free as possible? Nice to have there though..
- Configuration process. The option to use a custom hostname is now enabled in USER_SETTINGS.h and the name is set in USER_SETTINGS.cpp. But could as well be set directly at the #define HOSTNAME in USER_SETTINGS.h. I landed in the current setup as it follows the current settings for wifi and MQTT.

Software/USER_SETTINGS.h
```diff
@@ -118,6 +118,7 @@
 /* Connectivity options */
 #define WIFI
 //#define WIFICONFIG  //Enable this line to set a static IP address / gateway /subnet mask for the device. see USER_SETTINGS.cpp for the settings
+//#define HOSTNAME  //Enable this line to use a custom hostname for the device, hostname is set in USER_SETTINGS.cpp
```
kyberias: I suggest CUSTOM_HOSTNAME. HOSTNAME is too generic and may easily conflict with something else.

Software/USER_SETTINGS.cpp
```diff
@@ -21,11 +21,12 @@ volatile CAN_Configuration can_config = {
...
-std::string password = WIFI_PASSWORD;     // Set in USER_SECRETS.h
-const char* ssidAP = "Battery Emulator";  // Maximum of 63 characters, also used for device name on web interface
-const char* passwordAP = AP_PASSWORD;     // Set in USER_SECRETS.h
-const uint8_t wifi_channel = 0;           // Set to 0 for automatic channel selection
+std::string ssid = WIFI_SSID;               // Set in USER_SECRETS.h
+std::string password = WIFI_PASSWORD;       // Set in USER_SECRETS.h
+const char* ssidAP = "Battery Emulator";    // Maximum of 63 characters, also used for device name on web interface
+const char* passwordAP = AP_PASSWORD;       // Set in USER_SECRETS.h
+const uint8_t wifi_channel = 0;             // Set to 0 for automatic channel selection
+const char* hostname = "battery-emulator";  // Hostname for the device, needs to be enabled in USER_SETTINGS.h
```
kyberias, line 29: custom_hostname would be better

Software/USER_SETTINGS.h
kyberias: ... and is it not possible to have the actual hostname configure here as well:

#define CUSTOM_HOSTNAME "mybattery"

...instead of having to edit multiple files.

freddanastrom: > ... and is it not possible to have the actual hostname configure here as well:
> 
> #define CUSTOM_HOSTNAME "mybattery"
> 
> ...instead of having to edit multiple files.

Yes, that is a option I have thought of and mentioned in my earlier comment.
With this current approach I was following along the way some other settings for WiFi and MQTT is defined.
But setting the name directly in the #define surely makes it easier. Would this be preferable?

kyberias: IMHO yes. BTW, for the new common image, I need to implement a way to configure this at run-time.

dalathegreat (approved):
Thanks for the contribution!

PR #1301, opened 2025-07-11
GM BEV2: Add SOC candidate

dalathegreat:
### What
This PR removes many estimations/calculations from the GM BEV2 integration

### Why
Estimated is grossly inaccurate

### How
- Use periodically sent cellvoltages instead of using slow polling method
- Removes the find min/max cellvoltage function, use periodically sent min/max instead
- Removes the calculation to find min/max temperature, use periodically sent min/max instead
- Fully decode CAN messages 0x200, 0x202, 0x204, 0x206, 0x208
- Add sensed values in CAN message 0x216
- Add isolation resistance from CAN message 0x20C
- Add inlet/outlet coolant temperature in 0x460
- Stop sending 0x7E4 polls, since battery does not reply to these anyways

### DIFF
--- a/Software/src/battery/BOLT-AMPERA-BATTERY.cpp
+++ b/Software/src/battery/BOLT-AMPERA-BATTERY.cpp
@@ -35,26 +35,42 @@ TODOs left for this implementation
 0x460 Energy Storage System Temp HV (Who sends this? Battery?)
 */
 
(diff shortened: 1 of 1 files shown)


dalathegreat:
Merging this, tested OK by HV on the Discord! :raised_hands:

PR #1305, opened 2025-07-13
Preliminary changes for better MG HS PHEV integration

jonny5532:
### What
This PR improves MG HS PHEV battery integration:
- Uses CAN1 broadcast messages for SoC, voltages, current and temperatures, rather than relying on OBD
- Implements optional voltage-based SoC scaling to allow the (NMC) cells to be charged over 4.1V/cell.
- Tidies various things up.

### Implementation details
- It uses update_values as a 1Hz tick (is this acceptable?)
- It will only allow voltage-based SoC scaling if the voltage reading is less than 10s old, to avoid a being stuck on a stale voltage due to a partial CAN connection failure.

### DIFF
--- a/Software/src/battery/MG-HS-PHEV-BATTERY.cpp
+++ b/Software/src/battery/MG-HS-PHEV-BATTERY.cpp
@@ -281,16 +281,32 @@ void MgHsPHEVBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
   }
 }
 void MgHsPHEVBattery::transmit_can(unsigned long currentMillis) {
-  // Send 70ms CAN Message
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):
Nice work!

dalathegreat:
Can you modify this PR to only change the files

- Software/src/battery/MG-HS-PHEV-BATTERY.cpp
- Software/src/battery/MG-HS-PHEV-BATTERY.h

And we can merge this right away!

jonny5532:
Ah I forgot I had those in - removed.

(They're still in the testing fork which I'm using on the actual hardware here: https://github.com/jonny5532/Battery-Emulator/tree/feature/mg-hs-phev-updates-testing if anyone wants to try. )

PR #1314, opened 2025-07-15
Tesla: Simulate entire vehicle towards battery

dalathegreat:
### What
This PR simulates the entire vehicle CAN network towards the battery

### Why
Reduce the number of active fault codes on the battery

### How
Massive thanks to greenoem over on the Discord server for these findings! :raised_hands:

### DIFF
--- a/Software/src/battery/TESLA-BATTERY.cpp
+++ b/Software/src/battery/TESLA-BATTERY.cpp
@@ -326,6 +326,263 @@ inline const char* getFault(bool value) {
   return value ? "ACTIVE" : "NOT_ACTIVE";
 }
 
+// Clamp DLC to 0–8 bytes for classic CAN
+inline int getDataLen(uint8_t dlc) {
+  return std::min<int>(dlc, 8);
+}
+
+// Fast bit‐field writer: writes 'bitLen' bits of 'value' starting at 'startBit'
+inline void setBitField(uint8_t* data, int bytes, int startBit, int bitLen, uint64_t value) {
+  int bit = startBit;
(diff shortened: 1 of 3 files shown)


Software/src/battery/TESLA-HTML.h
```diff
@@ -183,15 +183,46 @@ class TeslaHtmlRenderer : public BatteryHtmlRenderer {
     static const char* noYes[] = {"No", "Yes"};
     static const char* Fault[] = {"NOT_ACTIVE", "ACTIVE"};
 
-    //0x20A 522 HVP_contatorState
-    content += "<h4>Contactor Status: " + String(contactorText[datalayer_extended.tesla.status_contactor]) + "</h4>";
-    content += "<h4>HVIL: " + String(hvilStatusState[datalayer_extended.tesla.hvil_status]) + "</h4>";
+    //Main battery info
+    char readableBatterySerialNumber[15];  // One extra space for null terminator
```
kyberias, line 187: We're repeating the same string conversion here multiple times. Perhaps a helper method would be suitable?

String safe_array_to_String(const char* str, int len);

dalathegreat: Let's do it for next Tesla PR, this is atleast functional now, and we need to get this to .dev as soon as possible to get more testers.

dalathegreat: For instance for next PR, having the GTW_ modifiable on the fly will be needed, since I am very sure most people will miss this new requirement for Tesla :sweat_smile:

PR #1315, opened 2025-07-15
Atto 3: Add ramp down of discharge power

freddanastrom:
### What
This adds a ramp down of the maximum discharge power when SOC is estimated by voltage.

### Why
To avoid false estimation of SOC due to voltage drops cased by heavy discharge.

### How
By ramping down max discharge power from 10% estimated SOC.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -172,17 +172,20 @@ void BydAttoBattery::
   datalayer_battery->status.remaining_capacity_Wh = static_cast<uint32_t>(
       (static_cast<double>(datalayer_battery->status.real_soc) / 10000) * datalayer_battery->info.total_capacity_Wh);
 
-  datalayer_battery->status.max_discharge_power_W = MAXPOWER_DISCHARGE_W;  //TODO: Map from CAN later on
+  if (SOC_method == ESTIMATED && battery_estimated_SOC * 0.1 < RAMPDOWN_SOC && RAMPDOWN_SOC > 0) {
+    // If using estimated SOC, ramp down max discharge power as SOC decreases.
+    rampdown_power = RAMPDOWN_POWER_ALLOWED * ((battery_estimated_SOC * 0.1) / RAMPDOWN_SOC);
+
+    if (rampdown_power < MAXPOWER_DISCHARGE_W) {  // Never allow more than MAXPOWER_DISCHARGE_W
+      datalayer_battery->status.max_discharge_power_W = rampdown_power;
+    } else {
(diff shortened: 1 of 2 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -176,6 +176,14 @@ void BydAttoBattery::
 
   datalayer_battery->status.max_charge_power_W = BMS_allowed_charge_power * 10;  //TODO: Scaling unknown, *10 best guess
 
+  if (SOC_method ==
+      ESTIMATED) {  // If we are using estimated SOC, maximum discharge power is ramped down towards the end.
+    if (battery_estimated_SOC * 0.1 < RAMPDOWN_SOC) {
+      datalayer.battery.status.max_discharge_power_W =
+          RAMPDOWN_POWER_ALLOWED * ((battery_estimated_SOC * 0.1) / RAMPDOWN_SOC);
```
dalathegreat: I am always scared of division by 0, but I think there should be no way for the code to enter this branch and divide by 0 if RAMPDOWN_SOC was accidentally tuned to 0. But if we are really scared of this crash scenario, we could add a && RAMPDOWN_SOC > 0 check.

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -176,6 +176,14 @@ void BydAttoBattery::
 
   datalayer_battery->status.max_charge_power_W = BMS_allowed_charge_power * 10;  //TODO: Scaling unknown, *10 best guess
```
dalathegreat: Would be good to wrap the writing of datalayer.battery.status.max_discharge_power_W in an if/else statement, so that we dont write it twice and get some odd race condition (there should not be due to the single threaded nature of core0, but good practice. Not mandatory, but if you have time, I would suggest making it something like;

if estimated && if < rampdown
max_discharge_power_W = calc
else 
max_discharge_power_W = MAXPOWER_DISCHARGE_W

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
freddanastrom: Division by zero shouldn't really be possible unless the user screws up the configuration before building, and there are potentially a lot of other things that can be set to wrong values. But it's an easy safety check to do and also adds the possibility to disable the ramp-down completely by setting the value to 0, if someone wishes to have it that way.

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
freddanastrom: Will do!
I should probably also add a condition so ramp down power never can be higher than MAX_DISCHARGE_POWER_W. Which currentely is a fixed value, but having it controlled by the BMS is more or less ready for another PR.

dalathegreat (approved):
Nice work! :+1:

PR #1320, opened 2025-07-18
Fix RS485 inverters by proper registration of a receiver

kyberias:
### What
This fixes Kostal RS485 inverter communication.

### How
RS485 inverter object needs to be registered as a receiver for the communication to work.


kyberias:
Let's wait for some Kostal owners to test this PR before merging.

PR #1325, opened 2025-07-19
Library :scroll: : Update eModbus and ESP32 core

dalathegreat:
### What
This PR updates the ESP32 version from 3.2.0 -> 3.2.1

We also update eModbus from v1.7.3 stable -> 1.7.4 stable

### Why
Using latest ESP32 core is always a good idea, since users always download the latest in Arduino IDE

Using the latest eModbus is also a good idea, since the release contains fixes that prevent out-of-bounds writes and potential crashes.

### How
ESP32 3.2.0 -> 3.2.1
eModbus v1.7.3 stable -> 1.7.4 stable


kyberias (approved):
LGTM!

PR #1328, opened 2025-07-21
Atto 3: Max discharge power, charge power scaling and unknown values

freddanastrom:
### What
The main purpose of this one is to use the correct min and max charge/discharge power reported by the BMS.
It also have some other minor improvements:

-Using max discharge power reported by the BMS.
-Correction of max charge power scaling.
-Mapping two more unknown values.
-Tidying up the more battery info page.

### Why
To utilize the correct values of max charge/discharge power reported by the BMS which improves safety and also making the more info page a bit easier to read.

### How
By mapping unknown values and using the correct scaling.
The values on more info page is scaled and labeled to more human friendly format.


dalathegreat (approved):
Very nice reverse engineering! :raised_hands:

PR #1332, opened 2025-07-22
Update datalayer_extended.h

Gerwaz3:
### What
This PR implements extended info.


kyberias:
Let's not create more extended datalayer structs since we have tried to remove them. See earlier PRs. #1291

PR #1333, opened 2025-07-22
Fix: Ignore erroneous 53.3c CPU temperature value that ESP32 often returns

jonny5532:
### What
Ignore the specific float value 53.3c (hex value 0x42555555) that temperatureRead often returns.

### Why
With a ESP32 lilygo at 78c, the temperature keeps bouncing between 53.3c and 78c - the lower value is incorrect. Ignoring it results in a smoothly changing temperature.

Hopefully this is not device specific, but is a common issue with all ESP32s (https://github.com/esphome/issues/issues/6543 suggests that it is).

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -361,7 +361,14 @@ void check_interconnect_available() {
 
 void update_calculated_values() {
   /* Update CPU temperature*/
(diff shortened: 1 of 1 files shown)


kyberias (approved):
Nice! I've encountered the same issue and I bet many others as well.

PR #1334, opened 2025-07-22
Feature: Allow powerRemovalDuration to be set via USER_SETTINGS, for shorter BMS resets

jonny5532:
### What
Adds a BMS_POWER_REMOVAL_TIME_MS define which can be set in USER_SETTINGS.h to reduce powerRemovalDuration from the default 30000 milliseconds.

### Why
This allows quicker BMS resets and less downtime, where the BMS tolerates it. The MG BMS resets fine within 2s.


dalathegreat:
Having the BMS_RESET configurable is a great idea! Since we are moving towards live-editable settings on the fly, I suggest we make this time configurable via the Webserver. I can implement that in another PR if you want? Or do you want to have a go at adding it yourself to the Webserver?

dalathegreat:
Closing this PR, it is superseeded by PR #1343

PR #1335, opened 2025-07-22
Feature: Allow BMS_POWER to be used on a Lilygo if CONTACTOR_CONTROL isn't used

jonny5532:
### What
Add an extra check so that it does not error if you have enabled BMS_RESET as well as CAN_ADDON on a Lilygo, but not CONTACTOR_CONTROL. Normally there would not be enough GPIOs for three contactors as well as BMS_RESET, but you might not need the contactors.

In this instance, also set BMS_POWER to reuse the positive contactor pin by default.

### Why
This allows BMS_RESET to be used with CAN_ADDON on a Lilygo, if you are using a battery with CAN-driven contactors.


kyberias:
GPIO conflicts are now checked at runtime based on functionality and actual need for specific GPIO. This PR may not be needed.

PR #1336, opened 2025-07-22
WIP: Add a DISCOURAGE_CONTACTOR_OPENING define to stop Solax integration opening battery contactors during idle

jonny5532:
**Edit - am still having issues with this!**

Note - is DISCOURAGE_CONTACTOR_OPENING a good name? It could be:

AVOID_CONTACTOR_OPENING
PREVENT_IDLE_CONTACTOR_OPENING

or something...

### What
Add an optional DISCOURAGE_CONTACTOR_OPENING define, which if set, stops the Solax integration asking for battery contactors to be opened during idle or CAN timeout.

There is also a change to the state machine, that will jump straight from ANNOUNCE to WAITING_FOR_CONTACTOR/CONTACTOR_CLOSED if the define is set. This is necessary if the battery contactors are not going to be opening (or have been bypassed), since the Solax will otherwise wait forever to see 0V on the battery input, which will never come.

### Why
Some batteries hate contactor opening requests, which can cause the battery to go into fault state and require a power cycle to recover. We don't change the FAULT handling, which can open the contactors if the battery implements that.

Also, with bypassed contactors, the Solax integration is currently unusable (with an X1-AC, at least).

### TODO

Other inverter integrations (Kostal, SMA?) may need updating too (although we would need to check that they are OK with having the contactors closed all the time - the Solax X1-AC doesn't mind, at least).

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -18,7 +18,9 @@ void SolaxInverter::
     update_values() {  //This function maps all the values fetched from battery CAN to the correct CAN messages
   // If not receiveing any communication from the inverter, open contactors and return to battery announce state
   if (millis() - LastFrameTime >= SolaxTimeout) {
+#ifndef DISCOURAGE_CONTACTOR_OPENING
     datalayer.system.status.inverter_allows_contactor_closing = false;
+#endif
     STATE = BATTERY_ANNOUNCE;
   }
   //Calculate the required values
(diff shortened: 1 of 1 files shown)


Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -151,6 +155,13 @@ void SolaxInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
         // Byte 4 changes from 0 to 1
         if (rx_frame.data.u64 == Contactor_Close_Payload)
           STATE = WAITING_FOR_CONTACTOR;
+
+#ifdef DISCOURAGE_CONTACTOR_OPENING
```
dalathegreat, line 159: This is a  really good idea, BUT, for this new v9.0.0 version we are trying to move away from ifdefs, and instead have live configurable settings. Think you could convert this PR to be able to achieve that? 

Example code: https://github.com/dalathegreat/Battery-Emulator/pull/1343

jonny5532: Yeah this commit is a bit of a mess, I was going to tidy it up, and then had a lengthy discussion with Kyberias about a better way to handle this kind of thing - involving adding interfaces to the Battery and Inverter, to distinguish between 'allow' (which the contactors must obey), and 'request' (which the battery contactors can ignore if they want), plus a mechanism for the inverter to discover whether its requests will be ignored or not.

Will probably keep the hacky version in my MG branch, but sort out something more substantial with these new interfaces.

PR #1343, opened 2025-07-22
Feature: Add user configurable BMS reset off time

dalathegreat:
### What
This PR implements user configurable BMS reset off duration

### Why
Some BMS does not require a long cooldown when the reset occurs. Another usecase for setting a short off duration is frequency control, when user might require battery to be online as much as possible in order to not miss any frequency stabilization requests from the grid provider

### How
Value is now configurable in Webserver. 1-59 seconds. Default 30s.

image


dalathegreat:
@kyberias , while testing this new feature, I noticed that all the old configuration options are now causing PANIC RESET when you try to edit the values. Ideas?

image

kyberias:
> @kyberias , while testing this new feature, I noticed that all the old configuration options are now causing PANIC RESET when you try to edit the values. Ideas?
> 
> image

Let me check...

kyberias:
@dalathegreat Fixed by #1344

PR #1345, opened 2025-07-23
Reduce direct usage of global can_config

kyberias:
### What
This PR greatly reduces direct usage of global can_config structure and simplifies can frame transfer API.

### Why
Reducing dependencies to global config variables allows easier refactoring of how settings are managed at run-time.

### How
Base classes for battery, inverter, charger and shunt implement the necessary methods that don't require the caller to know the exact can interface used.

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -533,7 +533,7 @@ void init_webserver() {
 
   auto update_string = [](const char* route, std::function<void(String)> setter,
                           std::function<bool(String)> validator = nullptr) {
-    def_route_with_auth(route, server, HTTP_GET, [&](AsyncWebServerRequest* request) {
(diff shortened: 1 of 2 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -674,7 +674,7 @@ void NissanLeafBattery::transmit_can(unsigned long currentMillis) {
 
       //Only send this message when NISSANLEAF_CHARGER is not defined (otherwise it will collide!)
```
dalathegreat, line 675: Edge case, Double LEAF batteries, with a charger connected to one of the CAN channels, will cause the second battery to not send any 1F2 messages towards it. Could we solve this somehow?

kyberias: Yes, I was thinking in a separate PR though.

PR #1347, opened 2025-07-24
Add a 64-bit millis64() function to avoid rollover issues

jonny5532:
Query - is millis64 a good name? Have I put the files in a sensible place?

### What
Add a new millis64() function which does the same thing as the Arduino millis() function, but keeps all 64 bits of the response rather than truncating.

Replace millis() with millis64() in events timestamp handling.

Remove redundant code (get_timestamp, etc).

### Why
This avoids the need for any rollover handling, as the new 64-bit microsecond resolution will mean the timestamps won't rollover for 600k years.


Software/src/devboard/utils/millis64.cpp
```diff
@@ -0,0 +1,6 @@
+#include "esp_timer.h"
+
+uint64_t millis64() {
+  /* Return ESP32's high-resolution timer in milliseconds, as a 64 bit value. */
```
kyberias: Move the documentation to the header file. I guess our convention is more C++ comments.

jonny5532: Sure, done - I copied the format in some of the other headers, is that OK?

Software/src/devboard/utils/millis64.cpp
```diff
@@ -0,0 +1,5 @@
+#include "esp_timer.h"
+
+uint64_t millis64() {
+  return esp_timer_get_time() / 1000ULL;
```
dalathegreat, line 11: Might be a good note to add that we are now using the micros counter for ESP32 (and dividing it by 1000 to convert it to ms)

Software/src/include.h
```diff
@@ -10,6 +10,7 @@
 #include "devboard/hal/hal.h"
 #include "devboard/safety/safety.h"
 #include "devboard/utils/logging.h"
+#include "devboard/utils/millis64.h"
```
dalathegreat: Semi-related, but the library Software/src/lib/YiannisBourkelis-Uptime-Library/src/uptime.cpp has some very specific rollover checks that could be deleted, if we change to usingn this nenw millis64 function. Want to do that in the same PR? :)

Software/src/devboard/utils/millis64.cpp
jonny5532: Sure, how do you mean? As a comment in this file?

(Amusingly the code is basically identical to millis(), with just the return type changed:

https://github.com/espressif/arduino-esp32/blob/c7520ccef0c343611d7f9eb0afb3576f612de2f2/cores/esp32/esp32-hal-misc.c#L208

)

jonny5532: Have added a comment - also put the ARDUINO_ISR_ATTR attribute on which you apparently need if you want to use it in an ISR, which might happen.

Software/src/include.h
jonny5532: Have done! Mostly just deleting things.

dalathegreat:
Very nice work! Love seeing PRs that improve while also removing tons of lines of code.

We have one final blockcer before being able to merge this, the include.h has been removed due to PR https://github.com/dalathegreat/Battery-Emulator/pull/1353 just being merged. Can you merge main into your branch and fix this?

jonny5532:
Now rebased and import fixed.

dalathegreat:
@jonny5532 , I think we forgot to change all the currentMillis handlers to use the new datatype

  virtual void transmit_can(unsigned long currentMillis);

Could you fix this in a new PR?

jonny5532:
@dalathegreat - I guess it's optional whether we do this? The batteries probably don't need never-wrapping timestamps as they're generally just counting off seconds at most? The 64 bit should truncate down to a 32 bit currently so nothing has changed as far as the batteries are concerned.

I guess it might reduce potential bugs if people do timestamp-checking wrong (I saw you caught a bug like that recently).

The biggest downside is that we're now doing 64-bit instead of 32-bit arithmetic with all the timestamp checks in the batteries, and the Xtensa is a 32-bit proc, so that'll be noticably slower, although probably still extremely fast.

What do you think?

PR #1353, opened 2025-07-25
Eliminate include.h

kyberias:
### What
This PR removes references to include.h.

### Why
Having a common file that many modules include when they actually need only a small amount of functionality conflicts with many good C/C++ practices. Replacing this with more accurate module references helps to keep the code base organized, provides faster builds and allows building modules in smaller subsets for, for example, unit testing.

### How
The file include.h is removed and references replaced with more fine-grained and accurate header file references.


Software/src/battery/Battery.h
```diff
@@ -64,7 +64,7 @@ class Battery {
   virtual void update_values() = 0;
 
   // The name of the comm interface the battery is using.
-  virtual String interface_name() = 0;
+  virtual const char* interface_name() = 0;
```
kyberias, line 67: Changed to minimize using Arduino-types.

dalathegreat (approved):
:+1:

PR #1356, opened 2025-07-26
Tidy up MG-HS-PHEV include/ifdefs and add to BATTERIES.cpp

jonny5532:
### What

MgHsPhev was missing from BATTERIES.cpp, added.

Also a rogue #ifndef in MG-HS-PHEV-BATTERY.cpp stopped it building properly, removed.


dalathegreat:
Sync with main, tests should work then

dalathegreat (approved):

PR #1357, opened 2025-07-26
Make web log output show in correct order.

jonny5532:
### What
When rendering the `logged_can_messages` rolling buffer in the web log, show the old part from the offset before the newer part, so logs show consecutively.

Also preallocates a large enough String to avoid reallocations.

### Why
To make the logging output more readable.


dalathegreat (approved):
Very good improvement! :+1:

Software/src/devboard/webserver/debug_logging_html.cpp
```diff
@@ -4,8 +4,46 @@
 #include "index_html.h"
 
 #if defined(DEBUG_VIA_WEB) || defined(LOG_TO_SD)
+char* strnchr(const char* s, int c, size_t n) {
+  // Like strchr, but only searches the first 'n' bytes of the string.
+
+  // Check for a null pointer for safety.
```
kyberias: Perhaps we could be slightly less verbose with the comments. We can see from the code that there is a null pointer check.

jonny5532: Sure, I removed that one - the others seem somewhat helpful though? (Such pointer arithmetic is usually pretty incomprehensible...)

I admit I may have got Gemini to help with that, though I did check the logic myself. And removed a lot of other comments (the LLMs do love comments...)

PR #1358, opened 2025-07-26
Unit tests

kyberias:
### What
This PR adds some pure unit tests to the project.

### Why
Good unit tests coverage protects from regressions. Tests need to be easy to run and run fast so that there's a low barrier for developers.

### How
Tests are written using Google Test C++ unit testing framework. Select source code is built using CMake for Linux with Arduino dependencies emulated or stubbed to the extend needed by the tests. The tests are pure in a sense that they do not require specific hardware (namely ESP32) but can be run on a PC or as part of the build.


Software/src/inverter/ModbusInverterProtocol.h
```diff
@@ -15,14 +15,14 @@ class ModbusInverterProtocol : public InverterProtocol {
 
  protected:
   // Create a ModbusRTU server instance with 2000ms timeout
-  ModbusInverterProtocol() : MBserver(2000) { mbPV = ::mbPV; }
+  ModbusInverterProtocol();
```
kyberias, line 18: Revert, this is probably unnecessary

PR #1364, opened 2025-07-27
Safety: Improve stopping charge/discharge when battery is full/empty

dalathegreat:
### What
This PR implements a new safety check for stopping charge/discharge

### Why
With the introduction of negative scaled SOC (to be able to discharge to true 0%), we now only checked the safety on the reported SOC, and stopped if that was 0. This is a bit risky, so we now also stop if the real SOC% value is 0

This prevents undercharging batteries.

### How
Instead of stopping when
//Scaled SOC% value is 100.00%

We now stop when
//Either Scaled OR Real SOC% value is 100.00%

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -130,7 +130,8 @@ void update_machineryprotection() {
 
     // Battery is fully charged. Dont allow any more power into it
     // Normally the BMS will send 0W allowed, but this acts as an additional layer of safety
-    if (datalayer.battery.status.reported_soc == 10000)  //Scaled SOC% value is 100.00%
+    if (datalayer.battery.status.reported_soc == 10000 ||
+        datalayer.battery.status.real_soc == 10000)  //Either Scaled OR Real SOC% value is 100.00%
     {
       if (!battery_full_event_fired) {
         set_event(EVENT_BATTERY_FULL, 0);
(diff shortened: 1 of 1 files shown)


freddanastrom:
Generally this is a good idea. But for example the BYD Atto 3 batteries that currently suffers from soc drift you sometimes need to discharge real soc to 0% and continue the discharge to restore some of that drift. With this one it won’t be possible.

I don’t know if there are any other batteries that have a similar need, but maybe it would be possible to set a “flag” in the battery definition that makes it possible to override this extra safety condition?

dalathegreat:
Let's add a configuration option "Allow deep discharge", sounds like a good option for those that need to go below 0%

Alternatively, we could allow automatically for deep discharge if user sets negative SOC scaling?

freddanastrom:
On the other hand, maybe I didn’t think through my previous comment entirely..
To deep discharge the Atto 3 I used as example, you switch soc method to estimated soc. Which then becomes your real soc, right?
Then this new condition isn’t limiting anything after all.
Sorry for causing any confusion 🫣

dalathegreat:
@freddanastrom Yeah as long as the real soc is above 0%, then it should be fine!

freddanastrom (approved):

freddanastrom:
> @freddanastrom Yeah as long as the real soc is above 0%, then it should be fine!

All should be fine then 👍
For what it’s worth I gave it my approval, put don’t know if it helps when it says approved with read only permission.

PR #1365, opened 2025-07-27
Kostal inverter allows

gwk1:
### What
Change in inverter contactor closing.

Removed also some unused variables and added battery info frame status flag

### Why
Contactors must remain open during the startup sequence

### How
Set inverter_allows_contactor_closing to true after the inverter has requested 7 battery status messages.

Added a check to ensure that battery info is sent before battery status packets are transmitted.


dalathegreat (approved):
Nice work @gwk1 :raised_hands:

PR #1366, opened 2025-07-27
Kia/Hyundai 40/64: Respect inverter_allows_contactor_closing on startup

gwk1:
### What
Contactor closing allowed once the inverter is ready.

### Why
Some inverters doesn't like, if there is voltage at HV line during startup (eg. Kostal)

### How
Added a check for the inverter_allows_contactor_closing flag before sending certain CAN frames.


dalathegreat:
This PR should be renamed, since it essentially is only a fix for Kia/Hyundai 40/64kWh packs, making them wait until inverter closing is allowed.

I suggest renaming it to : Kia/Hyundai 40/64: Respect inverter_allows_contactor_closing on startup

PR #1367, opened 2025-07-27
New Inverter :zap: Add initial support for Growatt Wit CAN

dalathegreat:
### What
This PR implements basic support for [Growatt Wit inverters](https://github.com/dalathegreat/Battery-Emulator/wiki/Inverter:-Growatt-WIT). Paving the way for 100kW inverters for the Battery-Emulator

### Why
User requested feature on the Discord server

### How
TODO before testing. All CAN .ID definitions need to be set correctly. I could not figure out how to do this yet. See the code for TODOs

### DIFF
--- a/Software/src/inverter/InverterProtocol.h
+++ b/Software/src/inverter/InverterProtocol.h
@@ -12,6 +12,7 @@ enum class InverterProtocolType {
   Foxess,
   GrowattHv,
   GrowattLv,
+  GrowattWit,
   Kostal,
   Pylon,
   PylonLv,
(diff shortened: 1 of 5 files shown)


Software/src/inverter/InverterProtocol.h
```diff
@@ -12,6 +12,7 @@ enum class InverterProtocolType {
   Foxess,
   GrowattHv,
   GrowattLv,
+  GrowattWit,
```
kyberias, line 15: We should now have similar hard-coded numbers for all inverters as we have for batteries since we're using integer values when saving the selected inverter. Care to add them in this PR?

dalathegreat: I'd rather just merge this, and focus on another integration. But if it is as simple as just throwing an

enum class InverterType {
  None = 0,
etc
  Highest
};

into the INVERTERS.h, then I can do it

kyberias: Yes it's that simple, similarly as with batteries.

PR #1369, opened 2025-07-28
Library :scroll: Update ESPAsyncWebServer to v3.7.10

dalathegreat:
### What
This PR updates the ESPAsyncWebServer library we use

### Why
Mainly to fix the known vulnerability [GHSA-87j8-6f7g-h8wh](https://github.com/ESP32Async/ESPAsyncWebServer/security/advisories/GHSA-87j8-6f7g-h8wh)

### How
We update from v3.7.2 -> v3.7.10


jonny5532:
FYI BE isn't vulnerable to that vulnerability (`addHeader` is never called with untrusted input), but probably still worth upgrading!

PR #1378, opened 2025-07-31
Make Modbus inverters allocate their own register memory as required

jonny5532:
### What
Make Modbus inverters set up the Modbus register file when constructed, rather than having 25k of RAM statically allocated even if no Modbus inverter is in use.

Make inverters specify how many registers they use, meaning they don't need to allocate so much either.

### Why
Reduces COMMON_IMAGE memory usage by 25k, which fixes issues with the Log, amongst other things.

### Notes

The MBPV_MAX global defaults to zero, so the Modbus code shouldn't try to read/write the null pointer before an inverter sets up the buffer.

Previously there was a discrepancy between the old 13100 entry buffer array and the 30000 MBPV_MAX value that eModbus was using (an inverter could have overwritten other memory by writing to high registers).

**I have not tested this with a BYD-supporting inverter (I do not have one).** I tested that BE does at least start up with this inverter selected.


jonny5532:
One potential risk is that any Modbus activity to registers higher than 401 from the inverter will cause errors, where previously it'd just update unused registers. I have no idea whether the inverter does this - one option would be to add extra space.

Ideally this needs testing by someone with one of these inverters before merging.

dalathegreat:
So here is how the Modbus implementation for Modbus-BYD works; https://gitlab.com/pelle8/inverter_resources/-/blob/main/byd_registers_modbus_rtu.md

We use registers
101-167
201-213
301-324
401-419
12289-13057

This is why we allocated 13100 entries, to make room for that odd register writing at  those very high registers. Will this PR break this?

kyberias:
The best approach would probably to use a std::set or something instead if a huge array to dynamically allocate the memory needed. It's never 13k.

kyberias:
> The best approach would probably to use a std::set or something instead if a huge array to dynamically allocate the memory needed. It's never 13k.

Oh but the modbus library needs an array. Never mind.

kyberias:
> > The best approach would probably to use a std::set or something instead if a huge array to dynamically allocate the memory needed. It's never 13k.
> 
> Oh but the modbus library needs an array. Never mind.

Well it might not be too complicated to patch the library to allow something other than an explicit array for storage.

jonny5532:
Another option would be to silently ignore writes outside the range, and always return 0 for such reads.  This may or may not work depending on whether the inverter reads back these values. Plus it may cause future invisible issues.

Patching it to use std::set makes sense - eModbus itself doesn't actually access the registers, that's done by these BE-specific callbacks anyway.

I'll have a look.

jonny5532:
OK that ended up being a bigger change than I expected, I have:

- Changed `mbPV` from a `uint16_t[]` to a `std::map<uint16_t, uint16_t>`
- Moved the Modbus callbacks to be methods of `ModbusInverterProtocol` instead
- Tidied them up (removed excessive debugging)
- Made the `ModbusInverterProtocol` constructor set up the callbacks rather than the BYD setup method
- Moved various BYD globals into the class instead
- Changed the BYD register setup to not use `memcpy`

It works as best as I can tell - I can see my Solax trying to query it, and feed it back bogus data. I don't have a BYD-talking inverter, ideally someone who does can test this.

There is a significant memory improvement:

Before:
    Free heap: 79496, max alloc: 40948

After:
    Free heap: 101220, max alloc: 61428

and that will be even better if you're not instantiating this inverter.

dalathegreat:
There are some merge conflicts, could you fix these @jonny5532 , and I can take this PR for a spin with a Modbus inverter?

jonny5532:
@dalathegreat - I rearranged things a bit to fit in with @kyberias testing changes, and also added a few mocks to improve test coverage (very) slightly. Hopefully that hasn't broken anything (I can't easily retest with my Solax right now).

Software/src/lib/eModbus-eModbus/ModbusServerRTU.h
```diff
@@ -7,8 +7,6 @@
 
 #include "options.h"
 
-#if HAS_FREERTOS
```
kyberias: Is this modification necessary? I think we can define ESP32 for the tests to include this code.

jonny5532: I guess the choice is between `#ifdef` in the code/header files, and mock headers for testing?

I'd have thought mock headers were preferable, since it keeps the main code cleaner, and increases coverage (although mainly coverage of declarations than actual code...). A later integration test might mock the separate Modbus task (eg, using `std::thread` in place of xTaskCreate) or something?

Although I'm also not sure why tests.cpp is including the ModbusInverter header yet, as it doesn't actually test it?

jonny5532: Although I realise that does commit us to mocking all the ModbusRTU dependencies, which isn't even our library. Still, it would be neat to be able to end-to-end test it with actual messages.

The name `HAS_FREERTOS` seems misleading, as ModbusRTU doesn't have any FreeRTOS dependencies beyond starting its background task. It's mostly Arduino stuff, which you are mocking. Currently the only things that need disabling because of not-having-FreeRTOS are the task pointer and `begin()` call, which is still `#ifdef HAS_FREERTOS`.

kyberias: I have some work ongoing that succeeds in building all our code and libraries in tests but that isn't ready yet. In that I mock (or reimplement) most everything.

If ModbusInverter isn't yet covered with tests, I think you can just leave it out.

kyberias: Oh yes and the point absolutely should be that we don't modify our code or library code for tests.

jonny5532: Sounds good, but still unsure what you think I should do here. Options include:

1. Leave it as I have, with the `begin()` call gated, but the ModbusRTU stuff still in. With the two mock headers for the FreeRTOS includes.

2. Put the `#ifdef HAS_FREERTOS` back in, so the ModbusRTU stuff is largely removed from the test code.

3. Use a different define like `RUNNING_TESTS`, to exclude things like this, where you don't want them compiled into the test code? That seems clearer to me than `HAS_FREERTOS`, when the lack-of-FreeRTOS only affects a few lines of ModbusRTU.

If you don't want code or library code modified, then it seems better to me if we don't add these `#ifdef HAS_FREERTOS` lines if we don't need them?

kyberias: Sorry, I might be totally confused now.

I think #ifdef HAS_FREERTOS should be in library code as it was, since we don't want to modify library code. What that then means for our code, I'm not sure. If we need to add some more stubs or mocks in order to compile and run the tests, then let's do that.

jonny5532: Oh, my bad! When I rebased I assumed that you had added `#if HAS_FREERTOS` as part of your unit test changes. Infact Dala added it recently when upgrading to the latest eModbus library code (presumably it came from there).

I've now:
- changed the define from `__linux__` to `ESP32`, and added a bunch more stubs so that tests compile
- fixed a declaration bug in eModbus that stops it compiling (not sure why it doesn't affect the main build)

in addition to the original changes.

dalathegreat:
I took this PR into use today with a Modbus Fronius Gen24. It has been operational for 1hour, let's wait and see that it stays operational for a few more hours before we merge this PR! Nice work @jonny5532 

image

dalathegreat (approved):
Setting this as approved, inverter still operational after 24 hours! Nice work @jonny5532

PR #1379, opened 2025-07-31
New Battery :battery: : Add support for Samsung SDI LV

dalathegreat:
### What
This PR implements support for the 5kWh Samsung SDI Low Voltage 48V batteries

### Why
User requested feature

### How
Initial implementation should work good with 48V inverters. There are some TODOs regarding error mapping that can be further improved later on if this battery gets used by anoyne


kyberias (approved):
Looks good.

PR #1396, opened 2025-08-07
Readd watchdog (re)initialization code, disable idle timeouts, make AsyncTCP yield (discuss!)

jonny5532:
Edit: I'm an idiot, the current 5s timeout is fine (I initially suggested raising it to 10s).

### What
This code readds the watchdog initialization code that was removed in https://github.com/dalathegreat/Battery-Emulator/commit/600ef3751eaee80dc476c97f253a0815af171116 with some adjustments:

- Disable idle task watchdogs
- Improve AsyncTCP yielding

### Why
There are several issues:

1. Inconsistent initial watchdog state.

   PlatformIO and Arduino seem to have different defaults for the idle task watchdog setting, in their builds of ESP-IDF. You could also supply your own ESP-IDF sdkconfig with different watchdog settings. By resetting it it ensures that all build methods have the same behaviour. This requires a reinitialize rather than an init if ESP-IDF has already initialized them.

2. Idle watchdogs are redundant

   The idle watchdogs ensure that there is always some slack CPU time on the cores. It is possible to max out the CPU and block out the idle tasks by sending continuous HTTP traffic (you don't even need to auth) such that BE would hit the idle watchdog timeouts and reboot every 5s.

   We already have task watchdogs on the important tasks (core_loop, connectivity_loop, mqtt), so the idle watchdogs seem unnecessary.

3. AsyncTCP doesn't yield enough

   An explicit taskYIELD() in the webserver now means it no longer blocks other prio-3 tasks from running under heavy HTTP load. It was possible for the webserver to block other tasks for 23s (one example), which would force all the other core0 watchdogs to panic.

   The upstream seems to be a pinned version of AsyncTCPSock, hence making this change here.

### Potential issues

Disabling the idle timeouts will reduce the effectiveness of the watchdogs, although so far I think they may have caused more problems than they have solved, causing curious errors and nuisance panics. They are still essential for long term reliability though.

There are probably other solutions, like disabling all the watchdogs except for the core_loop one, and adding manual timeout handling code in that, so we can have different timeouts for different tasks (ESP32 only allows a single timeout for task watchdogs).

And maybe this is all unnecessary - most people aren't load-testing BE, or likely having any issues with this stuff...

### DIFF
--- a/Software/src/lib/mathieucarbou-AsyncTCPSock/src/AsyncTCP.cpp
+++ b/Software/src/lib/mathieucarbou-AsyncTCPSock/src/AsyncTCP.cpp
@@ -228,6 +228,11 @@ void _asynctcpsock_task(void *)
         sockList.clear();
 
         xSemaphoreGiveRecursive(_asyncsock_mutex);
+
+        // Battery-Emulator modification: Yield so that other same-priority
+        // tasks on the same core get a turn, otherwise heavy HTTP traffic will
+        // exclude them.
+        taskYIELD();
     }
 
     vTaskDelete(NULL);


jonny5532:
(initially I proposed increasing the timeout to 10s, but that's just because I accidentally hardcoded a 5s wait instead of 1s in my own code, 5s watchdog timeout is still fine)

dalathegreat (approved):

jonny5532:
(This also seems to have fixed my own experiments with custom OTA that used to abort with a stack-smashing error - presumably it was the idle watchdog crashing the chip somehow).

PR #1398, opened 2025-08-07
Add Goodwe CAN messages

dalathegreat:
### What
This PR implements ...

### Why
Why does it do it?

### How
How does it do it?


dalathegreat:
Not needed. Was user misconfiguration

PR #1402, opened 2025-08-07
Add static constexpr to lots of static battery CAN_frames to save RAM

jonny5532:
### What
Add lots of `static constexpr` or `const` to `CAN_frame` declarations in some of the larger battery header files.

Also tweak the CAN-frame-transmit functions to be `const`.

### Why
Frames that don't get modified by the integrations can be `const`/`constexpr`, which means they won't use up any RAM.

Saves 3-4kb RAM.

(I think that's worth it? Looks messier though! A much larger saving would be using a different struct for non-FD messages, which will save 56 bytes per frame, which I will add in a future PR...)


dalathegreat (approved):
Does not look so much messier to me, so I am approving this :+1:

PR #1403, opened 2025-08-07
Nissan LEAF: Add more ZE1 CAN message sending

dalathegreat:
### What
This PR simulates more of an entire 2018+ ZE1 LEAF vehicle towards the battery

### Why
There is a growing concern that 40/62kWh batteries are not balancing as they should. We speculate that we might not satisfy all CAN data needs that the ZE1 style LEAF batteries have.

### How
We add the following messages to the periodic sending, if we detect that we are on ZE1 battery:
- LEAF_355
- LEAF_3B8 // Sending 3B8 removes U1000 and P318E DTC
- LEAF_5C5 // Sending 5C5 removes U214E DTC
- LEAF_5EC 
- LEAF_626 // Sending 625 removes U215B DTC

CAN data taken from github.com/dalathegreat/Nissan-LEAF-Battery-Upgrade

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -296,6 +296,9 @@ void NissanLeafBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
         LEAF_battery_Type = AZE0_BATTERY;
       }
       break;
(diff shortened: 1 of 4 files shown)


dalathegreat:
Merging this, tested by one person on the Discord, and operational still after a few days! Merging to get more feedback easier!

PR #1417, opened 2025-08-14
Cleanup :broom: Remove obsolete uptime library

dalathegreat:
### What
This PR removes the YiannisBourkelis/Uptime-Library

### Why
The library was added as a good way for Arduino to handle millis() rollover after 49.7 days of uptime. With the introduction of our own millis64() call, we no longer have to worry about a rollover handler. We can reduce codesize by writing our own uptime formatter in the webserver.

### How
Uptime-Library is removed. 1000 lines of code removed from repository

This change reduces codesize slightly
Before:
RAM:   [==        ]  23.5% (used 76964 bytes from 327680 bytes)
Flash: [========  ]  76.0% (used 1493655 bytes from 1966080 bytes)

After:
RAM:   [==        ]  23.5% (used 76916 bytes from 327680 bytes)
Flash: [========  ]  76.0% (used 1493367 bytes from 1966080 bytes)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -794,21 +794,23 @@ String get_firmware_info_processor(const String& var) {
 
 String get_uptime() {
   static uint64_t milliseconds;
-  static uint64_t seconds;
-  static uint64_t minutes;
-  static uint64_t hours;
-  static uint64_t days;
+  static uint64_t remaining_seconds;
```
jonny5532: Btw I'm not sure any of these need to be static? Maybe it was from when it needed to track the rollover, but now they can just go on the stack (and save a tiny bit of RAM).

jonny5532 (commented):
Nice ratio: +22 −1,027

Software/src/devboard/webserver/webserver.cpp
dalathegreat: Fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1417/commits/dc01dea9b23981cdefce097cc22c5fcff2c9c011 , also reduced datatype size on some values

PR #1420, opened 2025-08-15
Zoe Gen2: Bugs related to two batteries

dalathegreat:
### What
This PR fixes a few bugs related to running 2x Zoe Gen2 batteries

### Why
Reported issues from Discord testing

### How
We do the following changes
- Speeds up cellvoltage reading. All cellvoltages are read after 3 seconds on boot. Previous version took ~20 seconds to read all cells after boot, so nice improvement
- Balancing for battery2 is now written correctly to datalayer


Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.cpp
```diff
@@ -163,33 +152,263 @@ void RenaultZoeGen2Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
-    case 0x5C9:
-    case 0x5CB:
-    case 0x5CC:
-    case 0x5D6:
-    case 0x5D7:
+      datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
+      break;
+    case 0x5F7:
+      datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
+      datalayer_battery->status.cell_voltages_mV[0] =
```
kyberias, line 159: Seems like this identical code is repeated a dozen times or so here. The code could be made also a little more readable (and save flash) by implementing a helper method, something like this:

void extract_cell_voltages(int index, CAN_frame frame) {
  datalayer_battery->status.cell_voltages_mV[index] =
    ((rx_frame.data.u8[0] << 4) | (rx_frame.data.u8[1] & 0xF0) >> 4) + 1000;

...

}

and then call that in each of the cases that extract 5 voltages.

dalathegreat: Good idea. 

Closing this PR, will re-open a new one once Common Image testing has settled. Critical fix included in PR https://github.com/dalathegreat/Battery-Emulator/pull/1469 for Zoe2

PR #1424, opened 2025-08-15
Make custom-BMS voltage limits configurable via settings

jonny5532:
### What
Add four new settings (pack/cell max/min voltage limits), and apply them to the custom-BMS batteries only.

I rejigged the settings JS/CSS/HTML a bit:
- tidying up some weird styling
- replacing JS-controlled hide/reveal behaviour with CSS
- enabled hide/reveal for dependent battery/inverter/charger/shunt/double-battery settings

### Why
Will be necessary for COMMON_IMAGE use of custom BMSes.

Also checking that I've added these settings correctly.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -152,6 +152,11 @@ battery_chemistry_enum user_selected_battery_chemistry = battery_chemistry_defau
 
 BatteryType user_selected_battery_type = BatteryType::NissanLeaf;
 bool user_selected_second_battery = false;
+// Defaults for the battery voltage limits, only applied with custom-BMS batteries
+uint16_t user_selected_max_pack_voltage_dV = 0;
+uint16_t user_selected_min_pack_voltage_dV = 0;
+uint16_t user_selected_max_cell_voltage_mV = 0;
+uint16_t user_selected_min_cell_voltage_mV = 0;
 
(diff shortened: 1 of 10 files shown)


dalathegreat:
Should we also make datalayer.battery.info.number_of_cells user customizable? Would be useful for some custom BMS integrations

jonny5532:
> Should we also make datalayer.battery.info.number_of_cells user customizable? Would be useful for some custom BMS integrations

All the custom BMS integrations also receive the cell count via CAN from the BMS, so I tweaked them to do that instead - there's no safety concern as it's just for cell monitor, and is one less thing to configure?

(see my other PR)

dalathegreat:
Testing with Daly and 250V - 500V allowed, nets these

image

dalathegreat (approved):
Approved! Tested OK :+1:

jonny5532:
@kyberias - I've excised all the COMMON_IMAGE stuff, and also removed the `static const` slots from the custom BMS batteries as they won't be usable anymore (people can just set those up via the web interface).

Is that what you had in mind?

kyberias:
> @kyberias - I've excised all the COMMON_IMAGE stuff, and also removed the `static const` slots from the custom BMS batteries as they won't be usable anymore (people can just set those up via the web interface).
> 
> Is that what you had in mind?

Yes, exactly, thank you!

PR #1425, opened 2025-08-16
Make more inverter settings configurable

jonny5532:
### What
Make the Pylon cells/modules/cells-per-module/voltage-level/capacity settings configurable on the three Pylon-ish inverters.

Make the two Solax modules/type settings configurable.

Add a 'ignore contactors' checkbox to Solax to stop it trying to open battery contactors.

### Why
Necessary for use of these inverters with COMMON_IMAGE.

This PR will probably need rebasing when you've merged the previous one...


Software/src/battery/CELLPOWER-BMS.cpp
```diff
@@ -231,8 +232,15 @@ void CellPowerBms::setup(void) {  // Performs one time setup at startup
   strncpy(datalayer.system.info.battery_protocol, Name, 63);
   datalayer.system.info.battery_protocol[63] = '\0';
   datalayer.system.status.battery_allows_contactor_closing = true;
+#ifdef COMMON_IMAGE
```
kyberias: We should minimize the use of  ifdevs such as COMMON IMAGE. Here the setting variable should be defined somewhere based on config and we would just use it with no additional conditions.

jonny5532: I thought the COMMON_IMAGE ifdef was a bit of a special case - it allows the new settings to be used for the new prebuilt images, but also means people can continue hacking USER_SETTINGS and the individual inverter files if they want.

At some point we'd presumably make COMMON_IMAGE the only option, and remove all the hardcoded constants at the same time?

(I could make the underlying settings vars configurable via USER_SETTINGS or something, but existing users will be annoyed if they can't just hack the battery/inverter .h file like they used to? At least for a version or two?

jonny5532: (One complication is the defaults - the settings are (somewhat) common across inverters, to avoid duplication, but the underlying defaults come from the inverter files themselves, which vary by inverter.)

jonny5532: Although the behaviour for the settings-with-defaults (eg, NUMBER_OF_CELLS) is to only use the user selected value if it's nonzero, otherwise you get the inverter's default. So you could lose the COMMON_IMAGE check for those, except that if there's a stored NVM variable which is nonzero, it'll take precedence, and if you don't have COMMON_IMAGE set, you can't change the settings to remove it (without factory resetting your device)...

kyberias: I'm not sure I follow what you mean, but let me try to clarify.

Consider CONTACTOR_CONTROL ifdef setting, for example. If not using common image, the user can define it in USER_SETTINGS.h or if we're using the common image, the user can switch it on/off in the UI. In both cases the **variable** contactor_control_enabled gets set accordingly.

When the setting is used, we no longer need or want to use neither ifdefs CONTACTOR_CONTROL nor COMMON_IMAGE, we just use the variable.

This is what I mean by minimizing the use of #ifdefs. Every ifdef we add is a code smell. There should be a very limited number of ifdefs especially for COMMON_IMAGE.

When we want to remove support for USER_SETTINGS.h, we simply remove usage of CONTACTOR_CONTROL (in exactly one place).

jonny5532: @kyberias - I've rearranged things, this should be better now?

kyberias: Yes, indeed, thanks!

PR #1437, opened 2025-08-19
Library :scroll: Update ACAN2517FD to version 2.1.16

dalathegreat:
### What
This PR updates the pierremolinaro-ACAN2517FD library from version 2.1.14 to version 2.1.16

### Why
Bug fix: incorrect data rate setting for x1 data bit rate factor and Data Bit Rate Switch. Before this release:

- CANFD sent frames always had the BSR bit clear, even if frame type is CANFDMessage::CANFD_WITH_BIT_RATE_SWITCH;
- CANFD network frames with the BSR bit set were not received.

### How
Note, after update we apply Battery-Emulator specific buffer settings, and a custom define DISABLEMCP2517FDCOMPAT to increase performance of the driver


dalathegreat:
Note before merging this. Needs testing on FD chip, both in FD and normal classic CAN mode

dalathegreat:
Tested successfully on bench with Stark CMR v2, sending both Normal CAN and CAN-FD

PR #1440, opened 2025-08-20
Update BYD-ATTO-3-BATTERY.cpp

bingflash2:
beta test 50kw battery

thanks to fredrik & johnny5532

### What
beta test 50kw battery

### Why
only 60kw is presently supported

### How
request lower voltage on contactor open with different can values from a log


dalathegreat:
This PR seems to have been based on an older version of main, so it is now trying to overwrite lots of previous Atto 3 fixes. Would it be possible for you to re-base towards latest main @bingflash2 ? Amazing work!

PR #1441, opened 2025-08-20
Enabler: External system limits into datalayer

dalathegreat:
### What
This PR paves the way for MQTT writing of allowed charge/discharge limits. External automation system could write allowed Ampere to the Battery-Emulator, to allow for smarter use of battery

### Why
User requested feature

### How
Note, this is only ½ the implementation. MQTT side is missing, since I do not have the skills or testing setup to get this working, maybe someone else can continue on that part?

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -1020,13 +1020,17 @@ String processor(const String& var) {
         content += formatPowerValue("Max discharge power", datalayer.battery.status.max_discharge_power_W, "", 1);
         content += formatPowerValue("Max charge power", datalayer.battery.status.max_charge_power_W, "", 1);
         content += "<h4 style='color: white;'>Max discharge current: " + String(maxCurrentDischargeFloat, 1) + " A";
-        if (datalayer.battery.settings.user_settings_limit_discharge) {
+        if (datalayer.battery.settings.MQTT_settings_limit_discharge) {
+          content += " (MQTT)</h4>";
+        } else if (datalayer.battery.settings.user_settings_limit_discharge) {
           content += " (Manual)</h4>";
         } else {
           content += " (BMS)</h4>";
         }
         content += "<h4 style='color: white;'>Max charge current: " + String(maxCurrentChargeFloat, 1) + " A";
-        if (datalayer.battery.settings.user_settings_limit_charge) {
+        if (datalayer.battery.settings.MQTT_settings_limit_charge) {
+          content += " (MQTT)</h4>";
+        } else if (datalayer.battery.settings.user_settings_limit_charge) {
           content += " (Manual)</h4>";
         } else {
           content += " (BMS)</h4>";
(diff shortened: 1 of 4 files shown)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -1020,13 +1020,17 @@ String processor(const String& var) {
         content += formatPowerValue("Max discharge power", datalayer.battery.status.max_discharge_power_W, "", 1);
         content += formatPowerValue("Max charge power", datalayer.battery.status.max_charge_power_W, "", 1);
         content += "<h4 style='color: white;'>Max discharge current: " + String(maxCurrentDischargeFloat, 1) + " A";
-        if (datalayer.battery.settings.user_settings_limit_discharge) {
+        if (datalayer.battery.settings.MQTT_settings_limit_discharge) {
```
kyberias: I think there should be a single setting that can be modified from multiple sources (MQTT and UI). Otherwise they may be in conflict and it's confusing for everyone.

kyberias: Also the setting name (internal variable) should not refer to MQTT. MQTT is just one protocol that may allow editing of this variable. We might in the future have more interfaces that support modifying the same variable.

dalathegreat: I thought about re-using user_set, BUT, this is stored to persistent, and for the external MQTT control we do not want to have it touch persistent memory. So I strongly believe this should be its own setting 

I agree with the other part, MQTT is just the protocol. Let's change it to (EXTERNAL) limit? :)

kyberias: Sorry I don't understand. Why we don't want the externally controlled MQTT setting to be stored to persistent memory as all other settings? I don't see any difference here.

It if has to be a different setting then there must be some fundamental reason for that to be different. Merely MQTT protocol isn't it.

dalathegreat: It causes unnecessary writing/wear. Persistent writing is not something we want to do every second.

Most external controllers work this way, Limits are only temporary. This is standard in EV charging / battery management

No-Signal: I would recommend a similar approach to how Solax inverters manage this where they split long term settings and temporary settings. I also think we need an internal service to enable changing these types of settings vis multiple channels e.g. UI, API, MQTT etc  https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/

dalathegreat: MQTT has been changed to External system in [a785a67](https://github.com/dalathegreat/Battery-Emulator/pull/1441/commits/a785a6765f6386c547c9d07a04a141dbad6c53ed)

I agree this could be refactored, but now we need a MVP for this feature. People keep asking me about this, so let's deliver something usable!

kyberias: Seems to me there should still be just one variable. When MQTT writes to it, it would update only the variable in memory. When UI updates it should also persist to NVM. When the system boots, it should read the variable from NVM.

dalathegreat: Ok, I tried :D Let's have the other person do that when they implement the MQTT. I close this then!

PR #1442, opened 2025-08-20
New Battery :battery: Add support for RELi³ON LV battery protocol

dalathegreat:
### What
This PR adds support for [48V ReLion batteries](https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Relion-LV), running via CAN at 250kbps (same as RJXZS BMS).

### Why
Feature requested via Discord

### How
image

Note: Since protocol runs at 250kbps instead of the industry standard 500kbps, the batteries need to be connected to the Native CAN port, since it is the only one that supports dropping the kbps


dalathegreat:
Testing cellvoltages OK: 
image

No-Signal (approved):
Successfully tested using CAN logs

PR #1443, opened 2025-08-20
CAN log based testing

jonny5532:
### What
This PR adds a testing mechanism that will replay CAN logs against individual batteries, currently checking that:
- All required values are set (and no safety events triggered) for `_good.txt` logs
- An overvoltage event is triggered for `_overvoltage.txt` logs

It is probably best to create these files from real CAN logs, for authenticity (the MG one is currently).

### Why
This should be an easy way of testing quite a lot of code at once:
- Battery initialization
- CAN frame processing
- Battery value updating
- Safety checks

and should also be relatively easy to populate if people can grab real CAN logs from their setups.


dalathegreat (approved):
Extremely nice work jonny! Feel free to merge whenever you feel this is ready! :heart_eyes:

PR #1444, opened 2025-08-20
BYD Atto 3: Send correct data in 441 message

freddanastrom:
### What
This PR sends the correct data in 441 message towards the battery, which is probably used as a confirmation from the car to the battery that it sees the same voltage as the BMS.

### Why
To fully support 50 kWh battery and solving the issue with contactors not staying closed.
Not sending the correct voltage could potentially cause issues with the 60 kWh if real voltage deviates too much from the previous static message.

### How
By sending the battery voltage and calculated checksum.


freddanastrom:
Needs testing on both 50 kWh and 60 kWh before merging!

freddanastrom:
Works fine on 60 kWh

freddanastrom:
Tested and also works with 50 kWh. contactors stay closed

dalathegreat (approved):

PR #1450, opened 2025-08-21
Adding event_level and event_level_color to mqtt information

No-Signal:
Adds event_level and event_level_color to mqtt payload and HA integration

event_level_color maps event levels to standard colors that can be used in HA


kyberias:
Can you explain how these colors are used on HA side? In principle it seems wrong to hard code something in BA MQTT to some other system's UI constants. I'm somewhat familiar with how HA works, so I'm thinking whether there might be some other options but we need to know more about the use case.

No-Signal:
The code is replicating the led color logic to enable users of the MQTT integration to visualise the BE status elsewhere. Ideally there should be a global BE status that drives the led color etc and this could be sent via MQTT but with all of the other changes I didn't want to add it for this.

kyberias:
What I'm trying to ask is that couldn't we report the event status (string?) INFO, DEBUG, WARNING etc. and do the mapping to colors at Homeassistant side where it belongs IMHO?

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -311,6 +313,31 @@ static bool publish_common_info(void) {
         set_battery_attributes(doc, datalayer.battery2, "_2", battery2->supports_charged_energy());
       }
     }
+
+    EVENTS_LEVEL_TYPE event_level = get_event_level();
+    doc["event_level"] = String(event_level);
```
kyberias: This results in a cast to integer and the event_level reported as number which might not be that useful. If you want to get a string representation of the event_level, you should call get_event_level_string().

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -311,6 +313,10 @@ static bool publish_common_info(void) {
         set_battery_attributes(doc, datalayer.battery2, "_2", battery2->supports_charged_energy());
       }
     }
+
+    doc["event_level"] = get_event_level_string(get_event_level());
+    doc["emulator_status"] = get_emulator_staus_string(get_emulator_status());
```
kyberias: typo: staus

No-Signal: Typo fixed and led_color enum removed as its now not used

PR #1453, opened 2025-08-21
Adding the ability to set a temporary maximum charge or discharge lim…

No-Signal:
Enables setting a temporary maximum charge and or discharge rate via MQTT

Follows a similar mechanism to SolaX inverters for temporary limits where the value is not persisted and only lasts for a short period of time as the expectation is for automations to push new values frequently. https://homeassistant-solax-modbus.readthedocs.io/en/latest/solax-mode1-modbus-power-control/

Topic: BE/command/set_limits

Message Structure:
{
	"max_charge": 30,
	"max_discharge": 10,
	"timeout": 50000
}

Charge and discharge rates are dA as per other settings
Timeout is milliseconds

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -292,7 +292,7 @@ void core_loop(void*) {
         battery2->update_values();
         check_interconnect_available();
       }
-      update_calculated_values();
+      update_calculated_values(currentMillis);
       update_machineryprotection();  // Check safeties
 
       // Update values heading towards inverter
@@ -377,7 +377,7 @@ void check_interconnect_available() {
   }
 }
 
-void update_calculated_values() {
+void update_calculated_values(volatile unsigned long currentMillis) {
   /* Update CPU temperature*/
   union {
     float temp;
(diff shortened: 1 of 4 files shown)


dalathegreat:
Did we not agree in https://github.com/dalathegreat/Battery-Emulator/pull/1441/files that MQTT implementation should just write to datalayer.battery.settings.max_user_set_charge_dA ? For an extremely simple solution?

No-Signal:
I don't like the idea of using the variable datalayer.battery.settings.max_user_set_charge_dA for multiple different purposes as it could lead to confusion.

If we agree to the temporary limit approach as used by Solax then we need both values as once the temporary limit is removed the user limit needs to restart as it may be set for safety reasons.

dalathegreat:
@No-Signal , could you solve the merge conflict, and we could get this merged?

No-Signal:
> @No-Signal , could you solve the merge conflict, and we could get this merged?

@dalathegreat merge conflicts resolved

dalathegreat (approved):

dalathegreat:
PR merged! Could you update the MQTT Wikipage with an example on how to use this feature?  https://github.com/dalathegreat/Battery-Emulator/wiki/MQTT

Issue #1449, opened 2025-08-20
[BUG] No CAN communication on Stark CMR


freddanastrom:
**Describe the bug**
Upgrading from 8.16 to 9.0.RC2 breaks CAN communication on Stark CMR.
Interface used is CAN FD native with classic CAN.
I can see frames being transmitted in the CAN logger, but doesn’t seem to reach the battery.
Unknown if regular native CAN is affected in the same way.

Tested on main with both common image settings and old fashioned file based settings, plus pre compiled bin with the same result.

Downgrade to 8.16 restores the communication.

**To Reproduce**
Use native CAN FD as classic CAN on Stark CMR with version 9.0.RC2

**Expected behavior**
Normal communication.

**Screenshots**

**Version and settings (please complete the following information):**
- Please attach your USER_SETTINGS files for easier debugging
- Software version: 9.0.RC2
- Battery used: BYD Atto 3
- Inverter communication protocol: None
- Hardware used for Battery-Emulator: HW_STARK
- CONTACTOR_CONTROL: no
- MQTT: yes

dalathegreat:
Unable to reproduce. Tested latest 9.0.RC2experimental

<img width="423" height="153" alt="Image" src="https://github.com/user-attachments/assets/d73ecb2d-6a9a-470c-8264-82d2948f9a92" />

Stark CMR v1 connected to Kvaser, sending Byd Atto 3 CAN messages into emulator. Code is accepting these, and comm works.

CAN-FD port is also outputting CAN messages as expected

<img width="478" height="154" alt="Image" src="https://github.com/user-attachments/assets/1e7512a3-3ec7-4ea3-ac86-7199b8b402ad" />

Are you perhaps using Stark CMR v1? If so HAL file needs editing for CAN-FD GPIO pin 17 -> 16 (easy to miss)

freddanastrom:
This is weird..
I'm using Stark V2 and has been flipping between 8.16 and 9.0.RC2, and RC2 refused to work.
Same builds, same bins files, same configs.
All of a sudden RC2 just works today. I have no clue why, same build as yesterday. Was my battery in a bad mood?

So yeah… sorry for the chaos 😅 But also kinda not sorry, because it kicked off the whole V1/V2 talk and landed in a really nice PR. Huge props to the devs for turning my confusion into something awesome! 🚀

PR #1454, opened 2025-08-21
Hardware: Add autodetection method for Stark V1 / V2

dalathegreat:
### What
This PR adds autodetection for Stark CMR v1 / v2 HAL

### Why
Stark CMR v1 has GPIO pin 16 for SCK, CMR v2 has GPIO pin 17. This makes the Stark CMR v1 fail to initialize the CAN-FD port for operation, unless user manually configures the HAL file

Fixes https://github.com/dalathegreat/Battery-Emulator/issues/1449

### How
We now autodetect, and based on the CPU used, we switch between pin 16 / 17

### DIFF
--- a/test/emul/Arduino.h
+++ b/test/emul/Arduino.h
@@ -25,12 +25,12 @@ void delayMicroseconds(unsigned long us);
 int max(int a, int b);
 
 class ESPClass {
-public:
-    size_t getFlashChipSize() {
-        // This is a placeholder for the actual implementation
-        // that retrieves the flash chip size.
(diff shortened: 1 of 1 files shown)


kyberias (commented):
👍

dalathegreat:
@kyberias , any idea on how to fix the Test that is failing? Stub the HAL.cpp file into the tests folder?

kyberias:
> @kyberias , any idea on how to fix the Test that is failing? Stub the HAL.cpp file into the tests folder?

I can fix that.

PR #1456, opened 2025-08-24
Tesla: Add SOC reset feature

greenoem:
### What
Adds Tesla pack SOC reset function.

### Why
On some batteries SOC drifts over time and becomes inaccurate, particularly LFP packs - the SOC reset function resets the BMS SOC based on Open Circuit Voltage (OCV), known as SocByOcv.

### How
Implements the same UDS commands Tesla's software does to reset SOC, only if current SOC <15% or >90%.

### DIFF
--- a/Software/src/devboard/webserver/advanced_battery_html.cpp
+++ b/Software/src/devboard/webserver/advanced_battery_html.cpp
@@ -25,6 +25,10 @@ std::vector<BatteryCommand> battery_commands = {
      [](Battery* b) {
        b->reset_BMS();
      }},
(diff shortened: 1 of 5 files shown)


Software/src/devboard/webserver/advanced_battery_html.cpp
```diff
@@ -25,6 +25,10 @@ std::vector<BatteryCommand> battery_commands = {
      [](Battery* b) {
        b->reset_BMS();
      }},
+    {"resetSOC", "SOC reset", "reset SOC?", [](Battery* b) { return b && b->supports_reset_SOC(); },
```
dalathegreat, line 28: Should we mention in the text shown to the user that the battery needs to be at the extreme ends? Either fully charged or discharged?

PR #1457, opened 2025-08-25
BYD Atto 3: Auto switch soc method

freddanastrom:
### What
This PR adds a function to automatically switch to estimated SOC when measured SOC reaches a predefined threshold.

### Why
To provide an automatic correction for SOC drift.

### How

- Switches to estimated SOC when measured SOC drops below 2%.
- Requires estimated SOC ≥ 5% to switch.
- Switches back to measured SOC when measured SOC rises above 2.5%.
This values can be adjusted.

To activate this feature you need to allow your inverter to have minimum discharge below 2%. If that isn't possible, use negative scaling.

Note:
This feature may allow the battery to reach 0% estimated SOC. This is intentional to correct drift as much as possible.
Safety is maintained by the existing BMS protections (min cell voltage, max discharge power) and the 0% voltage estimation seems to be quite conservative.
Automatic switching does not force full discharge; if always active it may only correct a few % during normal usage (e.g., overnight discharge).
If needed, we could however add a separate lower limit of estimated SOC.

Feedback is welcome.

> [!TIP]


dalathegreat:
Closing this as obsolete, since we have the new SOC calibration now in place! https://github.com/dalathegreat/Battery-Emulator/pull/2101

PR #1464, opened 2025-08-28
Ioniq 28: Correct SOH reading. Add isolation resistance to More Battery Info

dalathegreat:
### What
This PR fixes incorrect SOH% reading on Ioniq 28kWh packs

### Why
User reported bug on Discord

### How
We accidentally read bit 2/3 instead of 1/2

Bonus:
This PR adds Isolation Resistance in kOhm to the More Battery Info page
Code is also added with all optional PIDs, for future development if desired


dalathegreat:
Tested OK on the Discord server!

PR #1468, opened 2025-08-29
Improvement: Debug Web/USB as configurable setting

dalathegreat:
### What
This PR makes Web and USB a configurable setting in the Webserver

### Why
So users don't need to recompile software to get logging enabled. Target for 9.0 release to not have ifdefs

### How
Webserver options added under Settings page:

image

### DIFF
--- a/Software/src/devboard/utils/logging.h
+++ b/Software/src/devboard/utils/logging.h
@@ -4,6 +4,7 @@
 #include <Print.h>
 #include <inttypes.h>
 #include "../../../USER_SETTINGS.h"
+#include "../../datalayer/datalayer.h"
 #include "types.h"
 
 class Logging : public Print {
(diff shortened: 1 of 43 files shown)


Software/src/devboard/utils/logging.h
```diff
@@ -13,18 +14,23 @@ class Logging : public Print {
...
-#define DEBUG_PRINTLN(str) logging.println(str)
-#else
-#define DEBUG_PRINTF(fmt, ...) ((void)0)
-#define DEBUG_PRINTLN(str) ((void)0)
-#endif
+// Replace compile-time macros with runtime checks
+#define DEBUG_PRINTF(fmt, ...)                                                                  \
+  do {                                                                                          \
+    if (datalayer.system.info.web_logging_active || datalayer.system.info.usb_logging_active) { \
+      logging.printf(fmt, ##__VA_ARGS__);                                                       \
```
kyberias, line 27: The settings are now checked in two places: here in this macro and also in Logging.printf. I think it should be either, not both. I suggest we would log stuff consistently with these macros and the macros would perform the check and the Logging class would not check, it would just print.

To summarize:

- the macros should check whether (debug) logging is enabled.
- the Logging would check where to log (USB or web) and act accordingly.

In fact it would now make sense to define "levels" to logging. User could select one of these for "logging_level" with increasing number of messages:

- off = 0 (no logging)
- error = 1 (logs only errors)
- info = 2 (logs also info log messages)
- debug = 3 (logs also debug)

In addition, the user would select where to log: usb, web, ...

And then we define macros:

ERROR_PRINTF(
DEBUG_PRINTF(...)
INFO_PRINTF(

the ERROR macro would log if logging_level >= 1
the INFO macro would log if logging_level >= 2
and the DEBUG macro would log if logging_level >= 3

dalathegreat: As discussed on the Discord, lets absolutely do this, but we need to release functional software first!

dalathegreat:
@kyberias , any idea on how to add a mock logging class for tests? PR is IMO done once that is in :)

EDIT: I figured out how to run unit tests locally, added to CONTRIBUTING.md https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#local-unit-test-run

EDIT2: Tests are passing :heavy_check_mark:

PR #1479, opened 2025-09-01
Adding ESP32C6 hardware support

No-Signal:
Work in progress adding support for ESP32C6 hardware which supports dual native can

### DIFF
--- a/Software/Software.ino
+++ b/Software/Software.ino
@@ -31,7 +31,7 @@
 #include "src/devboard/wifi/wifi.h"
 #include "src/inverter/INVERTERS.h"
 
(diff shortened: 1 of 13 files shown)


Software/Software.ino
```diff
@@ -154,6 +157,7 @@ void setup() {
 // Loop empty, all functionality runs in tasks
 void loop() {}
 
+#ifdef ENABLE_SDCARD
```
dalathegreat, line 160: we have a new user configurable option for ENABLE_SDCARD. you can switch to using that instead to avoid IFDEFs

No-Signal: The ESP32-C6 does not have an SDMMC Host controller so I had to completely disable SDCard support to compile for it

dalathegreat:
Closing stale PR for now, feel free to re-open later!

PR #1484, opened 2025-09-03
Stellantis: Add initial support for 50/75kWh vans

dalathegreat:
### What
This PR implements basic support for 50/75kWh vans

### Why
Users on Discord started using these packs!

### How
For the mystery van platform (see wiki), we cannot use normal eCMP mappings, and instead need to rely on the values polled via OBD2 pid polling.


dalathegreat:
Tested OK on 75kWh battery!

PR #1494, opened 2025-09-06
Fix multiple bugs with native change_can_speed operation to fix BMW PHEV

jonny5532:
### What
The `change_can_speed` function was broken in a few ways:
- Didn't rerun the ACAN_ESP32_Settings constructor, so the actual bitrate register derivations didn't reoccur
- Returned the old speed without /1000 first, so after repeated calls the CAN speed would *1000 each time.
- Didn't reset the peripheral first (which seems to be necessary in my brief testing - needs more investigation a real setup to establish whether the reset is actually required)

### Why
To fix BMW PHEV battery (which changes CAN speeds).

### DIFF
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -1,6 +1,4 @@
 #include "comm_can.h"
-#include <algorithm>
-#include <map>
 #include "../../lib/pierremolinaro-ACAN2517FD/ACAN2517FD.h"
 #include "../../lib/pierremolinaro-acan-esp32/ACAN_ESP32.h"
 #include "../../lib/pierremolinaro-acan2515/ACAN2515.h"
@@ -10,6 +8,11 @@
 #include "src/devboard/safety/safety.h"
 #include "src/devboard/sdcard/sdcard.h"
 #include "src/devboard/utils/logging.h"
+
+#include <esp_private/periph_ctrl.h>
+
+#include <algorithm>
+#include <map>
 
 volatile CAN_Configuration can_config = {.battery = CAN_NATIVE,
                                          .inverter = CAN_NATIVE,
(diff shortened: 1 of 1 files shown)


Software/src/communication/can/comm_can.cpp
```diff
@@ -472,11 +472,41 @@ void restart_can() {
...
+    if (errorCode != 0) {
+      logging.print("Error Native Can: 0x");
+      logging.println(errorCode, HEX);
+    }
+    return oldSpeed;
   }
-  return speed;
+
+  // Return some arbitrary speed? (FIXME)
+  return CAN_Speed::CAN_SPEED_500KBPS;
```
dalathegreat: Could we just return 1 if the speed was changed successfully? and 0 if we failed?

jonny5532: The way this gets used is:

```
auto old_speed = change_can_speed(100kbit);
...
change_can_speed(old_speed);
```

So it can change it back to whatever it was once it's done the weird 100kbit stuff.

The battery doesn't actually know its own CAN speed (it isn't saved in the class instance, and most batteries just get the default 500kbit).

It would probably be sensible for batteries to have the CAN speed set explicitly, but that's a bit of work to update them all...

kyberias: Also please don't use integers to return success in a language with type safe bool.

jonny5532: I've now made it store the initial_speed on the CanBattery, and this is passed as a fallback default to `change_can_speed`.

It's still a bit weird, ideally change_can_speed could return some kind if INVALID speed, but I don't want to introduce that to the enum, and C++ doesn't have algebraic types or pattern matching...

jonny5532: Changed again, now returns a `bool` and there's a new `reset_can_speed()` method.

dalathegreat:
As discussed on the Discord, the BMW PHEV implementation is not 100% yet, but now when they continue it they will have better chances of success :) Let's merge this!

PR #1496, opened 2025-09-06
Fix RJXZS cell count population (and tidy)

jonny5532:
### What
The previous cell-count-removal broke RJXZS cell monitor - this should fix.

One potential issue is that it assumes the cell count is a multiple of 3. If not, the final cell(s) will show erroneous voltages. If we can find out what the voltage representation for a missing-cell is, this can be fixed.


puffnfresh:
I think a missing cell will come back as 0V. One thing which is important and doesn't seem addressed is that there could be zeros throughout the cells and we don't want to count those.

For example, I have 10S packs but one 12S module for each pack. This means you should see 3.7V * 10, 0V * 2 for each pack. The previous code handled this.

puffnfresh:
https://github.com/dalathegreat/Battery-Emulator/blob/52ffa5289bba68a8e5e81d8438cd55532b34b2f9/Software/src/battery/RJXZS-BMS.cpp#L77

jonny5532:
> I think a missing cell will come back as 0V. One thing which is important and doesn't seem addressed is that there could be zeros throughout the cells and we don't want to count those.
> 
> For example, I have 10S packs but one 12S module for each pack. This means you should see 3.7V * 10, 0V * 2 for each pack. The previous code handled this.

So there see to have been multiple things going on here:

- Dala removed the code you quoted, since the filtering-out-0-values is now done in the cellmonitor page
- However that broke it because prepopulated_cellvoltages was being left at 0 and not updated
- But my change should now fix it by setting prepopulated_cellvoltages to a large enough number to include all the cells

In theory, anyway, I don't have a RJXZS to test - does that sound right?

puffnfresh:
Sounds reasonable. I do have an RJXZS which I can test this on when I get time in the next few days.

The code does look a lot better. I vote to merge and raise an issue if RJXZS still has problems.

dalathegreat:
Yeah, it was already broken, so lets merge this and see if it works! Nice work @jonny5532 !

jellgrafik-oss:
Good evening gentlemen, unfortunately, no, only three cells out of 96 are displayed, the rest are missing! I can provide you with access to my system. That would be no problem! Please let me know. I work with Anydesk or Rustdesk. Good evening, Franz.

jonny5532:
There was a silly bug which should be fixed in https://github.com/dalathegreat/Battery-Emulator/pull/1507

PR #1509, opened 2025-09-08
Add tests to make sure batteries aren't renewing liveness on bogus CAN frames

jonny5532:
### What
This tests all CanBattery subclasses to make sure they don't acknowledge a deliberately-bogus CAN frame as being an indication of working CAN.

### Why
To check the CAN_STILL_ALIVE mechanism is reliable.


dalathegreat (approved):
Oh I'm liking where this testing is going! :raised_hands:

PR #1526, opened 2025-09-12
Switch T-2CAN back to dio, and try 16mb

jonny5532:
### What
Try and fix the T-2CAN factory image build yet again - one user reports that it might actually work this time.

The 2CAN should support qio, but this switches back to dio for now. Need to test against real hardware. Also switch back to 40mhz (even though the chip should be good for 80mhz).

It also uses 16mb (the true flash size) rather than 4mb - my assumption that it would work with the smaller size may be wrong.

Will experiment again if I can get my hands on an actual S3 to test.

### DIFF
--- a/.github/workflows/release-assets.yml
+++ b/.github/workflows/release-assets.yml
@@ -64,7 +64,7 @@ jobs:
 
     - name: 🛠 Build factory image for Lilygo 2-CAN
       run: |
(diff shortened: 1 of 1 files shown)


dalathegreat:
Was this OK to merge from testing?

jonny5532:
Yep apparently it worked - when I get my own board I'll investigate the settings a bit more thoroughly (can probably get a bit more performance).

PR #1530, opened 2025-09-12
Experiment: Cap Fronius max kWh as 3xHVM

dalathegreat:
### What
This PR...

### Why
To avoid this error

image

### How
Experiment: Dalathegreat — 22:07
We cap value btw to 60kWh (19.3*3=57.9), maybe we need to change the value to 57.9 to keep fronius happy and believe it has 3 HVM stacks in parallel?


dalathegreat:
Closing, does not help to cap to 57kWh, or 50kWh. Might need to go lower? But since this does not affect operation, let's not focus on it

PR #1537, opened 2025-09-14
Restore previous SMA TRIPOWER behaviour to fix regression/issues

jonny5532:
### What
A previous tidyup broke the SMA-TRIPOWER inverter integration by removing a flag-setting, and also changed the behaviour (sending the 12 init messages all-at-once rather than with 250ms between).

Revert back to the old behaviour until we can test against the real hardware and check the new behaviour would still work.

### Why
Fix actual and potential SMA-TRIPOWER breakage.


dalathegreat:
Now I remember why I simplified this. The Unit Tests :sweat_smile: Seems like the pushFrame needs to be stubbed in tests

jonny5532:
Despite the scary errors it was just a missing #include...

dalathegreat (approved):

PR #1541, opened 2025-09-15
Stop erroneous events when saving an empty SSID/pw

jonny5532:
### What
If you set a blank SSID or password, or set an SSID before setting a password, you get an erroneous save error. It does actually save, but putString returns 0 if the string is empty, which is currently interpreted as an error.

To be thorough it should probably getString and compare, since it can't tell the difference between saving an empty string and not saving (bad Arduino library design), but that's probably unnecessary for the moment.


dalathegreat (approved):
Nice quickfix!

PR #1543, opened 2025-09-16
Fix validation so that 8 character WiFi passwords work

jonny5532:
### What
Mirek P found a bug where exactly-8-character WiFi passwords didn't work.

This fixes a off-by-one so these now save.


dalathegreat (approved):
:+1:

PR #1554, opened 2025-09-20
Kostal revert startup counter.

mbuhansen:
### What
This PR revert startup counter back to 7, for close contactor on cyclic data.

### Why
contactor does not close, when 5E 02 message not is sent.

### How
How does it do it?


dalathegreat (approved):
:+1:

PR #1564, opened 2025-09-24
Improvement: Add configuration options for auth

dalathegreat:
### What
This PR implements configurable auth

### Why
Forgotten feature during migration to v9

### How

image


dalathegreat:
Thinking about this, maybe it should just be a password, and no username? And the password field will have to be password protected to edit. Hmm, this might require a larger refactor

PR #1566, opened 2025-09-24
Automatically set the version number during Actions builds

jonny5532:
### What
This replaces the version_number string in Software.cpp just before compiling in the Actions builds (both the on-push and on-release ones). The version number will be based on the current tag (with the leading "v" stripped), eg for releases:
`9.1.0`
and for random commits, it'll be the most recent tag, plus the number of commits since, plus the commit hash:
`9.1.0-2-g440bf2be`

### Why
Makes the automatically-built OTA images have sensible version numbers, and also saves having to push a commit that bumps the version number when releasing.

### DIFF
--- a/.github/workflows/compile-common-image-lilygo-2CAN.yml
+++ b/.github/workflows/compile-common-image-lilygo-2CAN.yml
@@ -30,6 +30,13 @@ jobs:
     steps:
     - uses: actions/checkout@v5
       name: Checkout code
+      with:
+        # Fetch a bunch of commits to pick up the most recent tag
+        fetch-depth: 200
+        # Include tags too
+        fetch-tags: true
+        # Specify the ref explicitly to avoid weird tag-related error
+        ref: ${{ github.ref }}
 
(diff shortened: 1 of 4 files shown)


jonny5532:
For example, this very commit will have the version `9.1.0-5-g1f94efde` (which is slightly confusing, since the commit itself has hash `440bf2b` - but GitHub is creating a merge commit for this PR, which is what gets used for PR builds...)

dalathegreat:
Will this affect releases also? Or just the builds inbetween?

jonny5532:
It should work for the release-assets workflow (I have tested in my own repo). The source zips on the release will still have the original `version_number` (but I guess that makes sense as it means that if people download and modify then the builds will have "9-dev" or whatever we set that to.)

(No rush on merging this if you want to test it more - I just had a go to see if it was feasible).

dalathegreat:
I think I want to close this PR, I feel like it is better to manage this manually for now, if that is OK? Thanks anyways Jonny!

PR #1567, opened 2025-09-24
Kia E-GMP: Update valid cellvoltage filter

dalathegreat:
### What
This PR improves cellvoltage validation for Kia EGMP

### Why
Incorrect values getting interpreted as actual cellvoltages

image

### How
Previous filter would filter out values below 1000mV, we now update this to below 1800 getting filtered outu


elco22:
Tested and updated cellvoltage filter 2600mV, solved issue.

PR #1568, opened 2025-09-24
Performance: Reduce likelyhood of TASK_OVERRUN when using Tesla

dalathegreat:
### What
This PR reduces the chances of getting TASK_OVERRUN when using the Debug options, especially with Tesla batteries

### Why
TASK overruns are bad, and can cause CAN message sending to get delayed

### How
USB logging in particular hurts performance the most.
- We remove some unnecessary stuff from the tesla logging, and shorten names on some other things.
- Periodically logged Wifi status also shortened
- Make it so that user can only configure one general logging method at once (None / Webserver Loging / USB Logging)
- Simplify sending of 10ms messages for increased perf

### DIFF
--- a/Software/src/battery/TESLA-BATTERY.cpp
+++ b/Software/src/battery/TESLA-BATTERY.cpp
@@ -113,33 +113,6 @@ inline const char* getHvilStatusState(int index) {
   }
 }
 
-inline const char* getBMSState(int index) {
-  switch (index) {
-    case 0:
-      return "STANDBY";
-    case 1:
-      return "DRIVE";
-    case 2:
-      return "SUPPORT";
-    case 3:
-      return "CHARGE";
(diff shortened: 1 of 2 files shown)


1ni3shi:
I'm getting CAN_BUFFER_FULL - MCP2515 message failed to send. Buffer full or no one on the bus to ACK the message!

I don't have any debug or any loggings enabled.

Is this issue related?

dalathegreat:
@1ni3shi Hard to say. The CAN_BUFFER_FULL can mean that there was actually nobody that replied on the CAN bus to ACK the message we just sent, which will not clear the buffer. But it can also mean that we tried to send too many messages at once, which overloads the buffer. So no real conclusion can be drawn from it unfortunately, but it is not a good message to be seeing!

PR #1575, opened 2025-09-27
Feature: Tesla - add initial/basic events

greenoem:
### What
This PR starts the implementation of adding key Tesla info/warnings/errors into native Events.

### Why
Most Tesla events and warnings are only viewable within the debug log. Some (such as pack SOC recalibration) are only active for a short period so will most likely be missed by the user.

### How
Extends and uses the existing/awesome BE events structure.


Software/src/devboard/utils/events.cpp
```diff
@@ -244,6 +249,12 @@ String get_event_message_string(EVENTS_ENUM_TYPE event) {
       return "Battery measurement unavailable. Check 12V power supply and battery wiring!";
     case EVENT_BATTERY_ISOLATION:
       return "Battery reports isolation error. High voltage might be leaking to ground. Check battery!";
+    case EVENT_BATTERY_SOC_RECALIBRATION:
+      return "The BMS updated the HV battery State of Charge (SOC) by more than 3% based on SocByOcv.";
```
dalathegreat: Please check if this triggers properly, printing % sign is dangerous in embedded systems, it can crash USB/Weblogging.

If possible, please rewrite the "3%" as "3percent" to make iti safe for all print variants!

greenoem: Good catch, thanks, changed to "3pct" for brevity.

PR #1589, opened 2025-10-01
Adding Modbus to Lilygo T-2CAN and moving WUP pins

pmajor021:
### What
This pull request introduces a Modbus addon to the first QWIIC connector and reassigns the WUP pins to GPIO 1 and 2, located on the second connector.

### Why
It makes possible to use Modbus on T-2CAN board.
Relocating the WUP pins to GPIO1-2 simplifies using dual i3 batteries since they only require two GPIOs.

### How
Updating the HAL file as attached.


dalathegreat:
Closing this, superseeded in PR #1597

PR #1590, opened 2025-10-01
Improvement: Increase CAN add-on buffer size to 32

dalathegreat:
### What
This PR implements ...

### Why
Why does it do it?

### How
How does it do it?


mbuhansen:
For some reason i cant compile this image in vscode, vscode set error. But i have try to make the same changes in v9.1.3 and make all buffer the same size, and now v9..1.3 is working on my kostal double BMW i3 setup without buffer event.
I also think this with buffer size was the reason that canFD runs better than mpc2515 thet only have 16 and canfd has 22, in the beginning i use mpc2515 but get alot of buffer event failure.

dalathegreat:
Nice, merging this fix then :raised_hands:

PR #1595, opened 2025-10-01
Improvement: Move SSID/Password to common image

dalathegreat:
### What
This PR refactors the SSID/Password handling to be part of the common image instead of freefloating Edit elements

### Why
Stark boards have issues with OTA. Maybe this helps?

### How
Proper password field, and part of the common image
image

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -32,18 +32,8 @@ void init_stored_settings() {
 
   esp32hal->set_default_configuration_values();
 
-  char tempSSIDstring[63];  // Allocate buffer with sufficient size
-  size_t lengthSSID = settings.getString("SSID", tempSSIDstring, sizeof(tempSSIDstring));
-  if (lengthSSID > 0) {  // Successfully read the string from memory. Set it to SSID!
-    ssid = tempSSIDstring;
-  } else {  // Reading from settings failed. Do nothing with SSID. Raise event?
-  }
-  char tempPasswordString[63];  // Allocate buffer with sufficient size
-  size_t lengthPassword = settings.getString("PASSWORD", tempPasswordString, sizeof(tempPasswordString));
-  if (lengthPassword > 7) {  // Successfully read the string from memory. Set it to password!
-    password = tempPasswordString;
-  } else {  // Reading from settings failed. Do nothing with SSID. Raise event?
-  }
+  ssid = settings.getString("SSID", "REPLACEME").c_str();
+  password = settings.getString("PASSWORD", "REPLACEME").c_str();
 
   temp = settings.getUInt("BATTERY_WH_MAX", false);
   if (temp != 0) {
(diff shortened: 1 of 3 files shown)


Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -132,8 +132,13 @@ String settings_processor(const String& var, BatteryEmulatorSettingsStore& setti
       return battery->interface_name();
     }
   }
+
   if (var == "SSID") {
-    return String(ssid.c_str());
+    return settings.getString("SSID", "REPLACEME");
```
jonny5532, line 137: Is it better to default to REPLACEME or just an empty field? Some people don't use the SSID/PASSWORD (they use the AP instead), so they'd have to remember to delete or it'll try and connect...

(you can always put a placeholder on the field if you want to encourage people to fill it in)

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -32,18 +32,8 @@ void init_stored_settings() {
...
-    ssid = tempSSIDstring;
-  } else {  // Reading from settings failed. Do nothing with SSID. Raise event?
-  }
-  char tempPasswordString[63];  // Allocate buffer with sufficient size
-  size_t lengthPassword = settings.getString("PASSWORD", tempPasswordString, sizeof(tempPasswordString));
-  if (lengthPassword > 7) {  // Successfully read the string from memory. Set it to password!
-    password = tempPasswordString;
-  } else {  // Reading from settings failed. Do nothing with SSID. Raise event?
-  }
+  ssid = settings.getString("SSID", "REPLACEME").c_str();
```
jonny5532, line 35: This makes it impossible to disable WiFi? (if you're using AP only)

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -983,18 +982,28 @@ const char* getCANInterfaceName(CAN_Interface interface) {
...
           <p>Settings saved. Reboot to take the new settings into use.<p> <button type='button' onclick='askReboot()'>Reboot</button>
         </div>
 
+        <div class="settings-card">
+        <h3>Network config</h3>
+        <div style='display: grid; grid-template-columns: 1fr 1.5fr; gap: 10px; align-items: center;'>
+
+        <label>SSID: </label>
+        <input type='text' name='SSID' value="%SSID%" 
+        pattern="[A-Za-z0-9!#*-]{8,63}"
```
jonny5532, line 998: We definitely need to support SSIDs less than 8 chars, and with spaces and underscores in. We could be more generous with the regex, maybe:

[ !#-;=?-~]{1,63}

which would match every printable ASCII char except ", < and >, which skirts the XSS issues the templating creates.

dalathegreat: I have been trying to get this better for a few hours unsuccessfully. The current implementation does not seem to block dangerous input (Like for instance Wifi24#¤%&>><"), and entering it breaks the page completely, even though the regex should block it.

Augh, I can't seem to thet this right :anguished: 

AI says;
The issue is that your regex pattern is not actually blocking the dangerous characters - it's just providing client-side validation that can be easily bypassed. When you enter invalid characters, the browser might still submit the form, and your server-side code is not sanitizing the input before rendering it back in the template.

But the suggested steps are massive to not only rely on client side validation...

dalathegreat:
Superseded by #1615

PR #1600, opened 2025-10-03
Fix three select-case-fallthrough bugs (and enable compile warning flag)

jonny5532:
### What
Turns out there's a compiler flag (`-Wimplicit-fallthrough`) to detect missing `break;` statements in select cases.

This enables it, and fixes three bugs that it detected:

- `setup_charger()` always chose `NissanLeafCharger` regardless of what you selected.  
- CMFA-EV-BATTERY would overwrite the first cellvoltage sometimes.  
- KiaHyundai64 wasn't usable as a second battery.

The Tesla 0x212 bug is left unfixed (as it is in a separate PR).


dalathegreat (approved):
Massive thanks @jonny5532 :raised_hands:

PR #1601, opened 2025-10-04
Feature/sungrow sbr096 can

aferrara:
### What
This PR implements the Sungrow SBRXXX battery protocol. Tested against an Australian SH10RS inverter.

### How
It implements the initial discovery/init handshake, and then sends the required messages to appear in iSolarCloud / WiNet-S2 interface. It still is not ready for prime time, but enough of a step to allow others to take a look.

The CAN buffer needed to be increased to allow the amount of messages to be sent on the wire.

I have implemented a mechanism to initialise the CAN interface to 250k as this is required for the Sungrow protocol.


Software/src/communication/can/comm_can.cpp
```diff
@@ -508,6 +539,7 @@ uint32_t init_native_can(CAN_Speed speed, gpio_num_t tx_pin, gpio_num_t rx_pin)
   settingsespcan->mRequestedCANMode = ACAN_ESP32_Settings::NormalMode;
   settingsespcan->mTxPin = tx_pin;
   settingsespcan->mRxPin = rx_pin;
+  settingsespcan->mDriverTransmitBufferSize = 64;
```
jonny5532, line 542: The default TX buffer size has been gently increasing due to some of the other integrations, but 64 seems quite high - I guess it sends all these messages in one big go? Is that how the Sungrow batteries do it, or could we maybe spread them out across a few 10ms iterations?

aferrara: Yeah, the Sungrow protocol sends a large number of messages in the 1000ms interval.

If there is consensus to keep the buffer size small, I could just iterate through the 1000ms list over several passes. What would be a sane buffer size to target?

jonny5532: Another PR seems to raise the default buffer to 32 messages so I guess that's the max so far.

The only real cost is RAM, at 72 bytes for an FD frame that's 2304 vs 4608 bytes, probably not a huge deal I guess (BE runs with about 89k free RAM normally), although we might have 3 active CAN interfaces. Sending the 1000ms messages in two batches is probably not too awkward?

(There might even be an advantage to buffering more frames - we should look at moving the CAN sender routines into IRAM which executes far faster than the flash. However this is somewhat irrelevant here given that it only needs to send at 1000ms intervals. I need to do some benchmarking...)

aferrara:
Ok, let me refactor my code and I will target the 32 message buffer size. I have also already refactored the code in CanInverterProtocol.h to set the CAN speed. I should have all this done in a few days.

dalathegreat:
32 buffer size PR has been merged: https://github.com/dalathegreat/Battery-Emulator/pull/1590

I see you had a better method of forcing the buffer size via settings! Very clever, will need to switch to that also :+1:

PR #1607, opened 2025-10-05
Bugfix: Make it possible to save 0 as minpercentage

### LINKED ISSUE #1542: [BUG] Scaled Minimum SOC gets restored to 20% on reboot
**Describe the bug**
Minsoc gets restored to default value when periodic reboot occurs

**To Reproduce**
Use Scaled SOC, set limits to 0-90, and enable periodic BMS reboot every 24h

**Expected behavior**
Minsoc stays at 0%

Actual behaviour, Minsoc reverts to 20% (default value)

**Screenshots**

- Software version: 9.0.0
- Battery used: [Nissan LEAF]
- Inverter communication protocol: [Any]
- Hardware used for Battery-Emulator: [Any]
- GPIO controlled contactors: [yes/no]
- MQTT: [yes/no]

dalathegreat:
### What
This PR makes it possible to use MINSOC as 0% when using Scaled SOC

### Why
User reported issue. Fixes #1542 

### How
We now allow values set to 0 to be used at boot

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -54,7 +54,7 @@ void init_stored_settings() {
     datalayer.battery.settings.max_percentage = temp * 10;  // Multiply by 10 for backwards compatibility
   }
   temp = settings.getUInt("MINPERCENTAGE", false);
-  if (temp != 0) {
+  if (temp < 499) {
     datalayer.battery.settings.min_percentage = temp * 10;  // Multiply by 10 for backwards compatibility
   }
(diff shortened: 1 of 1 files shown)


jonny5532 (approved):
I guess the max-50% is arbitrary? Someone might want to do something really weird like cycle their battery between 60-70% only, although that would be strange, and also bad for the battery...

dalathegreat:
Yes we do validation on the webserver that minsoc should be between 0-49%, and Maxsoc between 51-100%. If we do not validate, user can set "dangerous" values that break the calculation. Good enough for now :)

PR #1609, opened 2025-10-05
Bugfix: Make capacity writing work for Ferroamp

dalathegreat:
### What
This PR fixes the capacity writing for Ferroamp protocol

### Why
User reported feature broken on Discord

### How
The reported values entered in the Webserver is now mapped periodically to the 7321 message

image

Bonus: All Send0 messages removed from the Ferroamp protocol, to save system resources.


dalathegreat:
Tested OK by Pflodin on the Discord server! Merging this!

PR #1610, opened 2025-10-05
Precharge: Move evaluation of max precharge time to always be checked

dalathegreat:
### What
This PR moves the max precharge timeout check to always be evaluated

### Why
Avoid odd scenario where precharge would fail, but not turn off (110V issue)

### How
Now we always check the timeout while we are in active precharge state


dalathegreat:
Successfully tested by @StarkJohan on the Discord! :raised_hands:

PR #1614, opened 2025-10-06
Implement HTML escaping on index/settings, relax validation a bit

jonny5532:
### What
This PR makes a few changes:
- Specifies page encoding as UTF-8, to avoid ambiguity
- Wraps `settings_processor` to html-escape settings when printed back to the page (excluding options settings).
- Adds html escaping to the SSID and hostname on the index page.
- Changes some of the settings regexes:
  - AP name/pass and MQTT user/pass now allow all printable ASCII characters. Technically SSIDs can contain any UTF-8 chars (including emoji etc) but we don't allow that for now.
  - Fixes existing regexes that use `-` which needs to be escaped as `\-` apparently.

### Why
I don't believe the NVS-corruption issues are caused by special characters (I have successfully added < > " to my SSID without issue, but having < > " etc unescaped in the HTML causes the pages to break.

We will exclude all non-printable-ASCII chars from SSIDs as this seems safe and will avoid encoding confusion.

(Although that may need to change if people want, eg, chinese characters in SSIDs)


dalathegreat (approved):
Amazing, thanks for fixing this :raised_hands:

PR #1620, opened 2025-10-06
BMW iX: Fix incorrectly applied min/max voltages

dalathegreat:
### What
This PR makes it possible to charge/discharge 78S iX1 batteries properly

### Why
User reported bug on Discord, not possible to charge battery due to 0kW allowed even though it should allow

### How
We now define limits separately for 78S, 96S and 108S packs. This PR makes all iX integrations more robust


dalathegreat:
Tested OK by pfaraday on the Discord server! :raised_hands:

PR #1639, opened 2025-10-11
Bugfix: Make devkit build compile properly

dalathegreat:
### What
This PR makes it possible to build ESP32_DEVKIT

### Why
User reported bug on Discord

### How
We add the missing compiler flag


dalathegreat:
Tested OK :+1:

PR #1640, opened 2025-10-12
Update BMW PHEV Battery to support unified image

wjcloudy:
Patches ACAN library to properly clear mDriverIsSending state when reinitializing

### What
Patch to support the BMW SME magic wake packet
### Why
New CAN library requires update

### How
Patches the ACAN library to clear sending state on bus speed changes.

### DIFF
--- a/Software/src/battery/BMW-PHEV-BATTERY.cpp
+++ b/Software/src/battery/BMW-PHEV-BATTERY.cpp
@@ -211,7 +211,7 @@ void BmwPhevBattery::parseDTCResponse() {
     logging.println("DTC count exceeds buffer, truncating");
   }
 
-  int validDtcCount = 0;  // ✅ Track actual valid DTCs
+  int validDtcCount = 0;  // Track actual valid DTCs
 
   logging.print("Parsing DTCs (max ");
   logging.print(maxDtcCount);
@@ -232,8 +232,7 @@ void BmwPhevBattery::parseDTCResponse() {
                        (uint32_t)gUDSContext.UDS_buffer[offset + 2];
 
     uint8_t dtcStatus = gUDSContext.UDS_buffer[offset + 3];
-
-    // ✅ Skip invalid DTCs (0x000000 or status 0x00)
+    // Skip invalid DTCs (0x000000 or status 0x00)
     if (dtcCode == 0x000000 || dtcStatus == 0x00) {
       logging.print("  Skipping invalid DTC at offset ");
       logging.println(offset);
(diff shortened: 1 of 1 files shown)


Software/src/battery/BMW-PHEV-BATTERY.cpp
```diff
@@ -176,18 +208,17 @@ void BmwPhevBattery::wake_battery_via_canbus() {
...
+  delay(10);
   transmit_can_frame(&BMW_PHEV_BUS_WAKEUP_REQUEST);
-
-  // FIXME: This might not wait for the above frames to send before it changes
-  // the speed back. A state-machine controlled delay is likely necessary.
-
-  reset_can_speed();
-
   logging.println("Sent magic wakeup packet to SME at 100kbps...");
+  delay(50);
```
dalathegreat: delay() calls are usually a no-go in embedded, since it will halt the CPU entirely and not be able to process any other incoming tasks. Calling two consecutive delay(50) blocks will cause TASK_OVERRUN (since normal 10ms control loop wont update properly), and if the CAN load on any other component is high (Inverter/Charger/Shunt), we might drop messages if RX queue fills up.

I think we can approve this, BUT, a better approach would be to use a state machine, that we can exit in between, that just keeps track of the timestamps when the CAN change was requested etc. There are some examples of these in the code, I think the same approach as used in ECMP could be suitied  here? https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/src/battery/ECMP-BATTERY.cpp#L1196

Basically at the top of the transmit CAN section, we would have an if check that sees if we should wake up, and if we do, we halt all sending. Inside the if check, we would then perform the can-speed change, and delay via checking if millis has progressed enough after we started the process. So we would exit the code section, but keep coming back to it each time the transmit_can is called.

Again, not mandatory to do now (getting the PHEV working is more important), but I really think we should eliminate delay() block use completely for the entire Battery-Emulator project, since it messes up control loops and risks dropping messages.

Software/src/battery/BMW-PHEV-BATTERY.cpp
```diff
@@ -501,53 +652,72 @@ void BmwPhevBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
-              }
-              logging.print(gUDSContext.UDS_buffer[i], HEX);
-              logging.printf(" ");
-            }
-            logging.println();  // new line at the end
           }
         }
 
+        // DTC Response ($19 returns $02)
+        if (gUDSContext.UDS_moduleID == 0x02) {  // ✅ Changed from 0x02 to 0x59
```
dalathegreat: Let's remove emojis from raw .cpp files :sweat_smile:

PR #1643, opened 2025-10-13
Kia64FD: Bugfix, make startup voltage value sane

dalathegreat:
### What
This PR makes the startup voltage sane for Kia64FD

### Why
To avoid fake undervoltage event on startup before values have been read

### How
Instead of initializing  battery voltage to 0V, we default it to 370.0V


bilal-512:
very good appraoch

PR #1644, opened 2025-10-14
Initial Sungrow SBRXXX protocol implementation

aferrara:
While this is functional, there is still some testing to be done before it should be widely used.

### What
This PR implements the Sungrow SBRXXX protocol over CAN

### Why
Sungrow inverters are starting to limit the types of batteries that can be used. In the AU market, Sungrow already have dropped BYD on some inverters, necessitating the use of this protocol.

### How
It implements the initial discovery/init handshake, and then sends the required run messages to appear in iSolarCloud / WiNet-S2 interface.

I have implemented a mechanism to initialise the inverter CAN interface to 250k as this is required for the Sungrow protocol.

It still is not ready for prime time, but enough of a step to allow others to take a look.

### DIFF
--- a/Software/src/devboard/utils/types.h
+++ b/Software/src/devboard/utils/types.h
@@ -46,7 +46,6 @@ enum PrechargeState {
 #define INTERVAL_500_MS 500
 #define INTERVAL_640_MS 640
 #define INTERVAL_1_S 1000
-#define INTERVAL_1_5_S 1500
 #define INTERVAL_2_S 2000
 #define INTERVAL_5_S 5000
 #define INTERVAL_10_S 10000


Software/src/inverter/SUNGROW-CAN.cpp
```diff
@@ -312,42 +452,233 @@ void SungrowInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
...
+bool SungrowInverter::setup() {
+  // 10-digit unique from eFuse MAC
+  uint64_t mac = ESP.getEfuseMac();
+  uint32_t folded = (uint32_t)((mac >> 32) ^ (mac & 0xFFFFFFFFULL));
+  uint64_t unique10 = (uint64_t)folded % 10000000000ULL;  // 0..9,999,999,999
+
+  char uid10[11];
+  snprintf(uid10, sizeof(uid10), "%010llu", (unsigned long long)unique10);
+
+  // --- Serial number: "S" + 10-digit (e.g., S1234567890) ---
```
dalathegreat: Seems like a very complicated method to generate an UID for serial numbers. **Are we sure that is even needed?** Is there some technical reason we cannot just have a static message?
 
Our reasoning on other protocols is that we keep the serial number the same on all Battery-Emulator installs, and if there ever is a problem in the future we can ofcourse implement randomized serials.

Software/src/inverter/SUNGROW-CAN.cpp
```diff
@@ -312,42 +452,233 @@ void SungrowInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
...
-      transmit_can_frame(&SUNGROW_501);
-      transmit_can_frame(&SUNGROW_502);
-      transmit_can_frame(&SUNGROW_503);
-      transmit_can_frame(&SUNGROW_504);
-      transmit_can_frame(&SUNGROW_505);
-      transmit_can_frame(&SUNGROW_506);
-      transmit_can_frame(&SUNGROW_500);
-      transmit_can_frame(&SUNGROW_400);
-      alternate = false;
+  // ---- constants (local to function to keep header clean) ----
```
dalathegreat: If I am being really honest here, I think this transmit_can code section is unmaintainable. It has so much going on, so many different terms, checks, arrays, queues, that I don't think anyone is going to be able to build upon it in a meaningful way.

Software/src/inverter/SUNGROW-CAN.cpp
```diff
@@ -312,42 +452,233 @@ void SungrowInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
...
+
+  lastSliceMs = currentMillis;
+
+  // CAN sequence chunks
+  static CAN_frame* const SEQ_HEAD[] = {&SUNGROW_500, &SUNGROW_400, &SUNGROW_401};
+
+  static CAN_frame* const SEQ_TAIL[] = {&SUNGROW_512, &SUNGROW_501, &SUNGROW_502, &SUNGROW_503,
+                                        &SUNGROW_504, &SUNGROW_505, &SUNGROW_506};
+
+  static CAN_frame* const SEQ_RUN[] = {
```
dalathegreat: Instead of doing all this, we had an example of batch sending in the SMA protocol: https://github.com/dalathegreat/Battery-Emulator/blob/e7eb814b749d697f00ef3d1fe36cec4dbeb01d76/Software/src/inverter/SMA-BYD-HVS-CAN.cpp#L244

I think we could probably remove 200-300 lines of code by doing it in a similar way here

aferrara:
Thanks for the feedback. I have refactored my changes to make them as simple and as readable as possible, and removed the dynamic serial number code.

dalathegreat (approved):
Very nice work! Approved :heavy_check_mark:

PR #1645, opened 2025-10-15
Improvement: Fronius real-time capacity display

pmajor021:
Use summary values for remaining and maximum capacity when 2 batteries are used.

### What
This PR generates a summary value for batteries 1 and 2 and sends it to Fronius. This is then displayed correctly on the website under the real-time values.

### Why
It's fair to assume that AI uses it to determine charge/discharge times. If it wasn't interesting, they wouldn't display it, I think, like in older versions.

### How
Adding the 2 capacity values, if the 2nd battery exists, up to a maximum value of 60kWh.


pmajor021:
Before:
image
image

After:
image

PR #1656, opened 2025-10-17
Experimental: Add used-capacity nudging for RJXZS to correct SoC drift.

jonny5532:
### What
This PR adds a 60-second routine to the RXJZS battery integration that compares the reported SoC from a value derived from the cell voltage midpoint. If the discrepancy is more than 1Ah-equivalent of charge, it nudges the 'reported usage capacity' to correct for the discrepancy.

It also adds '07' messages to instruct the BMS to open contactors if BE wants to.

Note: untested on real hardware, use at your own risk and monitor carefully!

Potential issues:
- It tries to correct the entire discrepancy in one jump, it may be prefereable to have it make smaller corrections (however the minimum resolution is 1Ah).
- There is currently no smoothing/slow-start, it may issue under/overcorrections if the cell voltage values are noisy (eg, due to large charge/discharge currents).

### DIFF
--- a/Software/src/battery/RJXZS-BMS.cpp
+++ b/Software/src/battery/RJXZS-BMS.cpp
@@ -7,31 +7,45 @@
 // https://www.topbms.com/static/upload/file/20230804/1691122017124559.pdf
 // https://github.com/puffnfresh/Battery-Emulator/commit/7b0f817ef138b22902c27c5cff3fbe2df4ba341b
 
-const uint16_t voltage[101] = {4200, 4173, 4148, 4124, 4102, 4080, 4060, 4041, 4023, 4007, 3993, 3980, 3969, 3959, 3953,
-                               3950, 3941, 3932, 3924, 3915, 3907, 3898, 3890, 3881, 3872, 3864, 3855, 3847, 3838, 3830,
-                               3821, 3812, 3804, 3795, 3787, 3778, 3770, 3761, 3752, 3744, 3735, 3727, 3718, 3710, 3701,
-                               3692, 3684, 3675, 3667, 3658, 3650, 3641, 3632, 3624, 3615, 3607, 3598, 3590, 3581, 3572,
-                               3564, 3555, 3547, 3538, 3530, 3521, 3512, 3504, 3495, 3487, 3478, 3470, 3461, 3452, 3444,
-                               3435, 3427, 3418, 3410, 3401, 3392, 3384, 3375, 3367, 3358, 3350, 3338, 3325, 3313, 3299,
-                               3285, 3271, 3255, 3239, 3221, 3202, 3180, 3156, 3127, 3090, 3000};
+const uint16_t voltage_NMC[101] = {
+    4200, 4173, 4148, 4124, 4102, 4080, 4060, 4041, 4023, 4007, 3993, 3980, 3969, 3959, 3953, 3950, 3941,
+    3932, 3924, 3915, 3907, 3898, 3890, 3881, 3872, 3864, 3855, 3847, 3838, 3830, 3821, 3812, 3804, 3795,
+    3787, 3778, 3770, 3761, 3752, 3744, 3735, 3727, 3718, 3710, 3701, 3692, 3684, 3675, 3667, 3658, 3650,
+    3641, 3632, 3624, 3615, 3607, 3598, 3590, 3581, 3572, 3564, 3555, 3547, 3538, 3530, 3521, 3512, 3504,
+    3495, 3487, 3478, 3470, 3461, 3452, 3444, 3435, 3427, 3418, 3410, 3401, 3392, 3384, 3375, 3367, 3358,
+    3350, 3338, 3325, 3313, 3299, 3285, 3271, 3255, 3239, 3221, 3202, 3180, 3156, 3127, 3090, 3000};
 
 static uint16_t cell_voltage_to_soc(uint16_t v) {
   // Returns SOC in hundredths of a percent (0-10000)
 
-  if (v >= voltage[0]) {
+  if (v >= voltage_NMC[0]) {
     return 10000;
   }
-  if (v <= voltage[100]) {
+  if (v <= voltage_NMC[100]) {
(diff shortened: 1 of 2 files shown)


Software/src/battery/RJXZS-BMS.cpp
```diff
@@ -4,6 +4,36 @@
 #include "../datalayer/datalayer.h"
 #include "../devboard/utils/events.h"
 
+// https://www.topbms.com/static/upload/file/20230804/1691122017124559.pdf
+// https://github.com/puffnfresh/Battery-Emulator/commit/7b0f817ef138b22902c27c5cff3fbe2df4ba341b
+
+const uint16_t voltage[101] = {4200, 4173, 4148, 4124, 4102, 4080, 4060, 4041, 4023, 4007, 3993, 3980, 3969, 3959, 3953,
```
puffnfresh: Obviously this is hard-coded to a standard Li-Ion voltage range.

I personally prefer to limit my cells to 4V because research shows this can easily double the working number of cycles. I tell RJXZS that I have fewer Ah because of that, calculated from the SoC table in the datasheet of my cells. Something which makes this tricky

Even ignoring preferences, we can't ship with a hardcoded voltage table like this - RJXZS supports Li-Ion, LiFePO4 and LTO.

I don't really want a UI where you input 100 voltages to create a SoC curve, but that's kinda what we need.

jonny5532: Good point, I hadn't considered that RJXZS supported other chemistries. Adding tables for at least LFP and maybe NCA (pretty close to NMC) could correspond with the options in the settings dropdown. I don't know if anyone is using LTO cells with RJXZS currently?

Re: termination voltage, two options would be:
- Use BE's SoC scaling, although this gives you less control over the exact voltages (and means the SoC-drift-compensation needs to be good, or the termination voltages will be wrong)
- Truncate the voltage tables using the supplied min/max cell voltages in settings, and scale the SoC that way - this would probably be better, and shouldn't need anyone entering all the individual voltage steps.

I'm a bit concerned that the crude SoC estimation won't be good enough (with no account for temperature or current draw) - although the solution might be to have it make 1Ah step changes very infrequently (just enough to keep up with the drift) so it doesn't try to overcorrect...

The admin setting could be 'Maximum drift correction per day (Ah)', with zero being the default to disable it.

jonny5532: I've tweaked slightly:

- Now applies the settings-set cell voltage limits as a pre-scaler to the derived SoC (independent of BE's own SoC scaling). Currently the curve voltages only go down to 3V, maybe people like to run them lower, I dunno.
- Makes at most 1Ah correction every minute. This probably needs reducing a lot further - but it makes testing rather slow so I've left it like it is currently.

Software/src/battery/RJXZS-BMS.cpp
```diff
@@ -4,6 +4,50 @@
...
+    if (v >= voltage_NMC[i]) {
+      uint16_t t = (100 * (v - voltage_NMC[i])) / (voltage_NMC[i - 1] - voltage_NMC[i]);
+      return 10000 - (i * 100) + t;
+    }
+  }
+  return 0;
+}
+
+uint16_t RjxzsBms::cell_voltage_to_soc_scaled(uint16_t v) {
+  // Scales the SOC according to user set min and max cell voltage limits.
```
puffnfresh, line 38: Yeah, I like this, good idea :+1:

Software/src/battery/RJXZS-BMS.cpp
```diff
@@ -4,6 +4,50 @@
...
+// https://github.com/puffnfresh/Battery-Emulator/commit/7b0f817ef138b22902c27c5cff3fbe2df4ba341b
+
+const uint16_t voltage_NMC[101] = {
+    4200, 4173, 4148, 4124, 4102, 4080, 4060, 4041, 4023, 4007, 3993, 3980, 3969, 3959, 3953, 3950, 3941,
+    3932, 3924, 3915, 3907, 3898, 3890, 3881, 3872, 3864, 3855, 3847, 3838, 3830, 3821, 3812, 3804, 3795,
+    3787, 3778, 3770, 3761, 3752, 3744, 3735, 3727, 3718, 3710, 3701, 3692, 3684, 3675, 3667, 3658, 3650,
+    3641, 3632, 3624, 3615, 3607, 3598, 3590, 3581, 3572, 3564, 3555, 3547, 3538, 3530, 3521, 3512, 3504,
+    3495, 3487, 3478, 3470, 3461, 3452, 3444, 3435, 3427, 3418, 3410, 3401, 3392, 3384, 3375, 3367, 3358,
+    3350, 3338, 3325, 3313, 3299, 3285, 3271, 3255, 3239, 3221, 3202, 3180, 3156, 3127, 3090, 3000};
+
```
puffnfresh, line 17:
```suggestion

+static const uint16_t voltage_LFP[101] = {
    3650, 3645, 3640, 3635, 3630, 3625, 3620, 3615, 3610, 3605, 3600, 3595, 3590, 3585, 3580, 3575, 3570,
    3565, 3560, 3555, 3550, 3545, 3540, 3535, 3530, 3525, 3520, 3515, 3510, 3505, 3500, 3495, 3490, 3485,
    3480, 3475, 3470, 3465, 3460, 3455, 3450, 3445, 3440, 3435, 3430, 3425, 3420, 3415, 3410, 3405, 3400,
    3395, 3390, 3385, 3380, 3375, 3370, 3365, 3360, 3355, 3350, 3345, 3340, 3335, 3330, 3325, 3320, 3315,
    3310, 3305, 3300, 3295, 3290, 3285, 3280, 3275, 3270, 3265, 3260, 3255, 3250, 3240, 3230, 3220, 3210,
    3200, 3190, 3180, 3170, 3160, 3150, 3135, 3120, 3100, 3070, 3040, 3000, 2960, 2900, 2800, 2500
+};

```

jonny5532:
I think this probably needs a different approach (EKF etc), closing...

PR #1658, opened 2025-10-21
Library: :scroll: Update ESP32 version to v3.3.2

dalathegreat:
### What
This PR updates the ESP32 dependency from v3.3.1 -> 3.3.2

### Why
Bugfixes are nice, but the space savings are nicer!

### How
This PR reduces flash usage by 0.4% :raised_hands: 

> [!TIP]

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -63,7 +63,6 @@ build_flags =
     -Wimplicit-fallthrough
     -Wextra
     -Wall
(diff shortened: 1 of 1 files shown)


dalathegreat:
Merging this, as OTA seems way more stable with 3.3.2 compared to 3.3.1 that we recently swapped to!

PR #1668, opened 2025-10-24
Allow using the inverter protocol as a shunt

M4GNV5:
### What
This PR implements using voltage and current values provided by inverters as shunt values

### Why
Some inverters provide voltage and current measurements. This can be used for battery protocols like chademo which require a shunt. 
This removes the requirement of buying an expensive external shunt.

### How
Added a new shunt type "inverter", added two new functions `provides_shunt` and `enable_shunt` for inverters.
Currently only implemented for BYD CAN, but could probably implemented for others as well (namely SMA BYD CAN, ...))

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -558,7 +558,7 @@ void setup() {
   setup_charger();
   setup_inverter();
   setup_battery();
-  setup_can_shunt();
+  setup_shunt();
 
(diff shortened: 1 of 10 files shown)


dalathegreat (approved):
Oh this is excellent! Thanks @M4GNV5 for implementing this! :raised_hands: 

I know that this should be possible to port to some other inverter protocols also, but let's do that later!

PR #1679, opened 2025-10-26
Solax: Double battery fix; make solax ignore_contactors option wait for the first close request and then stay closed

frenetic1:
### What
This PR changes how the ignore_contactors option for solax inverters work.

### Why
On a double battery setup the current implementation doesn't work. The inverter checks the voltage on the 2nd battery after it requests the contactors to close. If contactors of both batteries are closed from the beginning the inverter refuses to go to normal operation mode.

### How
Wait until the inverter requests contactor closing instead of closing them on bootup. The contactors will then stay closed. This way the contactors of the 2nd battery will get closed when the inverter requests it.

It's not a perfect fix but it works at least for tesla batteries where we cannot switch contactors as we like.

### DIFF
--- a/Software/src/battery/TESLA-BATTERY.h
+++ b/Software/src/battery/TESLA-BATTERY.h
@@ -84,7 +84,7 @@ class TeslaBattery : public CanBattery {
   static const uint8_t CAR_DRIVE = 1;
   static const uint8_t ACCESSORY = 2;
   static const uint8_t GOING_DOWN = 3;
-  uint8_t powerDownSeconds = 0;  // Car power down (i.e. contactor open) tracking timer, 3 seconds per sendingState
+  uint8_t powerDownSeconds = 3;  // Car power down (i.e. contactor open) tracking timer, 3 seconds per sendingState
   //0x2E1 VCFRONT_status, 6 mux tracker
   uint8_t muxNumber_TESLA_2E1 = 0;
   //0x334 UI


greenoem:
Rather than powerDownSeconds = 0 have you tried testing at least powerDownSeconds = 1 ? Setting it to 0 when I was testing adding the contactor control/shutdown sequence initially meant that some CAN frames were missed (with high system load, which double battery almost certainly would have), so the pack went into odd states.

frenetic1:
Didn't try it but i think everything <= 3 will work. My intention was making this if not reset to any higher state than OFF:
https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/src/battery/TESLA-BATTERY.cpp#L850

Should I set it to 3?

greenoem:
On my latest test builds I run it set to 3 and works quickly/fine - when set to 0 I definitely had issues with the pack not correctly seeing the change of states, but equally haven’t tried it for a while. May be worth trying it set to 3 and check the Solax is happy with the second or two delay?

frenetic1:
I'm pretty sure that solax is happy with 3. It has a very generous timeout when it waits for contactor closing and will also restart the whole turning on procedure if something fails. I'll update it to 3 in this PR.

dalathegreat:
Nice work, shall we merge this to get more broad feedback from Tesla&Solax users?

frenetic1:
I was hoping that drlumsden will do some tests as he is facing this problem currently:
https://github.com/dalathegreat/Battery-Emulator/discussions/2154

But I think we can merge it for more feedback also from single bat users.

jonny5532:
As I mentioned on Discord, this breaks battery integrations that always close contactors on startup and leave them closed. the MG HS breaks, although I can now fix this since the UDS reset gives us full contactor control (will PR).

However for some where contactors can't be reliably toggled (one of the Kias?) this results in the following:

- Battery contactors close on boot
- Solax goes idle (eg, zero power demand)
- Power demand appears, Solax tries to start up
- Solax goes into Announce state, waits for contactors to open
- Contactors never open

(this was the reason for the state-machine bypass that this PR removes).

jonny5532:
Also, have you tested what happens with Tesla, if you let the Solax go idle (ie, zero charge/discharge demand), and then let it resume?

dalathegreat:
Sounds like we need to revert this PR then!

jonny5532:
I think there are a couple of options to fix:
- Stop the Solax state machine going from contactor-closed to announce if the avoid checkbox is ticked
- Make it check the battery_allows_contactor_closing flag to determine whether the battery is cooperating

but I haven't figured out which will work best...

frenetic1:
May solax never idles so that's a test case i haven't done. I think the first option is the safest and works for both cases. I'll make another pr.
EDIT: https://github.com/dalathegreat/Battery-Emulator/pull/2192/changes

Is there somebody who can easily test idle mode with a battery?

PR #1686, opened 2025-10-28
BMW IX - Handle Transport/energy saving mode

wjcloudy:
### What
This PR implements support for reading and resetting energy saving mode (or transport mode)

### Why
We've seen some battery go into this mode

### How
Uses UDS to read + set normal mode on request

> [!TIP]


wjcloudy:
Hi @dalathegreat  we just confirmed this can detect and remove transport mode on an issue pack - so good to merge if review passes

dalathegreat (approved):

PR #1687, opened 2025-10-28
Volvo SPA: Use dynamic pack voltage limits

dalathegreat:
### What
This PR changes the Volvo integration to use dynamic pack voltage limits 

### Why
Users reported not possible to charge batteries fully

### How
Instead of using hardcoded pack voltage limits, we now use the Max / Min voltages that the BMS sends. This will allow batteries to store more energy

Bonus: This PR also reduces usage of floats for a performance speedup

> [!TIP]


dalathegreat:
Closing, not tested enough

PR #1699, opened 2025-11-03
Bugfix: Make interconnect check more robust

### LINKED ISSUE #1694: [BUG] Too large voltage diff between the batteries error after OTA upgrade
**Describe the bug**
"Too large voltage diff between the batteries. Second battery cannot join the DC-link" event with the data 39  after an OTA upgade from 9.2.dev to 9.2.1

Turning the WUP2 off and on immediately restored the battery.

**To Reproduce**
OTA upgrade

**Expected behavior**
it is expected that the 2 batteries will work together, as they were only idle during the OTA upgrade, so it is unlikely that the difference would be greater than 3V

**Screenshots**
n/a

- Software version: 9.2.1
- Battery used: BMW i3
- Inverter communication protocol: BYD Modbus
- Hardware used for Battery-Emulator: LilyGo T-2CAN
- MQTT: yes

dalathegreat:
### What
This PR makes the Double-Battery interconnect check more reliable, especially for BMW i3

### Why
Fixes #1694

### How
- We now skip the interconnect check if Battery1 is 370.0V, OR, Battery2 is 0V
- We now skip reading BMW voltage if value unavailable (can be on boot, 0xFFFF)
- BONUS: Slight flash save of 32bytes by smarter packing of BMW struct
- Experimental: We refactor how i3 contactor closing is handled to make it smoother for second battery

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -122,8 +122,8 @@ void logging_loop(void*) {
 }
 
 void check_interconnect_available() {
-  if (datalayer.battery.status.voltage_dV == 0 || datalayer.battery2.status.voltage_dV == 0) {
-    return;  // Both voltage values need to be available to start check
+  if (datalayer.battery.status.voltage_dV == 3700 || datalayer.battery2.status.voltage_dV == 0) {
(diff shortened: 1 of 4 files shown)


dalathegreat:
PR is stale. Closing

PR #1704, opened 2025-11-06
Reduce redundancy in Bolt/Zoe2 cellvoltage handling

jonny5532:
### What
The cellvoltage handling in the Zoe2 and Bolt integrations was very verbose - have replaced with much shorter for-loops.

Saves 5296 bytes.

(ideally someone with those integrations would test this)

### DIFF
--- a/test/can_log_based/can_logs/29_RenaultZoe2_cv88.txt
+++ b/test/can_log_based/can_logs/29_RenaultZoe2_cv88.txt
@@ -1,0 +1,4 @@
+# Update cell voltage 88 via a faked poll response
+# (this is synthetic, please update with a real log!)
+
+(84215.921) RX0 18DAF1DB [8] 00 00 90 7B 0C 33 AA AA
--- a/test/can_log_based/can_logs/4_BoltAmpera_cv88.txt
+++ b/test/can_log_based/can_logs/4_BoltAmpera_cv88.txt
@@ -1,0 +1,3 @@
+# Actually comes out as 3122mV
+# (this is synthetic, please replace with real logs)
+(123.893) RX0 206 [8] 00 00 13 84 00 00 a0 00
--- a/test/can_log_based/canlog_safety_tests.cpp
+++ b/test/can_log_based/canlog_safety_tests.cpp
@@ -148,6 +148,22 @@ class CellUnderVoltageTest : public CanLogTestFixture {
   }
 };
 
+// Check that the parsed logs set a particular cell voltage value (3123mV)
+class CellVoltageTest : public CanLogTestFixture {
+ public:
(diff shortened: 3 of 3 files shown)


Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.cpp
```diff
@@ -449,295 +356,12 @@ void RenaultZoeGen2Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
-        case POLL_CELL_94:
-          datalayer_battery->status.cell_voltages_mV[94] = (rx_frame.data.u8[4] << 8) | rx_frame.data.u8[5];
-          break;
-        case POLL_CELL_95:
-          datalayer_battery->status.cell_voltages_mV[95] = (rx_frame.data.u8[4] << 8) | rx_frame.data.u8[5];
-          break;
-        default:  // Unknown reply
+        default:
+          // Handle cell voltages
+          if (reply_poll >= POLL_CELL_0 && reply_poll <= POLL_CELL_95) {
```
dalathegreat, line 361: Is this really working? There is a break in IDs, the last cell is ID 0x9083 , so when taking - POLL_CELL_0 you will map it to position 98 (even though it should be 96)

jonny5532: Well spotted, I hadn't noticed that! Also the Bolt has a similar discontinuity.

Both are now handled, plus I tested the derivation for every cell index so they should be correct (would still be good to get a real user to test!)

Saving is now 5248 bytes.

Software/src/battery/RENAULT-ZOE-GEN2-BATTERY.cpp
```diff
@@ -449,295 +356,24 @@ void RenaultZoeGen2Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
+            int cell_index = reply_poll - POLL_CELL_0;
+
+            // Three offsets are skipped in the polling sequence, account for that.
+            if (cell_index > POLL_CELL_30) {
```
dalathegreat: Cell index will be a number between 0-96

POLL_CELL_XX will be a hex decimal number between 0x9021 - 0x9083

You cannot compare if cell_index > POLL_CELL_XX , it will never be true

You need to compare reply_poll instead

jonny5532: You're right, typo'd, sorry!

As penance I have added a CAN-log-based test which would have caught this, and another for the Bolt Ampera too.

dalathegreat:
Merging this, I believe this will work great now!

PR #1709, opened 2025-11-07
Safety: Tighten DC-bus requirement to 1.5V

### LINKED ISSUE #1708: [BUG] BMW I3 Paralel   Voltage Difference and Contactor Activation
In our BMW i3 parallel battery project, it has been observed that if the batteries are disconnected and reconnected frequently — about once or twice a week — the battery contactors fail within 1/4 - half a year.

We believe this is caused by switching under load, when there is a voltage difference of up to 3 V between the batteries.

We kindly request that the voltage difference threshold for contactor activation be reduced from 3 V to 1 V.

Thanks.

Havrla

dalathegreat:
### What
This PR tightens the limit for allowing secondary battery to join the DC bus

### Why
Suspected by users that a 3VDC diff can cause arcing if contactors are opened repeatedly. Fixes #1708

### How
We tighten the limit down from 3.0VDC -> 1.5VDC.

3VDC limit originally comes from an OEM that allowed parallel/series connection of two 400VDC batteries. But that system was designed for constant switching, all EV packs might not be as tolerant. Since software 9.2.0 we also OPEN contactors if the batteries start to drift out of spec, for instance from blowing a fuse (https://github.com/dalathegreat/Battery-Emulator/pull/1635)

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -129,7 +129,7 @@ void check_interconnect_available() {
   uint16_t voltage_diff = abs(datalayer.battery.status.voltage_dV - datalayer.battery2.status.voltage_dV);
   uint8_t secondsOutOfVoltageSync = 0;
 
-  if (voltage_diff <= 30) {  // If we are within 3.0V between the batteries
+  if (voltage_diff <= 15) {  // If we are within 1.5V between the batteries
     clear_event(EVENT_VOLTAGE_DIFFERENCE);
     secondsOutOfVoltageSync = 0;
     if (datalayer.battery.status.bms_status == FAULT) {


havrla:
Whether the primary battery is under load or not does not matter. What is important is the voltage difference between the batteries at the moment of connection, so that the current between them is zero within the switching time limit.

dalathegreat:
The accuracy of voltage measurement is not accurate enough on most EV batteries to make the limit <0.5VDC. On the Nissan LEAF; I can observe 1.0 (and even 1.5V under quick load change) on two DC linked Nissan LEAF batteries. So reducing from 3.0V -> 1.5V is a safe change that will reduce arcing, and not cause issues with accidental opening due to thinking voltage differs too much

dalathegreat:
Merging this now to get more feedback on solution!

PR #1715, opened 2025-11-08
Allow the precharge to go from COMPLETED->IDLE if the BMS goes standby

jonny5532:
### What
The MEB integration doesn't re-precharge after a BMS reset. Hopefully this will cause the precharge state machine to allow a re-precharge if it detects the BMS has gone back to standby.

Needs testing on a real battery before merging (and someone to check that my logic is correct!)

Note: includes PR https://github.com/dalathegreat/Battery-Emulator/pull/1719 which would need merging first


uitgeslapen:
Tested, RESET via MQTT works now!

uitgeslapen:
Tested, fully works now

dalathegreat (approved):
Nice work!

PR #1717, opened 2025-11-09
Improvement: Optimize flash usage for Modbus library

dalathegreat:
### What
This PR reduces the flash usage by the Modbus library

### Why
We need to free up flash memory to add more features

### How
By removing the debug printouts (that we were not relying on in any way), we save -1572 bytes of flash

> [!TIP]


dalathegreat:
Tested OK on LEAF setup at my workshop!

PR #1719, opened 2025-11-09
Make BMS-reset delay between pause and contactor-open if needed

jonny5532:
### What
During a BMS reset, with BE-controlled contactors, it is fine if the BMS is powered down at the same instant as the inverter is paused.

However with BMS-powered contactors, this doesn't give long enough for the inverter to stop drawing current before the BMS (and the contactors) drop out, so arcing will occur.

This adds a proper state machine for the BMS reset process, and adds a new 'waiting for pause' state for BMS-controlled contactors, where it waits for the current to go to zero, or low after 5 seconds, before turning off the BMS.

If the current never drops low enough (threshold is 1A), it will abort the reset, to avoid damaging the contactors, and raises an error.

Have tested locally but it would be got to get feedback from others before merging.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -69,6 +69,7 @@ add_compile_definitions(ESP32 HW_LILYGO COMMON_IMAGE)
 add_executable(tests 
     tests.cpp 
     safety_tests.cpp 
+    bms_reset_tests.cpp
     battery/NissanLeafTest.cpp 
     battery/still_alive_tests.cpp
     can_log_based/canlog_safety_tests.cpp
--- a/test/bms_reset_tests.cpp
+++ b/test/bms_reset_tests.cpp
@@ -1,0 +1,157 @@
+#include <gtest/gtest.h>
+
+#include "../Software/src/communication/contactorcontrol/comm_contactorcontrol.h"
+#include "../Software/src/datalayer/datalayer.h"
+#include "../Software/src/devboard/safety/safety.h"
+#include "../Software/src/devboard/utils/events.h"
+
+#include "Arduino.h"
+
+const unsigned long bmsWarmupDuration = 3000;
+
+// Test a BMS reqest sequence from end to end. This is for the case where the
+// contactors are powered directly by BE, so the reset doesn't need to wait for
+// zero current before cutting the BMS power.
+TEST(BmsResetTests, BmsResetSequenceDirectSuccess) {
+  set_millis64(0xffffffffffffffff - 10);  // Test overflow handling
+  remote_bms_reset = true;
+  contactor_control_enabled = true;
+  datalayer.battery.settings.user_set_bms_reset_duration_ms = 30000;  // 30 seconds
+
+  for (int i = 0; i < 10; i++)
+    handle_BMSpower();
+
+  // Nothing has happened yet, should still be idle.
+  EXPECT_EQ(datalayer.system.status.bms_reset_status, BMS_RESET_IDLE);
+  start_bms_reset();
+
+  for (int i = 0; i < 10; i++)
+    handle_BMSpower();
+
+  // BMS should have gone straight to powered-off
+  EXPECT_EQ(datalayer.system.status.bms_reset_status, BMS_RESET_POWERED_OFF);
+  EXPECT_EQ(emulator_pause_request_ON, true);
+
+  for (int i = 0; i < 10; i++)
+    handle_BMSpower();
+
+  // Should still be powered off
+  EXPECT_EQ(datalayer.system.status.bms_reset_status, BMS_RESET_POWERED_OFF);
+
+  set_millis64(20000);  // Simulate 20 seconds passing
+
+  for (int i = 0; i < 10; i++)
+    handle_BMSpower();
+
+  // Should still be powered off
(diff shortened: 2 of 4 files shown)


Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -260,57 +260,100 @@ This makes the BMS recalculate all SOC% and avoid memory leaks
...
+      lastPowerRemovalTime = millis();
 
-      // Set emulator state to paused (Max Charge/Discharge = 0 & CAN = stop)
-      // We try to keep contactors engaged during this pause, and just ramp power down to 0.
+      // Issue a pause, which should stop charge/discharge whilst the reset is ongoing
       setBatteryPause(true, false, false, false);
 
-      digitalWrite(bms_power_pin, LOW);  // Remove power by setting the BMS power pin to LOW
+      if (contactor_control_enabled) {
+        // We control contactors, so we can avoid closing/opening them during reset.
```
dalathegreat:
```suggestion
        // We control contactors via GPIO, so we can avoid closing/opening them during reset process, and avoid welding
```

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -260,57 +260,100 @@ This makes the BMS recalculate all SOC% and avoid memory leaks
...
+
+        // Thus we can cut the BMS power now
+        bms_power_off();
+
+        // and jump straight to powered off state, no need to wait.
+        datalayer.system.status.bms_reset_status = BMS_RESET_POWERED_OFF;
+      } else {
+        // We do not control contactors, so we need to wait for the pause to
```
dalathegreat:
```suggestion
        // We do not have complete control over contactors (CAN control), so we need to wait for the pause to
```

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -260,57 +260,100 @@ This makes the BMS recalculate all SOC% and avoid memory leaks
...
-      set_event(EVENT_PERIODIC_BMS_RESET, 0);
-      clear_event(EVENT_PERIODIC_BMS_RESET);
+        bms_power_off();
+        lastPowerRemovalTime = currentTime;
+        datalayer.system.status.bms_reset_status = BMS_RESET_POWERED_OFF;
+      } else if (currentTime - lastPowerRemovalTime >= 10000) {
+        // There's still current, and we don't want to weld the contactors, so give up.
+
+        logging.printf("BMS reset: Unable to pause battery, cancelling reset.\n");
+
```
dalathegreat, line 302: Should we add an event here?

Periodic reset failed, inverter was still using battery so we are unable to open contactors

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -260,57 +260,100 @@ This makes the BMS recalculate all SOC% and avoid memory leaks
...
-    if (datalayer.system.status.BMS_startup_in_progress && currentTime - bmsPowerOnTime >= bmsWarmupDuration) {
+      int16_t battery_current_dA = datalayer.battery.status.current_dA;
+      int16_t battery2_current_dA = datalayer.battery2.status.current_dA;  // Should be 0 if no battery2
 
-      setBatteryPause(false, false, false, false);
+      if (
+          // No current, safe to cut power
+          (battery_current_dA == 0 && battery2_current_dA == 0)
+          // or reasonably low current and 5 seconds has passed
+          || (abs(battery_current_dA) < 10 && abs(battery2_current_dA) < 10 &&
```
dalathegreat, line 292: Optional: Could be simpler to use Power value instead of current, no need to worry about ABS()

jonny5532: Looking at it, it seems the power can be negative too, so would still need an abs? I figured current is more direct (as the power is calculated from the current). Although maybe there's an argument that power is more relevant to the risk of contactor arcing, I'm not really sure.

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
jonny5532: Have added an event (I didn't hijack the Tesla one as that seems to have a slightly different purpose).

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
dalathegreat: Oh you are right, forget this :)

Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
jonny5532: Have rejigged the comments (went with 'powered' since that's the issue rather than the control - if you cut BMS power the the contactors drop out too, even if you have direct CAN control over them...)

jonny5532:
Have also added some tests which run through the sequence. I do note that there's some overlap in functionality with the pause mechanism - which also has a similar PAUSING->PAUSED transition which checks that the inverter current has fallen (below 1.8A rather than 1A though). However the existing pause mechanism doesn't time out, and also doesn't check the battery2 current, so probably best to use a separate implementation for this which does.

PR #1723, opened 2025-11-10
Improvement: Make Kia64FD use user set charge/discharge values

dalathegreat:
### What
This PR makes Kia64FD integration use user set charge/discharge values

### Why
User requested feature, to be able to use more than hardcoded 10kW

### How
Same way Kia EGMP platform does it.

> [!TIP]


gednet:
Working in my setup very well. THX.

PR #1742, opened 2025-11-15
RJXZS: Add balancing indicator to cellmonitor page 

dalathegreat:
### What
This PR implements an indicator to RJXZS, so user can see when battery is balancing on the cellmonitor page

### Why
User requested feature on Discord

### How
Note, RJXZS does not send which cells are balancing, so we paint all as balancing when active

> [!TIP]

### DIFF
--- a/Software/src/devboard/webserver/cellmonitor_html.cpp
+++ b/Software/src/devboard/webserver/cellmonitor_html.cpp
@@ -6,7 +6,7 @@
 String cellmonitor_processor(const String& var) {
   if (var == "X") {
     String content = "";
-    // Page format
+    // Page formatH
     content += "<style>";
     content += "body { background-color: black; color: white; }";
     content +=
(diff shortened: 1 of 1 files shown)


puffnfresh (approved):

wjcloudy:
I also just raised a PR to support an overall balance status for monitoring. It allows for ready/error/active states instead of just boolean so wouldn't require us to flag all cells 1 or zero when we don't have per cell monitoring .

puffnfresh:
Works great:

image
image

dalathegreat:
@puffnfresh , I have updated the PR to use the full pack balancing status. Can you please re-test this? :)

puffnfresh:
image

dalathegreat:
Thanks for the feedback on the Discord @puffnfresh ,

Remade it so that the UI shows it more clearly that the battery is balancing:

image

PR #1745, opened 2025-11-15
Sign-extend Volvo SPA temps so negative is handled

jonny5532:
### What
Sign-extend Volvo SPA battery negative temps up to a 16 bit int so that negative values are interpreted correctly.

This assumes they are 13-bit twos-complement integers (seems likely?)

Also maybe it is worth having a library of utils like the `sign_extend_int16` function that can be used across BE?


dalathegreat (approved):
Nice work Jonny! :raised_hands: 

Yeah having a common library of functions used by many integrations would be nice, reduced codesize and improves maintainability (CRC functions, helpers etc.)

PR #1755, opened 2025-11-19
Stellantis CMP: Fixed an issue where power would always be positive

dalathegreat:
### What
This PR improves current handling on the CMP smart platform integration

### Why
The CMP Smart platform integration would report batteries as always discharging, even though they were charging

### How
We now switch to using the sign_extend_to_int16() function, same as Volvo SPA is using. This PR makes it a common function

> [!TIP]


dalathegreat:
Tested OK by Hackun on the Discord!

PR #1759, opened 2025-11-22
BMW-iX: Improve contactor control and add cold temperature charge limiter

wjcloudy:
### What
Major changes:
- Fixed contactor control to only send CAN commands when GPIO control is disabled
- Implemented startup sequence: 2s delay → BMS reset → enable UDS communication
- Added temperature-based charge power limiting (ramps to 0W from 5°C to 0°C)
- UDS requests now work with or without inverter
- Pause UDS requests during contactor closing sequence to prevent interference

### Why
Issues with inverters that didn't request contactors

### How
Changes to contactor close logic and charge limiter

> [!TIP]

### DIFF
--- a/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
+++ b/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
@@ -204,6 +204,7 @@ void handle_contactors() {
         dbg_contactors("NEGATIVE");
         prechargeStartTime = currentTime;
         contactorStatus = PRECHARGE;
+        datalayer.system.status.contactors_engaged = 3;
         break;
 
       case PRECHARGE:
@@ -212,6 +213,7 @@ void handle_contactors() {
           dbg_contactors("PRECHARGE");
           negativeStartTime = currentTime;
           contactorStatus = POSITIVE;
+          datalayer.system.status.contactors_engaged = 3;
         }
         break;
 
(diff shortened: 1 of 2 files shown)


dalathegreat:
Should we really add our own temperature ramp down to 0W at freezing temps? Most BMSes send a reduced charge/discharge limit when the temps go down, I am for instance using my NMC LEAF batteries still at -20°C (they just reduce kW heavily)

If this is really needed and iX does not derate automatically, and we need this temperature derating, maybe it would be best to put it in the SAFETY.cpp file, and allow other integrations to also access it?

wjcloudy:
> Should we really add our own temperature ramp down to 0W at freezing temps? Most BMSes send a reduced charge/discharge limit when the temps go down, I am for instance using my NMC LEAF batteries still at -20°C (they just reduce kW heavily)
> 
> If this is really needed and iX does not derate automatically, and we need this temperature derating, maybe it would be best to put it in the SAFETY.cpp file, and allow other integrations to also access it?

IX is not derating (remains at 247A even sub zero) Might have to find another derating value to catch, but this provides for in the meantime. Happy to add something in safety also (ideally configurable in web interface for min and max temp derates.
Won't hurt to have this in the IX code still (as most BMS will produce this ramp themselves).   I've changed the derating to be a ramp from 10kW limit at 5c, down to no charge at -10c. That should keep firmly out of the degradation state.

Also added fixes for PRECHARGE display on the web interface.

dalathegreat (approved):

PR #1771, opened 2025-11-26
Improvement: I2C display tidying and configurable port setting

jonny5532:
### What
This tidies the I2C display code a bit (and now includes the IP address/RSSI).

It also adds a new 'GPIOOPT1' setting on T-2CAN, which allows you to you choose between WUP1/WUP2, ESTOP/BMS_POWER, and the I2C Display on the second 'QWIIC' port on the top of the board. I think it makes sense for only this port to be configurable - the other should always be RS485 (since it sends chatter on boot which will upset I2C and contactors/BMS outputs).

Is this a sensible way to do it?


GoSmartEV:
Untitled

Issue #1764, opened 2025-11-23
[BUG] Cell Voltage not sent over MQTT


mccasian:
**Describe the bug**
I am using a 48cells custom LFP Battery with RJXZS BMS. Everything is technically working fine, but i've noticed in Home Assistant and then later one investigated with MQTT Explorer that the Battery Emulator is not publishing the spec_data containing the cell_voltages. 
However, i've temporary selected the "Fake battery for testing purposes" as Battery and the spec_data immediately popped up in MQTT Explorer, and of course in Home Assistant too.

At the same time, the Cell Monitor menu sees all the cells and voltages without problems, so the cell values are there, they are just not being published on MQTT broker

Of course, i've tried playing with "Send all cellvoltages via MQTT" option by deactivating and re-activating it. Also this issue was present in v9.2.2 too.

**To Reproduce**
Configure the battery to RJXZS BMS, DIY Battery, and also some other models like BYD Atto 3 Battery, BMW i3, etc. -> reboot Battery Emulator ->  Delete the BE Topic from MQTT(i am using MQTT Explorer for simplicity) -> restart the battery emulator again or wait a while to make sure all messages supposed to be published on MQTT are being published.

**Expected behavior**
Since the cell voltages information is already available on Battery Emulator, and the code is there to actually publish this info to MQTT Broker, the cell voltages info should be published to MQTT broker on all battery types, not only on the one dedicated for testing purposes.

**Screenshots**
a few screenshots attached

<img width="659" height="572" alt="Image" src="https://github.com/user-attachments/assets/989b6436-3c59-44c0-b470-662d67537199" />

<img width="593" height="856" alt="Image" src="https://github.com/user-attachments/assets/7e8d6ede-0520-47e7-91c3-5d785f622bcb" />

[battery_emulator_log.txt](https://github.com/user-attachments/files/23698630/battery_emulator_log.txt)

**Version and settings (please complete the following information):**
- Please attach screenshot of configuration for easier debugging
- Software version: 9.2.3
- Battery used: RJXZS BMS, DIY Battery
- Inverter communication protocol: BYD Battery-Box Premium HVS over CAN Bus
- Hardware used for Battery-Emulator:LilyGo T-2CAN
- GPIO controlled contactors: no
- MQTT: yes

havrla:
with BMW I3 :

balancing_data
Parameter 	Value
[...]

PR #1778, opened 2025-11-29
fix RJXZS cell count population

mccasian:
**Problem**
Cell voltages were not being published via MQTT for RJXZS BMS batteries despite the option being enabled. The root cause was that [datalayer.battery.info.number_of_cells](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) was being set incorrectly, causing the MQTT publishing check to always fail.

**Root Cause**
The RJXZS BMS sends cell voltage data via CAN messages with mux IDs 0x07-0x46, with each message containing 3 cell voltages. The BMS continues sending messages in this range even after all actual cells have been reported, filling the remaining messages with zeros.

The original code incremented [populated_cellvoltages](https://github.com/dalathegreat/Battery-Emulator/blob/3b81d3384ef9e0a74b2573d2af6293a2272ba625/Software/src/battery/RJXZS-BMS.cpp#L134) for every message received in the 0x07-0x46 range, regardless of whether the message contained actual cell data or just zeros. This resulted in an incorrect cell count being reported.

For example, a 48-cell battery (16 modules × 3 cells) would:

- Send valid data in messages 0x07-0x16 (cells 1-48)
- Send all zeros in messages 0x17-0x46 (would-be cells 49-192)
- Incorrectly count cells beyond the actual 48 present

**Solution**
Modified the cell voltage parsing logic in [RJXZS-BMS.cpp](https://github.com/dalathegreat/Battery-Emulator/blob/3b81d3384ef9e0a74b2573d2af6293a2272ba625/Software/src/battery/RJXZS-BMS.cpp#L137 to only update populated_cellvoltages](https://github.com/mccasian/Battery-Emulator/blob/124084a7cae487de5dd9951f7fd19263e25470e2/Software/src/battery/RJXZS-BMS.cpp#L145C12-L145C14) when the CAN message contains at least one non-zero cell voltage value. This ensures number_of_cells accurately reflects the actual number of cells in the battery pack.

Changes
Added validation to check if received cell voltage data is non-zero before updating the populated cell count
Prevents counting phantom cells from zero-filled CAN messages
Ensures `datalayer.battery.info.number_of_cells` is set correctly for MQTT and other consumers
Testing
Validated against real CAN bus logs showing a 48-cell RJXZS BMS battery configuration.

### DIFF
--- a/Software/src/battery/RJXZS-BMS.cpp
+++ b/Software/src/battery/RJXZS-BMS.cpp
@@ -134,13 +134,20 @@ void RjxzsBms::handle_incoming_can_frame(CAN_frame rx_frame) {
       } else if (mux >= 0x07 && mux <= 0x46) {
         // Cell voltages 1-192 (3 per message, 0x07=1-3, 0x08=4-6, ..., 0x46=190-192)
         int cell_index = (mux - 0x07) * 3;
+        bool has_valid_data = false;
         for (int i = 0; i < 3; i++) {
           if (cell_index + i >= MAX_AMOUNT_CELLS) {
             break;
           }
-          cellvoltages[cell_index + i] = (rx_frame.data.u8[1 + i * 2] << 8) | rx_frame.data.u8[2 + i * 2];
+          uint16_t cell_voltage = (rx_frame.data.u8[1 + i * 2] << 8) | rx_frame.data.u8[2 + i * 2];
+          cellvoltages[cell_index + i] = cell_voltage;
+          // Check if this cell has valid (non-zero) voltage data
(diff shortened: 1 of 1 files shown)


mccasian:
this fixes https://github.com/dalathegreat/Battery-Emulator/issues/1764

mccasian:
image

dalathegreat:
Massive thanks for fixing this bug @mccasian :raised_hands:

PR #1793, opened 2025-12-01
Use High Precision Current Value when avilable

rjsc:
### What
This PR sends the High precision current value when available to the datalayer instead of the coarse value in 1A steps

### Why
Inverts such as Solax do not measure the battery port current internally and relay on the battery reported value.
With such coarse power and current values, the inverter miscalculates the house load value (sum of battery + solar power)

### DIFF
--- a/Software/src/battery/ECMP-BATTERY.h
+++ b/Software/src/battery/ECMP-BATTERY.h
@@ -423,7 +423,6 @@ class EcmpBattery : public CanBattery {
   int16_t battery_lowestTemperature = 0;
   int16_t HV_BATT_COP_CURRENT = 0;
 
-  uint8_t HighPrecisionCurrentSampling = 1;
   uint8_t ContactorResetStatemachine = 0;
   uint8_t CollisionResetStatemachine = 0;
   uint8_t IsolationResetStatemachine = 0;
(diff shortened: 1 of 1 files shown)


Software/src/battery/ECMP-BATTERY.h
```diff
@@ -423,6 +423,7 @@ class EcmpBattery : public CanBattery {
   int16_t battery_lowestTemperature = 0;
   int16_t HV_BATT_COP_CURRENT = 0;
 
+  uint8_t HighPrecisionCurrentSampling = 1;
```
dalathegreat: Could be made a bool with true/false

Software/src/battery/ECMP-BATTERY.cpp
```diff
@@ -38,7 +38,12 @@ void EcmpBattery::update_values() {
 
     datalayer.battery.status.voltage_dV = battery_voltage * 10;
 
-    datalayer.battery.status.current_dA = -(battery_current * 10);
+    // If High Precision Curent is avilable, use it
+    if (pid_current != NOT_SAMPLED_YET && datalayer.battery.status.bms_status != FAULT) {
```
dalathegreat, line 42: Think we can remove the if not in fault check, that wont really add any value?

rjsc: It was to avoid sending wrong values to the inverter if we get a fault after reading some values.
In the current code, all polling is suspended in fault state and the last sampled value remains in the variable.

Software/src/battery/ECMP-BATTERY.h
rjsc: Nice point. Corrected.

PR #1805, opened 2025-12-04
Fronius: Safety :warning:  Force SOC% to never go below 1% to avoid overdischarge

dalathegreat:
### What
This PR makes it so that SOC% can never be 0% for BYD Modbus

### Why
Fronius has changed something recently when it comes to SOC% and charging. Previously it allowed forced charging between 0-5% to avoid overdischarged batteries. But now this only seems to work between 1-5%, and if the battery reports 0% it stops the attempt to force charge

### How
We fix this by never reporting below 1% SOC for Fronius Modbus protocol

> [!TIP]


dalathegreat:
Juanjo tested this on the Discord and got a 0% charge to start :+1:

dalathegreat:
Tested OK also on my Fronius. Now a sudden SOC drop to 0% will be reported as 1%, and emergency topping up starts as expected instead of the battery becoming disabled.

image

PR #1807, opened 2025-12-04
BMW i3: Bugfix, update SOC init to 50%

dalathegreat:
### What
This PR inits i3 SOC to 50% instead of 0%

### Why
Secondary battery not joined will cause SOC% to get stuck at 0%
image

### How
We fix this by initing soc to 50%

> [!TIP]


havrla:
not functional
try datalayer.h

/** The "real" SOC reported from the battery, in integer-percent x 100. 9550 = 95.50% */
  uint16_t real_soc = 5000 ;
  /** The SOC reported to the inverter, in integer-percent x 100. 9550 = 95.50%.
   * This value will either be scaled or not scaled depending on the value of
   * battery.settings.soc_scaling_active
   */
  uint16_t reported_soc = 5000;

PR #1544, opened 2025-09-17
WIP: New frontend

jonny5532:
### What
This adds an experimental JS frontend, with a first draft of:

- Dashboard (with some graphical widgets)
- Events
- Cell monitor
- Extended info
- Settings
- Infinite realtime CAN log
- CAN replay (with client-side log parsing and realtime streaming)
- System log
- OTA updating (with new version check)

Interesting features:
- Uses Preact and Typescript, which is concise and hopefully familiar to any React devs
- Tree-shakes and gzips well (~30kb blob currently, which includes some cruft)
- Hand-rolled CSS (to keep it small)
- Experimental extended info page that retrieves datalayer/Battery as a binary blob and parses in JS, to reduce backend size/complexity
- New settings loading/saving topology (preserves but doesn't persist defaults, handles volatile settings)

The JS can be built by installing Bun and doing:
```sh
cd frontend
bun i
bun run build
```

Currently saves 143,796 bytes, there is also an replacement webserver branch (TinyWebServer) which saves another 30kb or so (and also resolves some of the AsyncWebServer issues (excessive allocations and occasional data corruption)), but that can be considered later.

Further ideas:
- Finishing the battery 'actions' API so that battery-specific endpoints can go in battery classes, passing a `JsonDocument` in and out for the frontend to consume
- UDS client mode (which ties in with the now-merged `UdsCanBattery` to allow custom UDS commands to be sent from JS (allowing a nice UI for running routines, etc)).
- Isolating battery-specific behavior more (eg, plugin API for custom panels)

PR #1823, opened 2025-12-11
Improvement: Flash usage reduction by HTML via LittleFS

GrahamWillsher:
### What
This PR implements ...
Reading HTML from LittleFS NVM rather than Flash
### Why
Why does it do it?
To reduce the amount of memory used for the code. Program size after change RAM 25.2%
Flash 99.1%

### How
How does it do it?
Moved HTMl body code to the data folder to be stored in NVM

> [!TIP]
Pleas excuse any errors. This is my first Github PR.

### DIFF
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -1,17 +1,17 @@
 #include "settings_html.h"
 #include <Arduino.h>
+#include <string>
 #include "../../../src/communication/contactorcontrol/comm_contactorcontrol.h"
 #include "../../../src/communication/equipmentstopbutton/comm_equipmentstopbutton.h"
 #include "../../charger/CHARGERS.h"
 #include "../../communication/can/comm_can.h"
 #include "../../communication/nvm/comm_nvm.h"
 #include "../../datalayer/datalayer.h"
+#include "LittleFS.h"
 #include "html_escape.h"
 #include "index_html.h"
 #include "src/battery/BATTERIES.h"
 #include "src/inverter/INVERTERS.h"
-#include "LittleFS.h"
-#include <string>
 
 extern bool settingsUpdated;
 
@@ -21,7 +21,7 @@ String readFileFromLittleFS(const char* filename) {
   if (!file) {
     return String("<!-- Error: Could not open file: ") + filename + " -->";
   }
-  
+
   String content = file.readString();
   file.close();
   return content;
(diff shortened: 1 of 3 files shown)


jonny5532:
That's quite tidy code-wise (LittleFS seems pretty simple to integrate), although this general approach doesn't seem ideal to me (I know others may disagree!)

- Users now need to OTA two different files to do upgrades, given the HTML changes quite often
- The HTML and main firmware will be out of sync during the upgrade process, and potentially afterwards if they only upgrade partially. This could cause problems (and in the worst case, crashes, if something expects something to be there, that isn't).
- It means that the OTA is no longer unbrickable - currently, a bad flash to the second partition just means it'll reboot back onto the first partition if there's a problem. With a third NVM partition involved, a bad flash of that could stop it booting altogether, with no easy way to fix. We'd have to be extremely careful that we never deploy code that might cause BE to crash due to NVM partition corruption.
- The saving here isn't that great (you're getting <1% of the expected 2.1% Flash reduction, presumably due to the extra space taken up by LittleFS itself). I guess adding compression would help, but you'll then pay the cost of the decompression code too, and it'll slow things down.

(I admit I'm horribly biased, my preferred approach is in https://github.com/dalathegreat/Battery-Emulator/pull/1544 - using a gzip-compressed entirely-JS frontend and a much more lightweight webserver. However it's a pretty huge change.)

GrahamWillsher:
Hi @jonny5532,

Fully agree with all of the above :).
It was just to see what memory reduction could be achieved, as any reduction in memory usage might be useful to a code improvement elsewhere.
What lightweight webserver were you thinking of to use as an alternative?
I have some time so might be worth looking at.
The underlying issue is that we only have a finite amount of memory to play with.

Graham.

jonny5532:
The webserver itself is in https://github.com/jonny5532/Battery-Emulator/tree/experiment/new-frontend/Software/src/lib/TinyWebServer - it started off as an experiment that got slightly out of hand, but the idea is:
- zero runtime allocations (AsyncWebServer does quite a lot)
- minimal protocol parsing (it is intended to handle very basic GET/POST with primarily JSON bodies)
- zero dependencies (although is used with ArduinoJson for the JSON serialization)

It still needs a bit more tidying up and testing - I want to get fuzzing working robustly to stress it, as protocol-handling in C is very easy to get wrong. I also want to spend more time making sure the JSON API it exposes is as good as possible, since people will likely want to use it for other purposes.

That branch replaces the BE frontend HTML with a single HTML+JS+CSS blob, which gzips down to ~22kb currently, so, after removing AsyncWebServer and its dependencies, and a lot of the current webpage backends, and the existing OTA, it brings the flash usage down to 90.6%.

The biggest issue is that it totally changes the UI, so we need to make sure everyone is happy with the direction it goes in, and there'll be a lot of finding and fixing the inevitable glitches.

dalathegreat:
Closing this PR for now, still massive thanks for the exploratory work @GrahamWillsher :raised_hands:

GrahamWillsher:
As mentioned, it was just a thought, and it did have its down sides.

I will continue to have a look to see if there are any other things that I can think of that might help.

Regards,

Graham Willsher
07770 792296

PR #1849, opened 2025-12-18
Add support for MG5

j0rgst:
### What
This PR implements support for MG5. It is tested for MG5 52.5kWh battery using stark.

### Why
To use the MG5 battery

### How
Based on MG HS PHEV.
main changes:
1) The hybridCAN does not seem to contain SOC info in the 3AC can messages, so I only use the SOC info from the PTCAN.
2) Sending the 8A can message with byte 5=0x00 does close the contactors, but it also generate  a DTC error 0293:"Lost Communication with Vehicle Control
Unit(VCU)". This prevents the contactors from closing again when when sending 8A with byte  5=0x02. I can close the contactors however by clearing the DTC's after sending byte 5=0x02.
3) changed to amount of cells, max power, limits,...
4) Added support for reading and resetting the DTC's and opening/closing the contactors

> [!TIP]

### DIFF
--- a/Software/src/battery/MG-5-BATTERY.cpp
+++ b/Software/src/battery/MG-5-BATTERY.cpp
@@ -78,34 +78,32 @@ void Mg5Battery::update_soc(uint16_t soc_times_ten) {
     datalayer.battery.status.remaining_capacity_Wh = 0;
   }
 
-
-//#if MG5_USE_FULL_CAPACITY
+  //#if MG5_USE_FULL_CAPACITY
 
   // Calculate the maximum charge power. Taper the charge power between 90% and 100% SoC, as 100% SoC is approached
-    if (RealSoC < StartChargeTaper ) {
-      datalayer.battery.status.max_charge_power_W = MaxChargePower;
-    } else if (RealSoC >= 100) {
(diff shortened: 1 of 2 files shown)


dvbhannes:
Are there any new results?

dalathegreat:
Nice work @j0rgst , merging this to get more testers!

dvbhannes:
Isn't the same battery also used in the MG4?

j0rgst:
> Isn't the same battery also used in the MG4?

No, the MG4 does not work with this code. People are already trying to debug the MG4 battery but it is not yet fully working.

dvbhannes:
Thanks Dala for the information!

dvbhannes:
I found this in an MG forum: "MG ZS EV OBD PID Reverse Engineering"

Perhaps this will help with developing access to the MG battery!
https://www.mgevs.com/threads/pids-for-torque-pro-app.490/
https://docs.google.com/spreadsheets/d/1cs_8YQVGERspQiPMMY5yFpfLYoyH7UEP30oKzrMk35g/edit?gid=4521783#gid=4521783

j0rgst:
Yes we already know this data, the MG5 has the same adresses.

dvbhannes:
OK

PR #1867, opened 2025-12-23
FoxESS - fix Battery not showing in App / Web Portal

rjsc:
Mapped realistic ASCII values to battery serial numbers and real voltage per pack to get the battery to show in the FOX app and Web Portal


rjsc:
I already tested successfully.  Someone review in order to merge.
Maybe we can later randomize some digits in the serial numbers from the ESP32 MAC address.

PR #1871, opened 2025-12-25
MQTT: Increase buffer, move to stack, de-static various things

jonny5532:
### What
Several tweaks to the MQTT library:
- Increase the buffer size from 1024 to 4096 bytes
- Dynamically allocate it at the start of the mqtt routine and free at the end.
- Fail gracefully if allocation fails.
- Remove the `static`s on the various JsonDocuments (not sure why this is necessary, they all got cleared anyway)

### Why
Should increase the space for MQTT messages (which is causing issues currently), and reduce overall RAM usage.

The mqtt_loop task has a 4096 byte stack is probably unnecessarily large (can look at tuning that later).

Needs testing by someone with a complex MQTT setup!

> [!TIP]

### DIFF
--- a/Software/src/devboard/mqtt/mqtt.cpp
+++ b/Software/src/devboard/mqtt/mqtt.cpp
@@ -46,10 +46,10 @@ const char* ha_device_id =
 
 esp_mqtt_client_config_t mqtt_cfg;
 esp_mqtt_client_handle_t client;
-char mqtt_msg[MQTT_MSG_BUFFER_SIZE];
 MyTimer publish_global_timer(0);  // Will be configured with mqtt_publish_interval_ms on first use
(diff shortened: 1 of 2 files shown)


dalathegreat:
Note, just merged https://github.com/dalathegreat/Battery-Emulator/pull/1881 , might be good to sync main?

danpadure:
well, for me this pull was unusable.
it would reboot instantly when connecting to mqtt (in a loop)
had to stop mqtt server to be able to reflash it with a different version

fastbike:
I had issues with this and destroyed my laptop when connecting for USB debug

jonny5532:
Seems I may have made it worse rather than better! Ideally someone who uses MQTT will sort this out...

PR #1873, opened 2025-12-26
Fix missing includes

madymax:
### What
Adding missing includes

### Why
During a strip-down operation in the Battery Emulator code I have found
some missing includes that were preventing the correctly compile the
project. Most likely in the full project (all inverter and batteries compiled)
the build system is not complaining because the missing objects and functions
are indirectly included.

### How
Adding missing includes

Signed-off-by: madymax <darkmadymax@yahoo.com>

> [!TIP]


dalathegreat:
Massive thanks @madymax for the contribution! :raised_hands:

madymax:
I'm glad that was helpful Dala!

Issue #1880, opened 2025-12-28
[BUG] OTA Update fails if the upload takes more than the MQTT Publish interval


mccasian:
**Describe the bug**
OTA updates for the Battery Emulator consistently fail when the MQTT publish interval is set to a low value. The root cause was not immediately obvious until the MQTT publish interval became configurable (see PR #1846).
While upgrading from a development version using a 2-second MQTT publish interval to the official 9.3.2 release, the OTA update repeatedly failed. Increasing the MQTT publish interval to a higher value (for example, 60 seconds) allows the OTA update to succeed reliably, provided no mqtt_publish() execution occurs during the upload.
This indicates that OTA uploads are disrupted when mqtt_publish() runs while the update is in progress.

**To Reproduce**
1. Set the MQTT publish interval to 1–2 seconds.
2. Attempt an OTA update.
3. If the upload duration exceeds the publish interval (which is typical depending on WiFi conditions), the update fails.
As a control:
1. Set the MQTT publish interval to 60 seconds.
2. Reboot the device.
3. Start the OTA update immediately after reboot.
4. The update succeeds as long as no mqtt_publish() execution occurs during the upload window.

**Expected behavior**
OTA updates should not be interrupted or affected by the mqtt_publish() function.

**Screenshots**
[Youtube Link](https://youtu.be/j8JhKBv-I1Y)

**Version and settings (please complete the following information):**
- [youtube link](https://youtu.be/j8JhKBv-I1Y)
- Software version: 9.3.1-9.3.2
- Battery used: DYI
- Inverter communication protocol: Solis S6(BYD Battery-Box Premium HVS over CAN Bus)
- Hardware used for Battery-Emulator: [LilyGo T-2CAN]
- GPIO controlled contactors: [no]
- MQTT: [yes]

mccasian:
Not sure why I didn’t try fixing it myself first instead of opening an issue. Either way, it’s meantime resolved via #1881.
I am not closing the Issue yet, since you may prefer a different approach.

dalathegreat:
Massive thanks for the PR, merged it, closing this issue as complete!

PR #1881, opened 2025-12-28
Pause MQTT publishing during OTA updates to prevent failures

mccasian:
### What
This PR implements a fix to prevent MQTT publishing from interfering with OTA updates by pausing MQTT transmissions while an OTA update is in progress.

### Why
This change fixes issue #1880 
OTA updates were consistently failing when the MQTT publish interval was set to low values (1-2 seconds). The issue occurred because mqtt_publish() was competing with the OTA update process for network resources and CPU time. When MQTT publishes happened during the file upload phase of an OTA update, the update would get stuck and eventually fail. This was particularly problematic after PR #1846 made MQTT publish intervals user-configurable, allowing users to set very aggressive publish rates that would reliably trigger this failure mode.

### How
The fix works by:
1. Exposing the existing ota_active flag (set by onOTAStart() and cleared by onOTAEnd()) through webserver.h
2. Checking this flag in mqtt_client_loop() before attempting to publish MQTT messages
3. Skipping all MQTT publishing operations while ota_active is true
4. Adding a safeguard that clears the flag when loading the main page, ensuring recovery even if onOTAEnd() fails to execute

This ensures that during the critical OTA upload and flashing phases, no MQTT traffic competes for resources. MQTT publishing automatically resumes when the OTA update completes (whether successful or not), or when the user navigates to the main page.

Changes:
- Added extern bool ota_active; declaration to webserver.h
- Added webserver.h include to mqtt.cpp
- Modified mqtt_client_loop() to check !ota_active before calling publish_values()
- Added ota_active = false; in the main page handler as a failsafe

Tests:
The changes were tested on non-CAN LillyGo dev board and in real life scenario using a Lilygo T-2CAN board - both confirmed to be working as expected.

> [!TIP]

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -187,8 +187,11 @@ void init_webserver() {
   });
 
   // Route for root / web page
-  def_route_with_auth("/", server, HTTP_GET,
-                      [](AsyncWebServerRequest* request) { request->send(200, "text/html", index_html, processor); });
+  def_route_with_auth("/", server, HTTP_GET, [](AsyncWebServerRequest* request) {
+    // Clear OTA active flag as a safeguard in case onOTAEnd() wasn't called
+    ota_active = false;
+    request->send(200, "text/html", index_html, processor);
+  });
 
   // Route for going to settings web page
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Massive thanks for this fix @mccasian , very nice detective work! :raised_hands:

PR #1893, opened 2025-12-30
MQTT: Replace watchdog timer with 60s check from connectivity loop

jonny5532:
### What
This removes the ESP32 watchdog timer setup/reset from the MQTT loop, and instead adds an atomic last-looped-time variable, which is set during the loop, and checked from the connectivity loop.

This allows the timeout to be increased to 60s, avoiding nuisance resets due to wdt timeouts under heavy load (since the mqtt_loop runs with a low priority, it can easily be blocked for 5s.

The code is a bit messy, there's probably a neater way of doing it...

(Also increases the mqtt_loop delay from 1ms to 100ms, given that it publishes at most every 800ms anyway).

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -26,6 +26,7 @@
 #include "src/devboard/utils/timer.h"
 #include "src/devboard/utils/types.h"
 #include "src/devboard/utils/value_mapping.h"
+#include "src/devboard/utils/watchdog.h"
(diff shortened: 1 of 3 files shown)


danpadure:
Uptime: 6 days, 4 hours, 4 minutes, 29 seconds
with no antenna
merge it

dalathegreat (approved):
👍

PR #1897, opened 2026-01-01
BYD Atto 3: Experimental cell balancing

bingflash2:
### What
This PR implements ... experimental cell balancing -  requests a software reset of the bms, if the conditions to trigger balancing are present, only once every 12hours (assumes one complete charge from solar, per day. nigh-time energy arbitragers can alter the frequency)

### Why
Why does it do it? byd has an unusual mechanism where, the logic to trigger cell balancing is as such : if cell voltage >=3.66v then flag for balancing on the next power on event, normally this would be the ign on event after the bms wakes up, after a charge cycle, which does not happen in the BE, a subtle blocking condition.

### How How does it do it? add uds command to reset bms - 0x11 0x03, should testing fail, ill change this to 0x11 0x02 (keyoff)

after many many hours of observation, the exact voltage at which balancing is flagged is 3.66v - Best regards, Billy

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -594,7 +594,7 @@ void BydAttoBattery::transmit_can(unsigned long currentMillis) {
 
     if (uds_reset == true) {                      // initially enabled
       if (uptime_ticks > 50) {                    // bms running for 10s min, ensure contactors closed etc
-        if (BMS_lowest_cell_voltage_mV > 3599) {  // if minimum cell voltage >=3.6v
+        if (BMS_highest_cell_voltage_mV > 3650) {  // if max cell voltage >=3.6v
           uds_reset = false;                      // reset to false, wait 12hrs
           transmit_can_frame(&ATTO_3_RESET);
         }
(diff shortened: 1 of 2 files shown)


bingflash2 (commented):
can you make a binary, for someone to test ?

jonny5532:
Have let it (try to) build, but you have typos that will need fixing first...

TheTillerMan:
Hi all do you have a  test file for a 50kw byd I can trial? Guessing the lillygo can reset the battery BMS

demon1300:
Since I've found the frame that increments the charge times counter, couldn't we now force the BMS to consider a charge being completed, hence triggering this balancing code without a restart?

bingflash2:
> Hi all do you have a test file for a 50kw byd I can trial? Guessing the lillygo can reset the battery BMS

im presently trying to get platformio python and vscode running on windows and then we can test some things.

demon1300:
I've just compiled and flashing now, will let you know how it goes.  Attached is the built image for any other Lilygo T-CAN485 users that want to test this.

[LilygoT-CAN485_PR1897_firmware.zip](https://github.com/user-attachments/files/24425364/LilygoT-CAN485_PR1897_firmware.zip)

dalathegreat:
@bingflash2 , could you fix the code? https://github.com/dalathegreat/Battery-Emulator/actions/runs/20793261033/job/59726787579?pr=1897#step:6:130

TheTillerMan:
Hi Guys, 

Just wondered if you had worked out a fix for the BYD 50KW pack SOC issue and if cell balancing should fix this issue. 


Thanks

dalathegreat:
PR is stale, the BYD integratioon has been reworked significantly since this was posted. I will close this PR, but feel free to re-base and open a new one!

PR #1902, opened 2026-01-02
T-CAN485: Configurable BMS Power pin option

### LINKED ISSUE #1570: nissan leaf battery BMS power cycling GPIO conflicts with IO for external CAN-board
nissan leaf battery BMS power cycling GPIO conflicts with IO for external CAN-board.
For the pre 9-series releases this could be solved by editing the HAL-definitions and setting the #define BMS_POWER to a usefull pin, maybe this can be solved by control of which GPIO pins functions are assigned to in the web interface?

regards
/Linda

### LINKED ISSUE #1896: [BUG]  GPIO 18 pin assign conflict between bms_power and MCP2515
**Describe the bug**
System was running on 8.8.0 with battery on native can and inverter on addon mcp2515 can

Upgraded to 9.4.dev to test new battery integration and got error for GPIO pin 18 conflict.

Image

CAN between inverter and BE did not work.

Temporary fix:
I have changed BMS_power GPIO from 18 to 25 which works for me now.

Maybe not a bug but it doesn't seem to be documented that the GPIO pins could be in conflict depending on which options are used?

Image

- Software version: 9.4.DEV
- Battery used: EMUS BMS   (I New, basically a copy of Pylon hv with minor changes)
- Inverter communication protocol: [GROWATT HV CAN]
- Hardware used for Battery-Emulator: [LilyGo T-CAN+MCP2515
- GPIO controlled contactors: [no]
- MQTT: [yes]

dalathegreat:
### What
This PR implements a Configurable BMS Power pin option for the LilyGo T-CAN485 board

### Why
Flexibility required for T-CAN485 board. Some people need add-on CAN + BMS power, others need contactor control + BMS power. This PR allows for both combos. Fixes #1570 and also fixes #1896

### How
We add a configuration option where user can select Pin18 or Pin25 for BMS power.

image

> [!TIP]

### DIFF
--- a/test/can_log_based/canlog_safety_tests.cpp
+++ b/test/can_log_based/canlog_safety_tests.cpp
@@ -178,7 +178,7 @@ void RegisterCanLogTests() {
   //     cuv:  test that normal and critical cell undervoltage events are triggered
   //     cv88: test that cell 88 (or another) voltage is correctly set (to 3123mV)
 
-std::string directoryPath = TEST_CAN_LOG_DIR;
+  std::string directoryPath = TEST_CAN_LOG_DIR;
 
   for (const auto& entry : fs::directory_iterator(directoryPath)) {
     if (!entry.is_regular_file() || entry.path().extension().string() != ".txt") {


jonny5532:
Looks good, my only thought is whether the var setting could be better named - maybe just GPIOBMSPWR or whatever.

(I called the 2CAN one GPIOOPT1 just because there was the possibility of having more multi-GPIO-affecting options in future (it has another 2-pin QWIIC connector, plus all the IOs on the bottom), and I couldn't think of a better name...)

But doesn't really matter I guess, whatever you think makes sense!

dalathegreat:
After going thru and renaming, I think it made the code harder to follow. A list of GPIO Options (1-2-3-4..) looks cleaner and easier to manage. So I think we'll keep it as-is!

PR #1941, opened 2026-01-10
MG HS/MG5 UDS refactor

jonny5532:
### What
This PR rebases the MG HS integration on the new UdsCanBattery subclass. It also renames it MG-GEN1 since it now supports ZS/MG5/MarvelR.

Features:
- Change from pow() based to linear charge/discharge tapering to save a bunch of code
- Adding cellvoltage and temperature based tapering
- Adding double/triple battery support
- Supports keeping contactors closed during OTA/reboots (max 30s)

The MG5 integration is untouched, it remains large-flash only. A future PR may contain a small fix (to make it work on the ZS packs) and possibly making it legacy in favour of this integration.

Previous versions of this PR have been tested and work on ZS/MG5/MarvelR 41/52/61kWh packs as well as the original 16.6kWh HS.

Double battery has also been tested.

Potential improvements:
- It includes a ignore-SoC mode, where charge/discharge taper is based entirely on cellvoltage/temp, to allow more of the battery to be used (some batteries top out at less than 4.2V/cell). This is a compile time setting though, disabled by default.
- The top-end SoC taper (but not the bottom) is somewhat duplicated by the new BE built-in taper
- The voltage/temp tapers could be made a wider feature for other batteries (which can suffer SoC drift too).

> [!TIP]


Software/src/devboard/utils/common_functions.h
```diff
@@ -13,3 +13,8 @@ extern int16_t sign_extend_to_int16(uint16_t input, unsigned input_bit_width);
 extern const uint8_t crc8_table_SAE_J1850_ZER0[256];
 extern const uint8_t crctable_nissan_leaf[256];
 extern const uint8_t crctable_geely_geometryC[256];
+
+uint32_t taper_charge_power_linear(uint32_t soc, uint32_t max_charge_power, uint32_t trickle_charge_power,
```
dalathegreat: Very nice, we might be able to switch all estimated integrations to these functions!

dalathegreat (approved):

Software/src/lib/uds_isotp/uds.h
```diff
@@ -70,7 +70,10 @@ enum SID : uint8_t {
   RequestDownload               = 0x34,  ///< Section 14.2 (p. 270) - Init download
   RequestUpload                 = 0x35,  ///< Section 14.3 (p. 275) - Init upload
   TransferData                  = 0x36,  ///< Section 14.4 (p. 280) - Data blocks
-  RequestTransferExit           = 0x37   ///< Section 14.5 (p. 285) - End transfer
+  RequestTransferExit           = 0x37,  ///< Section 14.5 (p. 285) - End transfer
+
+  // === Negative Response (Section 7.4) ===
+  NegativeResponse              = 0x7F   ///< Section 7.4 (p. 26) - Always starts with 0x7F
```
malec1984: This should be removed, it's not a service, the negative responses codes are down below in enum NegativeResponseCode. Hmm ok, I see you wanted the use the negative response 0x7F which is not defined anywhere, perhaps you could just use a define for this scope.

Software/src/battery/MG-GEN1-BATTERY.cpp
```diff
@@ -0,0 +1,830 @@
...
+
+  char buf[128];
+
+  ret += "UDS address: ";
+  ret += String(uds_address, 16);
+  print_chars_or_hex(buf, pid_f190, 17);
+  ret += "<br>VIN: ";
+  ret += buf;
+  print_chars_or_hex(buf, pid_f18a, 8);
+  ret += "<br>F18A: ";
```
malec1984: What do you think about using names here instead of the diagnostic ID?

Software/src/battery/MG-GEN1-BATTERY.cpp
```diff
@@ -0,0 +1,830 @@
...
+
+    case POLL_BATTERY_TYPE:  // Battery type
+      if (value == 0) {
+        // Retry until we get a valid battery type (0 is invalid)
+        return POLL_BATTERY_TYPE;
+      }
+      got_battery_type(value);
+      memcpy(pid_f18a, data, length > sizeof(pid_f18a) ? sizeof(pid_f18a) : length);
+      break;
+    case 0xF120:
```
malec1984, line 645: Same for these PIDs as you might know the meaning behind would be easier to read the code if you have names instead of hex numbers.

Software/src/battery/MG-GEN1-BATTERY.cpp
jonny5532: Fair point, although I don't know what any of the other identifiers mean - this is intended for autodetection, so I can reliably tell the different batteries apart. I guess people might be interested in the VIN in the frontend though, I'll pass that through properly.

(There are a couple(?) more batteries that we haven't seen yet, so by leaving this code in, I hope to more easily get the required data off other users. It can go away eventually).

malec1984: Perhaps some of the definitions (generic) are available here:
https://udsoncan.readthedocs.io/en/latest/udsoncan/helper_classes.html#dataidentifier

jonny5532: Thanks, have added the ones that match, MfrDate is indeed interesting (the others rather more random)

Software/src/battery/MG-GEN1-BATTERY.cpp
jonny5532: Added where known

PR #1943, opened 2026-01-11
Fix SD Init bug when SD missing or is corrupted

madymax:
### What
This fix is checking if the SD card was correctly initialized , otherwise it will delete the logging buffers, disable the SD logging and delete the logging task to save some memory

### Why
There are cases when SD logging is enabled without a phisical SD card present or the SD card is corrupted. In these cases a crash could happen because the SD init routine is not checking the result of init_sdcard() function and this could lead into different crashes depending on the case.

### How
Checks the init_sdcard() function return code.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -111,11 +111,19 @@ void connectivity_loop(void*) {
 }
 
 void logging_loop(void*) {
+  bool sd_initialized = false;
 
   init_logging_buffers();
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Massive thanks for this contribution @madymax , looks like a great improvement! :raised_hands:

PR #1948, opened 2026-01-13
Feature: Recovery charge for undercharged batteries

dalathegreat:
### What
This PR implements a recovery charge mode for undercharged batteries

### Why
Recovering an undercharged battery is hard. It requires compiling source code manually, with a lower allowed voltage value for the specific battery you are working with. It requires help from developers to recover batteries, which make it very inconvenient if you run into a scenario where your battery dipped too low and is no longer possible to charge. Fixes #1948

> [!CAUTION]
> Recovering a too low cell can result in fire :fire: Always limit charging power to a single digit value, and pay close attention to temperatures. Charging too fast can result in cell rupture and fire. Recovering a 2.7V cell is generally safe, 2.5V needs caution, below 2.0V is usually not possible and will always result in a permanently damaged battery that is no longer safe for stationary use. You have been warned!

### How
A scary looking option is added to the Settings page

image

If user presses Start, they have to enter "1" to confirm going into the emergency recovery mode

image

User gets notified also via Events that this mode started

image

In this mode, charging up to 5.0A is allowed. Current can be lowered manually via the max charge A setting. After 30 minutes, the mode is exited.

Special edge case, if one cell is below 2000mV, the function disables instantly. It is not safe to recover such an overcharged cell.

> [!TIP]

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -13,6 +13,8 @@ static bool battery_empty_event_fired = false;
 
 #define MAX_SOH_DEVIATION_PPTT 2500
 #define CELL_CRITICAL_MV 100  // If cells go this much outside design voltage, shut battery down!
+#define LOWEST_ALLOWED_CELLVOLTAGE_RECOVERY_CHARGE_MV 2000  //If cells are below this, recovery charge not allowed
+#define MAX_CHARGEPOWER_RECOVERY_CHARGE_DA 50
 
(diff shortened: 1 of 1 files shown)


abdyousef60-alt:
Good job,well test it and feeding you back

Simply-Johann:
I've revived/charged undercharged high-voltage batteries using a variable transformer/isolation transformer and a rectifier.
Of course, the weakest cell must be intact!
FYI

PR #1951, opened 2026-01-15
Add runtime battery model configuration for Sungrow SBRXXX inverter

aferrara:
- Add SungrowBatteryConfig struct with nameplate capacity and module count
- Support battery models SBR064 through SBR256 (2-8 modules)
- Dynamically set battery model ID in 0x705 based on module count
- Initialize module-dependent CAN frames in setup():
  - 0x71A: cell type per module
  - 0x71B-E: production dates
  - 0x71F: serial numbers
  - 0x715-718: module voltage overview
- Add web UI dropdown for battery model selection (Sungrow inverter only)
- Conditionally transmit 0x71F frames based on configured module count

> [!TIP]

### DIFF
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -112,6 +112,11 @@ static const std::map<int, String> tesla_chassis = {{0, "Model S"}, {1, "Model X
 
 static const std::map<int, String> tesla_pack = {{0, "50 kWh"}, {2, "62 kWh"}, {1, "74 kWh"}, {3, "100 kWh"}};
 
+static const std::map<int, String> sungrow_models = {
+    {0, "SBR064 (6.4 kWh, 2 modules)"},  {1, "SBR096 (9.6 kWh, 3 modules)"},  {2, "SBR128 (12.8 kWh, 4 modules)"},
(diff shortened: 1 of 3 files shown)


Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -773,6 +773,34 @@ String raw_settings_processor(const String& var, BatteryEmulatorSettingsStore& s
     return String(settings.getUInt("INVBTYPE", 0));
   }
 
+  if (var == "INVBTYPE_SBR064") {
```
dalathegreat: To save memory, use a single variable for getUINT, and store each selection as an enum
0 = SBR064
1 = SBR096
2 = SBR128
etc.

dalathegreat (changes requested):

Software/src/devboard/webserver/settings_html.cpp
aferrara: Thanks for the feedback @dalathegreat. I've made some changes which hopefully resolves your valid concern.

dalathegreat:
Nice work! Can you update the Sungrow Wiki with info on what model to use depending on your battery voltage range?

aferrara:
> Nice work! Can you update the Sungrow Wiki with info on what model to use depending on your battery voltage range?

Of course. I'll get that sorted today.

PR #1956, opened 2026-01-16
Exit Extended Diagnostic Session  after opening contactors

rjsc:


Software/src/battery/KIA-HYUNDAI-64-BATTERY.cpp
```diff
@@ -182,6 +182,11 @@ void KiaHyundai64Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
             KIA64_7E4_OPEN_CONTACTOR_SEQUENCE.data.u8[1] = 0x2F;
             KIA64_7E4_OPEN_CONTACTOR_SEQUENCE.data.u8[2] = 0xF0;
             KIA64_7E4_OPEN_CONTACTOR_SEQUENCE.data.u8[3] = 0x31;
+            // Exit Elevated Mode
```
dalathegreat, line 185: Should we only exit elevated mode and allow re-closing of contactors if

inverter allows = true, and not in FAULT mode?

rjsc: My addition doesn't reclose the contactors.
Only exits the elevated mode to allow for reclosing on other code sections such as line 477.
At line 477 the code checks if the battery is in fault mode  and only then proceeds to close contactors.

dalathegreat: Reviewing this again, looks like this change will write the bits back too fast before the state=4 has sent the message. So we need to add a state=5, or do a check after the sending if we should exit.

Issue #1959, opened 2026-01-16
[BUG] Battery Emulator reports "Battery-Undervoltage" at 360Vdc (SMA SBS2.5 connected to Think-Battery)


Simply-Johann:
**Describe the bug**
A clear and concise description of what the encountered bug is

**To Reproduce**
Please try to explain the steps required to reproduce the issue

**Expected behavior**

**Screenshots**

**Version and settings (please complete the following information):**
- Please attach screenshot of configuration for easier debugging
- Software version: X.Y.Z
- Battery used: [Nissan LEAF]
- Inverter communication protocol: [BYD Modbus]
- Hardware used for Battery-Emulator: [LilyGo T-CAN/Stark CMR/LilyGo T-2CAN]
- GPIO controlled contactors: [yes/no]
- MQTT: [yes/no]

dalathegreat:
Fix for the Undervoltage event on startup: https://github.com/dalathegreat/Battery-Emulator/pull/1960

Simply-Johann:
I made several pairing attempts at a battery voltage of 324V. I navigate to the battery configuration in the SBS server form wizard, but nothing happens after that. The inverter does not recognize the battery. The battery emulator reports "Battery_Undervoltage" and "Inverter missing". I've currently even raised the battery voltage to 360V - BE is reporting "Undervoltage"...
Then I connected the SBS2.5 to the mains, and when the enable line went high, I started the BE and the battery (about 2-3 seconds after enabling). This time I had a pairing attempt.

Simply-Johann:
The battery emulator sends the CAN messages to the inverter as soon as the enable line is set to HIGH.
The high voltage at the inverter input is low at this time. My main contactor only closes after 1200 ms. Now HV reach its value. Emhal reports that a delay for CAN Messages after Enable_High solve the pairing problem at SBS2.5. And, apart from the Battery_UnderVoltage problem, my inverter is behaving the same way as his. Could you implement the SBS2.5-inverter protocol of Emhal to check it? What else could be causing the BE error message "Battery_Undervoltage" at real 324V?

Hi Daniel, why can I not describe the BUG? I can not edit my report...

![Image](https://github.com/user-attachments/assets/cb2be200-0ac7-492f-ba8e-a8f005403abd)

![Image](https://github.com/user-attachments/assets/da3b1cd4-4dd8-4d55-b783-3d561d38d73c)

dalathegreat:
[...]

PR #1960, opened 2026-01-16
Bugfix: Think startup voltage initialization

dalathegreat:
### What
This PR fixes the Think startup voltage initialization

### Why
Value is initialized to 0, which means we might trigger Battery-Undervoltage event incase CAN data is not read the first second after poweron. Reported in #1959

### How
We init the value to 360.0V instead of 0.0V

> [!TIP]


lenvm:
Fixes #1959

PR #1966, opened 2026-01-17
Testing the maximum capacity and voltage readings with the Fronius Primo and BYD Atto3

pmajor021:
The update prevents BE from reporting a maximum or remaining capacity above 44.16 kWh to the inverter, which matches the total of two 22.08 kWh BYD HVM batteries.

What
This PR updates the implementation to report a reduced capacity value.

Why
The maximum allowed configuration by design for Fronius Gen24 inverters is 2x22.08 BYD HVM batteries.

How
Updating the previous 60kWh limit to 44.16kWh.

> [!TIP]

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -54,7 +54,7 @@ void BydModbusInverter::handle_update_data_modbusp201_byd() {
   }
   mbPV[205] =
       std::min(datalayer.battery.info.max_design_voltage_dV,
-               static_cast<uint16_t>(4549u));  // Max Voltage, if higher Gen24 forces discharge, cap to 454.9V for Primo
+               static_cast<uint16_t>(4500u));  // Max Voltage, if higher Gen24 forces discharge, cap to 450.0V for Primo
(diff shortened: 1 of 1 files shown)


pmajor021:
It might be a good idea to update the code to include Battery 3’s remaining capacity in the total capacity reported to the inverter.

pmajor021:
With the latest update, the maximum design voltage of 454.9V and total capacity 57.96kWh are the peak values reported to Fronius, with the total capacity coming from 3 x 19.3kWh. The maximum design voltage is just under the allowed limit for the Fronius Primo Gen24.

pmajor021:
Further lowered the maximum voltage to 450V.

dalathegreat (approved):
Approved! Only downside of this PR is that Fronius will no longer flag too high voltage batteries with a warning. For instance the BYD Atto3 with a max voltage of 460V is technically over the design limit of the Fronius. But the warning message that Fronius uses when voltage is too high is quite bad, they say "Invalid battery size detected" even though it should mention voltage. So I think this change is fine as a start.

Later on we could have a dropdown menu of BYD batteries OR a selection between Symo and Primo inverters.

So to recap, this PR shifts responsibility of system design more towards end user, less warnings.

PR #1967, opened 2026-01-18
BYD Atto 3 - Add gradual charge taper

demon1300:
### What
You can still control the “Manual charge current limit”, but as you approach ~3.42–3.50V max cell or 40–80mV delta, BE will proactively reduce the allowed charge power so it will taper like the car, instead of waiting for the BMS hard clamp.

### Why
In its current form, BE is acting on BMS messaging and charging limits that appear to be hard set safety limits to slow charging.  This results in large current input right until a hard clamp, once the safety limit is hit.  This doesn't allow a more controlled taper of current at the top end which is what has been whitnessed by AC charging logs from a vehicle, so this makes the charging behave more like it would in the car.

### How
This change honours the currently set manual charigng limit up until the below thresholds are met.  It computes a progressive value from 0-1 between the start and end thresholds then linearly reduces the cap from the manual limit set, to 0.2A.  It also incorporates a slew-rate limiter so it ramps smoothly (which is more car-like).

Taper Start - 3.420v OR cell delta of 40mv
Taper End - 3.500v OR cell delta of 80mV
Tail Current - 0.2A

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -42,6 +42,9 @@ jobs:
           - board_name: "ESP32 DevKit V1"
             pio_env: "esp32devkit_330"
             artifact_name: "battery-emulator-esp32devkit"
+          - board_name: "BECom"
+            pio_env: "BECom_330"
+            artifact_name: "battery-emulator-becom"
 
     steps:
     - uses: actions/checkout@v6
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -30,12 +30,13 @@
 #include "src/devboard/wifi/wifi.h"
 #include "src/inverter/INVERTERS.h"
 
-#if !defined(HW_LILYGO) && !defined(HW_LILYGO2CAN) && !defined(HW_STARK) && !defined(HW_3LB) && !defined(HW_DEVKIT)
+#if !defined(HW_LILYGO) && !defined(HW_LILYGO2CAN) && !defined(HW_STARK) && !defined(HW_3LB) && !defined(HW_BECOM) && \
+    !defined(HW_DEVKIT)
 #error You must select a target hardware!
 #endif
 
 // The current software version, shown on webserver
-const char* version_number = "9.4.dev";
+const char* version_number = "9.3.5";
 
 // Interval timers
 volatile unsigned long currentMillis = 0;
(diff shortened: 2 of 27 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -199,6 +199,103 @@ void BydAttoBattery::
 
   datalayer_battery->status.cell_min_voltage_mV = BMS_lowest_cell_voltage_mV;
 
+  // AC-like top-of-charge taper
```
dalathegreat, line 202: This could be its own function, since the section is very long

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -199,6 +199,103 @@ void BydAttoBattery::
...
+    const uint16_t step = (up_step >= diff) ? diff : (uint16_t)up_step;
+    cap_slewed_dA += step;
+  }
+
+  // Convert current cap (dA) -> power cap (W): P = I(dA) * V(dV) / 100
+  const uint32_t power_cap_W = (uint32_t(cap_slewed_dA) * uint32_t(datalayer_battery->status.voltage_dV)) / 100;
+
+  // Apply taper by capping the allowed charge power reported to the rest of BE/inverter logic.
+  if (datalayer_battery->status.max_charge_power_W > power_cap_W) {
+    datalayer_battery->status.max_charge_power_W = power_cap_W;
```
dalathegreat, line 295: Should there be a note that max_charge_power_W behaves differently based on if you are using actual SOC or estimated SOC? And should we always use the taper?

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -199,6 +199,103 @@ void BydAttoBattery::
...
+      return 0.0f;
+    if (x > 1.0f)
+      return 1.0f;
+    return x;
+  };
+
+  float vprog = 0.0f;
+  if (cell_max_mV > V_TAPER_START_mV) {
+    const uint16_t denom = (V_TAPER_END_mV > V_TAPER_START_mV) ? (V_TAPER_END_mV - V_TAPER_START_mV) : 1;
+    vprog = float(cell_max_mV - V_TAPER_START_mV) / float(denom);
```
dalathegreat, line 238: I like using protection checks for avoiding division by 0. Not really a problem here though!

PR #1975, opened 2026-01-19
New battery, Emus G1 BMS

ErikssonPer:
This adds integration for Emus G1 BMS, [www.emusbms.com](http://www.emusbms.com/)

Emus works with the pylon integration for the basic stuff but I have made it's own so cellmonitoring and calculation of capacity from the BMS
Also so further implementation of wanted data, error handling and such doesn't interfer with the pylon integration.

It has been running for couple of days during charge and discharge on my system with Growatt SPH inverter and the emus G1 BMS installed on a 1 string 119 cell LFP diy batterypack.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: configure
+      run: ./configure
+    - name: make
+      run: make
+    - name: make check
+      run: make check
+    - name: make distcheck
+      run: make distcheck


Software/src/inverter/GROWATT-HV-CAN.cpp
```diff
@@ -23,12 +23,45 @@ BMS - Battery Information Collector */
...
-    ampere_hours_remaining =
-        ((datalayer.battery.status.reported_remaining_capacity_Wh / datalayer.battery.status.voltage_dV) *
-         100);  //(WH[10000] * V+1[3600])*100 = 270 (27.0Ah)
-    ampere_hours_full = ((datalayer.battery.info.total_capacity_Wh / datalayer.battery.status.voltage_dV) *
-                         100);  //(WH[10000] * V+1[3600])*100 = 270 (27.0Ah)
+    // 0x3140 expects capacity in 10mAh units.
+    // capacity_10mAh = Wh * 1000 / dV   (because V = dV/10, and 10mAh units = Ah*100)
+    const uint16_t v_dV = datalayer.battery.status.voltage_dV;
+
+    uint32_t full_10mAh = (uint32_t)((uint64_t)datalayer.battery.info.total_capacity_Wh * 1000ULL / v_dV);
```
dalathegreat, line 47: if v_dV would be 0, division would crash the CPU

Software/src/inverter/GROWATT-HV-CAN.cpp
```diff
@@ -23,12 +23,45 @@ BMS - Battery Information Collector */
 void GrowattHvInverter::
     update_values() {  //This function maps all the values fetched from battery CAN to the correct CAN messages
 
+  // ---- Cell/module topology (Growatt frames 0x3150 and 0x3180) ----
+  // Previous revisions hard-coded TOTAL_NUMBER_OF_CELLS=300 and NUMBER_OF_MODULES_IN_SERIES=20.
+  // EMUS provides the actual series cell count in datalayer.battery.info.number_of_cells (e.g. 119).
+  // Use that when it is sane. Keep the hard-coded values as a fallback for integrations that do not populate
+  // number_of_cells (e.g. some Pylon paths set it to a placeholder value).
+  uint16_t total_cells = TOTAL_NUMBER_OF_CELLS;
```
dalathegreat, line 31: These Growatt-HV-CAN changes look great! Could they be added in as a separate PR?

Software/src/battery/EMUS-BMS.cpp
```diff
@@ -0,0 +1,267 @@
+#include "EMUS-BMS.h"
```
dalathegreat, line 1: This whole EMUS-BMS seems to be 80% PYLON-BATTERY.cpp

Since we are limited on memory, we should instead use Pylon-Battery, and add the separate Emus CAN IDs there in the CAN handler.

We can also add a "250kbps :white_check_mark: " option to the Pylon battery, to gain even more compatibility with more batteries.

P76E:
Ok, keeping it compact is a good point.
Added number of cells and  cellmonitoring into the pylon integration, removed the emus integration completely and added a 250/500k setting on the config page instead.
Been running it for a day now so I think it's good to go in a while.
image

dalathegreat:
Nice work! Feel free to open a new PR, or update this one!

PR #1978, opened 2026-01-25
Experimental Feature: Double batteries in Series

dalathegreat:
### What
This PR adds support for running double battery in series connection instead of the normal parallel mode

### Why
User requested feature. Charge faster on current limited inverters

### How
To use the feature, connect batteries in Series Operation mode

image

Then, enable the "Series operation, EXPERIMENTAL" under the Double-Battery config

For now, only Fronius Gen24 is supported during testing phase

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -143,27 +143,64 @@ void check_interconnect_available(uint8_t batteryNumber) {
     if (datalayer.battery.status.voltage_dV == 0 || datalayer.battery2.status.voltage_dV == 0) {
       return;  // Both voltage values need to be available to start check
     }
-    uint16_t voltage_diff_battery2_towards_main =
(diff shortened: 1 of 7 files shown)


Alikoman:
Hello. How metering SOC and balancing every battery?

dalathegreat:
This feature has been scrapped due to safety concerns.

- No control over CAN controlled contactors makes this feature hard to use safely
- How to ensure balancing, that each battery reaches 100%? In parallel operation this is easy, in series it becomes next to impossible.
- None of the isolation is designed for double the working voltage. Yes, each battery only sees it's own voltage, but the isolation to earth and in the BMS comms suddenly sees twice. As do any internal contactors, which is probably the more immediate issue.
- I do not have a test setup to develop this feature safely

PR #1886, opened 2025-12-29
Library :books: : Update ESP32 to v3.3.5

dalathegreat:
### What
This PR updates the ESP32 backbone to v3.3.5

### Why
Smaller flash usage which we desperately need, and fixes #1884

Before:
RAM:   [===       ]  25.2% (used 82432 bytes from 327680 bytes)
Flash: [==========]  99.6% (used 1957519 bytes from 1966080 bytes)
After:
RAM:   [===       ]  25.4% (used 83100 bytes from 327680 bytes)
Flash: [==========]  99.4% (used 1954255 bytes from 1966080 bytes)

### How
We update from v3.3.0 -> v3.3.5

> [!TIP]
> [You can help test this PR with this guide](https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#downloading-a-pull-request-build-to-test-locally-)

PR #1986, opened 2026-01-28
Feature: Update Stark to 8MB + Update ESP32 to 3.3.6

dalathegreat:
### What
This PR updates the partitioning layout on Stark CMR to become 8MB. It also bumps up the ESP32 version from v3.3.0 -> v3.3.6

### Why
We have started to wrap some features to exclusively work on 8MB boards (mDNS and some upcoming new integrations), because the 4MB is starting to become a limitation. So to ensure that Stark's get all new features, this change is needed. It is also needed for future HW expansion add-ons for the Stark CMR

### How

> [!IMPORTANT]
> Factory image flash required via USB after this change

This update will change the partitioning layout AND flash frequency for factory images, **which means it will require a USB flash instead of OTA update**

> [!TIP]

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -47,7 +47,7 @@ board_build.f_flash = 80000000
 board_build.arduino.memory_type = qio_qspi
 board_build.partitions = default_8MB.csv
 framework = arduino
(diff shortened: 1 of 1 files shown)


dalathegreat:
Note, comment in https://github.com/dalathegreat/Battery-Emulator/pull/1886#issuecomment-3697318164 needs to be followed to successfully integrate this change

PR #1988, opened 2026-01-29
Prevent voltage bouncing due to repeadetly charge/discharge with Pylon LV

M4GNV5:
### What
This PR implements less harsh charging/discharging rules. The goal is to just stop discharging when reaching minimum voltage or soc rather than immediately running into a force charge request.

### Why
With the current code, when the battery gets discharged below the user min voltage BE immediately requests a "force charge" resulting in it being charged slightly over the min voltage and the cycle continues.
image

### How
Only request force charge when battery below design voltage. Do not request when voltage below user limit. This way the two voltages act as a kind of hysteresis.

draft for now until I am able to test this with the real setup.

### DIFF
--- a/Software/src/battery/DALY-BMS.cpp
+++ b/Software/src/battery/DALY-BMS.cpp
@@ -31,6 +31,17 @@ void DalyBms::update_values() {
   datalayer.battery.status.max_discharge_power_W =
       (datalayer.battery.settings.max_user_set_discharge_dA * voltage_dV) / 100;
 
+  // limit power when reaching discharge voltage limit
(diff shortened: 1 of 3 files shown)


M4GNV5:
manually reflashed my lilygo today (v10 updates) and tested the changes from this PR. Had to make some slight adjustments, but it works as intended with only minimal current bouncing now.

image

Are there any other LV / Pylon / Daly users besides myself btw?

dalathegreat (approved):
Nice work! There aren't that many LV users yet, so I don't think we'll get any testers on this. But code looks good, so let's merge this! 👏

Issue #1884, opened 2025-12-29
[BUG] platform-espressif32 55.03.30 leads to failed installation of idf_tools.py


lenvm:
**Describe the bug**
`platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.35/platform-espressif32.zip` causes an issue with `pdf_tools.py` not installing properly:

```
idf_tools.py installation failed
idf_tools.py installation failed
idf_tools.py installation failed
idf_tools.py installation failed
idf_tools.py installation failed
idf_tools.py installation failed
```

**To Reproduce**
Clone the repository to a Mac, open VSCode, install PlatformIO, try to build.

**Expected behavior**
No issues during the build.

Issue can be fixed using:
`platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.35/platform-espressif32.zip`

Let's check if we can upgrade to this newer version with bug fixes.

**Version and settings (please complete the following information):**
- Software version: `84fdba5`

dalathegreat:
We had a massive issue with updating ESP32 to v3.3.1 / 3.3.2 some month ago, which led to PSRAM / crashes & other various issues on the Stark CMR. So I am not so eager to update the ESP32 this time. I can ofcourse open a PR with 3.3.5, and we can start stability testing it properly, but yeah, after the countless emails I got last time of the failed ESP32 update, not so eager 😅

lenvm:
Thanks @dalathegreat! I appreciate the PR #1886!

I just checked in detail: the bug reported is fixed once updating from [v3.3.0-2](https://github.com/pioarduino/platform-espressif32/releases/tag/55.03.30-2) to [v3.3.1](https://github.com/pioarduino/platform-espressif32/releases/tag/55.03.31).

PR #1997, opened 2026-02-01
Improvement: Inform user of low memory situation + ESP downgrade

dalathegreat:
### What
This PR informs user when memory is running low

### Why
We have had no clue how on edge the heap allocation has been. Let's inform users in the future if we run low on RAM

### How
- Event added if we go below 62k free heap
- Profiling page now displays free heap
- ESP32 downgraded to v3.3.5 to avoid heap issue (https://github.com/espressif/arduino-esp32/issues/12284) (36k allocated more in v3.3.6, some regression)

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -37,7 +37,7 @@
 #endif
 
 // The current software version, shown on webserver
-const char* version_number = "10.0.RC3";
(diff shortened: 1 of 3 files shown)


lenvm:
Fixes #1884

PR #2001, opened 2026-02-02
Growatt hv capacity fix

ErikssonPer:
Fixed capacity reporting to inverter and added actual cellcount if avaliable from BMS.

In some cases they came in conflict with eachother and the inverter disabled charge and discharge.

> [!TIP]


P76E:
Tested on my system.

PR #2004, opened 2026-02-03
Bugfix: Think City CAN values

### LINKED ISSUE #2003: [BUG]Battery Emulator always shows SOC 50% (Think-City battery connected to SMA SBS2.5)
**Describe the bug**
Battery Emulator always shows SOC 50% (Think-City battery connected to SMA SBS2.5)

**To Reproduce**
While the actual SOC (determined via BMS) fluctuates widely, the BE SOC always remains at 50%. The SBS server also always displays a SOC of 46%

**Expected behavior**
SOC is a dynamic value…

**Screenshots**

Image

- Software version: [9.3.5 from @emhal with modified Inverter protocol]
- Battery used: [Think City]
- Inverter communication protocol: [SMA compatible BYD Battery-Box HVS], modified by @emhal
- Hardware used for Battery-Emulator: [LilyGo T-2CAN]
- GPIO controlled contactors: [yes]
- MQTT: [no]

After receiving the software with a slightly modified inverter protocol from @emhal, everything went smoothly:
I set "SMA compatible BYD Battery-Box HVS" in BE and in the SBS Wizard, and restarted the system. After a few minutes, the inverter had already started charging/discharging. After a full day of operation, I noticed that the SOC value wasn't changing.

Daniel-san, could you please implement the SOC in the Think City Battery protocol?
And, while you're at it, also the min/max cell voltages?
Unfortunately, the Min/Max_CV value has an annoying float factor of 0.00244141, which significantly increases processing power. I therefore suggest updating this values only once per minute.
Attached the ID-Messages 0x301 and 0x610 of Think City BMS (it uses DOD instead of SOC)
Thank you!

Image

Image

dalathegreat:
### What
This PR improves several CAN mappings for the Think City

### Why
Better support for Think City batteries. Fixes #2003

### How
- Change SOC to be based on DOD
- Swap charge/discharge, these were accidentally wrong way
- Fix cell temperatures to use proper min/max instead of mean temp
- Add min/max cellvoltages

> [!TIP]

### DIFF
--- a/Software/src/battery/THINK-BATTERY.cpp
+++ b/Software/src/battery/THINK-BATTERY.cpp
@@ -25,13 +25,13 @@ void ThinkBattery::update_values() {
 
   datalayer.battery.status.temperature_max_dC = (int16_t)(max_pack_temperature * 10);
 
-  datalayer.battery.info.min_design_voltage_dV = sys_voltageMinDischarge;
+  if (max_cellvoltage > 2800) {  //Value is low when unavailable
+    datalayer.battery.status.cell_max_voltage_mV = max_cellvoltage;
+  }
 
-  datalayer.battery.info.max_design_voltage_dV = sys_voltageMaxCharge;
-
-  datalayer.battery.status.cell_max_voltage_mV = max_cellvoltage;
-
-  datalayer.battery.status.cell_min_voltage_mV = min_cellvoltage;
+  if (min_cellvoltage > 2800) {  //Value is low when unavailable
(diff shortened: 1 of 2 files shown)


Simply-Johann:
Dala, you're the best!!

I was expecting that after the battery protocol update I'd get the old (non-functional) inverter protocol back... But no, after the OTA update, the inverter started up perfectly, and the SOC and Min/Max CV are also real. Only the "Battery Undervoltage"-event is still there – it probably has the same cause as the previous Min/Max CV – but it doesn't bother me!

Good luck with the renovation!
Best regards from Germany

Now_OK

PR #2021, opened 2026-02-08
Feature: Add more HIA4V1precharge configuration options

dalathegreat:
### What
This PR implements more configuration options for HIA4V1 precharge board

### Why
Some MEB packs would not work without further tuning

### How
Three new parameters added to the Webserver, with the following defaults:
- uint16_t Precharge_min_PWM_Freq = 5000;
- uint16_t Precharge_default_PWM_Freq = 11000;
- uint8_t Precharge_coarse_narrow_multiplier = 3;

> [!TIP]


dalathegreat:
Should we develop better hardware for the precharge instead? User reports on broken BMS due to faulty H1V4AI boards

PR #2048, opened 2026-02-20
Add support for DYNESS Tower

tekampe-oss:
### What
This PR implements support of the DYNESS Tower based on Pylon

### Why
To support new BMS

### How
Implemented by using Pylon as base and adapt some information based on CAN-logs.

image


dalathegreat:
Hi @tekampe-oss , nice work getting the Dyness to work!

Since we are a bit low on flash memory, would it be possible to instead modify the Pylon integration directly, and just add the Dyness optionals there? Or how much changes were needed?

tekampe-oss:
Hi @dalathegreat ,

I think it is possible ... let me try.

dalathegreat:
I think this PR needs to be opened from scratch, since it is now modifying 17 files!

tekampe-oss:
Will create a new one

PR #2081, opened 2026-02-28
Add simple CT for Chademo

drc38:
### What
Replaces the CAN current and voltage shunt with a simple CT clamp measurement for current . Assumes voltage is the target voltage from the vehicle.

### Why
CAN shunts are difficult to source and costly in comparison to a CT clamp

### How
Uses the CHADEMO_CT build flag to swap out sensors

> [!TIP]

### DIFF
--- a/Software/src/battery/CHADEMO-CT.cpp
+++ b/Software/src/battery/CHADEMO-CT.cpp
@@ -21,17 +21,17 @@ typedef enum { ADC_0db = 0, ADC_2_5db, ADC_6db, ADC_11db } adc_attenuation_t;
 uint16_t ct_clamp_offset_mV = 0;
 uint16_t ct_clamp_nominal_voltage_dV = 40;
 uint16_t ct_clamp_nominal_current_A = 100;
-adc_attenuation_enum  ct_clamp_pin_atten = adc_attenuation_enum::ADC_11db;
+adc_attenuation_enum ct_clamp_pin_atten = adc_attenuation_enum::ADC_11db;
 extern const char* name_for_adc_attenuation(adc_attenuation_enum type) {
   switch (type) {
     case adc_attenuation_enum::ADC_0db:
-      return "a) 0dB 0-950mV";
+      return "a)   0dB 0-950mV";
     case adc_attenuation_enum::ADC_2_5db:
       return "b) 2.5dB 0-1250mV";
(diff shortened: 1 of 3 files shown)


drc38:
Appreciate views on whether this is the best way to implement a CT clamp, the risk of not directly measuring the voltage seems low as it is being measured by both the bms and inverter.

platformio.ini
```diff
@@ -84,33 +66,11 @@ build_flags =
     -Wimplicit-fallthrough
     -Wextra
     -Wall
+    ;-D CHADEMO_CT            ; uncomment this flag to use CT clamp support
```
dalathegreat: Suggestion to instead have CT clamp as an option under the Shunt config, instead of being an IFDEF

Image

Alternatively it's own entry entirely in this "Optional components config"

drc38: Thanks, where in the codebase is the optional components config done?

dalathegreat: See this PR for an example: https://github.com/dalathegreat/Battery-Emulator/pull/1668

drc38: I've updated it to use the optional shunt config, but I seem to have broken the saving?! Appreciate any pointers on what I've missed.

drc38: Figured it out, there was an issue with the legacy code

drc38:
Should now build for all boards and pass test suite

drc38:
> Nice work, and thanks for fixing the Shunt legacy issue!

No problem, a while ago I asked in the wiki if anyone had some modbus logs for the Gen24 inverter and can logs for the chademo interface as I plan to simulate the start / charge / discharge / discharge process (using a separate Lily 485) prior to connecting everything up to the real inverter and my leaf. I don't suppose you have any you can share?

dalathegreat:
@drc38 
- Modbus logs: https://gitlab.com/pelle8/inverter_resources/
- Chademo logs: https://github.com/dalathegreat/EV-CANlogs/tree/main/Nissan%20LEAF/V2X

drc38:
> @drc38
> 
> * Modbus logs: https://gitlab.com/pelle8/inverter_resources/
> * Chademo logs: https://github.com/dalathegreat/EV-CANlogs/tree/main/Nissan%20LEAF/V2X

Awesome, thanks! 👍

PR #2087, opened 2026-03-01
MQTT payload improvements

mkunsch:
### What
MQTT payload Improvements
1. Added "emulator_uptime" (in seconds) to the BE/info topic
2. "cpu_temp" moved from batterySensorConfigTemplate to globalSensorConfigTemplate

### Why
1. To allow implementation of an application-level watchdog at the EMS/SCADA level;
    the Battery Emulator now publishes: { ..., "emulator_uptime": <seconds> }
    within the BE/info MQTT topic.
2. "cpu_temp" parameter has been relocated from batterySensorConfigTemplate to globalSensorConfigTemplate, as during 
    a BMS CAN communication failure, data defined in batterySensorConfigTemplate is not transmitted, resulting in the cpu_temp
    value not being updated (tested on BMWi3).

### How
changes in mqtt.cpp, code memory increased by 244 bytes

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -71,7 +71,7 @@ jobs:
       run: pio run -e ${{ matrix.pio_env }}
 
     - name: Upload artifact for ${{ matrix.board_name }}
-      uses: actions/upload-artifact@v6
+      uses: actions/upload-artifact@v7
       with:
         name: ${{ matrix.artifact_name }}.bin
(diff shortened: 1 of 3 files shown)


fastbike:
I agree this refactoring is needed as the existing  template mixes battery info with system info. E.g. CPU temp appears twice even though it is a system variable.
Ideally the output should be refactored to support an array of batteries as triple battery setups are on the development path.

Issue #1868, opened 2025-12-23
[BUG] MQTT values corrupted with double battery


fastbike:
**Describe the bug**
Values for "info" sent by MQTT for double battery are corrupted at the end of the string

**To Reproduce**
Requires a double battery installation. 
Using MQTT Explorer, copy the value from the "info" attribute
Paste it into a JSON validator such as jsonlint.com

**Expected behavior**
A valid JSON packet is present

**Screenshots**
Here is the value that was copied
`{"bms_status":"ACTIVE","pause_status":"RUNNING","SOC":63.38,"SOC_real":56.2,"state_of_health":88,"temperature_min":18,"temperature_max":19.6,"cpu_temp":42.22222,"stat_batt_power":-4375,"battery_current":-12.5,"battery_voltage":357,"cell_max_voltage":3.732,"cell_min_voltage":3.715,"cell_voltage_delta":17,"total_capacity":33880,"remaining_capacity_real":17787,"remaining_capacity":27914,"max_discharge_power":125000,"max_charge_power":68000,"balancing_active_cells":43,"balancing_status":"Unknown","SOC_2":0,"SOC_real_2":59.1,"state_of_health_2":87,"temperature_min_2":18.3,"temperature_max_2":20,"cpu_temp_2":42.22222,"stat_batt_power_2":-4375,"battery_current_2":-12.5,"battery_voltage_2":357,"cell_max_voltage_2":3.734,"cell_min_voltage_2":3.717,"cell_voltage_delta_2":17,"total_capacity_2":33495,"remaining_capacity_real_2":18480,"remaining_capacity_2":13957,"max_discharge_power_2":125000,"max_charge_power_2":63000,"balancing_active_cells_2":31,"balancing_status_2":"Unknown","event_level":"EVENT_LEVEL_DEBUG","emulatoH�?`

<img width="828" height="96" alt="Image" src="https://github.com/user-attachments/assets/bb0d54e1-bdf2-4d38-9cef-6994597ae7f7" />

**Version and settings (please complete the following information):**
- Please attach screenshot of configuration for easier debugging
- Software version: 9.3.2
- Battery used: Double Nissan LEAF (2x 40kWh)
- Inverter communication protocol: BYD Battery-Box Premium HVS over CAN Bus 
- Hardware used for Battery-Emulator: Stark CMR
- GPIO controlled contactors: yes
- MQTT: yes

fastbike:
It is a classic buffer overrun issue !
[...]

PR #2092, opened 2026-03-02
MQTT: Remove decimals from CPU temperature to save space

dalathegreat:
### What
This PR removes the decimals on CPU temperature

### Why
MQTT buffer overflow could be observed in #1868

### How
This is not a proper fix, but a quickfix until MQTT can be refactored further

> [!TIP]


mkunsch:
Correct. Decimals in CPU temperature readings have no real meaning due to the limited precision of the internal temperature sensor.

PR #2094, opened 2026-03-02
Afore: Add LFP cellvoltage spoofing

dalathegreat:
### What
This PR implements cellvoltage spoofing for Afore CAN protocl

### Why
When using Afore CAN, we notice odd charge limits. This might be due to LFP chemistry always being assumed, and if NCM is used we will overvoltage and get limited. Working theory.

### How
We switch to using the same LFP spoofing mechanism as Ferroamp is using!

> [!TIP]


dalathegreat:
Merging this fix, this solved the problem noted on Afore protocol with the low charging speed :partying_face: 

image

PR #2101, opened 2026-03-04
BYD: Elevated diagnostic session + Calibration feature

### LINKED ISSUE #1673: Issue with the state of charge (SOC) in a BYD battery.
I need help diagnosing an issue with the battery's state of charge (SOC). Initially, everything worked smoothly: the battery would charge up to 100%, and discharge gradually from 100% without any sudden jumps in SOC.

However, after about a month, I started noticing abnormal jumps in the SOC. It would suddenly jump from 90% to 100%, and then begin to drop. As of today, the SOC has dropped to 66%.

It’s also worth noting that at 66% SOC, the battery voltage reaches the maximum configured limit of 440V

- Battery used: [BYD 60k]
- Inverter communication protocol: [ Deye]

dalathegreat:
### What
This PR implements the seed-key challenge for elevated BYD commands, the first being Calibrate SOC & AH

### Why
To be able to recalibrate SOC in order to gain lost capacity. Example, if user is not able to charge to 100% anymore, you can re-calibrate the BMS this way. Fixes #1673

### How
Calibration data can be set via the "More Battery Info" page. After configuring the target SOC & AH, the user can press the button "Calibrate SOC" to start the process.

image

Bonus;  the following info is also added to the More Battery Info page
- Charging battery state:
- Min cell voltage number:
- Max cell voltage number:
- Min temp module number:
- Max temp module number:

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -637,8 +637,14 @@ void BydAttoBattery::transmit_can(unsigned long currentMillis) {
         break;
       case RUNNING_STEP_3:
         // WriteDataByIdentifier dataIdentifier=1F FC (calibrate SOC), data = 10 27 98 3A
-        ATTO_3_7E7_RESET_SOC.data = {0x07, 0x2E, 0x1F, 0xFC,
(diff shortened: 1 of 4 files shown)


dalathegreat:
Please log a Github issue, with Events, Screenshots from BE, etc. Impossible to say why charging stopped otherwise.

It is very bad practice to comment your issues on a pull request for a feature. Please never do this again.

PR #2115, opened 2026-03-07
Auto CT offset and invert CT value

drc38:
### What
A few usability improvements to CT clamp implementation.

### Why
Simplifies setup of the CT clamp

### How
Adds auto offset option and invert CT clamp values checkbox

> [!TIP]

### DIFF
--- a/Software/src/battery/CHADEMO-BATTERY.cpp
+++ b/Software/src/battery/CHADEMO-BATTERY.cpp
@@ -5,8 +5,8 @@
 #include "CHADEMO-SHUNTS.h"
 #include "Shunt.h"
 
-uint16_t (*get_measured_current_ptr)();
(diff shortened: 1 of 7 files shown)


drc38:
Just realised the current is being returned unsigned, how is charge/discharge being determined? Would it be useful for the CT to also track current flow direction?
```
uint16_t get_measured_current() {
  return (uint16_t)Amperes;
}
```

dalathegreat:
If possible that would be great. //Negative value = Discharging , Positive value = Charging, on the current_dA parameters

drc38:
> If possible that would be great. //Negative value = Discharging , Positive value = Charging, on the current_dA parameters

ok great, I see at the moment the CT current is not even mapped to `current_dA` in `void ChademoBattery::update_values()` for the Chademo battery

PR #2120, opened 2026-03-09
Added support for Dyness Tower

tekampe-oss:
### What
This PR implements support of the DYNESS Tower based on Pylon

### Why
To support new BMS

### How
Implemented by using Pylon as base and adapt some information based on CAN-logs.
image

### Open Issues:
* cell voltages might be off
* contactor is not supported (does not know the correct can signal)


dalathegreat:
I think we can close this, as people are successfully usingn Dyness with the standard pylon protocol https://github.com/dalathegreat/Battery-Emulator/discussions/2083

tekampe-oss:
It added support for Cell-Voltages, total and remainig capacity for dyness towers. These informations are in differend can messages than the standard pylon protocol.

dalathegreat (approved):

PR #2124, opened 2026-03-10
Chademo improvements

drc38:
### What
Fixes can bus message reading to match Big Endian in dbc QC file, 
adds extra status reporting,
refines sequence logic, and
improves logging to avoid spamming log with repeat messages
switches to user defined maximum charging power, current and voltage

### Why
Why does it do it?

### How
How does it do it?

> [!TIP]


drc38:
Hi Dala, I'm assuming the dbc files are correct and the emulator should be using big endian? 

Also is the factor on rated battery capacity a typo? 0.01, 0.1 or is it actually 0.11
image

dalathegreat:
No don't assume the DBC is correct :D crosscheck with some log files that capacity seems OK, and also byteorder.

drc38:
Thanks, are you aware of another other QC log file in addition to the link you sent previously, it would be good to have two files to verify with

drc38:
I rechecked the `v2h_startup__charge_ffrom_solar_excess_for_a_while_then_turn_on_oven_to_create_load_then_switch_off_load__then_end_session` log with savvycan and the QC dbc files. If I swap the byte order to little endian (Intel) as per original emulator code in the dbc file the can messages are not logical eg available output voltage of 5V. Do you know what the rated battery capacity was on the car? As that is the only one that looks off in the dbc file, looks to me like the factor should be 1 and it reports back Wh.

Intel (little endian)
image

Motorola (big endian)
image

dalathegreat:
Here is from another source, content from my Kempower charger when I charge my Nissan LEAF, maybe helps?

publish: { "content": "carData", "data": { "maxBatteryVoltage": 435, "maxChargeTimeSec": 255, "maxChargeTimeMin": 96, "targetBatteryVoltage": 410, "chargedRateRefConstant": 100, "estChargeTime": 0, "protocolNumber": 2, "chargingCurrentRequest": 1 "faultFlag": "0x00", "statusFlag": "0x81", "chargedRate": 68 }}

{ "content": "chargerData", "data": { "limitVolt": 435, "limitTime": 5760, "h109MsgSendPending": 0, "state": 21, "protocolNumber": 2, "startAllowed": false, "logicalDiscrepancy": false, "carNotDetected": false, "normalStop": false, "emergencyStop": false, "internalFault": false, "stopReason": "none", "canSendingEnabled": true, "status": "PRECHARGE" }}

drc38:
Will leave this PR draft until I can connect emulator to my car and directly log the can bus with the emulator or find another QC bus log file to verify changes with 😉

drc38:
Hi @dalathegreat I'm trying to understand how the resume charging works after pressing the pause charge button, how do the max charge/discharge values get reset to prior to the pause?

dalathegreat:
It signals a pause requestin power usage by setting both the maximum and minimum charge values to zero, thereby preventing any inverter operation. After pause is lifted, battery set values can again pass thru

drc38:
> It signals a pause requestin power usage by setting both the maximum and minimum charge values to zero, thereby preventing any inverter operation. After pause is lifted, battery set values can again pass thru

Thanks, I've found and fixed the issue with values not coming back through when unpausing I hope.

drc38:
Hi @dalathegreat, to use the CT clamp reliably requires two ADC pins. From what I've read ADC2 should be avoided due to the wifi use, which will require using IO4 which is being used for pin4. How likely is it anyone is using the chademo implementation and could be impacted by any pin change?

dalathegreat:
Hi @drc38 , I have never heard or seen anyone use it, so feel free to change pins as you please!

drc38:
Hi @dalathegreat, I have the unit to the point where it can log the QC can bus (but a few odd things happening including the charge lock toggling on/off). Let me know if you want a copy of the logs uploaded somewhere. The capacity factor needs to be 0.0005 to match my 30kWh leaf, interestingly the logged messages start at 30kWh and then drop the value to 23kWh so assume it is taking into account the SOH.

dalathegreat:
Nice! I don't have much input on the Chademo implementation at all, but nice to hear there is progress :zap:

drc38:
Hi @dalathegreat, without connecting to an inverter the emulator now gets to the point the contactors are opened and voltage is present at the emulator fuse terminals. Yay, progress! Without a voltage measurement the TargetBatteryVoltage is used, interestingly this is about 20V higher than that measured (410V vs 393V measured with voltmeter which is also the same as that reported by OVMS), which is odd as it is provided by the car. The can messaging dbc has been corrected from motorola to intel and the dbc updated here https://github.com/dalathegreat/leaf_can_bus_messages/pull/14 . Appreciate your review for further improvements or understanding, the x200 MinimumBatteryDischargeLevel and MaxRemainingCapacityForCharging (150 when SOC is 98%) still don't make much sense.

drc38:
I noticed when using the can log feature it can be a bit hit and miss as to what is captured, can the buffer be increased to capture a larger log sequence?

dalathegreat:
@drc38 , CAN logging via webserver will be worked on later this year. For now if you need reliable logging, USB is the way to go! https://github.com/dalathegreat/Battery-Emulator/wiki/CAN-logging#usb-can-logging

PR #2134, opened 2026-03-13
BYD: Refactor BYD handling to be pack agnostic

dalathegreat:
### What
This PR refactors the BYD handler to become pack agnostic

### Why
The BYD Atto 3 implementation is starting to become used by many different BYD packs. The integration was never designed to be used by anything other than Atto3 (Standard/Extended 50/60kWh), so when we use other BYD batteries the amount of cells, voltage limits, all break.

### How
We fix this by waiting 60 seconds after bootup, to count how many cells are detected. User can track this progress by looking at the "More Battery Info" page. Based on this cell count, we set pack voltage limits.

image

At the same time, this PR removes lots of older Atto3 functionality:
- No more estimated SOC based on voltage (Since we will support so many various batteries, we do not have space to make specific estimation tables for all various BYD packs. Removed all traces of SOC estimation
- We can now unlock batteries, and recalibrate SOC, so this old estimate is no longer needed.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -179,7 +179,7 @@ void BydAttoBattery::
 
   //After some time has passed after startup, we assume we have read all cellvoltages at least once, so we can trust the cell count and calculated design voltage limits.
   //Before that, we keep the limits wide to avoid cutting off the battery in case cell count is not read correctly yet.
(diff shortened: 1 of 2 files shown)


TheOfficialBatteryMan:
Detected 126 cells correctly for me as quick as I could load the page after upload

dalathegreat:
Merging, tested OK also with a 172S BYD battery! Wow :partying_face:

PR #2149, opened 2026-03-18
Support for external WS2812 programmable LED with GRB color order for HW_LILYGO2CAN

mkunsch:
### What
This PR implements support for an external WS2812 programmable LED with GRB color order for HW_LILYGO2CAN. It is configurable via the Status LED pattern, with new GRB LED modes added.
There are several WS2812 variants with different packages and data transmission orders (GRB or RGB) that can be used as an external LED for LILYGO2CAN.

### Why
The LilyGO T-485CAN board uses a WS2812B LED with RGB data order for BE status indication.
There is no Status LED on the LilyGO T-2CAN, so an external WS2812-compatible programmable LED can be connected (pin IO35).
Using a WS2812D LED (THT package) with GRB color order results in the green and red colors being swapped. Additional configuration is required to correct this color mismatch.

### How
Implemented for HW_LILYGO2CAN: the enum led_mode_enum { CLASSIC, FLOW, HEARTBEAT, GRB_CLASSIC, GRB_FLOW, GRB_HEARTBEAT } has been extended to address color swapping. Corresponding led_mode strings — {3, "GRB Classic"}, {4, "GRB Energy Flow"}, {5, "GRB Heartbeat"} — have been added for the Status LED pattern and are configurable via the web interface.
(The implementation is minimalistic with performance in mind.)
Tested on the LilyGO T-2CAN with the Worldsemi WS2812D-F5, which uses GRB color order.

> [!TIP]

### DIFF
--- a/Software/src/lib/adafruit-Adafruit_NeoPixel/Adafruit_NeoPixel.cpp
+++ b/Software/src/lib/adafruit-Adafruit_NeoPixel/Adafruit_NeoPixel.cpp
@@ -24,9 +24,11 @@ void Adafruit_NeoPixel::setPin(int16_t p) {
   }
 }
 
+#ifdef HW_LILYGO2CAN
 void Adafruit_NeoPixel::setColorOrder(uint32_t o) {
   color_order = o;
 }
+#endif
 
 void Adafruit_NeoPixel::show(void) {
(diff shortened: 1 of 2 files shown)


mkunsch:
Thank You.

PR #2151, opened 2026-03-19
Solax: min/max cell voltage fix/rescale for IE102 error

jonny5532:
### What
The latest Solax firmware is causing Bat Mismatch IE102 errors.

From a bit of Ghidra-ing, this seems to be due to the CAN-reported cell voltages in 1874 not meeting its requirements. They are actually expected in dV, not mV, and also to fit the LFP range (2.9 to 3.6V).

This fix the reporting of these voltages and scales if necessary.

> [!TIP]

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -49,10 +49,33 @@ void SolaxInverter::
   SOLAX_1874.data.u8[1] = (datalayer.battery.status.temperature_max_dC >> 8);
   SOLAX_1874.data.u8[2] = (int8_t)datalayer.battery.status.temperature_min_dC;
   SOLAX_1874.data.u8[3] = (datalayer.battery.status.temperature_min_dC >> 8);
-  SOLAX_1874.data.u8[4] = (uint8_t)(datalayer.battery.info.max_cell_voltage_mV);
-  SOLAX_1874.data.u8[5] = (datalayer.battery.info.max_cell_voltage_mV >> 8);
-  SOLAX_1874.data.u8[6] = (uint8_t)(datalayer.battery.status.cell_min_voltage_mV);
-  SOLAX_1874.data.u8[7] = (datalayer.battery.status.cell_min_voltage_mV >> 8);
+
+  int32_t cell_max_voltage_mV = datalayer.battery.status.cell_max_voltage_mV;
+  int32_t cell_min_voltage_mV = datalayer.battery.status.cell_min_voltage_mV;
+
+  // Fake values during startup?
+  if (cell_max_voltage_mV == 0) {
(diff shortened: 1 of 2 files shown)


Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -49,10 +49,25 @@ void SolaxInverter::
...
-  SOLAX_1874.data.u8[4] = (uint8_t)(datalayer.battery.info.max_cell_voltage_mV);
-  SOLAX_1874.data.u8[5] = (datalayer.battery.info.max_cell_voltage_mV >> 8);
-  SOLAX_1874.data.u8[6] = (uint8_t)(datalayer.battery.status.cell_min_voltage_mV);
-  SOLAX_1874.data.u8[7] = (datalayer.battery.status.cell_min_voltage_mV >> 8);
+
+  uint32_t cell_max_voltage_mV = datalayer.battery.status.cell_max_voltage_mV;
+  uint32_t cell_min_voltage_mV = datalayer.battery.status.cell_min_voltage_mV;
+
+  // Rescale to the range 2.9->3.6V
+  cell_max_voltage_mV =
```
dalathegreat, line 65: Consider duplicating this function: https://github.com/dalathegreat/Battery-Emulator/blob/85acfad961d77906bef6dfabd872bb19930fc0e5/Software/src/inverter/FERROAMP-CAN.cpp#L13

It checks if we are not on LFP, and only then re-scales the voltage range.

Even better, throw it in the Common Functions, and point all protocols that do LFP rescaling to use the same function. (Foxess, Ferroamp, Solax)

jonny5532: Definitely makes sense to put it in common-functions.

Re: rescaling, my thought was to use the integration min/max, even for LFP, since the Solax won't let you charge LFP above 3.6V, which prevents, eg, triggering BYD balancing (which is believed to happen at 3.65V).

But could limit to non-LFP only.

dalathegreat (approved):
Massive thanks for investigating and fixing this @jonny5532 🙌

jonny5532:
To add to the initial description - the actual problem wasn't the cellvoltages (the firmware check isn't actually working), but the battery type - decimal 80 (the BE default) is not a valid battery type - 81 is the new default.

roy13:
Hi this change seems to have affected my existing setup running maybe older firmware, when i switch to 10.8 my Solax reports Error code 6 Battery voltage exceeding the normal range of allowable operation.
I'm using it with a Tesla NCA

dalathegreat:
Hi @roy13 , we updated the Wiki on this with info on how to do the new configuration https://github.com/dalathegreat/Battery-Emulator/wiki/Inverter:-Solax#battery-type-information-pre-2026

PR #2177, opened 2026-03-29
Fronius: Use actual battery voltage for calculating limits

dalathegreat:
### What
This PR makes the Fronius use battery voltage instead of theoretical max voltage to calculate charge/discharge limits

### Why
Old implementation could lead to confusing state, where selecting 20A would still allow inverter to charge with 22A. First when going down to 19A a change was registered on the inverter side.

### How
We switch to using actual battery voltage for charge limit calculation instead of theoretical max pack voltage limit

> [!TIP]


dalathegreat:
Tested OK on the Discord server :raised_hands:

PR #2181, opened 2026-03-30
Feature: Double BOLT-AMPERA battery support 🔋🔋

vitalii-ch:
### What
This PR implements support for double Bolt/Ampera batteries.

### Why
To allow using two Bolt/Ampera battery packs in a single Battery-Emulator setup. Personal need — running two Bolt/Ampera packs in my setup.

### How
Same approach as double-ECMP/double-Kia64. Refactored BoltAmperaBattery to use instance pointers (datalayer_battery, datalayer_boltampera, allows_contactor_closing) instead of directly accessing globals.
Added a second constructor that accepts datalayer_battery2 and datalayer_extended.boltampera_2 for the second battery instance. Registered the BoltAmpera case in BATTERIES.cpp for battery2 creation.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -38,7 +38,7 @@
 #endif
 
 // The current software version, shown on webserver
-const char* version_number = "10.4.0_BDB_v1";
+const char* version_number = "10.4.0";
 
 // Interval timers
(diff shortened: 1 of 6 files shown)


dalathegreat (approved):
Nice work!

PR #2184, opened 2026-03-30
Fix voltage drift timeout never triggering for battery2/battery3

vitalii-ch:
### What

This PR fixes the voltage drift timeout for battery2/battery3 in multi-battery setups and extracts the logic into a testable core module.

### Why

The `secondsOutOfVoltageSyncBattery2` and `secondsOutOfVoltageSyncBattery3` counters were declared as local variables inside `check_interconnect_available()`.
Since this function is called once per second, the counters reset to 0 on every invocation. The 10-second timeout condition (`counter >= 10`) could never be reached, meaning battery2/battery3 contactors would never open on voltage drift.

### How
- Mark both counters as `static` so they persist between calls
- Extract the function from `Software.cpp` into a new `Software/src/core/parallel_safety.h/.cpp` module
- Rename `check_interconnect_available()` to `check_parallel_battery_safety()` for clarity and to accommodate future safety checks (temperature, bms_status)
- Add `test/voltage_sync_tests.cpp` with 6 gtest cases covering:  timeout triggering, re-sync recovery, battery3 parity, battery1 fault propagation, and zero-voltage early return 

> [!TIP]


Software/src/core/interconnect.cpp
```diff
@@ -0,0 +1,64 @@
+#include "interconnect.h"
+#include "../datalayer/datalayer.h"
+#include "../devboard/utils/events.h"
+
+void check_interconnect_available(uint8_t batteryNumber) {
```
dalathegreat: This whole file can be deleted, check_interconnect_available is never called anymore, since you added a check_parallel_battery_safety function

vitalii-ch: You're right, thanks for catching that. `interconnect.cpp` is a leftover from an earlier iteration — I renamed the files but didn't clean up before pushing. Will delete it.

vitalii-ch: Done

PR #2185, opened 2026-03-31
New battery :battery: Add initial support for Microvast battery

dalathegreat:
### What
This PR implements initial support for Microvast batteries

### Why
Requested in https://github.com/dalathegreat/Battery-Emulator/discussions/2182

### How
New option added to enable support for Microvast

> [!TIP]

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -100,6 +100,8 @@ const char* name_for_battery_type(BatteryType type) {
 #endif
     case BatteryType::MgHsPhev:
       return MgHsPHEVBattery::Name;
+    case BatteryType::MicrovastBattery:
+      return MicrovastBattery::Name;
     case BatteryType::NissanLeaf:
       return NissanLeafBattery::Name;
     case BatteryType::Pylon:
@@ -219,6 +221,8 @@ Battery* create_battery(BatteryType type) {
 #endif
     case BatteryType::MgHsPhev:
       return new MgHsPHEVBattery();
+    case BatteryType::MicrovastBattery:
+      return new MicrovastBattery();
     case BatteryType::NissanLeaf:
       return new NissanLeafBattery();
     case BatteryType::Pylon:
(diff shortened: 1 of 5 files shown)


jasonday171973-eng:
I'm getting an error during the compile
I'm compiling for a lillygo t2can

image

dalathegreat:
Hi @jasonday171973-eng , you should **not** compile, you should just download the precompiled file and update to it.

Direct link: https://github.com/dalathegreat/Battery-Emulator/actions/runs/23797237727/artifacts/6199172042
Instructions: https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#downloading-a-pull-request-build-to-test-locally-

jasonday171973-eng:
I would say I'm an idiot. but I'm also sleep deprived.

I've now applied the bin file and here is a log.

Thanks for your speedy response

image


[canlog_05-22-43.txt](https://github.com/user-attachments/files/26388205/canlog_05-22-43.txt)

dalathegreat:
Looks like the battery is in some sleep mode, or we are not sending the correct info towards it. It is only repeating one message over and over again (1AAAF303). The message it sends is NOT in the DBC file.

CAN logs from a fully functional vehicle using this battery may be required, or lots of tinkering with the VCU messages to get it to wake up. I estimate a 5-10hours is required to fully exhaust the DBC possibilities, and there is no guarantee that it will work. So CAN logs from a vehicle with same battery would be next step.

jasonday171973-eng:
I've been provided with a couple more bdu files that are related to this battery. single string would be most applicable 

[MV_Multi string CAN_Protocol V1.5_20221027.dbc.txt](https://github.com/user-attachments/files/26420432/MV_Multi.string.CAN_Protocol.V1.5_20221027.dbc.txt)
[MV_SingleString_CAN_V1.5_3(V1.7)+20230213.dbc.txt](https://github.com/user-attachments/files/26420434/MV_SingleString_CAN_V1.5_3.V1.7.%2B20230213.dbc.txt)

jasonday171973-eng:
I have 2 other batteries.
1 reads 
1AAAF302
the other 
1AAAF301

So I'm guessing the battery is announcing it's uniq ID/serial number.
These are deisgned to be used in parallel with other batteries and then connect to an external contactor.
I'm also guessing it's expecting a response before it sends more data.

As there is no contactor, the battery is live and can feed power and be charged.
I've emailed Microvast for more information.

PR #2192, opened 2026-04-03
solax: fix idle mode

frenetic1:
### What
Fixes an idle mode problem introduced by https://github.com/dalathegreat/Battery-Emulator/pull/1679

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -16,9 +16,8 @@ void SolaxInverter::
   // If not receiveing any communication from the inverter, open contactors and return to battery announce state
   if (millis() - LastFrameTime >= INTERVAL_2_S) {
     if (!configured_ignore_contactors) {
-      datalayer.system.status.inverter_allows_contactor_closing = false;
+      STATE = BATTERY_ANNOUNCE;
     }
-    STATE = BATTERY_ANNOUNCE;
   }
   //Calculate the required values
   temperature_average =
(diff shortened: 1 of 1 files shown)


jonny5532:
With a contactors-stay-closed battery, I got this to work once - where the battery happened to start up slowly, so during the Solax's ANNOUNCE phase there was a brief moment with the contactors open which was enough for the Solax to proceed to the next state. However, subsequently, it doesn't work - the contactors remain closed, and Solax stays in the initial ANNOUNCE phase.

The issue is that for contactors-stay-closed batteries, it can't start in the ANNOUNCE state, as the Solax will wait forever for the contactors to be open (for a weld check) and they won't, hence the ignore-contactors checkbox which skipped this state.

I think we therefore need to handle both cases separately - an 'ignore contactors' mode to make the Solax skip the weld check (for contactors-stay-closed batteries), and a delayed-close for your Tesla double-battery situation.

We could just add another checkbox or a dropdown, although it would be nicer if this was built in - since batteries generally know if they will be able to open contactors or not, which the Solax integration can take heed of, to avoid people having to understand the subtleties of Solax inverters...

frenetic1:
Hmm, yes, the implementation will only work for batteries that honor inverter_allows_contactor_closing at least on boot. I didn't expect that most battery types are not using this at all. Currently only those are using it:
BMW-IX-BATTERY
BMW-SBOX
KIA-HYUNDAI-64-BATTERY
MG-5-BATTERY
MG-HS-PHEV-BATTERY
RIVIAN-BATTERY
TESLA-BATTERY
TESLA-LEGACY-BATTERY


Maybe it's good idea to introduce a battery property that tells us how contactors can be controlled by the inverter. Something like
CONTROL_NONE,
CONTROL_ONCE,
CONTROL_FULL

Then we could completely get rid of the user menu. @dalathegreat : What do you think?

Btw. @jonny5532: what does your solax do when there's no power demand and it wants to go to standby but we ignore its contactors open request and always keep answering contactors closed? Will it go into standby?

dalathegreat:
I personally think a "Delayed contactor closing" checkbox might be the best way forward

dalathegreat:
FYI @frenetic1 , I reverted PR https://github.com/dalathegreat/Battery-Emulator/pull/2205 , so that we can release 10.5.0. Please re-add the changes into this PR!

frenetic1:
New PR is up: https://github.com/dalathegreat/Battery-Emulator/pull/2211

PR #2209, opened 2026-04-07
perf: Zero-RAM WebServer optimizations & CAN buffer stability

DevNar:
This PR significantly reduces heap memory usage, prevents memory fragmentation crashes (`LoadProhibited` core panics) when accessing heavy web pages, and improves overall CAN bus stability.

### 1. WebServer Memory Fixes (Zero-RAM HTML)
* **Removed massive RAM allocations:** Eliminated `String.reserve()` and massive string building in `/canlog`, `/canreplay`, and `/events`.
* **Implemented `PROGMEM` & `AsyncResponseStream`:** Heavy HTML content is now sent directly from Flash, and the `/events` page dynamically streams content chunk-by-chunk.
* **UI/UX Refactoring:** Extracted inline C++ colors into dedicated CSS classes (`row-warning`, `row-error`). Maintained the project's native Dark Mode aesthetic while drastically improving mobile responsiveness (transforming the Events grid into neat cards for mobile screens).

### 2. Isolated CAN Streaming
* Added the `is_can_streaming` flag in `comm_can.cpp` to properly separate the new real-time web streaming logic from the legacy RAM-based `can_logging_active` flag. 
* This prevents wasted CPU cycles and unexpected string buffering when users access the `/canreplay` page or during normal background CAN operations.
* Handled page unloading gracefully via JS `navigator.sendBeacon()` to ensure the stream connection closes cleanly when leaving the page.

### 3. CAN Driver Stability
* While refactoring `comm_can.cpp` for the streaming output, I updated the default CAN buffers to handle higher bus loads safely.
* Increased `gBuffer.initWithSize` from `25` to `128`.
* Explicitly set `mDriverReceiveBufferSize` and `mDriverTransmitBufferSize` to `256` to prevent CAN frame drops during heavy traffic.

**Files modified:**
- `can_logging_html.h` / `.cpp`
- `can_replay_html.h` / `.cpp`
- `events_html.h` / `.cpp`
- `webserver.cpp`
- `comm_can.cpp`


dalathegreat:
Hi @DevNar 
Thanks for the PR, this is quite a lot to cover, and it looks like AI was heavily involved.

- I am uneasy about increasing the buffer size, especially since the tx/rx is all in one thread. Just ballooning the buffer will hide performance issues, some integrations require tight CAN loops, and with a large buffer you hide any traces of things taking too long to execute
- Events page changes seem a bit arbitrary, not sure what the end result really benefits from
- The streaming CAN log thing is a good idea, the ongoing PR for a new webserver does the same, see https://github.com/dalathegreat/Battery-Emulator/pull/1544
- The number1 change also seems quite OK, the new webserver we are working on kinda achieves the same.

Would it be possible to still split this PR up further? I would not really want to do the Change number3; CAN buffer size changes, as this has caused issues in the past.

DevNar:
Hi @dalathegreat,

Thanks for taking the time to review! You're right, I used AI to help
refactor the repetitive HTML/CSS payloads quickly, but the architectural
changes and memory issues were thoroughly tested on my actual hardware to
fix real crashes.

To address your concerns:

1. **CAN Buffer Size:** You are 100% right. Increasing the buffer masks
underlying latency issues. I will revert the changes made to
`mDriverReceiveBufferSize` and `gBuffer` back to their original values to
keep the CAN driver untouched.
2. **Events Page Changes:** The main goal wasn't just UI styling. The
original code used `content += ...` to build a massive String in the heap.
On my board, refreshing the Events page occasionally caused a
`LoadProhibited` core panic due to heap fragmentation. The core of this
change was converting it to use `AsyncResponseStream` (`response->print()`)
to stream the page directly from PROGMEM without allocating large RAM
blocks.
3. **The New Webserver (PR #1544):** I completely missed that you guys are
already working on a major webserver rewrite in #1544! That's awesome.

Since #1544 achieves similar goals, I'm happy to adapt. Your message cut
off at "Would..." at the end — were you going to suggest something?

**How would you like to proceed?** Should I strip this PR down to ONLY the
isolated CAN streaming flag and the zero-RAM Events page (and revert the
buffers)?
Or should we just close this PR entirely?

Let me know what works best for the project's roadmap!

dalathegreat:
Yes please do smaller PRs for separate improvements, and we can better review them!

DevNar:
Understood! I will close this mega-PR and break the improvements down into smaller, focused pull requests (e.g., separating the isolated CAN stream fix from the zero-RAM HTML fixes). I'll also drop the buffer size changes. Thanks for the guidance!

PR #2211, opened 2026-04-07
Feat/solax multiple tesla bats

frenetic1:
### What
This PR is a replacement for https://github.com/dalathegreat/Battery-Emulator/pull/2192

### Why
The old PR breaks other batteries.

### How
Implementing a dropdown with 3 different modes for contactor control.


dalathegreat (approved):
Looks like a good solution to me 👍

PR #2215, opened 2026-04-10
feat(ui): Add next gen frontend (On Testing)

herberthobregon:
# WORK IN PROGRESS
## What

- **Single-page application (SPA)** built with **Lit** (Web Components), replacing HTML generated inside C++ (`processor`, string-built pages, battery-specific HTML snippets).
- **REST-style JSON APIs** for dashboard, settings, battery status, CAN log, events, and “extra” battery data as structured rows—not HTML fragments.
- **Static shell + bundle**: (WIP) the browser loads a minimal HTML shell and one JavaScript bundle; the firmware serves the bundle (precompressed with **gzip**) from the embedded web server alongside the existing routes.
- **Settings and advanced views** (WIP) use the same `POST /saveSettings` and command endpoints as before, but the **UI labels and selects** are driven by data from the device (e.g. `select_options` on `/api/settings`) so enums stay in sync with the firmware.

---

## Why

1. **Maintainability**  
   Mixing UI markup with C++ made every UI tweak a firmware rebuild and encouraged copy-pasted HTML across battery types. Moving the UI to TypeScript/Lit keeps view code in one place and uses normal web tooling (TypeScript, Vite, linting).

2. **Correct separation of concerns**  
   The device should expose **state and actions** (JSON, form posts), not **presentation** (tables, sections, styling). Battery-specific “extra” info is now **label/value rows** (and optional section titles); the SPA decides how to render them. That scales better as more batteries and screens are added.

3. **Bandwidth and flash**  
   Embedded devices benefit from small payloads. The production bundle is roughly **~60 kB raw** and, with **gzip**, on the order of **~17–20 kB** over the wire (exact numbers shift slightly each build). The web server sets **`Content-Encoding: gzip`** so clients download the compressed asset only, which keeps OTA and flash budgets predictable.

4. **Future-proofing**  
   A component-based SPA is easier to extend (new routes, mocks for local dev, i18n, accessibility) without touching the real-time C++ paths except to add or extend APIs.

---

## How

1. **Build**  
   Vite bundles `frontend/src` into `dist/app.bundle.js` (and a tiny `index.html` shell where applicable). A small script emits C++ sources (`app_bundle.cpp` / `app_bundle.h`) that embed the **gzipped** bytes so the firmware can call `request->beginResponse(..., data, len)` with `Content-Encoding: gzip`.

2. **Serve**  
   The web server registers routes for the SPA entry and `/app.bundle.js` (gzip), while authenticated JSON routes power the Lit app. Legacy HTML pages are removed or reduced so there is a single UI story.

3. **Data flow**  
   Lit components call `fetch` (via a thin API layer) → JSON → reactive `@state` → declarative `html` templates. Forms that mirror the old POST body field names preserve compatibility with existing NVM handling in `webserver.cpp`.

---

## Benefits of Lit (LitElement)

- **Small runtime**  
  Lit keeps a minimal and efficient core, reducing bundle size and improving performance.

- **Standards-based**  
  Components are real **custom elements**, compatible with all modern browsers and easy to integrate with vanilla JS or other tools.

- **Declarative templates**  
  `html` tagged templates enable efficient re-rendering and reduce boilerplate compared to manual DOM manipulation.

- **TypeScript-friendly**  
  Strong typing for properties and state, with clear component boundaries, improves safety and readability in large projects.

- **Scoped styles**  
  `static styles` or `adoptedStyleSheets` isolate CSS within the component, preventing global style collisions in multi-screen SPAs.

- **Framework-agnostic**  
  Lit is built on native `HTMLElement` and does not rely on proprietary syntaxes like JSX or custom DSLs. This ensures that JavaScript written today will remain valid and understandable decades into the future, independent of framework trends.

---

## Summary

| Topic              | Outcome                                                                 |
| ------------------ | ----------------------------------------------------------------------- |
| **What**           | Lit SPA + JSON APIs + gzip bundle served from the ESP32 web server.     |
| **Why**            | Easier maintenance, separation of UI vs device logic, small wire size.  |
| **How**            | Vite build → gzipped embed → routes; Lit consumes JSON and forms.       |
| **Lit**            | Lightweight, standards-based components, good DX for a small footprint. |

This architecture trades a bit of upfront tooling for **clear boundaries**, **smaller transfers on the wire**, and a UI that can evolve **without** turning every change into a C++ string-concatenation exercise.


nagyrobi:
https://github.com/dalathegreat/Battery-Emulator/pull/1544

nagyrobi:
Any chance to see a preview of this?

herberthobregon:
Sure! Give a chance this weekend!

ngehrsitz:
We could probably save a few more bytes if we use brotli. There is full support since 2017 so I see no reason to stay with gzip: https://caniuse.com/brotli
Also instead of generating `app_bundle.cpp` we should evaluate putting it into spiffs/littlefs. I suspect the webserver might be better optimized for serving files from there than doing it manually.

dalathegreat:
FYI all here, we will be looking at merging the new webserver in https://github.com/dalathegreat/Battery-Emulator/pull/1544 later in Q3

herberthobregon:
Here are some screenshots and a video of what the frontend looks like! And all in just 40kb of gzipped data.

I think the UI looks quite pro and polished; I hope this is the next-generation UI the project needs!

image
image

https://github.com/user-attachments/assets/03c6048c-9b1c-45ea-ba36-4554c54db94c

please check @dalathegreat and @nagyrobi 

Next steps: **Create a CPP API. Working on 🔨**

Thanks @ngehrsitz I will integrate `brotli` ✅

herberthobregon:
## New Web UI on `lilygo_330` with `-DNEW_WEBUI_ONLY`

### What changes

Adds the `NEW_WEBUI_ONLY` build flag, which flips the strategy on 4 MB devices: instead of excluding the SPA due to limited flash, it excludes the **legacy UI** and embeds the full modern UI (SPA + `/api/*`).

- **`platformio.ini`**: adds `-DNEW_WEBUI_ONLY` to the `lilygo_330` environment.
- **Added `brotli`**
- **`webserver.cpp`**: under the flag, legacy HTML page routes are not registered (`/`, `/settings`, `/advanced`, `/canlog`, `/canreplay`, `/log`, `/cellmonitor`, `/events`, `/clearevents`). Legacy code is not deleted or modified: since we already build with `-ffunction-sections -fdata-sections -Wl,--gc-sections`, once unreferenced the linker automatically strips the processors and their HTML templates.
- **`generate-cpp.mjs` / `api_routes.cpp`**: the bundle guard changes from `#ifndef SMALL_FLASH_DEVICE` to `!defined(SMALL_FLASH_DEVICE) || defined(NEW_WEBUI_ONLY)`, so the SPA is embedded in this mode.
- **`api_routes.cpp`**: `new_webui_enabled_flag` is forced to `true` (NVM flag is ignored) and `/enableNewWebUI?value=0` responds with `400 Legacy web UI is not included in this build`.

Action endpoints used by the SPA remain registered: `/saveSettings`, `/update*`, `/pause`, `/reboot`, replay, exports, and battery commands. Other environments (`stark_330`, `esp32devkit_330`, …) are unchanged: they keep both UIs and the NVM switch.

### Build size (`pio run -e lilygo_330`)

**Before** (legacy UI, no embedded SPA):

```
RAM:   [===       ]  26.4% (used 86368 bytes from 327680 bytes)
Flash: [==========]  98.5% (used 1936304 bytes from 1966080 bytes)
```

**After** (full modern UI: embedded SPA + `/api/*`):

```
RAM:   [===       ]  26.4% (used 86576 bytes from 327680 bytes)
Flash: [========= ]  94.4% (used 1856012 bytes from 1966080 bytes)
```

**Result:** −73,012 bytes of flash (−3.7 pts) even with the embedded SPA, because the C++-generated legacy HTML is substantially heavier than the new UI’s `brotli`  bundle. RAM stays essentially the same, with ~100 KB of flash headroom left.

gegere:
Amazing!

dalathegreat:
The frontend rewrite by jonny takes down the firmware size by -172483 bytes, which is **-10%** of total flash. Just FYI this is the one that we are on track of merging: https://github.com/dalathegreat/Battery-Emulator/pull/1544

herberthobregon:
@dalathegreat Could this one be considered? I think it has a more modern and polished look. The other one also uses Preact with TSX, which isn't a traditional syntax; in 20 years, you don't know if it's still fashionable. Do you think you could put it to a vote?

ngehrsitz:
> The frontend rewrite by jonny takes down the firmware size by -172483 bytes, which is **-10%** of total flash. Just FYI this is the one that we are on track of merging: #1544

The other PR also includes a webserver change which is probably a significant chunk of the savings. Thus it´s not fair to compare these side by side.

PR #2219, opened 2026-04-12
Feature/low pass filter

mbuhansen:
### What
This PR implements a low pass filter on charge rate that is sent to inverter, to prevent oscillation when battery is charging

### Why
Why does it do it?

### How
Low pass filter only allow increasing values (10% new, 90% old)


Software/Software.cpp
```diff
@@ -171,10 +231,28 @@ void update_calculated_values(unsigned long currentMillis) {
   /* Calculate allowed charge/discharge currents*/
   if (datalayer.battery.status.voltage_dV > 10) {
     // Only update value when we have voltage available to avoid div0. TODO: This should be based on nominal voltage
-    datalayer.battery.status.max_charge_current_dA =
-        ((datalayer.battery.status.max_charge_power_W * 100) / datalayer.battery.status.voltage_dV);
-    datalayer.battery.status.max_discharge_current_dA =
+    int32_t target_charge = ((datalayer.battery.status.max_charge_power_W * 100) / datalayer.battery.status.voltage_dV);
```
dalathegreat, line 234: This change will affect every single battery. Would it not be better to have this in the BMW i3 integration?

mbuhansen: Yes it will, but i can move it to bmw i3 battery if it is better.

dalathegreat: Yes I think since this is such a "massive" change, we can either;

- Move it to BMW i3 integration since that is where the need is
- Have it stay in global scope, but make it configurable on/off as a setting

mbuhansen:
I have move it to bmw i3 protocol.

Software/Software.cpp
dalathegreat, line 234: Can you revert all changes to Software.cpp?

mbuhansen:
I have revert change back to software.cpp
I think it would be a good idea to make it an on/off setting, so other battery protocols can use it if they experience oscillation in charging current due to fast regulation.

dalathegreat:
Yes can you add a configurable option? Or do you need help with it?

mbuhansen:
I have changed it so that it can be turned on and off in inverter config on the setting page.

dalathegreat (approved):
Nice work 👍

PR #2229, opened 2026-04-14
Whitebeet Battery Implementation for Bidirectional Charging

M4GNV5:
### What
This PR implements a new "battery" called "CCS-WHITE-BEET", which implements the [SPI white beet protocol from FreeV2G](https://github.com/Sevenstax/FreeV2G/) into Battery Emulator.
The [Whitebeet modules](https://www.codico.com/en/white-beet-ei-1-1-evse-module-with-iso-15118-stack-pwm) are small PCBs implementing the DIN SPEC / ISO15118 protocol, allowing to communicate with most electric vehicles in europe and north america (CCS and NACS connectors).

### Why
The goal is to be able to build bidirectional charging stations using solar inverters, battery emulator and white beet modules.

### How
Basically porting the white beet python reference implementation to the battery emulator world.

Note: I did not yet obtain a white beet module. I wanted to implement this before signing any NDA associated with ordering one, to make sure all information contained within this code is public knowledge.
This is why I will keep this PR as a draft before real world testing has been carried out.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -54,6 +54,8 @@ const char* name_for_battery_type(BatteryType type) {
       return BoltAmperaBattery::Name;
     case BatteryType::BydAtto3:
       return BydAttoBattery::Name;
+    case BatteryType::CcsWhitebeet:
+      return WhitebeetCcsBattery::Name;
     case BatteryType::CellPowerBms:
       return CellPowerBms::Name;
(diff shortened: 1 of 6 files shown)


M4GNV5:
I wanted to order a white beet module just after finishing this PR. Turns out they must have discontinued the white beet module just a few days before the creation of this PR and the starter kit is no longer available for order :disappointed: 

I will check for alternatives, but maybe we can just close this.

PR #2248, opened 2026-04-21
SAFETY: :warning: Treat missing inverter as FAULT

dalathegreat:
### What
This PR changes how a missing inverter event is categorized, instead of a Warning this is now a FAULT

### Why
We had a close call [where an user forgot to power off their battery](https://www.patreon.com/posts/near-disaster-of-156107034) before starting maintenance on a system. This change could have potentially had time to notice the missing inverter, and stopped the system

### How
Missing inverter for longer than 60 seconds (CAN & RS485) are now treated as FAULT. This will on some integrations cause contactors to open (GPIO controlled ones always, CAN controlled are not ensured to open on all integrations!). Good safety improvement

> [!TIP]


dalathegreat:
Merging to gather feedback on trigger sensitivity in 10.6.0 release

PR #2261, opened 2026-04-26
BYD Atto 3 - Fix crash reset to include security access handshake before ClearDTC

demon1300:
### What
This PR fixes the BYD Atto 3 crash reset function.

### Why
The crash reset was sending ClearDTC without first authenticating with the BMS, causing it to silently fail every time.

### How
Adds the same security access handshake already used by the SOC calibration feature to the crash reset state machine, so the BMS accepts the ClearDTC request.

> [!TIP]


demon1300:
Closing as this has already been tested and it was established that it broke the crash reset feature.  No escalation is required.

Issue #2201, opened 2026-04-06
Tesla 12V_low error


roy13:
**Describe the bug**
Constant 12V_low warnings

**To Reproduce**
Closing or opening contacts, sitting with contacts open.

**Expected behavior**
The best solution to this i've found is using HVP_battery12V instead of battery_dcdcLvBusVolt.
However the code states 11.7v this won't work for those with lithium 16v, but they should never get this low so they won't be bothered.

**Screenshots**
See #1885 

**Version and settings (please complete the following information):**
- Please attach screenshot of configuration for easier debugging
- Software version: 10.2
- Battery used: Tesla NCA
- Inverter communication protocol: Solax
- Hardware used for Battery-Emulator: LilyGo T-CAN
- GPIO controlled contactors: No
- MQTT: Yes

roy13:
Solution line 508 TESLA-BATTERY.cpp

// New HVP_battery12V logic
  if (HVP_battery12V > 0) { 
    if ((HVP_battery12V * 0.1) < 11.7) {
        set_event(EVENT_12V_LOW, 0);
    } else {
        clear_event(EVENT_12V_LOW);
    }

Also a bypass this loop while contactor closing.

gegere:
Adding data to this, my 12VDC coming off the Tesla Battery is 14.2 VDC. The Tesla Pack is currently charged to 85% with a bus voltage of 391 VDC

<img width="1200" height="2670" alt="Image" src="https://github.com/user-attachments/assets/a4c078e0-a16f-4e67-9d27-3c0933a4a2a1" />

<img width="853" height="1096" alt="Image" src="https://github.com/user-attachments/assets/f647330c-a42b-4599-8e67-22f2a1f52609" />

roy13:
Even changing it to HVP_battery12V still needs a fine tune as when the contactor closes it deliveres one 12_LOW must read the dip as the contactors draw inrush current.

<img width="206" height="23" alt="Image" src="https://github.com/user-attachments/assets/a6e215c9-6a94-43df-92a1-4b367415b761" />

<img width="175" height="23" alt="Image" src="https://github.com/user-attachments/assets/efd4a56d-ad6d-41a4-ad62-c452b08c4e8d" />

gegere:
I’m seeing this behavior even **after the contactors are closed and the system is running**, which is what makes this a bit confusing.

For context, I was running `16.X.dev` until this morning and have since reverted back to `15.5` to compare behavior.

[...]

PR #2263, opened 2026-04-26
Tesla S/3/X/Y: Change how low 12V is triggered

roy13:
### What
This PR implements code to look at a different sensor for the 12V battery & only during contactors open. #2201 

### Why
As the sensor better reflects Lv battery reading & stops the contactor inrush from dipping the voltage.

### How
Utilises a different sensor & adds a check.

> [!TIP]


Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -505,12 +505,21 @@ void TeslaBattery::
...
-    if ((battery_dcdcLvBusVolt * 0.0390625) < 11.7) {
-      set_event(EVENT_12V_LOW, 0);
-    } else {
-      clear_event(EVENT_12V_LOW);
+  // Only monitor 12V health while contactors are OPEN to avoid false triggers during energization
+  if (HVP_packNegativeV == 0 && HVP_packPositiveV == 0) {
+
+    if (HVP_battery12V > 0) {
+      // Scale is 0.1, so HVP_battery12V * 0.1 is the physical voltage
+      if ((HVP_battery12V * 0.1) < 11.7) {
```
dalathegreat, line 513: Would IMO be better performant to take 
`if (HVP_battery12V < 117) {`
Instead of multiplying by fractal and checking towards decimal value. But not really mandatory, if it works it works!

dalathegreat (approved):
Nice to get Tesla improvements from Tesla users! 🙌

Software/src/battery/TESLA-BATTERY.cpp
roy13: I like it , less CPU load, i'll test and release it with another pull request some time.

PR #2267, opened 2026-04-27
Pylon RS485: Make integration always send current set charge/discharge limits

dalathegreat:
### What
This PR makes Pylon RS485 integration always send current set charge/discharge limits

### Why
Current value was scaled too large, and not updated when value was 0A allowed

### How
We remove the incorrect code that allowed charging when current was 0A allowed.

> [!TIP]


dalathegreat:
Tested OK on the Discord, nice safety improvement

PR #2278, opened 2026-05-01
Replace ACAN2515 with a new MCP2515_Lite CAN library

jonny5532:
### What

This PR adds a new MCP2515 CAN library replacing ACAN2515, with several advantages:
- Much more efficient SPI usage (one-transaction-per-transaction rather than one-transaction-per-byte)
- Triple-buffered transmit and double-buffered receive
- More efficient status/interrupt flag querying/clearing
- Ability to change CAN speed (so now usable for BMW PHEV)
- Slightly lower code (-2972) and RAM (-768) usage

### Why

I think ACAN2515 has been causing performance issues, since each SPI byte transfer causes a context switch and an interrupt. Due to the SPI being done in the background from a high-privilege thread, this appears as general slowdown rather than extra time spent anywhere specifically.

The new library uses the minimum possible number of SPI transactions and is observably much lower overhead. Measuring the average tick duration, with both Tesla and SolArk integrations transmitting frames:

ACAN2515: 16us
MCP2525_Lite: 7us

Adding in 4000 dummy frames per second being received:

ACAN2515: 28us
MCP2525_Lite: 9.5us

### How

The core mechanism is similar to ACAN2515 - using a queue to handle inbound and outbound packets, and a separate high-priority task reading the queue and sending packets, and receiving packets into the queue.

This obviously needs testing with different batteries and inverters to check it hasn't broken anything. It is currently operational with a Solax/T-CAN485 and a test T-2CAN.

(This resulted from first trying the ESP32 DMA SPI functions (which added 20kb extra code), and then a separate adventure using ESP32 high-level interrupts and Xtensa assembly for an even lighter weight library, but this has downsides (awkward to build, and Xtensa assembly...).

> [!TIP]


dalathegreat (approved):
Hands down one of the coolest PRs of 2026 so far. Tested on bench OK, merging to main to get more feedback during this weeks integration tests :raised_hands:

PR #2281, opened 2026-05-02
Fix: stop Preferences reporting bogus errors

jonny5532:
### What
The Arduino-ESP32 Preferences library prints errors when a key is read from NVM that doesn't exist yet.

The error is bogus, there's even a defaultValue arg for it to return if the key isn't found, but it doesn't look like it is going to be fixed anytime soon.

So call `isKey` first on every read to avoid triggering the message.

### Why
Confuses people, might also be causing perf issues (due to it always being printed to serial regardless of whether you have serial debugging enabled).

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.h
+++ b/Software/src/communication/nvm/comm_nvm.h
@@ -53,32 +53,36 @@ class BatteryEmulatorSettingsStore {
     settingsUpdated = true;
   }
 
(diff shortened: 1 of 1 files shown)


dalathegreat:
Awesome, everytime someone posts a USB log they get so worried about these. Massive thanks for fixing it @jonny5532 :raised_hands:

dalathegreat (approved):

PR #2285, opened 2026-05-04
Optimization: -15680 bytes RAM usage

jonny5532:
### What
Three tweaks to save some RAM (-15680 bytes):

- Reduce AsyncTCP stack size (the default is a huge 16kb for some reason, it says 4kb should be enough)
- Narrow enum fields in the event struct
- Const some Tesla CAN frames

> [!TIP]


dalathegreat (approved):
Nice findings! 👍

jonny5532:
(for the record, the current AsyncTCP stack-high-water-mark is 1232 bytes (of the now 4096 bytes) remaining, which seems a decent margin).

PR #2286, opened 2026-05-04
Tesla: split transmit_can into phases to reduce worst-case runtime

jonny5532:
### What
The Tesla `transmit_can` can take 7.5ms to run (when several of the timeouts coincide) which is enough to trigger TASK_OVERRUNs.

Use a loop phase counter to ensure only one runs at once.

I did also consider an if/else-if tree, although that has the (very small) risk of the more regular tasks completely starving the slower ones (under extremely heavy load), which this doesn't.

> [!TIP]


dalathegreat (approved):
Good triage solution for now, nice find 🙌

PR #2290, opened 2026-05-04
BMW i3: Split up transmit CAN to avoid overloads

dalathegreat:
### What
This PR splits up transmit CAN to avoid overloads in the BMW i3 integration

### Why
To avoid getting TASK_OVERRUN events

### How
Same approach as in https://github.com/dalathegreat/Battery-Emulator/pull/2286

> [!TIP]


odlg:
I guess that is very similar to add an else clause in the end of every if-clause in the transmit_can function. Except that else clauses would give higher priority to the first if clauses. 
I have been running with else clauses for a while, it works fine. I have been considering to split up the 1000ms in two. It is very large.

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -313,10 +313,15 @@ void BmwI3Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
 }
 
 void BmwI3Battery::transmit_can(unsigned long currentMillis) {
+  // Ensure we only send one message branch at a time, to reduce worst-case runtime
+  static int transmitPhase = -1;
```
dalathegreat, line 317: Does not work for double bat, static def

jonny5532: Good point! Would need to be a class member instead I guess.

@odlg - indeed an if/else-if tree is easier to read and doesn't need the extra variable.

I did that initially but had the concern that it then makes it possible for high-rate branches to starve lower ones, eg:

```
if(...every 10ms...) {
   ... something that takes longer than 10ms ...
} else if(...every 100ms...) {
   ...
} else if(... every 1s...) {
   ...
}
```

In this example only the first branch would ever run, the others would never get a chance. Using the transmitPhase thing means that it'd be fairer and they'd all get a turn.

(Maybe this is a total non-issue, but some batteries can take quite a few ms in the sending functions due to checksums (and ESP32 flash contention)).

PR #2297, opened 2026-05-07
Checks: add smoothing to free-heap checks

### LINKED ISSUE #2240: Low-heap-warning, Inverter rises communication error same as #2113 but I am on 10.5.0
**Describe the bug**
I have the same message as on Bug #2113. Fronius Symo Gen24 report battery communication error and at the same time I have the low heap warning in the BE events. However, the BE events occur more frequently than the error message from Fronius. So far, communication has always resumed automatically.

**To Reproduce**
Just wait :-)

**Expected behavior**
no low heap, no communication error

**Screenshots**

Image

Image

Image

- Software version: 10.5.0
- Battery used: Dyness Stack 100 (Pylontech protocol)
- Inverter communication protocol: BYD Modbus
- Hardware used for Battery-Emulator: LilyGo Can2RS485
- MQTT: yes

Image

Image

Image

Image

jonny5532:
### What
The free-heap warning is triggering randomly for people. I think this is because it compares the exact value at the moment that it is sampled. If you happen to load the settings page (which itself gobbles 40-50kb of RAM briefly) at the same time as the value is checked, then you will get a warning.

This adds some basic smoothing, so that it will only generate the warnings if the level is consistently too low, not just a brief blip.

### Why
Reduce phantom warnings which cause unnecessary concern. Fixes #2240

> [!TIP]

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -41,7 +41,7 @@ void update_machineryprotection() {
     clear_event(EVENT_CPU_OVERHEATED);  //Hysteresis on the clearing
   }
 
-  if (datalayer.system.info.CPU_free_heap < 62000) {
+  if (datalayer.system.info.CPU_free_heap_smoothed < 62000) {
     set_event(EVENT_LOW_HEAP_MEMORY, (datalayer.system.info.CPU_free_heap / 1000));
   } else {
     clear_event(EVENT_LOW_HEAP_MEMORY);
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -159,6 +159,8 @@ void update_calculated_values(unsigned long currentMillis) {
 
   /*Update free heap*/
   datalayer.system.info.CPU_free_heap = ESP.getFreeHeap();
+  datalayer.system.info.CPU_free_heap_smoothed =
+      (datalayer.system.info.CPU_free_heap_smoothed * 3 + datalayer.system.info.CPU_free_heap) / 4;
 
   /* Check is remote set limits have timed out */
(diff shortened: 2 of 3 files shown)


dalathegreat:
I think this might not be required. The heap usage is reduced significantly in main now (~20k more free than before), so any dips will be way rarer

jonny5532:
I believe Rik still got the low-heap-warning even with the extra heap - although I guess we'll know better when more people have tried it...

(I can still recreate the warning myself, although I do have to mash F5 on the settings page).

Software/src/devboard/safety/safety.cpp
```diff
@@ -41,7 +41,7 @@ void update_machineryprotection() {
     clear_event(EVENT_CPU_OVERHEATED);  //Hysteresis on the clearing
   }
 
-  if (datalayer.system.info.CPU_free_heap < 62000) {
+  if (datalayer.system.info.CPU_free_heap_smoothed < 62000) {
```
dalathegreat, line 44: Instead of adding more to the datalayer (especially uint32), could we not just do the following;

static uint8_t hysteresisSeconds = 0;

if (datalayer.system.info.CPU_free_heap < 62000) {
   hysteresisSeconds++;
   if(hysteresisSeconds > 5)
      set_event(EVENT_LOW_HEAP_MEMORY, (datalayer.system.info.CPU_free_heap / 1000));
} else {
   hysteresisSeconds = 0;
   clear_event(EVENT_LOW_HEAP_MEMORY);
}

dalathegreat (changes requested):

dalathegreat:
I am going with this IMO simpler integration that only adds an uint8_t https://github.com/dalathegreat/Battery-Emulator/pull/2404

Issue #2193, opened 2026-04-03
[BUG] Disabling internal AP and setting Custom Wifi hostname results in bootloop


Stewart555:
**Describe the bug**
Title says it all, when disabling the built in WiFi AP and setting a Custom WiFi hostname in the same "save" action BE will go in to a boot-loop and not recover when a reset/reboot command is issued.

**To Reproduce**
BE running 10.4.0, connect BE to "outside WiFi" (your own WiFi), connect to BE via your own WiFi, navigate to settings page, remove tick after "Broadcast Wifi access point" and set a "Custom Wifi hostname", click Save, click Reboot and confirm.

**Expected behavior**
Internal WiFi being disabled, and a hostname being set.

**Screenshots**
<img width="820" height="1954" alt="Image" src="https://github.com/user-attachments/assets/26aeff09-fd3a-4dfa-ad5f-37c77d260325" />

**Version and settings (please complete the following information):**
- Software version: 10.4.0
- Battery used: NONE
- Inverter communication protocol: NONE
- Hardware used for Battery-Emulator: Stark CMR
- GPIO controlled contactors: N/A
- MQTT: NO

[12:29:48]Rebooting...
[12:29:48]ets Jul 29 2019 12:21:46
[12:29:48]
[12:29:48]rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
[12:29:48]configsip: 153911750, SPIWP:0xee
[12:29:48]clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[12:29:48]mode:DIO, clock div:1
[12:29:48]load:0x3fff0030,len:4876
[12:29:48]ho 0 tail 12 room 4
[12:29:48]load:0x40078000,len:16560
[12:29:48]load:0x40080400,len:3500
[12:29:48]entry 0x400805b4
[12:29:49][    88][E][Preferences.cpp:506] getString(): nvs_get_str len fail: MQTTSERVER NOT_FOUND
[12:29:49][    97][E][Preferences.cpp:506] getString(): nvs_get_str len fail: MQTTUSER NOT_FOUND
[12:29:49][   104][E][Preferences.cpp:506] getString(): nvs_get_str len fail: MQTTPASSWORD NOT_FOUND
[...]

PR #2305, opened 2026-05-10
Bugfix: WiFi crash

malec1984:
This is solution for issue https://github.com/dalathegreat/Battery-Emulator/issues/2193

### What
After disabling the WiFi AP sometimes there is a crash at boot. Move the WiFi.onEvent calls before the WiFi.mode call. This ensures the callback list is fully populated before the arduino_events FreeRTOS task is created, eliminating the race condition.

> [!TIP]


dalathegreat (approved):
Massive thanks for investigating this @malec1984 🙌

PR #2311, opened 2026-05-11
Bugfix: Pylon CAN, Fix temperature sending

dalathegreat:
### What
This PR fixes the temperature measurement when using Pylon HV protocol

### Why
On Deye inverters running Pylon CAN, temps where showing up as -70*C, even though battery was at +25

image

### How
Correct offset applied
Bonus, also added decoding of module number 

> [!TIP]


dalathegreat:
Tested OK on the Discord, Deye inverters now show OK temperature when using Pylon HV CAN

image

PR #2326, opened 2026-05-14
Dacia Spring: Add override charge/discharge power setting for CMFA platform

### LINKED ISSUE #2298: Goodwe inverter + Spring Dacia Battery communicating but ...
I've got 3 Dacia spring batteries, 2 Inverters Goodwe (ET serie 10kW triphase  and EH serie 5kW onephase)
Looks all good, cells are all good, voltage, tem and percentage all correct, Inverter gives no errors but still it refuses using the battery, looks like the inverter sees current limit set to 0, and I cannot overwrite them.
If, from BE settings I do activate the very last option: Undercharged emergency recovery mode: START 
then it starts charging the battery with a 5A limit.

In order to use the battery, I am temporary using the Inverter with no communication to the battery, and using BE just to keep the battery ON (and to manage the precharge and the contacts relays), in this way it is working, but of course I would prefer to solve the main issue, any idea?

I am not using the batteries all together for the moment, just one by one, batteries and one by one the inverters

dalathegreat:
### What
This PR implements override charge/discharge power setting for CMFA platform

### Why
Some BMS send 0W allowed  charge, and 0W allowed discharge all the time, making the battery unusable. Fixes #2298

### How
We fix this with a new config option, 
- Manual charging power, watt
- Manual discharge power, watt
- Rampdown SOC, pptt

image

If values are set to 0 W, we use BMS sent value like before

If the battery is unavailable and always report 0W, user can set their own values

image

> [!TIP]

### DIFF
--- a/Software/src/battery/CMFA-EV-BATTERY.cpp
+++ b/Software/src/battery/CMFA-EV-BATTERY.cpp
@@ -42,9 +42,28 @@ void CmfaEvBattery::
   datalayer_battery->status.remaining_capacity_Wh = static_cast<uint32_t>(
       (static_cast<double>(datalayer_battery->status.real_soc) / 10000) * datalayer_battery->info.total_capacity_Wh);
 
-  datalayer_battery->status.max_discharge_power_W = discharge_power_w;
+  //Some packs are locked and do not report allowed charge power, in this case we need to use the user specified override value
+  //instead of the value sent from the BMS
+  if (datalayer.battery.status.override_charge_power_W > 0) {
+    if (datalayer_battery->status.real_soc > 9900) {
+      datalayer_battery->status.max_charge_power_W = 0;
+    } else if (datalayer_battery->status.real_soc >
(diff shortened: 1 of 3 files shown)


bellapanko:
It works!

dalathegreat:
@bellapanko We will be removing this fix. The proper way to fix the battery is to follow the WIki https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Dacia-Spring-%E2%80%90-Renault-K%E2%80%90ZE#note-on-missing-cell-voltages

@bellapanko , do you see all 72 cells on cellmonitor page?

dalathegreat:
Oh, yeah it appears that we actually do need this fix on some CMFA batteries. Even though all cells are reported OK, it for some reason sends 0A allowed.

Keep testing the PR, and let me know if it is a good enough solution to estimate values for now!

dalathegreat:
Tested OK by Courage2000 on the Discord! Merging this :+1:

PR #2330, opened 2026-05-15
Add Modbus TCP-to-RTU gateway

spartandingo:
## Summary

This adds an optional Modbus TCP-to-RTU gateway mode intended for ESP32 boards with RS485, especially the LilyGo T-CAN485. The goal is to let a Modbus TCP client, such as Home Assistant's Sunsynk integration, talk to an RS485 inverter through the emulator over WiFi without needing a separate USB-RS485 adapter and mbusd host.

## What changed

- Adds a small, self-contained Modbus gateway module that translates Modbus TCP MBAP frames to RTU frames on `Serial2`, then wraps valid RTU responses back into TCP responses.
- Adds settings page controls for enabling the gateway, TCP port, RTU baud rate, and RTU response timeout.
- Persists the new settings through the existing Preferences/NVS settings flow.
- Prevents RS485 double ownership at runtime by disabling gateway startup when another selected inverter, battery, charger, or shunt protocol already owns RS485/Modbus.
- Shows a warning in the settings UI when the selected configuration conflicts with gateway mode.
- Adds a small unit test for the Modbus CRC helper.

## Notes

The gateway deliberately avoids adding a new Modbus library dependency. It is a transparent frame bridge rather than a Modbus endpoint, which keeps the change localized and avoids increasing flash use more than necessary.

The UI currently warns about RS485 conflicts rather than preventing the user from selecting conflicting protocol combinations. At runtime, the gateway refuses to start if another selected protocol owns RS485, so the existing RS485 stack remains in control.

## Verification

- `rtk pio run -e lilygo_330`

This build passes locally. The host CMake unit-test build is currently blocked by an unrelated existing `String(...)` conversion ambiguity in `Software/src/battery/BMW-PHEV-HTML.h`, but the new gateway test/source compile before that pre-existing failure.


docs/superpowers/plans/2026-05-15-modbus-tcp-rtu-gateway.md
```diff
@@ -0,0 +1,30 @@
+# Modbus TCP-to-RTU Gateway Implementation Plan
```
dalathegreat, line 1: AI generated documents should not be attached to a PR

docs/superpowers/specs/2026-05-15-modbus-tcp-rtu-gateway-design.md
```diff
@@ -0,0 +1,85 @@
+# Modbus TCP-to-RTU Gateway Design
```
dalathegreat, line 1: AI generated documents should not be attached to a PR

Software/Software.cpp
```diff
@@ -650,4 +661,6 @@ void setup() {
 }
 
 // Loop empty, all functionality runs in tasks
-void loop() {}
+void loop() {
+  loop_modbus_gateway();
```
dalathegreat, line 665: Should run as a task, not in the old arduino loop function

Software/src/communication/modbus_gateway/modbus_gateway.cpp
```diff
@@ -0,0 +1,263 @@
+#include "modbus_gateway.h"
```
dalathegreat, line 1: Got any flash usage example? How much does it increase flash usage on a 4MB board like the LilyGo T-CAN485?

dalathegreat (changes requested):

jonny5532:
I had a look at this - a Modbus-RS485 bridge isn't really directly related to Battery Emulator's role (you could just use an extra hardware gadget), but enough people seem interested that it is probably worth it, as long as we can keep the code size and performance impact low enough. (In fact I use my own similar thing in my custom build, this could replace that).

The branch above compiles to 11kb extra code size on the T485, which is rather a big hit (it is very tight on code space, we keep finding things to tidy up but then the space disappears again!). The main culprit seems to be the WiFiServer. I (and Gemini ;)) had a go at rewriting with non-blocking BSD sockets instead, and have got it down to 4.3kb which seems more reasonable.

Also as Dala pointed out, this shouldn't go in loop() as it'll run with arbitrary priority and probably not on the correct core (all low-prio stuff like this should be pinned to Core 0). It can be further optimized by making it non-blocking so it can go in the connectivity loop, rather than needing its own process (and stack, which eats precious RAM).

Also there's already a Modbus CRC16 implementation linked in, so no need to add another (this isn't the only culprit, I can see at least one more in BE that needs rationalizing...)

Also, I wonder if we actually need the configuration? I hope that we can just stick to the standard port 502 and hardcode the timeout. Also I am very curious as to whether anyone would use it on a baud other than 9600, I'd be tempted to hardcode that until someone complains (again, this saves significant flash).

Finally, I can see why you'd want the RS485 conflict detection, but that `conflict_protocol_selected` function is fragile as people are going to forget to update it if new RS485-using integrations are added. I'm tempted to either just get rid of it, and make users responsible for avoiding conflict, or find another way of handling it that is more robust (I am not sure).

Actually-finally, there's another PR for new hardware which needs DE on the RS485 - so this will probably need tweaking if it gets merged after (shouldn't be a big deal though).

My (well, Gemini + my editing) 4.3kb version is attached FYI. It compiles but I haven't tested it (but can do if we want to go nonblocking-sockets instead, I think the more complex code is worth the significant size reduction).

[modbus_gateway.cpp](https://github.com/user-attachments/files/27988427/modbus_gateway.cpp)

ngehrsitz:
@jonny5532 I could make use of this to connect to my ABB energy meters. I run that bus on the highest supported baud rate 115200 so that I can query the parameters from all four meters in real time. Therefore having the baud rate configurable is a requirement for me.

dalathegreat:
This hasn't moved for 2 months, so I am closing this for now. If someone really feels strongly for adding this, the improvement points set by jonny should be implemented before the PR gets re-opened!

PR #2337, opened 2026-05-17
fix/byd-calibration-ah-default-from-bms

demon1300:
### What
When a user opens the "More Battery Info" page and hits Calibrate Now, the calibration also writes the pack AH (capacity) value. The AH field defaults to 150AH hardcoded in the struct, so if the user doesn't change it, their real pack capacity gets overwritten.

### Why
The BMS already knows the current packs capacity based on its history and the manual calibration fields should show this value by default; so users aren't silently writing a default value and overwriting their true SOH value.

### How
Added a calibrationAH_seeded flag. The first time BMS_capacity_current_calibration comes back non-zero from a poll (within a few seconds of boot), calibrationTargetAH is set from the live BMS value instead of the 150AH default. The flag ensures subsequent polls don't overwrite any value the user has manually changed.

> [!TIP]


demon1300:
works as intended screenshot is from a fresh boot confirming the default is no longer 150AH :)
image

dalathegreat (approved):

Issue #2334, opened 2026-05-16
[Feature request] Add Waveshare ESP32-S3-RS485-CAN as supported hardware target


cray9000:
Hi everyone,

I would like to make a feature request to add the **Waveshare ESP32-S3-RS485-CAN** board as an supported hardware target.

From my point of view, this board would be a very useful addition next to the existing **LilyGo T-CAN485 / CAN-RS485** target. It fills a similar practical role — one CAN interface plus one RS485/Modbus interface — but uses a newer ESP32-S3 platform with significantly more memory resources.

The Waveshare board is more of a modern ESP32-S3-based complement to the LilyGo CAN-RS485 style hardware.

## Why this board is interesting

The Waveshare ESP32-S3-RS485-CAN is an industrial ESP32-S3 control board with onboard isolated RS485 and isolated CAN interfaces, USB-C, 7–36 V input, digital/power isolation, TX/RX/CAN communication indicators and a rail-mount enclosure. Waveshare lists it as SKU `32154`, part number `ESP32-S3-RS485-CAN`, with onboard RS485 and CAN and a default ESP32-S3R8 module.

The official Whitepaper contains front-view, interface-description, onboard-component and DIN-rail enclosure images for the ESP32-S3-RS485-CAN board.
Source: https://www.waveshare.com/esp32-s3-rs485-can.htm

## Technical comparison: LilyGo T-CAN485 vs Waveshare ESP32-S3-RS485-CAN

| Area | LilyGo T-CAN485 | Waveshare ESP32-S3-RS485-CAN |
|---|---|---|
| Role in project | Existing CAN + RS485 target | Proposed additional CAN + RS485 target |
| MCU class | ESP32 | ESP32-S3 |
| Flash / PSRAM | 4 MB flash, no large PSRAM class | 16 MB flash + 8 MB PSRAM |
| CAN | Native CAN/TWAI via onboard transceiver | Native ESP32-S3 TWAI via onboard CAN transceiver |
| RS485 | AutoDirection-style RS485 transceiver behavior | RS485 direction-control pin on GPIO21 |
| RS485 firmware handling | Static enable is sufficient | Firmware should toggle TX/RX direction |
| Status LED | Programmable WS2812/NeoPixel on GPIO04 | No programmable WS2812/NeoPixel, only fixed PWR/RS485/CAN indicators |
| Mechanical / install | Dev-board style | Industrial enclosure / DIN-rail style |
[...]

PR #2339, opened 2026-05-17
Add Waveshare ESP32-S3 RS485 CAN hardware target

cray9000:
### What
This PR implements ...
[Feature request] Add Waveshare ESP32-S3-RS485-CAN as supported hardware target #2334

### Why
I think this board would be a very useful addition for the project community: it is affordable, easy to source, roughly in the same practical price class as other ESP32 CAN/RS485 boards, has CAN and RS485 onboard, offers much more memory headroom through ESP32-S3R8 with 16 MB flash and 8 MB PSRAM, and is already documented/tested in a similar open-source BMS context.

The required code changes are may also generally useful because they make the RS485 layer to support: AutoDirection boards such as LilyGo keep their existing behavior, while boards with explicit DE/RE control such as Waveshare can be handled correctly.

This would make waveshare_330 a good standard target next to lilygo_330, not a replacement for Stark or other higher-end multi-CAN hardware.

### How
https://github.com/dalathegreat/Battery-Emulator/issues/2334

> [!TIP]

### DIFF
--- a/Software/src/communication/rs485/comm_rs485.cpp
+++ b/Software/src/communication/rs485/comm_rs485.cpp
@@ -4,14 +4,33 @@
 
 #include <list>
 
+static gpio_num_t rs485_de_pin = GPIO_NUM_NC;
+static bool rs485_de_active_high = true;
+static bool rs485_control_pins_initialized = false;
+
+static uint8_t rs485_tx_level() {
+  return rs485_de_active_high ? HIGH : LOW;
+}
+
+static uint8_t rs485_rx_level() {
+  return rs485_de_active_high ? LOW : HIGH;
+}
+
 bool init_rs485() {
 
   auto en_pin = esp32hal->RS485_EN_PIN();
   auto se_pin = esp32hal->RS485_SE_PIN();
   auto pin_5v_en = esp32hal->PIN_5V_EN();
+  rs485_de_pin = esp32hal->RS485_DE_PIN();
+  rs485_de_active_high = esp32hal->RS485_DE_ACTIVE_HIGH();
 
   if (!esp32hal->alloc_pins_ignore_unused("RS485", en_pin, se_pin, pin_5v_en)) {
-    DEBUG_PRINTF("Modbus failed to allocate pins\n");
+    DEBUG_PRINTF("RS485 failed to allocate static enable pins\n");
+    return true;  //Early return, we do not set the pins
+  }
+
+  if (rs485_de_pin != GPIO_NUM_NC && !esp32hal->alloc_pins("RS485 DE", rs485_de_pin)) {
+    DEBUG_PRINTF("RS485 failed to allocate DE pin\n");
     return true;  //Early return, we do not set the pins
   }
 
@@ -30,10 +49,54 @@ bool init_rs485() {
     digitalWrite(pin_5v_en, HIGH);
   }
 
-  // Inverters and batteries are expected to initialize their serial port in their setup-function
-  // for RS485 or Modbus comms.
+  if (rs485_de_pin != GPIO_NUM_NC) {
+    pinMode(rs485_de_pin, OUTPUT);
+    digitalWrite(rs485_de_pin, rs485_rx_level());
+  }
 
+  rs485_control_pins_initialized = true;
+
+  // Inverters and batteries initialize the serial port in their setup-function.
   return true;
+}
+
+bool rs485_begin(const char* owner, HardwareSerial& serial, uint32_t baud, uint32_t config) {
+  auto rx_pin = esp32hal->RS485_RX_PIN();
+  auto tx_pin = esp32hal->RS485_TX_PIN();
+
+  if (!esp32hal->alloc_pins(owner, rx_pin, tx_pin)) {
+    return false;
+  }
+
+  serial.begin(baud, config, rx_pin, tx_pin);
+  rs485_set_direction(false);
+  return true;
+}
+
+void rs485_set_direction(bool transmit) {
+  if (!rs485_control_pins_initialized || rs485_de_pin == GPIO_NUM_NC) {
+    return;
+  }
+
+  digitalWrite(rs485_de_pin, transmit ? rs485_tx_level() : rs485_rx_level());
+}
+
+size_t rs485_write(const uint8_t* data, size_t len) {
+  if (rs485_de_pin == GPIO_NUM_NC) {
+    return Serial2.write(data, len);
+  }
+
+  rs485_set_direction(true);
+  delayMicroseconds(10);
+  size_t written = Serial2.write(data, len);
+  Serial2.flush();
+  delayMicroseconds(10);
+  rs485_set_direction(false);
+  return written;
+}
+
+size_t rs485_write(const char* data, size_t len) {
+  return rs485_write(reinterpret_cast<const uint8_t*>(data), len);
 }
 
 static std::list<Rs485Receiver*> receivers;
--- a/Software/src/communication/rs485/comm_rs485.cpp
+++ b/Software/src/communication/rs485/comm_rs485.cpp
@@ -1,20 +1,19 @@
 #include "comm_rs485.h"
 #include <Arduino.h>
 #include "../../devboard/hal/hal.h"
+#include "driver/uart.h"
 
 #include <list>
 
+// All current Battery-Emulator RS485 users use Serial2/UART2.  For boards with
+// a real DE/~RE pin (for example Waveshare ESP32-S3-RS485-CAN with SP3485EN),
+// we let the ESP-IDF UART driver control RTS in UART_MODE_RS485_HALF_DUPLEX.
+// This keeps the existing Serial2.write() path and the Arduino/ESP32 UART
+// locking behavior intact instead of wrapping writes in a separate function.
+static constexpr uart_port_t RS485_UART_NUM = UART_NUM_2;
+
 static gpio_num_t rs485_de_pin = GPIO_NUM_NC;
 static bool rs485_de_active_high = true;
-static bool rs485_control_pins_initialized = false;
-
-static uint8_t rs485_tx_level() {
-  return rs485_de_active_high ? HIGH : LOW;
-}
-
-static uint8_t rs485_rx_level() {
-  return rs485_de_active_high ? LOW : HIGH;
-}
 
 bool init_rs485() {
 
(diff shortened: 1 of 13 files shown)


platformio.ini
```diff
@@ -113,4 +113,38 @@ build_flags =
...
+board = esp32s3_flash_16MB
+monitor_speed = 115200
+upload_speed = 921600
+monitor_filters = default, time, log2file, esp32_exception_decoder
+
+board_build.arduino.partitions = default_16MB.csv
+
+; Waveshare ESP32-S3-RS485-CAN:
+; ESP32-S3R8 = 8MB Octal/OPI PSRAM, board has 16MB Flash.
+board_build.arduino.memory_type = qio_opi
```
dalathegreat, line 131: Might be best to set this this **qio_qspi** to avoid any unforseen consequences. We had so many odd startup glitches/crashes due to memory type being handled differently in the past!

Software/src/communication/rs485/comm_rs485.cpp
```diff
@@ -30,12 +49,56 @@ bool init_rs485() {
...
+
+  digitalWrite(rs485_de_pin, transmit ? rs485_tx_level() : rs485_rx_level());
+}
+
+size_t rs485_write(const uint8_t* data, size_t len) {
+  if (rs485_de_pin == GPIO_NUM_NC) {
+    return Serial2.write(data, len);
+  }
+
+  rs485_set_direction(true);
```
dalathegreat: Hmm. Not sure I like this changing of the serial2.write into the rs485_write section. 

The original code ultimately calls the ESP32 standard function

void uartWriteBuf(uart_t *uart, const uint8_t *data, size_t len) {
  if (uart == NULL || data == NULL || !len) {
    return;
  }

  UART_MUTEX_LOCK();
  uart_write_bytes(uart->num, data, len);
  UART_MUTEX_UNLOCK();

The original uartWriteBuf holds UART_MUTEX_LOCK() for the duration of the write. The new function has no such protection around the DE pin toggle + write + flush sequence. If two sections of code calls rs485_write() concurrently, you can get interleaved DE pin state and corrupted bus output.

Software/src/communication/rs485/comm_rs485.cpp
```diff
@@ -30,12 +49,56 @@ bool init_rs485() {
...
+  delayMicroseconds(10);
```
dalathegreat: Where did 10µs come from? What transceiver chip is being used, and what does its datasheet say about DE propagation delay and bus turnaround time?

Software/src/communication/rs485/comm_rs485.cpp
```diff
@@ -30,12 +49,56 @@ bool init_rs485() {
...
+  size_t written = Serial2.write(data, len);
+  Serial2.flush();
+  delayMicroseconds(10);
```
dalathegreat: At what baud rates was this tested at? The post-flush guard time should be at least one full character period at the lowest baud rate. Let's add some comments here

dalathegreat (commented):
Very nice to get this hardware into the mix. I have some comments on the RS485 refactoring, most likely my fear is unwarranted, but would be good to test this PR on atleast a Fronius Modbus RTU install, and a Deye RS485 install, before we merge it.

platformio.ini
cray9000: Hi, Waveshare defines it in their WIKI as PSRAM Mode: Octal SPI (8 lines). 
https://www.waveshare.com/wiki/ESP32-S3-RS485-CAN. 
Works fine with my device in qio_opi mode, no issues and PSRAM speed is at full supported speed there.

Software/src/communication/rs485/comm_rs485.cpp
cray9000: Good point, I agree with this concern and reworked the RS485 part accordingly.

The updated version no longer wraps or replaces the existing Serial2.write() calls. Instead, boards with a real RS485_DE_PIN() configure UART2 in UART_MODE_RS485_HALF_DUPLEX during RS485 initialization. This keeps the existing Arduino/ESP32 UART write path and its internal locking behavior intact, while the UART driver handles RTS/DE.

I also removed the fixed 10 µs guard delays. They were not well justified as a generic bus turnaround value.

I added a small hardening follow-up as well:

keep the UART2/Serial2 assumption explicit,
do not let uart_set_pin() claim a DE pin if the HAL pin allocator rejected it,
document that waveshare_330 intentionally does not define the legacy RS485_DE_PIN build flag because DE is handled through the HAL + UART RS485 half-duplex mode.

The remaining functional changes are the Waveshare HAL, the waveshare_330 PlatformIO target, central rs485_begin() initialization, and the no-LED safety fix for boards without a programmable WS2812/NeoPixel LED.

PS.
Pls note since hw_becom.h already has set:

virtual gpio_num_t RS485_DE_PIN() { return GPIO_NUM_16; }

BECom may benefit from the same RS485_DE_PIN handling, but should ideally be regression-tested.

Software/src/communication/rs485/comm_rs485.cpp
cray9000: pls see above comment (reworeked and removed)

Software/src/communication/rs485/comm_rs485.cpp
cray9000: tested at 9600 Baud, but It´s completly reworked in "Harden RS485 half-duplex UART setup"

cray9000:
I pushed an updated version to the same PR branch.

The RS485 part has been reworked: the custom rs485_write() wrapper was removed, existing Serial2.write() calls stay unchanged, and boards with a real DE/~RE pin now use the ESP-IDF UART RS485 half-duplex mode instead.

tested successfully with:
waveshare_330 board against Fronius Verto Plus, PylonHV Bat@Canbus
Lillygo_330 board against Fronius Gen24 Plus, PylonHV Bat@Canbus

dalathegreat:
Are you able to fix the unit tests? Instructions here: https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#local-unit-test-run-

cray9000:
> Are you able to fix the unit tests? Instructions here: https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#local-unit-test-run-

I fixed the unit-test build issue by guarding the ESP-IDF UART driver include and RS485 half-duplex setup with #ifndef UNIT_TEST. The firmware build still configures UART2 with UART_MODE_RS485_HALF_DUPLEX, but the host gtest build skips the ESP-IDF-specific hardware setup because driver/uart.h is not available outside the ESP32 build environment.

dalathegreat:
I have installed the currently built file ([2d67e6a](https://github.com/dalathegreat/Battery-Emulator/pull/2339/commits/2d67e6a53b1d648c52c3ebef2d19a9a3a01b61c8)) onto a Stark CMR connected via RS485 to a Fronius Inverter. It started up OK, will run it for a while and verify stability!

dalathegreat:
Wiki page created for the board: https://github.com/dalathegreat/Battery-Emulator/wiki/Hardware:-Waveshare-ESP32%E2%80%90S3%E2%80%90RS485%E2%80%90CAN

Software/src/communication/rs485/comm_rs485.cpp
```diff
@@ -30,8 +59,60 @@ bool init_rs485() {
...
+  auto rx_pin = esp32hal->RS485_RX_PIN();
+  auto tx_pin = esp32hal->RS485_TX_PIN();
+
+  if (!esp32hal->alloc_pins(owner, rx_pin, tx_pin)) {
+    return false;
+  }
+
+  serial.begin(baud, config, rx_pin, tx_pin);
+
+  if (rs485_de_pin != GPIO_NUM_NC) {
```
jonny5532, line 84: Unfortunately this clause adds 2.4kb of code usage on the T485, even though the T485 doesn't need it - it still gets compiled in because the C++ optimizer isn't clever enough to walk the class virtuals and detect this. The T485 is very tight on space (we have spent time freeing a few 100 bytes here and there) so it would be a shame to blow 2.4kb unnecessarily here.

A slightly-hacky fix is to wrap this if branch with `#ifndef SMALL_FLASH_DEVICE` to stop it being compiled on T485 - I can't think of a better way?

Software/src/devboard/utils/led_handler.cpp
```diff
@@ -17,24 +17,35 @@
...
 uint16_t map_int(uint16_t x, uint16_t in_min, uint16_t in_max, uint16_t out_min, uint16_t out_max) {
   return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
 }
 
 bool led_init(void) {
-  if (!esp32hal->alloc_pins("LED", esp32hal->LED_PIN())) {
+  gpio_num_t led_pin = esp32hal->LED_PIN();
+
+  if (led_pin == GPIO_NUM_NC) {
+    DEBUG_PRINTF("No LED configured for selected HW. Skipping LED setup.\n");
```
jonny5532, line 30: I think this DEBUG_PRINTF is unnecessary - it doesn't really need to log that there's no LED, it'll just confuse people?

(plus removing this saves a few more bytes on the T485 ;)

jonny5532 (commented):
Hi, a couple things I noticed (that will hopefully save some code space too)

Software/src/devboard/utils/led_handler.cpp
dalathegreat: Agreed, will handle this in #2348

Software/src/communication/rs485/comm_rs485.cpp
dalathegreat: Agreed, will handle this in #2348

dalathegreat (approved):
Massive thanks for integrating this hardware @cray9000 🙌 Merging this as code has been stable on RS485 devices!

sandersr:
Hi,

I just tried flashing this and I'm getting an error: "Your ESP32-S3 board is not supported."

Is there anything else that I need to do?

Alternatively, could you please add the instruction how to flash it using esptool?

dalathegreat:
@sandersr You are using the BE Websinstaller yes? https://dalathegreat.github.io/BE-Web-Installer/

Screenshots?

sandersr:
Yes I was.

I ended up doing this and it worked fine:
esptool.py --chip esp32-s3 --port /dev/ttyACM1 write_flash 0x10000 ~/Downloads/BE_v10.9.0_Waveshare-ESP32-S3-RS485-CAN.ota.bin

sandersr:
error

dalathegreat:
Massive thanks @sandersr , I have put up a fix towards the BE Web installer repository! 
https://github.com/dalathegreat/BE-Web-Installer/pull/6

PR #2353, opened 2026-05-20
Improve MEB code readability

malec1984:
### What
- Replace all message ID numbers with names
- Fix one wrong bit check for transport mode
- Don't advance to next PID request if there is no CAN bus available.
- Init the cell voltage array with 0.

### Why
Make the code more user friendly.

### How
Names are easier to remember than hex numbers.

> [!TIP]


dalathegreat (approved):
Nice improvements all around. Thanks for the contribution @malec1984 🙌

PR #2356, opened 2026-05-20
BMW I3 Contactor close change in logic

mbuhansen:
### What
This PR implements a new contactor close logic to BMW I3, this change do so battery not close if inverter not allow, and when contor is open we still send 10b message så battery dont set dtc, and precharge is  blocking if 10B message is missing for to long time.


odlg (approved):
Looks good.

PR #2367, opened 2026-05-24
Add external precharge via I2C + G05

bradlrein:
Feature:
This adds a new method for precharging the MEB Battery:

- External precharge via TI TPS55288 evaluation board that makes a variable voltage controlled by I2C bus and a XPPOWER(emco) G05 high voltage source

- TPS55288 support on second I2C bus
- Dynamic voltage regulation based on intermediate HV feedback
- Hardware support for STARK and LilyGo T-CAN485 ... more can be added when we know which free IO-pins to use
- We only need to connect the supply of the TI-Board parallel to the "Inverter disconnect contactor" and also SCL, SDA,GND

Todo:
design a cheaper board, that we do not need to buy a evaluation board directly from Texas Instruments for almost 100€
[overview.odg](https://github.com/user-attachments/files/28687116/overview.odg)


dalathegreat:
Which precharge hardware does this new method work with?

bradlrein:
overview


What we need:
- Texas Instruments TPS55288EVM-045 Evaluation Module
- XPPOWER (emco) G05 high voltage source
- Diode 1N4007
- Contactor Normally Closed
- Fuseholder and DC-Fuses 10x38 2A for DC use

dalathegreat:
Nice! I added a mention of this new HV source solution, feel free to further add to the Wiki (real life pictures?) https://github.com/dalathegreat/Battery-Emulator/wiki/High-Voltage-source#option-a-tps55288evm-045--xppower-emco-g05-high-voltage-source

bradlrein:
tpstest_web

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -469,8 +469,14 @@ String raw_settings_processor(const String& var, BatteryEmulatorSettingsStore& s
...
-  if (var == "EXTPRECHARGE") {
-    return settings.getBool("EXTPRECHARGE") ? "checked" : "";
+  if (var == "EXTPRECHARGE_DISABLED") {
+    return settings.getUInt("EXTPREMODE", 0) == 0 ? "selected" : "";
+  }
+  if (var == "EXTPRECHARGE_HIA4V1") {
+    return settings.getUInt("EXTPREMODE", 0) == 1 ? "selected" : "";
+  }
+  if (var == "EXTPREG05") {
+    return settings.getUInt("EXTPREMODE", 0) == 2 ? "selected" : "";
```
M4GNV5, line 521: Probably should move this to `settings_processor` with `options_for_enum`

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -422,6 +421,15 @@ void init_webserver() {
             [boolSettingNames, stringSettingNames, uintSettingNames](AsyncWebServerRequest* request) {
               BatteryEmulatorSettingsStore settings;
 
+              if (request->hasParam("EXTPREMODE", true)) {
+                auto modeParam = request->getParam("EXTPREMODE", true);
+                int mode = atoi(modeParam->value().c_str());
+
+                settings.saveUInt("EXTPREMODE", mode);
+                settings.saveBool("EXTPRECHARGE", mode != 0);
+                settings.saveBool("EXTPREG05", mode == 2);
```
M4GNV5: do we need the last two variables? `EXTPREMODE` should be sufficient by itself?

Software/src/communication/precharge_control/precharge_control.cpp
```diff
@@ -3,10 +3,12 @@
 #include "../../datalayer/datalayer.h"
 #include "../../datalayer/datalayer_extended.h"
 #include "../../devboard/hal/hal.h"
+#include "precharge_control_i2c_g05.h"
 #include "src/battery/BATTERIES.h"
 
 // Parameters adjustable by user in Settings page
 bool precharge_control_enabled = false;
+bool precharge_control_i2c_g05_enabled = false;
```
M4GNV5: This is fine for now, but at a later stage I would add an enum or even a `Precharge` class which can be implemented for different hardware, similar how the `Battery` and `Inverter` classes work

Software/src/communication/precharge_control/precharge_control.cpp
```diff
@@ -24,6 +26,33 @@ static unsigned long prechargeStartTime = 0;
...
+static unsigned long last_i2c_update = 0;
+static float g05_tps_voltage = 4.0f;
+
+static float g05_target_to_tps_voltage(int32_t target_dV) {
+  float target_V = target_dV / 10.0f;
+
+  // Gemessene Kennlinie:
+  // TPS 4.0 V  -> ca. 175 V extern
+  // TPS 11.0 V -> ca. 500 V extern
+  float tps = 4.0f + ((target_V - 175.0f) * 7.0f / 325.0f);
```
M4GNV5: Should be translated german to english.

Also I am pretty sure these values are only correct for one specific setup and vary based on capacitance (cable length, thickness, etc.).
The best approach would be to make this scaling configurable via the webinterface.

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -170,7 +170,10 @@ void init_stored_settings() {
   user_selected_gpioopt6 = (GPIOOPT6)settings.getUInt("GPIOOPT6", 0);
 #endif
 
-  precharge_control_enabled = settings.getBool("EXTPRECHARGE", false);
+  uint8_t extprecharge_mode = settings.getUInt("EXTPREMODE", 0);
+  precharge_control_enabled = extprecharge_mode != 0;
+  precharge_control_i2c_g05_enabled = extprecharge_mode == 2;
```
M4GNV5: having only one global variable which just stores the mode as an enum would be more resilient to bugs (e.g. cannot be control disabled, while g05_enabled).

M4GNV5 (approved):

Software/src/communication/precharge_control/precharge_control.cpp
bradlrein: Yes, I forgot translate the comments to english. I will correct that soon.

  // Gemessene Kennlinie:
  // TPS 4.0 V  -> ca. 175 V extern
  // TPS 11.0 V -> ca. 500 V extern

This is the voltage without load. The cable thickness and the capacitance do not affect the output voltage. The internal resistance of the precharge circuit inside the battery is the only thing that pushes down the voltage a little bit.
The starting point for the precharge regulation is almost the target voltage, so it is not necessary to enter a scale factor. 
If we set the output voltage correctly from the start, we could completely eliminate the need for regulation, but this is prone to errors because we don't know the load on the output
For all the other points (DropdownMenu...) I agree. Feel free to make this more simple. I am still very new to vscode, git, c++, coding at all....

Software/src/communication/nvm/comm_nvm.cpp
bradlrein: A lot of try and error has been done with the dropdown menu. This still needs to be cleaned up.

bradlrein:
Hi, 
thanks a lot for the review of the code.
I tried to address all the issues

PR #2372, opened 2026-05-25
BMW i3: Fetch cellvoltages more often, split 1000ms batch in two to reduce overrun

odlg:
### What
This PR implements:
- Get min/max voltage way more often, every 5 seconds, before it was every 105 seconds. It is used for safety, it should be fetched at a higher rate.
- Split the large 1000ms batch of CAN messages in two. To reduce Task overruns.
- Only do one batch of CAN messages at a time. 

### Why
Why does it do it?
For increased safety and to reduce task overruns.

### How
How does it do it?
Fetch min/max voltage several times during fetching cell voltages, splitting the largest batch in two and only doing one batch at a time.

> [!TIP]


dalathegreat (changes requested):
I think this PR modifies way too much at once.

It is possible to request cellvoltages faster than they already are by modifying the switch (cmdState) { section only.

We have another PR up for splitting up the CAN transmits to improve performance on the i3 integration here: https://github.com/dalathegreat/Battery-Emulator/pull/2290/changes , which I think should be prioritized  over this change.

Could  you try the above PR, and if it works good, just make this PR read cellvoltages more often?

odlg:
@dalathegreat The PR is not as large as it seems. The pre-commit hook did some indenting, that I have not looked into. If you switch do not show white space changes in the diff, it is a lot smaller.

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -351,220 +352,242 @@ void BmwI3Battery::transmit_can(unsigned long currentMillis) {
...
-      transmit_can_frame(&BMW_3FC);  //Order comes from CAN logs
-      transmit_can_frame(&BMW_3C5);
-      transmit_can_frame(&BMW_3A0);
-      transmit_can_frame(&BMW_592_0);
-      transmit_can_frame(&BMW_592_1);
+            alive_counter_500ms = increment_alive_counter(alive_counter_500ms);
 
-      alive_counter_5000ms = increment_alive_counter(alive_counter_5000ms);
+            transmit_can_frame(&BMW_30B);
+          } else
```
kyberias: These elses with out { } look suspicious.

odlg: @kyberias I reformatted in a way clang likes better.

dalathegreat:
We currently seem to have a 50day millis() issue with i3 batteries https://github.com/dalathegreat/Battery-Emulator/issues/2390

Would need to investigate if we need to swap over to millis64, or if the isssue is on the battery side instead. Due  to this, I am not sure if we want to do this change yet before we get to the bottom of the issue?

odlg:
@dalathegreat Any news regarding this branch?
Just as a sidenote, I have had BMW i3 batteries combined with BE above 50 days a number of times. My experience is that it mostly continues.

dalathegreat:
@odlg , there is a merge confligt in BMW-I3-BATTERY.cpp, can you resolve it and we can get this merged?

dalathegreat (approved):

PR #2378, opened 2026-05-28
Tesla Double Battery Support + cellmonitor fix

nizmot:
### What
This PR fixes the problem that only one of the battery's cells are shown under Cellmonitor when dual/triple-tesla

### Why
Originily the first battery that would fetch the data from BMS would show in Cellmonitor

### How
Now the cell variables are not shared between all instances of the battery

> [!TIP]


dalathegreat (approved):
Nice work, thanks for testing the double-battery feature on Tesla 🙌

PR #2382, opened 2026-05-28
Add I2C display pin definitions to StarkHal

StarkJohan:
## Summary
- Adds `DISPLAY_SDA_PIN()` and `DISPLAY_SCL_PIN()` virtual methods to `StarkHal`
- Returns GPIO 12 (SDA) and GPIO 14 (SCL) when `GPIOOPT1::I2C_DISPLAY_SSD1306` is selected, otherwise `GPIO_NUM_NC`

## Test plan
- [ ] Verify I2C display (SSD1306) initializes correctly on Stark hardware using the defined pins
- [ ] Verify `GPIO_NUM_NC` is returned when the option is not selected

### DIFF
--- a/Software/src/devboard/hal/hw_stark.h
+++ b/Software/src/devboard/hal/hw_stark.h
@@ -104,6 +104,20 @@ class StarkHal : public Esp32Hal {
   virtual gpio_num_t WUP_PIN1() { return GPIO_NUM_25; }
   virtual gpio_num_t WUP_PIN2() { return GPIO_NUM_32; }
 
+  // I2C display pins
+  virtual gpio_num_t DISPLAY_SDA_PIN() {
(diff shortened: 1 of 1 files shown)


jonny5532:
The GPIOOPT1 dropdown thing isn't shown on the Stark currently, it's just on the 2CAN. Thus you would either need to:

1. Make it show the dropdown on Stark also, and probably modify the options (eg, you don't want to include WUP1/WUP2 or estop as they already have different Stark assignments? In which case you might have the 'default' just be NC, with the screen selectable as an option).
2. Just hardcode the pins for the Stark (but would permanently use up those GPIOs).

(Would be worth someone with both a Stark and a screen testing this before merging).

StarkJohan:
Option 1 is certainly preferable. I wrongly assumed the selection was generically available. 

Testing should absolutely be done before merging any new code handling peripherals.

PR #2383, opened 2026-05-28
Solax: Add MAC-derived battery identities

Dibblah1:
### What
This PR replies to the Solax inverter identity query (0x1871 command 0x05) with a full burst of 0x1881 / 0x1882 CAN frames — one pair per battery slot — instead of a single static pair.

Each frame carries a slot index and seven uppercase hex characters that identify that battery in the stack.

### Why
Solax Cloud expects distinct identifiers per slot so it can tell which physical battery is which in a multi-module stack. The previous implementation always sent one fixed 1881/1882 pair, which is not enough for stacks with more than one battery module.

### How
When command 0x05 is received, the emulator sends (configured_number_of_modules + 1) slot pairs. For each slot, bytes D2–D8 of 1881 and 1882 are filled with seven-character strings derived from the ESP eFuse MAC, the slot index, and which frame half is being sent (1881 vs 1882). Each emulator therefore gets stable, unique IDs without hard-coded serial strings.

### Testing
This has been tested with a VAST X1 connected to Solax Cloud. 
Before the batteries were shown as a single entity which randomly picked one or the other battery for statistics. 
After:
image
image

> [!TIP]

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -127,6 +127,21 @@ void SolaxInverter::transmit_can(unsigned long currentMillis) {
   // No periodic sending used on this protocol, we react only on incoming CAN messages!
 }
 
+// Write 7 uppercase hex ASCII chars (D2..D8) from eFuse MAC, slot index, and frame half (0=1881, 1=1882).
+void solax_pack_identity_ascii(const uint8_t mac[6], uint8_t slot, uint8_t half, uint8_t out[7]) {
+  static const char hex[] = "0123456789ABCDEF";
+  uint8_t mix[4] = {
+      (uint8_t)(mac[0] ^ slot ^ (half * 0x11u)),
+      (uint8_t)(mac[1] ^ slot ^ (half * 0x22u)),
+      (uint8_t)(mac[2] ^ slot ^ (half * 0x33u)),
+      (uint8_t)(mac[3] ^ slot ^ (half * 0x44u)),
+  };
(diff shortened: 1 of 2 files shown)


Software/src/inverter/SOLAX-CAN.cpp
```diff
@@ -217,8 +232,22 @@ void SolaxInverter::map_can_frame_to_variable(CAN_frame rx_frame) {
   }
 
   if (rx_frame.ID == 0x1871 && rx_frame.data.u64 == __builtin_bswap64(0x0500010000000000)) {
-    transmit_can_frame(&SOLAX_1881);
-    transmit_can_frame(&SOLAX_1882);
+    int slot_count = (int)configured_number_of_modules + 1;
```
dalathegreat: We should make a sanity check here that if the configured_number_of_modules > 254 , we cap it. Otherwise if user accidentally configures this wrong, it will send an insane amount of CAN messages!

dalathegreat (changes requested):

Software/src/inverter/SOLAX-CAN.cpp
Dibblah1: Thank you for the review! Added the sanity check here — configured_number_of_modules is capped at 254 before computing slot_count as from your patch.

Possible follow-up: enforce a stricter, battery-type-aware module limit in setup() based on real inverter voltage specs (480–800 V); I looked at that separately and is probably worthwhile but it’s out of scope for this PR.

dalathegreat (approved):

PR #2358, opened 2026-05-20
CMFA: Improve current resolution, tidy FP math

jonny5532:
### What
The CMFA integration was rounding current (down) to the nearest amp. Now it is to the nearest 0.25A.

Also change some double math to float.

(saves 104 bytes!)

Needs testing by someone with a CMFA.

> [!TIP]
> [You can help test this PR with this guide](https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#downloading-a-pull-request-build-to-test-locally-)

PR #2385, opened 2026-05-28
Zoe Gen1: preserve 0.25 A shunt resolution in reported current

Dibblah1:
### What
This PR preserves the native 0.25 A shunt resolution from the Zoe Gen1 BMS when reporting battery current.

### Why
The shunt value is decoded as 12-bit × 0.25 A/LSB with a −500 A offset. Storing it in an `int16_t` after scaling to whole amps meant |I| < 0.5 A became zero on the inverter. Same `0x155` format as CMFA/Dacia in [#2358](https://github.com/dalathegreat/Battery-Emulator/pull/2358).

### How
Store the raw `0x155` value as `LB_Current_raw` and set `current_dA` to `(((int32_t)LB_Current_raw * 10) / 4) - 5000` (2.5 dA steps), matching the merged CMFA approach.

## Testing
Tested on a Solax X1 VAST with a 22kWh Zoe Gen1.

> [!TIP]

### DIFF
--- a/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
+++ b/Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
@@ -18,7 +18,7 @@ void RenaultZoeGen1Battery::
   datalayer_battery->status.real_soc = SOC_polled;
   //datalayer_battery->status.real_soc = LB_Display_SOC; //Alternative would be to use Dash SOC%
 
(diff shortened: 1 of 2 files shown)


Dibblah1:
Reworked to match #2358: store raw `0x155` counts and use `((raw * 10) / 4) - 5000` for dA instead of float. Same 0.25 A resolution, less code and no FPU on this path.

jonny5532 (approved):
LGTM!

PR #2399, opened 2026-06-02
Add optional CYD screen support

pauLTU3:
### What
This PR adds optional CYD screen support on the Battery Emulator side.

It adds support for a CYD touchscreen client over ESP-NOW.
It supports both single and dual battery setups.

The CYD screen project is kept in a separate repository:
https://github.com/pauLTU3/CYD-Battery-Emulator

This is still a testing version and has not been fully tested on a real system yet.

### Why
This allows Battery Emulator data to be shown on a small local touchscreen display.

### How
This adds a separate CYD transport path for sending data over ESP-NOW.


dalathegreat:
Is there a technical reason why the CYD screen cannot use the standard ESP-Now checkbox?

pauLTU3:
> Is there a technical reason why the CYD screen cannot use the standard ESP-Now checkbox?

Yes. The CYD screen currently uses its own ESP-NOW packet format and data flow, separate from the existing ESP-Now feature. Keeping it as a separate option avoids changing behavior for current ESP-Now users and reduces the risk of breaking existing setups.

Software/Software.cpp
```diff
@@ -98,6 +99,9 @@ void connectivity_loop(void*) {
   if (espnow_enabled) {
     init_espnow();
   }
+  if (cyd_enabled) {
```
dalathegreat, line 102: This approach makes it possible for users to accidentally enable both espnow and cyd at the same time, should probably be made as if-elseif

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -1807,6 +1811,9 @@ const char* getCANInterfaceName(CAN_Interface interface) {
         <label>Enable ESPNow: </label>
         <input type='checkbox' name='ESPNOWENABLED' value='on' %ESPNOWENABLED% />
 
+        <label>Enable CYD: </label>
```
dalathegreat, line 1814: Could be renamed to "Enable CYD screen via ESPNow"

jonny5532:
(Note that the existing ESP-NOW code is unreliable and needs substantial changes anyway)

Is there a need to keep to the v1 ESP-NOW version? I note the 250 byte max packet size - the newer supporting 1470 byte packets.

Why does it send both 8-bit and 16-bit cell voltage values?

The remote commands (reboot, open/close contactors etc) are slightly scary but I suppose that's little different to doing the same through the webserver.

Do you have a plan for versioning? Will the protocol ever change, and what will happen to older clients if it does?

What's the effect on code size? It might need to be a larger-device-only (Stark, 2CAN) feature.

Also how long does sending take? Hopefully it will just enqueue the packets and send them later, but if not it might be necessary to stagger the sending so it doesn't unnecessarily stall the comm loop.

dalathegreat:
Closing this PR, migration towards V2 ESP-Now is ongoing in PR https://github.com/dalathegreat/Battery-Emulator/pull/2557

PR #2068, opened 2026-02-25
Add second MCP2518 interface and BECom config

jonny5532:
### What
This adds support for a second MCP2518 on the same SPI bus as the first.

It also adds the Github Actions file to build the BECom firmware.

> [!TIP]
> [You can help test this PR with this guide](https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#downloading-a-pull-request-build-to-test-locally-)

PR #2401, opened 2026-06-03
Support for LilyGo T-2CAN FD variant

jonny5532:
### What
LilyGo are now selling the FD variant of the 2CAN:
[https://lilygo.cc/products/t-2can?variant=52597538259125](https://lilygo.cc/products/t-2can?variant=52597538259125)
This code attempts to add support for this board.

### How
Attempts to read MCP2515's CANSTAT register after a reset, which will differ depending on whether it is a MCP2515 or MCP2518FD. (Note: previously tried using GPIO level tricks but this wouldn't have worked).

**The first person to get one of these boards can test this PR and report back on whether it is successful.**

This doesn't currently enable the use of an external MCP2515 on the FD board - I think it would be better to incorporate the BECom dual-FD changes and allow a second MCP2518 on the external pins, to allow 2xFD instead.

> [!TIP]

### DIFF
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -196,12 +196,24 @@ bool init_CAN() {
 
     auto cs_pin = esp32hal->MCP2517_CS();
     auto int_pin = esp32hal->MCP2517_INT();
+    auto int0_pin = esp32hal->MCP2517_INT0();
+    auto int1_pin = esp32hal->MCP2517_INT1();
 
-    if (!esp32hal->alloc_pins("CANFD", cs_pin, int_pin)) {
+    if (!esp32hal->alloc_pins("CANFD", cs_pin)) {
       return false;
     }
+    if (int_pin != GPIO_NUM_NC) {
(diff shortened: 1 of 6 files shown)


dalathegreat:
https://github.com/dalathegreat/Battery-Emulator/pull/2068 was just merged, will it affect this PR?

jonny5532:
It turned out to be slightly more complicated to get the 2CAN FD to work, since it doesn't wire the normal `INT` pin, but instead `INT0` and `INT1`. This requires:
- Modifications to the ACAN2517FD library (which didn't previously support this)
- The 'RX' DIP switch (1) on the board turning on (to connect the `INT1` pin to the ESP32's `IO3`).
- Not using BMS POWER (as this is also mapped to `IO3`) - we should move this somewhere else for the FD version.

So far one person reports success with a loopback test.

dalathegreat (approved):
Good thing we have the libraries locally for quick editing :+1: 

Feel OK to merge this, or more testers needed?

jonny5532:
Closing in favour of 2469

PR #2412, opened 2026-06-04
Waveshare: add GPIO 1/2 configurable option for status LED or I2C display

demon1300:
### What
Adds a "GPIO 1/2 function" dropdown to the Settings page for the Waveshare ESP32-S3-RS485-CAN board, letting users choose between a status LED or an I2C SSD1306 display on those two pins.  This is situated directly behind the USB C connector.

### Why
GPIO 1 and 2 on the Waveshare board can serve two different purposes — a status LED (GPIO 2) or an I2C display (GPIO 1 = SDA, GPIO 2 = SCL) — but previously the assignment was hard-coded with no way to switch without recompiling. This gives users the flexibility to use whichever hardware they have connected.

### How
Follows the existing GPIOOPT pattern used by other boards (LilyGO2CAN, Stark). Adds a new GPIOOPT6 setting stored in NVM, exposed as a dropdown in the Hardware config section of the Settings page. The Waveshare HAL reads the setting at runtime and returns the appropriate GPIO numbers for LED_PIN, DISPLAY_SDA_PIN, and DISPLAY_SCL_PIN accordingly.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -166,6 +166,9 @@ void init_stored_settings() {
 #ifdef HW_STARK
   user_selected_gpioopt5 = (GPIOOPT5)settings.getUInt("GPIOOPT5", 0);
 #endif
+#ifdef HW_WAVESHARE
+  user_selected_gpioopt6 = (GPIOOPT6)settings.getUInt("GPIOOPT6", 0);
+#endif
 
   precharge_control_enabled = settings.getBool("EXTPRECHARGE", false);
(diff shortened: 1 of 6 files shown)


demon1300 (commented):
Tested as working :)

dalathegreat (approved):

dalathegreat:
Awesome! Can you update the Waveshare wiki page with some example pic and description?

demon1300:
> Awesome! Can you update the Waveshare wiki page with some example pic and description?

@dalathegreat sure can.  I'll drop this in also :)
image

PR #2416, opened 2026-06-05
New Battery: Fisker Ocean :battery: 

dalathegreat:
### What
This PR implements support for the Fisker Ocean 113/106kWh battery

### Why
User requested feature

> [!TIP]


Sascha-T:
~~If I understand your last commit message correctly, if you're looking for the BMS on UDS, the request header is `7E1`, and a response can be expected on `7E9`~~

edit: nvm I must be misunderstanding something, you wouldn't have the PIDs without knowing this, mb

TheOfficialBatteryMan:
Can we add in all the known PIDs and the can message to 'More battery info'

I'm hoping to work on the charge and discharge current allowed comparing up with battery pack state of charge so that we can reverse engineer and have this important safety metric.

I want to make it easier for those that will be working with physical batteries rather than needing separate logging tools they can directly get the car message reported along with the state of charge and share this with us

TheOfficialBatteryMan:
IIt is suspected that the following CAN messages are responsible for keeping the battery contactors closed:

* 0x093 — xorout: 0xBB
* 0x333 — xorout: 0x34

Testing has shown that removing power from the VCU wake-up signal causes the contactors to open.

We will continue investigating to identify the CAN message or sequence that commands the contactors to open. In the meantime, the VCU wake-up supply can be controlled via a relay, providing a reliable method of opening the contactors for safety.

This finding could significantly reduce the number of CAN messages that need to be transmitted to keep the battery awake and the contactors closed, simplifying the Fisker battery emulation considerably.

TheOfficialBatteryMan:
VAGMqbEvo = 55,
Akasol = 56,
FiskerOcean = 57,
Highest

Needs updating to merge but can wait for now

PR #2427, opened 2026-06-07
Fix crash when BYD Atto3 dual-battery is enabled

demon1300:
### What
Fix crash when BYD Atto 3 dual-battery is enabled

### Why
After merging the BYD auto-calibration PR, enabling the double battery checkbox causes an immediate crash loop on every boot, taking down the web UI and MQTT. The second BydAttoBattery instance was being constructed with nullptr as its extended datalayer pointer — something that was harmless before, but handle_auto_soc_calibration now dereferences that pointer unconditionally on every update cycle, so it crashes immediately on startup.

### How
 Adds bydAtto3_2 to DataLayerExtended and passes it to the battery2 constructor, bringing BYD in line with the existing pattern already used by BoltAmpera and KiaHyundai64. A defensive null guard is kept in handle_auto_soc_calibration as a safety net. With a real pointer in place, battery2 will auto-calibrate using the struct defaults (enabled, 5% drift threshold) — secondary battery settings are not yet exposed via webserver/MQTT but will likely come with a later update.

> [!TIP]


dalathegreat (approved):
Massive thanks for the quick fix @demon1300 🙌

Issue #2422, opened 2026-06-06
BYD Double Battery Crashes 10.10.1


maddengithub:
Emulator  failed after update from 10.9.0 to 10.10.1

Fresh Install to 10.10.1 and all goes well until double battery check box is checked & saved, No webui or mqtt working

I have tested this on 3 x Lilygo boards with same result

I have now reverted back to 10.9.0 & all is good

Software version: 10.10.1
Battery used: 2 x BYD Atto 3/Seal/Dolphin ( 60kWh )
Inverter communication protocol: BYD Battery-Box Premium HVS Over CAN Bus
Hardware used for Battery-Emulator: [LilyGo T-2CAN]
GPIO controlled contactors: No
MQTT: Yes

demon1300:
@maddengithub I've just put up a PR that should solve this, if you'd like to test it?
https://github.com/dalathegreat/Battery-Emulator/pull/2427

maddengithub:
Installed PR 10.11 and it solved the problem, Thank you !

PR #2431, opened 2026-06-08
Feature: Extend BYD Atto3 auto-cal and diagnostics to second battery

demon1300:
### What
Extends the BYD Atto3's auto-calibration controls and extended diagnostics to the second battery. Each pack now has its own auto-cal enable/drift settings, calibration targets, live status panel, MQTT entities, and NVM-persisted settings, instead of those being battery-1 only.

### Why
After my auto-calibration PR went live, dual-battery BYD users hit a crash (#2422) and had no way to see or calibrate their second pack — the advanced page just showed a "limited to main battery" notice. The SOC-reset command was also hardcoded to battery 1's targets, so calibrating battery 2 would have sent the wrong values to its BMS.

### How
* Gave battery 2 its own bydAtto3_2 struct and a real pointer (was nullptr — the crash).
* Made the renderer, web routes, NVM keys, and MQTT metrics suffix-aware.
* Pointed the SOC-reset frame at the per-battery struct instead of battery 1.
* Fixed drift % being saved as UChar but read as UInt, so it survives reboot.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -653,10 +653,10 @@ void BydAttoBattery::transmit_can(unsigned long currentMillis) {
                                      0x2E,
                                      0x1F,
                                      0xFC,
-                                     (uint8_t)(datalayer_extended.bydAtto3.calibrationTargetSOC * 100),
-                                     (uint8_t)((datalayer_extended.bydAtto3.calibrationTargetSOC * 100) >> 8),
-                                     (uint8_t)(datalayer_extended.bydAtto3.calibrationTargetAH * 100),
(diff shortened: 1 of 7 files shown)


demon1300:
@maddengithub this might be worth you testing?  I've added some more functionality for 2 x BYD battery users :)

maddengithub:
I installed #2431 And all looks good, I now have 2 x Battery info & calibration on "More Battery Info" page. It's a great improvement, I will monitor operation 👍Thanks

dalathegreat (approved):

maddengithub:
image

I notice the SOC values might have a decimal place issue

demon1300:
> image
> I notice the SOC values might have a decimal place issue

Yeah those variables have always been displayed that way, not something i've changed in this round - its just the way its coming from the BMS without manipulation as its read in centi-percent (×100).  I can definitely fix it in a separate PR.

PR #2449, opened 2026-06-12
Add web interface password protection

DeadmaroZ-TLOTL:
## Summary
- add optional HTTP Basic authentication settings for the web interface and OTA endpoints
- require repeated web password entry, with a show-password toggle
- implement Basic auth handling in the vendored ESPAsyncWebServer copy

## Tests
- `platformio run -e lilygo_2CAN_330 -e lilygo_330`
- `platformio run -e compiler_warning_check`

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -5,6 +5,7 @@
 #include "../../charger/CanCharger.h"
 #include "../../communication/can/comm_can.h"
 #include "../../devboard/mqtt/mqtt.h"
+#include "../../devboard/webserver/webserver.h"
 #include "../../devboard/wifi/wifi.h"
 #include "../../inverter/INVERTERS.h"
 #include "../contactorcontrol/comm_contactorcontrol.h"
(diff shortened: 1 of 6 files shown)


dalathegreat:
Nice, we do have a large upcoming webserver refactor which deletes quite a lot of code (will also purge this change), but for the time being we could go for an interim solution

rjsc:
We had this before, but somehow got deleted during the upgrades.
Very nice to see it back. My kids are browsing stuff around the network and may change something they shouldn't.

dalathegreat (commented):
This change increases flash usage quite significantly

Before
Flash: [==========]  97.2% (used 1910828 bytes from 1966080 bytes)
After
 Flash: [==========]  97.5% (used 1916432 bytes from 1966080 bytes)

Not sure if we can reduce this in any way, I atleast don't have any suggestions on how to make the PR more streamlined

rjsc:
> This change increases flash usage quite significantly
> 
> Before Flash: [==========] 97.2% (used 1910828 bytes from 1966080 bytes) After Flash: [==========] 97.5% (used 1916432 bytes from 1966080 bytes)
> 
> Not sure if we can reduce this in any way, I atleast don't have any suggestions on how to make the PR more streamlined

If that is a concern, we just exclude this from the boards that have the SMALL_FLASH_DEVICE flag if flash space is needed later for legacy installs.

DeadmaroZ-TLOTL:
I'm thinking of getting excited and migrating this wonderful project to STM32 (STM32H723ZET6). The project is too big for the cheap ESP32.

[POWERSET_Datasheet_EN_DARK.pdf](https://github.com/user-attachments/files/28871485/POWERSET_Datasheet_EN_DARK.pdf)

dalathegreat (approved):

dalathegreat:
@DeadmaroZ-TLOTL Next generation hardware is coming with ESP32-S3 16MB HW. But there are 1000s of people using the basic ESP32, so we need to ensure backwards compability

DeadmaroZ-TLOTL:
I don't like ESP32 😂😂😂😂😂😂 but this is my problem ;-)

nagyrobi:
There are many of us who really don't like STM family so welcome onboard :)

PR #2451, opened 2026-06-12
Feature: ISO-TP library

malec1984:
This is preparation phase for enhancing MEB implementation as first step.

### What
ISOTP library and some UDS definition as helper for UDS client.
Extended addressing is also supported so it can be used in BMW
Impact on flash code size is around 1.7k

### Why
Almost all the batteries use UDS so the ISOTP is a must have to avoid implementing in each battery the same thing.

> [!TIP]


dalathegreat (approved):
I have tried to compare this with the OVMSv3 version, and both are equally confusing to me :sweat_smile: But looking forward to seeing an example integration using this :+1:

malec1984:
You can check https://github.com/dalathegreat/Battery-Emulator/pull/1941 it's already used.
My next PR is going to be for MEB, no "logical" changes, only make use of this library.

PR #2453, opened 2026-06-13
Feature: BYD Contactor Control

demon1300:
### What
Adds real software control of the BYD Atto 3 (other models to be confirmed!) contactors over CAN. You can now open and close the pack's contactors from the emulator — via the equipment-stop button or the advanced-page buttons — instead of just leaving it to the BMS. The battery web page also shows the live contactor state now.

### Why
Before this, the 0x12D frame just spat out a fixed pattern with a faked counter and checksum, so the emulator couldn't actually command the pack open or closed in a controlled way. I wanted proper open/close control — for safe shutdown, the e-stop, and bringing up crashed packs — with feedback so we actually know the contactors did what we asked.

### How
* Drives a small state machine that replays the same 0x12D payload sequence the real car uses (standby → active-ack → close → drive-ready → shutdown), captured from CAN logs.
* Reads the 0x344 feedback byte to confirm closed / precharge / HV / drive / charge state.
* Fixed the checksum to the real inverted-sum over bytes 0–6 (was a fake counter before).
* Safety: power held at zero until the pack confirms closed, won't open until current drops under ~2.5 A (with a timeout fallback), confirms both open and close with timeouts, ignores stale CAN frames, and handles booting while an e-stop is still saved.
* Opens the contactors on a system FAULT, not just the e-stop.
* Honors the inverter's contactor-closing permission, so it won't close until the inverter is ready (Solax/SMA)
* Added events and a web UI section showing the contactor command, state, mode, and a diagnostic code.

> [!TIP]


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -535,6 +562,172 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
+  // the pack state.
+  if (!equipmentStopInitialized) {
+    previousEquipmentStop = datalayer.system.info.equipment_stop_active;
+    equipmentStopInitialized = true;
+    if (previousEquipmentStop) {
+      set_12D_payload(0x50, 0x18, 0x02, 0x20, 0x04, 0x31);  // Active-ack pattern
+      contactorState = CONTACTORS_BOOT_ESTOP;
+    }
+  }
+  if (datalayer.system.info.equipment_stop_active != previousEquipmentStop) {
```
dalathegreat: Should be easy to squeeze in inverter open request here?
(datalayer.system.status.inverter_allows_contactor_closing)

Handling this also would make the system even safer, and more compatible with protocols like Solax/SMA

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -353,6 +377,8 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       break;
     case 0x344:
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
+      contactor_feedback = rx_frame.data.u8[0];
+      lastContactorFeedbackMillis = (uint32_t)millis64();
```
dalathegreat: Should be OK to just call millis() instead of millis64() here, since you are anyway comparing value towards normall millis later on

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
demon1300: should be fixed now!

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
demon1300: fixed, must have been an oversight...

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -466,6 +493,7 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
           break;
         case POLL_FOR_BATTERY_CURRENT:
           BMS_current = ((rx_frame.data.u8[5] << 8) | rx_frame.data.u8[4]) - 5000;
+          lastCurrentSampleMillis = (uint32_t)millis64();
```
dalathegreat: Here is another millis64() call

demon1300: fixed.

dalathegreat (approved):
Tested OK on the Discord on the following packs:
- Seal 82kWh 👍 
- Seal 91kWh 👍 
- Atto 3 60kWh P48 👍 

Massive success, and huge step forward in making the BYD integration safer in a FAULT scenario. Massive thanks to @demon1300 for implementing this! 🔥

PR #2459, opened 2026-06-14
Improvement/node js20deprecated

dalathegreat:
### What
This PR removes the skip-duplicate action runner

### Why
Github posted this and while investigating I think we can skip the duplicate-actions entirely

image

### How
Running on bare push and pull_request is redundant for PR branches since you'll get double runs. The skip-duplicate-actions workaround handles this, but it's cleaner to just be explicit about when you want builds.

Replaced this section
```
on:
  # The workflow is run when a commit is pushed or for a
  # Pull Request.
  - push
  - pull_request
```

With this
```
on:
  push:
    branches: [main]
  pull_request:
```

BONUS: This PR also
- Removes obsolete build becom YML file
- Adds failfast skip to build system to allow it still building even though some check fails

> [!TIP]


dalathegreat:
Looks like it works for the Common Images, but the Unit Tests and Pre-Commit are run twice

image

dalathegreat:
Before we had;
All checks have passed
12 successful checks

Now we have
All checks have passed
10 successful checks

Increases build speed due to removing need to run Unit tests / pre-commit twice :raised_hands:

PR #2460, opened 2026-06-15
Add support for bidirectional charging via CCS

M4GNV5:
### What
This PR implements a new "battery" talking to a [chargebyte CME](https://chargebyte.com/products/controllers-modules/evse-controllers/charge-module-e) or [CCF](https://chargebyte.com/products/controllers-modules/evse-controllers/charge-control-f), which handles communication with electric vehicles according to DIN SPEC 70121 and ISO15118.

### Why
This could also be used by DIYers to build bidirectional charging stations and charge/discharge "first-life" vehicles.
Additionally it could be used to work with crashed vehicles for second life use cases (needs vehicle charge controller, but removes the need of reverse engineering the battery protocol).
For us the goal of this code was to discharge vehicles as part of our research paper [published at USENIX VehicleSec'25](https://www.usenix.org/conference/vehiclesec25/presentation/low).

### How
In our test setup we used a Stark CMR, a chargebyte CCF and a 4kW Fronius Primo.
photo_2026-06-15_11-31-31
photo_2026-06-15_11-36-06

For precharging the vehicle side a MCP4725 DAC is used to generate 0-3.3V which is turned to 0-500V by a XPPower Q05-5C.
photo_2026-06-15_11-35-16

### Open Points / TODOs
- [ ] The vehicle side precharge is currently weirdly integrated. Maybe this could be combined with the MEB precharging codebase to a generic battery-side precharging
- [x] The MCP4725 library is currently included using platformio. I would say this is the way to go, but all other libs seem to be included in the repo directly?
- [x] lilygo flash size is now 99.3%. Maybe disable the new "battery" on small flash devices.

personal note: While this might sound scary at first, I got approval from both my supervisor at THI as well as chargebyte to publish this code. The above just has to be stated for funding rule reasons.

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -472,6 +472,9 @@ void core_loop(void*) {
         if (precharge_control_enabled) {
           handle_precharge_control(currentMillis);  //Drive the hia4v1 via PWM
         }
+        if (battery) {
+          battery->handle_precharge();
+        }
         END_TIME_MEASUREMENT_MAX(10ms, datalayer.system.status.time_10ms_us);
       } else {  //Run 10ms tasks without timing it
         monitor_equipment_stop_button();
@@ -479,6 +482,9 @@ void core_loop(void*) {
         handle_contactors();  // Take care of startup precharge/contactor closing
         if (precharge_control_enabled) {
           handle_precharge_control(currentMillis);  //Drive the hia4v1 via PWM
+        }
+        if (battery) {
+          battery->handle_precharge();
         }
       }
     }
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -148,8 +148,10 @@ const char* name_for_battery_type(BatteryType type) {
       return VolvoSpaBattery::Name;
     case BatteryType::VolvoSpaHybrid:
       return VolvoSpaHybridBattery::Name;
+#ifndef SMALL_FLASH_DEVICE
     case BatteryType::ChargebyteCCSBattery:
       return ChargebyteCCSBattery::Name;
+#endif
     default:
       return nullptr;
   }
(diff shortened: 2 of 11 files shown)


dalathegreat:
Even though this is feature gated away from the SMALL_FLASH_DEVICE, it somehow increases flash usage quite a bit

Before PR
Flash: [==========]  98.8% (used 1941600 bytes from 1966080 bytes)
After PR
Flash: [==========]  99.1% (used 1948736 bytes from 1966080 bytes)

Is there some IFDEF missing somewhere? Could you do some digging @M4GNV5 ?

dalathegreat:
/home/runner/work/Battery-Emulator/Battery-Emulator/test/battery/../../Software/src/battery/CHARGEBYTE-CCS.h:8:10: fatal error: Wire.h: No such file or directory
    8 | #include <Wire.h>

Wire.h needs to be stubbed in the tests, instructions on how to run unit tests locally here; https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#local-unit-test-run-

jonny5532:
The Adafruit libraries are very heavyweight for a single short I2C transaction - ripping them out and replacing with the ESP-IDF i2c_* funcs saves a lot of space (eg, https://github.com/jonny5532/Battery-Emulator/commit/229fc9043952db6b801eca8f2406790dd641ccf6 )

M4GNV5:
I updated the PR utilizing the esp-idf I2C code provided by @jonny5532, rebased it on top of `main` and made sure the unit tests compile and pass.

I assume the previous flash size increase was also caused by the Adafruit libraries. With the new update size increases by only ~280 bytes, which I assume is caused by the introduction of the new `handle_precharge` function for Batteries.

dalathegreat (approved):
Awesome work @M4GNV5 , Let's merge this and fire up a Wiki page once the hardware is available! EDIT: Maybe start something here? https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-CCS-bidirectional-charging

PR #2464, opened 2026-06-16
Feature: 24 batteries in parallel (Controller/Node system)

mbuhansen:
What
Adds an Inter-Unit Master/Slave CAN protocol that lets multiple Battery Emulator units cooperate as one logical battery toward a single inverter.

Master (battery type = Inter-Unit Master) talks to the inverter, aggregates data from all slaves, and controls each slave's contactor. Runs the protocol on its battery interface (BATTCOMM).
Slave (inverter protocol = Inter-Unit Slave) talks to one real battery, reports its data to the master, and opens/closes its contactor on master command. Runs the protocol on its inverter interface (INVCOMM).
Up to 24 slave nodes on a shared 500 kbps CAN bus — no separate inter-unit bus required.
Full documentation in [INTER-UNIT-PROTOCOL-README.md](vscode-webview://1uf0l7gn7an7bjv68lbf0c6c6jldlf94de3c5o8ukglhno8b7nb3/INTER-UNIT-PROTOCOL-README.md) (CAN ID map, message layouts, timing, safety).
Master web UI gains a colour-coded Slave Nodes section with per-slave cards (SOC, V, A, power, temp, capacity, SOH, cell voltages, contactor status, IP link, OTA state).
Why
Some installations need more pack capacity or want to combine several physically separate packs that the inverter should see as a single battery. Doing this safely requires coordinated contactor control, voltage matching before paralleling, and a single aggregated view to the inverter — which a standalone unit cannot provide. This PR makes that possible while keeping the existing standalone behaviour unchanged (role is NODE_STANDALONE unless the inter-unit battery/inverter type is selected).

How
A lightweight CAN protocol coordinates the units:

Messages: Master broadcasts a 1 s heartbeat + per-slave contactor commands. Each slave replies with STATUS, POWER, CELL, INFO, IP and IDENT frames at staggered intervals (NodeID × 5 ms delay for collision avoidance).
Safety chain (any block wins over allow):
20 s startup grace — all contactors held open so slaves can announce voltage first.
60 s offline detection — missing slave → contactor blocked + event.
Stale-data toggle bit — frozen slave (3 s) → contactor blocked.
IDENT check — firmware version + battery type must match, else blocked.
Fault-flag → event mapping — BMS/contactor/CAN-timeout faults block; cell/temp warnings are advisory.
Voltage sync before close — direct-close (diff ≤ 1.5 V for 10 s) or a direction-aware prejoin path (1.5–1.8 V while inverter is loaded) that lets current pull the idle pack into range before closing.
Aggregation: capacity/current summed, charge/discharge power gated on every slave, lowest SOC reported (with a smooth blend toward 100% near full), conservative design-voltage limits.
Config: role is derived from the battery/inverter type selection — no separate NODE_MODE setting. Slave node ID (1–24) set in the web UI.
Standalone code paths are untouched; the new sources are added to the CI host-test build (test/).

Dependencies
Requires the Kostal RS485 fix from #2463 (fix(kostal-rs485): don't call battery->update_values() on cyclic poll) to be merged first. This branch is rebased on top of current main, which already includes that commit.
Tested hardware / Limitations
This has only been tested on the BMW i3 battery, which uses a CAN-bus controlled contactor.
For any other battery to work in this system it must support contactor control via the inverter's "allow contactor close" signal, and the battery driver must report the actual contactor status back to the datalayer. Batteries that don't expose contactor control/feedback this way are not (yet) supported as slaves.


Software/src/battery/BATTERIES.cpp
```diff
@@ -1,4 +1,5 @@
 #include "BATTERIES.h"
+#include "../communication/can/MASTER-CAN.h"
```
dalathegreat: Sorry to nitpick, but we removed all master-slave references in this repo. Would it be possible to use some other terminology such as;

- Leader - Follower
- Primary - Replica
- Controller - Worker
- Server - Clients
- Leading - Subordinate
- MainBattery - OptionalBattery
- Whatever you want :)

Software/src/communication/can/MASTER-CAN.cpp
```diff
@@ -0,0 +1,867 @@
...
+    }
+    default:
+      break;
+  }
+}
+
+// ---- CAN TX --------------------------------------------------------
+
+void MasterCan::transmit(unsigned long currentMillis) {
+  if (currentMillis - _last_heartbeat_ms >= IU_HEARTBEAT_INTERVAL_MS) {
```
dalathegreat: Unnecessary to define IU_HEARTBEAT_INTERVAL_MS , just use INTERVAL_1_S

Software/src/communication/can/BATTERY-NODE-CAN.cpp
```diff
@@ -0,0 +1,280 @@
...
+  if (!datalayer.system.status.master_online) {
+    datalayer.system.status.inverter_allows_contactor_closing = false;
+  }
+
+  // === Send reply frames if due ===
+  if (_reply_pending && currentMillis >= _reply_due_ms) {
+    _reply_pending = false;
+    send_status_frame();
+    send_power_frame();
+    if (_heartbeat_count % IU_INFO_INTERVAL_HEARTBEATS == 0) {
```
dalathegreat, line 100: Same here, just use INTERVAL_10_MS

Software/src/communication/can/MASTER-CAN.cpp
```diff
@@ -0,0 +1,867 @@
...
+  uint8_t sub = (uint8_t)(offset & 0x0Fu);   // low nibble = sub-message (0..2)
+
+  if (node_id < 1 || node_id > MAX_SLAVE_NODES) {
+    return;
+  }
+
+  SLAVE_NODE_TYPE& node = datalayer.system.slave_nodes[node_id - 1];
+
+  // Reset still_alive counter (60s timeout)
+  node.still_alive = IU_OFFLINE_TIMEOUT_S;
```
dalathegreat: Why not use CAN_STILL_ALIVE which is 60 s already? Saves memory not having to define IU_OFFLINE_TIMEOUT_S

Software/src/communication/can/BATTERY-NODE-CAN.cpp
mbuhansen: IU_INFO_INTERVAL_HEARTBEATS isn't a time value — it's a count. _heartbeat_count increments once per received master heartbeat (every INTERVAL_1_S), and we do _heartbeat_count % IU_INFO_INTERVAL_HEARTBEATS to send INFO every 10th heartbeat. INTERVAL_10_MS means 10 milliseconds, so reusing it here would be misleading even though both happen to equal 10. Also, it's a #define, so it compiles to an immediate constant and uses no RAM/flash — there's nothing to save by dropping it. I'd prefer to keep the named constant for clarity.

Software/src/communication/can/MASTER-CAN.cpp
mbuhansen: Good call — applied. node.still_alive uses the exact same still-alive pattern as CAN_battery_still_alive/CAN_master_still_alive: set on message receipt, decremented at 1 Hz in the same loop, offline at 0. So CAN_STILL_ALIVE (60) is the correct shared constant and IU_OFFLINE_TIMEOUT_S was a true duplicate.

Software/src/battery/BATTERIES.cpp
mbuhansen: I have now rename the project to inter-unit controller and Battery node.

Software/src/communication/can/INTER-UNIT-PROTOCOL.h
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * IDENT Message Content (sent at startup, same timing as IP frame):
+ *   [0..1] : uint16_t  fw_version_num   — (major << 8) | minor, e.g. 10.6 -> 0x0A06
+ *   [2..3] : uint16_t  battery_type_id  — BatteryType enum cast to uint16_t
+ *   [4..7] : reserved
+ */
+
+/* ---- Firmware version encoding ---- */
+// Update these when version_number in Software.cpp changes (e.g. "10.11.dev" -> major=10, minor=11)
+#define IU_FW_VERSION_MAJOR 10u
```
jonny5532: I think the version should only be in one place, or it'll forget to be updated. Either this could parse the main version tag, or the main version tag could be split up into numeric components that this could reference.

mbuhansen: Well, that was a mistake. The intention has always been for it to use the firmware version from software.cpp, I have made a correction.

jonny5532:
I note that the charge/discharge power limits are simply the sum of all of the batteries. Notably BE doesn't do this with double/triple battery, it instead uses the maximum, since it can't guarantee that they'll share current equally.

I think this either needs to do the same, or more thought needs to be given for how this can manage a scenario such as:
- All batteries have the same charge power and are sharing the current
- One battery gets hot/develops a bad cell, and drops its current limit to 1A
- The total charge power limit is therefore reduced a bit
- However the current into each battery isn't directly controlled, so if they continue to share the current roughly equally, the failing battery is now being significantly overcharged

(A error event may catch this eventually, but it might need to react quite quickly to avoid escalation)

jonny5532:
Also, how do you start offline balancing on a battery? There doesn't seem to be anything to set the current limit to zero when you start balancing (which opens the contactors), so unless you manually stop the inverter, it will open the contactors under load?

(Similarly, I note that if the controller decides a node should have contactor_allowed=FALSE, it will set it whilst current is still flowing, which may cause it to open under load? (I'm not sure the exact sequencing here, plus it presumably won't happen in normal circumstances)).

mbuhansen:
> I note that the charge/discharge power limits are simply the sum of all of the batteries. Notably BE doesn't do this with double/triple battery, it instead uses the maximum, since it can't guarantee that they'll share current equally.
> 
> I think this either needs to do the same, or more thought needs to be given for how this can manage a scenario such as:
> 
> * All batteries have the same charge power and are sharing the current
> * One battery gets hot/develops a bad cell, and drops its current limit to 1A
> * The total charge power limit is therefore reduced a bit
> * However the current into each battery isn't directly controlled, so if they continue to share the current roughly equally, the failing battery is now being significantly overcharged
> 
> (A error event may catch this eventually, but it might need to react quite quickly to avoid escalation)

Good catch, and you're right that the naive sum isn't safe. But I dug into how BE actually handles this today, and I don't think it's quite max() either — worth getting straight before we settle on a fix.

As far as I can tell, BE doesn't combine the power/current limits at all for double/triple battery. The inverter only reads datalayer.battery.status (battery 1). If you follow the merge in Software.cpp, battery 2/3 contribute summed capacity, summed remaining Wh, summed reported current and the most-extreme SOC — but their max_charge_power_W / max_discharge_power_W are never read into the inverter path (the only places they're read are the web UI and the zeroing in safety.cpp). So in practice it just uses battery 1's limit and assumes the packs are identical. That's conservative for two equal packs, but it isn't really a max() across packs, and it uses the system poorly: three packs that each take 1.2 kW would still only charge at 1.2 kW, which triples the charge time.

Rather than copy that, I'd prefer to bound the total by the weakest pack: total_limit = active_pack_count × min(per-pack limit). The nice properties:

When the packs are healthy and roughly equal, N × min equals the sum — full capacity, no penalty.
When one pack degrades and reports its limit down (your 1A case), the total collapses immediately and proportionally — no assumption about current sharing, and no feedback loop, so it can't oscillate.
It preserves the existing "if a pack reports 0 → block all power" behaviour.
On the idea that "a faulting pack can just open its own contactor" — I'd be careful leaning on that as a safeguard, because when you look at the code it's weaker than it sounds:

The e-stop path doesn't sequence anything. When equipment_stop_active becomes true, the contactor state machine jumps straight COMPLETED → DISCONNECTED (comm_contactorcontrol.cpp, lines 199-205) and the pins are dropped immediately (lines 187-191) — there's no wait for the inverter to actually stop drawing current, so a fault-triggered open happens under load.
The code does protect against exactly this, but only in the BMS-reset path: it waits for the current to fall below ~1 A before cutting power, and otherwise aborts to avoid welding (lines 318-331), with the comment "the contactors might drop out under load and be damaged" (lines 379-381). That same protection isn't applied on the e-stop/open path.
So a pack opening its own contactor means opening under load with no handshake — risk of arcing/welding, and not something I'd want as a first line of defence. The N × min limit handles the degraded pack earlier and more gently. Happy to implement it in this PR.

mbuhansen:
> Also, how do you start offline balancing on a battery? There doesn't seem to be anything to set the current limit to zero when you start balancing (which opens the contactors), so unless you manually stop the inverter, it will open the contactors under load?
> 
> (Similarly, I note that if the controller decides a node should have contactor_allowed=FALSE, it will set it whilst current is still flowing, which may cause it to open under load? (I'm not sure the exact sequencing here, plus it presumably won't happen in normal circumstances)).

Good points, and you're right that the balancing case was a genuine gap as it stood — I've now fixed it. Some background first, since it explains the mechanism.

Offline balancing is a BMW i3-specific feature (the only driver that implements it), and it's started on the battery node itself (that node's user_requests_balancing setting), not commanded by the controller. The IU_NODE_PFLAG_BALANCING flag the node sends up is purely informational — it tells the controller to stop using that node's data and to show the state in the UI; it isn't a control path.

The bug you spotted: when a node starts balancing it does zero its own max_charge_power_W / max_discharge_power_W (BMW-I3-BATTERY.cpp lines 89-96), but the controller never honoured that, because the aggregation loop excluded balancing nodes immediately (CONTROLLER-CAN.cpp, the old if (node.balancing) continue;) before reaching the max_charge_W == 0 → charge_blocked check. So the node's zero was dropped, the aggregate was computed from the other nodes which still reported full power, the inverter kept driving current, and the contactor opened under load when the node went to sleep ~20-30 s later.

The fix: I removed that early exclusion, so a balancing node now stays in the aggregation while it still reports its contactor engaged. Its zero trips charge_blocked / discharge_blocked, which drives the whole-pack limit to 0 and lets the inverter wind the current down. Only once the node reports its contactor open (or goes offline) is it dropped — by the existing !contactor_engaged check — and the remaining healthy nodes resume. Because the zero is now honoured from the moment balancing starts, current is gone well before the node physically opens its contactor, so it's a no-load open.

The trade-off (called out in a code comment): the whole pack briefly pauses charge/discharge while a node winds down for balancing. That's intentional — the packs share one DC bus with passive current sharing, so the only way to guarantee zero current through the leaving node when its contactor opens is to stop the inverter entirely.

On contactor_allowed = FALSE mid-current (stale / voltage-mismatch / fault): you're right that this isn't sequenced either, so that node could still open under load. As you note it shouldn't happen in normal operation, and it's the same weakness as the e-stop path — there's no pause → wait-for-current-to-fall → open handshake like the BMS-reset path uses (comm_contactorcontrol.cpp lines 318-331). I've left that as-is for now since it's an abnormal-condition path, but happy to extend the same wind-down approach to it if you'd prefer.

dalathegreat:
Since this PR will essentially encourage users to build massive CAN networks with many devices, and many users have a hard time getting all the CAN requirements perfect
- Shielded wiring
- Twisted pair
- Termination only in ends
- Keeping patches short
- Avoiding CAN wiring next to AC/DC wiring
- Etc.

All this will become even more important when constructing a larger CAN network. If the users fail to construct this properly, CAN glitches/corruptions will most likely appear quite easily.

A great way to mitigate this is to add a CRC byte to each CAN message. CRC signing each sent message, and CRC checking it before reading data from it would be an awesome safety feature. I made a youtube video on this topic earlier (https://www.youtube.com/watch?v=oENNNfy5GSM) showing how OEMs use either counter/checksum/crc , and for this application a full CRC byte makes the most sense.

mbuhansen:
> OEMs use either counter/checksum/crc , and for this application a full CRC byte makes the most sense.

Yes, the biggest problem for people is to understand at canbus network, but you are right the best solution will be CRC check. I will look at it.

mbuhansen:
I have added this to Interunit protokol.

Per-frame CRC8
Every frame in the inter-unit protocol now carries a CRC8 checksum in the last data byte (data[DLC-1]):

Algorithm: CRC8 (SAE J1850, poly 0x1D) over all data bytes preceding the checksum byte.
ID seeding: The CRC is seeded with the low byte of the CAN ID, so a frame delivered on the wrong ID (aliasing / mis-delivery) also fails validation — not just bit errors in the payload.
Transmit: iu_crc_stamp() writes the checksum into the last byte before transmission.
Receive: iu_crc_valid() validates the frame before any node state is updated. On mismatch the frame is dropped, and a counter + the decoded node ID is logged. Importantly, still_alive/online are not reset on a drop, so persistent noise lets the node run down to offline and opens its contactor (fail-safe).
New max charge/discharge calculation
Previously all connected nodes' max_charge_W / max_discharge_W were summed before being sent on to the inverter. That let a strong node mask a weak one and overstated what the parallel packs can actually sustain.

Now it uses the lowest value among the connected nodes × the number of connected nodes:

Find the lowest non-zero max_charge_W / max_discharge_W among the contactor-engaged nodes.
Multiply by the number of connected nodes.
charge_blocked / discharge_blocked (a node reporting 0 W) and equipment_stop still force the whole pack to 0 — unchanged safety logic.
Example: 3 nodes at 1000 / 1200 / 900 W → before 3100 W, now 900 × 3 = 2700 W.

eriksson25:
If any extra testers is needed i can help.

Ser-Olmy:
Is this being worked on? I am putting 24 packs into a 20ft shipping container so this would be great!

PR #2469, opened 2026-06-17
Feature: Additional FD add-on on 2CAN FD, freq autodetection

jonny5532:
### What
This allows a second MCP2518FD on the 2CAN FD, connected via the same add-on pinout as with the non-FD 2CAN. 

Also adds crystal autodetection and removes the CAN/CANFD crystal frequency setting. Boards with known values (eg, Stark) skip the autodetection and use those.

Adds a second 'Classic CAN' checkbox for the second FD interface.

### Why
Should allow double-CAN-FD on the 2CAN, and make things simpler to use.

Needs testing thoroughly!

+32 bytes on T485

> [!TIP]


dalathegreat:
Tested OK on Stark+2xLEAF+Fronius setup!

dalathegreat (approved):

PR #2474, opened 2026-06-18
Experimental dynamic voltage limits for BYD (for Deye primarily)

jonny5532:
### What
This PR adds dynamically calculated voltage limits to the BYD inverter CAN integration, primarily for Deye (which otherwise overcharges the batteries).

The limits are dynamically calculated such that the inverter will hit the max pack voltage as the highest cell hits its max cell voltage (and vice versa at the min end).

Cell voltages are hardcoded at the design max-100mV (and min+100mV), so 2.8/4.2V for NMC and 2.9/3.55V for LFP, these might want revisiting.

Also there are arbitrary offsets currently, but these might want to be editable.

Also it's enabled for all BYD integration users, maybe it needs to be selectively enabled.

**TEST WITH CAUTION!** This hasn't been tested on real hardware, keep an eye on it!

> [!TIP]

### DIFF
--- a/Software/src/inverter/BYD-CAN.cpp
+++ b/Software/src/inverter/BYD-CAN.cpp
@@ -6,6 +6,19 @@
 
 /* Do not change code below unless you are sure what you are doing */
 
+#define MIN(a, b)           \
+  ({                        \
+    __typeof__(a) _a = (a); \
+    __typeof__(b) _b = (b); \
+    _a < _b ? _a : _b;      \
+  })
+#define MAX(a, b)           \
+  ({                        \
+    __typeof__(a) _a = (a); \
+    __typeof__(b) _b = (b); \
+    _a > _b ? _a : _b;      \
+  })
+
 BydVoltageLimits BydCanInverter::calculate_dynamic_voltage_limits() {
   BydVoltageLimits limits = {0, UINT16_MAX};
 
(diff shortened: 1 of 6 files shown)


Software/src/inverter/BYD-CAN.cpp
```diff
@@ -5,6 +5,98 @@
...
+
+  // for (uint8_t cell_num = 0; cell_num < datalayer.battery.info.number_of_cells; cell_num++) {
+  //   int32_t cell_voltage_mV = datalayer.battery.status.cell_voltages_mV[cell_num];
+  //   if (cell_voltage_mV == 0 || cell_voltage_mV > datalayer.battery.info.max_cell_voltage_mV) {
+  //     // Invalid cell voltage reading, bail
+  //     return UINT16_MAX;
+  //   }
+  //   cell_total_mV += cell_voltage_mV;
+  // }
+  int32_t mean_cell_voltage_mV = cell_total_mV / datalayer.battery.info.number_of_cells;
```
dalathegreat, line 44: Some integrations do;

`datalayer.battery.info.number_of_cells = 0;  //Startup with no cells, populates later when we know packsize`

Might be good to add a div0 guard here

jonny5532: There's already one at the top of the func

Issue #2359, opened 2026-05-21
[Feature request] Improve security: login, and encrypt keys and passwords


lars-boerdijk:
I think it would be a good thing to implement some security to the emulator.
Starting with a proper login (create username(s) and password for the gui-interface
Maybe encrypt the config or at least encrypt the keys and passwords.

The Emulator is all about safety on the battery and inverter-side of things , but the interface is exposed for anyone on the network.
I have already placed the emulator in a dedicated iot vlan , but that is not for everyone.

fastbike:
This should  be an optional configuration. We have a LAN that is only accessed by two users.
Maybe if you have a LAN which has a wider audience this would be better handled by subnet configuration?

lars-boerdijk:
I disagree ; such an important device as the battery emulator should have some basic security in place.

I agree that someone with more advanced networking knowledge could isolate the device on a separate subnet , etc. 
But not everyone has the required knowledge and equipment.

dalathegreat:
Ethernet only hardware is being looked into for max cybersec!

Battery-Emulator should never be connected to a wireless network that exposes it to the internet and/or users that can harm it.

rand12345:
Unless your ethernet hardware option contains a separate MCU for networking and web services it's not anything close to secure or resilient. Look at OpenEVSE hardware for ideas.

dalathegreat:
Thanks @rand12345 , I will forward this feedback to the hardware designers!

DeadmaroZ-TLOTL:
This recommendation will not in any way impair the security of BE, everyone is doing their own thing, such a feature would be very useful to me!

nagyrobi:
+1 for this as a first step to zero trust security model

nagyrobi:
https://github.com/dalathegreat/Battery-Emulator/pull/2449

nagyrobi:
Tested with release 10.11.0, which includes the user/password set on Waveshare. Set user and password, save, reboot - I can still enter the webif without it asking for password.
Power-cycled, tried private mode, it doesn't ask for a password.

DeadmaroZ-TLOTL:
Thanks for testing this. Can you confirm whether the **Enable password protection** checkbox was also checked before saving?

In v10.11.0, setting only the username/password stores the credentials, but it does not enable authentication by itself. Authentication becomes active after reboot only when:

- `Enable password protection` / `WEBAUTH` is checked
- username is non-empty
- password is non-empty

[...]

PR #2480, opened 2026-06-19
Improve web interface logout handling

DeadmaroZ-TLOTL:
## Summary
- make `/logout` issue its own HTTP Basic auth challenge and bypass global auth middleware
- navigate the Logout button to `/logout` instead of using a background XHR
- reuse a single auth realm constant and keep spacing consistent between Reboot and Logout buttons

## Testing
- `platformio run -e waveshare_330`
- `platformio run -e lilygo_330`
- tested live LilyGo T-CAN485 endpoint:
  - unauthenticated `/` and `/settings` return `401 Unauthorized`
  - authenticated `admin:bmwi3` requests to `/`, `/settings`, and `/GetFirmwareInfo` return `200 OK`
  - `/logout` returns `401 Unauthorized` with `WWW-Authenticate: Basic realm="Battery Emulator"`, `Cache-Control: no-store`, and `Connection: close`

Refs #2359

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -198,9 +198,8 @@ void init_webserver() {
   server
       .on("/logout", HTTP_GET,
           [](AsyncWebServerRequest* request) {
-            AsyncWebServerResponse* response =
-                request->beginResponse(401, "text/plain",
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
LGTM! 👍

Issue #1660, opened 2025-10-22
[Feature request] In-UI update notification


marknisk:
**Is your feature request related to a problem? Please describe.**
Currently, users may not notice when a new version of Battery Emulator is released unless they check GitHub or Discord. By showing a version update notification directly in the UI, users can easily see when an update is available and stay up to date with the latest improvements and fixes.

**Describe the solution you'd like**
Add an in-UI notification similar to EVCC, for example, next to the current version number:

“New version x.xx.x available”
This message could appear when a newer release is detected (e.g., via the GitHub API or a version file) and optionally include a link to the release page. This ensures users who regularly use Battery Emulator but don’t follow GitHub or Discord are still informed about updates.

**Describe alternatives you've considered**

Manually checking the GitHub releases page

Subscribing to GitHub release notifications
Both work but are less user-friendly and easy to miss compared to an in-app notice.

**Additional context**
The EVCC project shows a “new version available” message next to its version indicator — a proven and user-friendly way to keep users informed about updates.

dalathegreat:
Implemented in v10.12.0

PR #2481, opened 2026-06-20
feature: add webUI check/display for new versions

demon1300:
### What
Adds an in-UI "New version available" notification. When a newer release exists, a small pill appears at the top of the web UI linking to the GitHub release page.

### Why
Users who run Battery Emulator but don't follow GitHub or Discord currently miss new releases (requested in #1660). This surfaces updates where people already look — the web UI.

### How
The check runs entirely in the browser, not on the ESP32:

* The served page fetches the GitHub releases/latest API, compares the tag against the running version_number, and shows the pill only if newer.
* Result is cached in localStorage for 6h, so the 15s page auto-refresh doesn't hit GitHub's rate limit.
* Skips dev builds entirely (no fetch, no pill).
* Fails silently when offline / AP-only.
This keeps it lightweight: ~750 bytes of static HTML/JS, no TLS, no extra heap/CPU on the device. The notification inherits the page's existing responsive layout, so it scales fine on mobile.

image

> [!TIP]

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -948,6 +948,7 @@ String processor(const String& var) {
 
     // Start content block
     content += "<div style='background-color: #303E47; padding: 10px; margin-bottom: 10px; border-radius: 50px'>";
+    content += "<div id='bxUpd' style='text-align:center'></div>";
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Extremely nice work @demon1300 , massive thanks for this contribution!  🙌

PR #2483, opened 2026-06-20
SMA: startup grace period for SMA inverter CAN watchdog

lenvm:
### What
This PR adds a grace period for the SMA inverter CAN watchdog. This ensures the Battery Emulator does not throw a fault during the start-up of the SMA. The SMA may not send CAN messages during the first couple of minutes (most likely 3 minutes) of start up.

### Why
SMA inverters can take a while to boot before they start sending CAN, so the inverter watchdog (CAN_inverter_still_alive, 60s) would raise the EVENT_CAN_INVERTER_MISSING error at startup before the inverter is even up.

### How
By adding a needs_can_startup_grace() predicate to InverterProtocol (default false) and override it to true in SmaInverterBase, covering all SMA variants.

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -12,7 +12,10 @@ static bool battery_full_event_fired = false;
 static bool battery_empty_event_fired = false;
 
 #define MAX_SOH_DEVIATION_PPTT 2500
-#define CELL_CRITICAL_MV 100  // If cells go this much outside design voltage, shut battery down!
+// Some inverters take a while to boot and start sending CAN. Suppress the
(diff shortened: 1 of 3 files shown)


dalathegreat:
Was it not enough to use the "Long inverter CAN timeout" checkbox?

image

lenvm:
We don't need the timeout to be long (> 60s) all the time, it just needs to be long during startup. Making it long all the time would change how quickly the battery emulator sets a timeout for the inverter during normal operation, which I wanted to prevent.

dalathegreat (approved):
Massive thanks for the improvement @lenvm 👍

PR #2485, opened 2026-06-20
Add BYD Atto3 DTC List

demon1300:
### What
A DTC list of codes for BYD Atto 3

### Why
Offer future decoding of DTC's as per other vehicles

### How
Adds byd_atto3_dtc.json into repo location as per other manufacturers.

> [!TIP]

### DIFF
--- a/web_data/dtc/byd_atto3_dtc.json
+++ b/web_data/dtc/byd_atto3_dtc.json
@@ -3511,6 +3511,12 @@
   },
   {
     "code": null,
+    "dtc": "U0073",
+    "s_dsc": null,
+    "l_dsc": "Control Module Communication Bus \"A\" Off (generic SAE)"
+  },
+  {
+    "code": null,
     "dtc": "U011000",
     "s_dsc": null,
     "l_dsc": "Communication with Motor Control Unit Failed"
--- a/web_data/dtc/byd_atto3_dtc.json
+++ b/web_data/dtc/byd_atto3_dtc.json
@@ -1,3932 +1,657 @@
 [
-  {
-    "code": null,
-    "dtc": "B110811",
-    "s_dsc": null,
-    "l_dsc": "PM2.5 Quick Tester Short-circuited"
-  },
-  {
-    "code": null,
-    "dtc": "B110913",
-    "s_dsc": null,
-    "l_dsc": "PM2.5 Quick Tester Circuit Broken"
-  },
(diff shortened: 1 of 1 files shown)


dalathegreat (commented):
Awesome! I see there are some DTCs in this document that are not related to the battery BMS, but I guess it wont hurt to have them here. If you feel like it, removing some of the Body control module, Airbag, ABS codes, etc. might slim the file down and make it faster to load.

demon1300:
> Awesome! I see there are some DTCs in this document that are not related to the battery BMS, but I guess it wont hurt to have them here. If you feel like it, removing some of the Body control module, Airbag, ABS codes, etc. might slim the file down and make it faster to load.

Yeah i did consider this, however some of the DTC's on my own pack related to ABS module, HVAC etc not being seen etc.  I figured it would just give people as complete a picture as possible of what their pack is complaining about.

malec1984:
This JSON is incorrect, the fields that aren't used can be simply omitted or must be empty, using "NULL" is not correct.

demon1300:
> This JSON is incorrect, the fields that aren't used can be simply omitted or must be empty, using "NULL" is not correct.

point taken, however ford_machE uses JSON null for code/s_dsc; volvo_SPA has a numeric code but stuffs the literal string "NULL" into dtc/s_dsc (which is arguably worse — it actually creates a bogus "NULL" lookup key). nissan uses "" for code, which fails the validator. So across the three "no clean precedent" files, all three do something different.

Maybe we should be writing a standard for these first before adding any?  At least that way we can ensure they're consistent?

dalathegreat:
@demon1300 , sorry for that, I added lists with incorrect NULL formatting also.

Would be great to have a README.md inside the DTC folder with guide how to add more

malec1984:
I will prepare some documentation

dalathegreat:
@demon1300 , readme.md has been added in main, care to update the DTC file in this PR to match it?

demon1300:
@dalathegreat Done — updated byd_atto3_dtc.json to follow the new README.md/dtc.schema.json: dropped the unused code/s_dsc fields so each entry is just {"dtc": "...", "l_dsc": "..."}. Validates clean against both validate_dtc_json.py and dtc.schema.json. Thanks for adding the standard — much clearer now.

dalathegreat (approved):

PR #2486, opened 2026-06-21
Safety: add 20 mV hysteresis to per-cell overvoltage charge cut

demon1300:
### What
Add 20 mV hysteresis to the per-cell overvoltage charge cut in safety.cpp.

### Why
The cut zeroed charge power the instant cell_max ≥ max_cell_voltage and re-allowed it the moment it dipped below — so when a pack is held at the cell ceiling, charge toggles on/off every cycle (start/stop chatter). The pack-voltage limit right below it already uses hysteresis; the per-cell cut didn't. Applies to every battery, not just one chemistry.

### How
Latch a cell_overvoltage_charge_blocked flag when the cell hits the limit, and only release it once the max cell drops a full 20 mV below the limit — mirroring the existing HYSTERESIS_OFFSET_DV pattern for the user-set pack limit. While latched, charge power is held at 0. The CRITICAL latching event and the warning event itself are unchanged.

> [!TIP]


dalathegreat (approved):
Good improvement 👍

PR #2489, opened 2026-06-21
EOTA: pause battery before restarting

jonny5532:
### What
It seems that the automatic reboot after an OTA upgrade doesn't pause the inverter, so can open contactors under load if you don't manually pause first.

(Unless I've misunderstood how it works).

Now pauses before the 2 second delay until it reboots.

> [!TIP]


dalathegreat (approved):
Good safety improvement 👍

PR #2500, opened 2026-06-23
Nissan LEAF: Skip reading temperatures until both min&max is available on AZE0

dalathegreat:
### What
This PR skips reading temperatures until both min&max is available on AZE0 LEAF

### Why
On AZE0 the value is fetched from 0x5C0 message that is multiplexed, so it can take a long time to  get both min&max.
When ambient temperatures are close to 30*C, the default startup value can make the software incorrectly flag a BATTERY_TEMP_DEVIATION_HIGH event on poweron

### How
We wait until both values are available until we take them into use

> [!TIP]


nagyrobi:
Tested on Waveshare + LEAF generation: AZE0 battery, BATTERY_TEMP_DEVIATION_HIGH is not triggered anymore at batter emulator startup.

PR #2509, opened 2026-06-27
`SET_LIMITS` discharge limit compared against wrong (charge) setpoint

nagyrobi:
### What
The discharge branch has a copy-paste error in the condition.

### Why
The threshold compares against `max_remote_set_charge_dA` (charge) while the clamp assigns `max_remote_set_discharge_dA` (discharge). The charge branch (lines 205–206) is correct. In the common case where you set both, or set only discharge (leaving charge at 0), it usually still clamps — but if you send a high `max_charge` and a low `max_discharge`, the discharge clamp may not trigger when it should. The condition should read > `max_remote_set_discharge_dA`.

### How
The SET_LIMITS discharge branch compared max_discharge_current_dA
against max_remote_set_charge_dA instead of max_remote_set_discharge_dA,
so a high max_charge with a low max_discharge could skip the discharge
clamp. Use the discharge setpoint, matching the charge branch.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -217,7 +217,7 @@ void update_calculated_values(unsigned long currentMillis) {
 
   /* Apply remote restrictions if set*/
   if (datalayer.battery.settings.remote_settings_limit_discharge) {
-    if (datalayer.battery.status.max_discharge_current_dA > datalayer.battery.settings.max_remote_set_charge_dA) {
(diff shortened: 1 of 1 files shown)


nagyrobi:
Tested on LilyGo T-CAN485 and working properly now.
I used [manually configured number entities](https://github.com/dalathegreat/Battery-Emulator/wiki/Home-Assistant#set_limits-user-interface) in HA in order to thoroughly test behavior.

image
image

dalathegreat (approved):

PR #2510, opened 2026-06-27
MQTT: Set suggested display precision to 3 and proper icon for cell voltages in HA discovery

nagyrobi:
### What
With "Send all cellvoltages via MQTT" enabled, each cell is auto-discovered as a voltage sensor in volts (e.g. 3.779). Home Assistant's default display precision for these entities shows no decimals, so every cell reads as 4 V, making the values useless at a glance. Same situation with `cell_max_voltage` and `cell_min_voltage`.
Additionally, the entities errorneously show a sine wave icon because for voltage, if unspecified, that's the default in Home Assistant.

### Why
At cell-voltage transmission with HA auto-discovery; `homeassistant/sensor/<topic>/cell_voltageN/config` now contains `"suggested_display_precision": 3` and cells display as e.g. 3.779 V in HA on fresh discovery. Adding icon = "mdi:current-dc" for proper display.

### How
Add `suggested_display_precision: 3` and `mdi:current-dc` icon to the cell-voltage discovery payload in `set_battery_voltage_attributes()`. 
Match only `cell_max_voltage` and `cell_min_voltage` (and their _2 double-battery variants, since those entity IDs are built by appending _2). No SensorConfig struct change, so no need to touch the other ~25 template rows — keeps the diff to this one block and avoids accidental scope creep.
strncmp/strlen need no new includes; already used throughout the file.

Before:
image

After:
image

### Note
One thing to note: suggested_display_precision only sets the default display rounding for new entities. For cells that already exist in HA, the retained discovery config gets re-published on the next MQTT connect, but HA won't retroactively change an entity whose display precision a user has already touched. For untouched existing entities the new default applies once the updated config is processed; if any look stuck, deleting the device/entities and letting them re-discover forces a clean pickup.
Icons similarly because these entities have device_class: voltage, Home Assistant normally picks the icon from the device class. A custom icon in the discovery config does override that default, but only for entities that don't already have a user-set icon — and as with precision, it applies cleanly on fresh discovery; existing entities pick it up when the retained config is re-published unless a user has customized the icon.

> [!TIP]

### DIFF
--- a/Software/src/devboard/mqtt/mqtt.cpp
+++ b/Software/src/devboard/mqtt/mqtt.cpp
@@ -234,6 +234,7 @@ void set_battery_voltage_attributes(JsonDocument& doc, int i, int cellNumber, co
   doc["state_class"] = "measurement";
   doc["state_topic"] = state_topic;
   doc["unit_of_measurement"] = "V";
+  doc["suggested_display_precision"] = 3;
   doc["value_template"] = "{{ value_json.cell_voltages[" + String(i) + "] }}";
 }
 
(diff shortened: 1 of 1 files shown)


nagyrobi:
Tested on a LilyGo T-CAN485 works as expected.

Quick test: to ensure no previous instance data was cached in HA and the broker, use a new _MQTT topic name_, _Prefix for MQTT object ID_, _HA device name_ and _HA device ID_ in BE _Customized MQTT topics_ settings, it will be autodiscovered as a new device showing proper precisions.

To fix for a device already in HA: 
1. Clear the retained discovery messages on the broker for that prefix. Eg. MQTT Explorer: connect to the broker, expand homeassistant/sensor/<original_topic> and homeassistant/button/<original_topic>, and delete those retained topics (select → delete; it publishes an empty retained payload). Empty config payloads tell HA to remove the entities.
2. In HA: Settings → Devices & Services → MQTT → delete the device, and restart HA
3. Reboot BE (or power-cycle). This resets the ha_*_published flags so BE republishes fresh discovery configs (now with suggested_display_precision: 3 and the icon) under the original IDs. HA recreates the entities correctly.

dalathegreat (approved):
Massive thanks for this improvement, and great to have the clear steps, will include those in the release notes!

nagyrobi:
#2508 is a similar fix worth considering

PR #2514, opened 2026-06-28
Add secondary battery contactor pin

dalathegreat:
## What
This PR adds secondary battery contactor pin definition for Waveshare boards

## Why
User requested feature. Enables double gpio battery setups with waveshare


nagyrobi:
And how about GPIO_NUM_1 for the third battery? It’s free when using the RGB Led. Safe to use, not even a strapping pin.

dalathegreat:
@nagyrobi Third battery will never be possible on this hardware, since it only has max 2 CAN channels

PR #2519, opened 2026-06-30
Feature: BYD live DTC read & erase

demon1300:
### What
Adds live DTC read & erase for the BYD Atto 3/Seal/Dolphin in the web UI. Reads the fault codes off the BMS, shows them with descriptions, and lets you clear them.

### Why
So you can actually see (and clear) what your pack is complaining about straight from the web UI, same as the BMW packs already do. Pairs up with the DTC list added in #2485.

### How
* "Read DTC" button asks the BMS for stored codes (UDS 19 02), reassembles the multi-frame reply and decodes each one back to its P/C/B/U string.
* Shows them in a table (active first, then stored, sorted by code) and pulls the descriptions from byd_atto3_dtc.json — same loader the BMW pages use.
* "Erase DTC" button clears them on the BMS (UDS 14 FF FF FF) then auto re-reads so the table updates itself.
* Both buttons are the existing generic battery commands — just flipped on supports_read_DTC / supports_reset_DTC for BYD, no new web routes.
* Added a small reload_after flag to the generic command buttons so the page refreshes itself after a read/erase (BMW's Read DTC picks this up too).

Heads up: the BYD erase uses the same 14 FF FF FF clear that the "Unlock crashed BMS" button already sends, so it'll also clear a crash code — they overlap on purpose, not worth a separate clear.  Can look to remove this at a later stage if required.

image

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -281,6 +281,25 @@ void BydAttoBattery::
       stateMachineCalibrateSOC = STARTED;
       datalayer_bydatto->UserRequestCalibrateSOC = false;
     }
+
+    if (datalayer_bydatto->UserRequestDTCreadout && stateMachineReadDTC == NOT_RUNNING) {
+      stateMachineReadDTC = STARTED;
+      datalayer_bydatto->dtc_read_in_progress = true;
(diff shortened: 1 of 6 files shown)


dalathegreat (approved):
Extremely nice work @demon1300 , merging this 🙌

PR #2525, opened 2026-07-02
Safety: ⚠️ hold BMS_POWER across reset/OTA reboot

nagyrobi:
### What
Improves safety by adding a generic per-board mechanism to latch BMS_POWER GPIO pads at their driven level across a firmware-initiated reboot/OTA, so they don't float during the boot window. This relies on this property of the ESP32 and ESP32-S3 (the RTCIO block).

[Espressif documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/gpio.html#_CPPv412gpio_hold_en10gpio_num_t) statement:

> "When a GPIO is set to hold, its state is latched at that moment and will not change when the internal signal or the IO MUX/GPIO configuration is modified... This function can be used to retain the state of GPIOs when the power domain of where GPIO/IOMUX belongs to becomes off. For example, chip or system is reset (e.g. watchdog time-out, Deep-sleep events are triggered), or peripheral power-down in Light-sleep."

### Why
If "Periodic BMS reset every 24h" or "Remote BMS reset via MQTT allowed" is enabled, `BMS_POWER` floats during a reboot/OTA, potentially briefly power-cycling the BMS. BMS_POWER is often placed on RTC-capable on the ESP32 and ESP32-S3, so its level can be latched across a CPU reset — this is especially useful in an eventual bootloop, where the BMS power could be toggled off/on indefinitely - potentially damaging it.

[Espressif datasheet](https://documentation.espressif.com/esp32-s3_datasheet_en.pdf) page 18 Table 2-2, **Power-Up Glitches on Pins**:
>  Low-level glitch: the pin is at a low level output status during the time period

### How
- Before `graceful_restart();` / an OTA reboot, `gpio_hold_en(GPIO_NUM_x)` (+ `gpio_deep_sleep_hold_en()`) latches the pad at its current level.
- On the next boot, after `init_contactors()` re-drives the pin to its intended level, `gpio_hold_dis(GPIO_NUM_x)` releases the hold so the pad adopts that level cleanly.

### Notes
- The hold latches **whichever level is currently driven** (HIGH or LOW) — it's not a "pull HIGH" mechanism. Enable the hold only after the pin is at the desired level; the hold overrides any write until released.
- Persisting through a reset (not just deep sleep) requires the RTC domain to stay powered, so `gpio_deep_sleep_hold_en()` is needed alongside `gpio_hold_en()`.
- **Does not survive a full power-cycle** — only the RTC domain being kept alive across a CPU/OTA reset.
- **Known limitation:** if a reboot happens mid-way through the 30 s BMS-off cycle, BMS_POWER pin is latched the other way across the reboot, then `init_contactors()` drives it back again — so the BMS ends up powered back on with no glitch, but the in-progress reset is silently truncated/abandoned (its state lives in RAM and doesn't survive the reboot). This is a pre-existing limitation of the reset state machine, not something this PR introduces or worsens.

[Periodic BMS reset](https://github.com/dalathegreat/Battery-Emulator/wiki/Periodic-BMS-reset) wiki page will need to be updated with details for pin choice recommendation and explanation of the functionality.

> [!TIP]


nagyrobi:
Refactored so that hold is conditional, release is unconditional. A pin is only latched when the firmware is actively driving it (for BMS_POWER, when periodic/remote BMS reset is enabled or the board forces it). Every candidate pin is released on every boot, so disabling the feature and rebooting can never leave GPIO6 stuck latched from a prior session.

nagyrobi:
Feasibility comparison for all the boards:

| Board (ESP32 variant) | BMS_POWER GPIO(s) | Strapping role | Real effect if latched HIGH across reset |
|---|---|---|---|
| Waveshare (ESP32-S3) | 6 | none | none — safe |
| BECom (ESP32-S3) | 1 | none | none — safe |
| LilyGo T-2CAN (ESP32-S3) | 3 (default)<br>2 (option) | 3: JTAG select strap<br>2: none | 3: does not block boot, only affects JTAG routing — effectively safe<br>2: none — safe |
| LilyGo T-CAN485 (ESP32) | 18 (default)<br>25 (option) | 18: none<br>25: none | 18: not RTC-capable — cannot be latched<br>25: safe |
| Stark CMR (ESP32) | 23 (default)<br>25 (option) | 23: none<br>25: none | 23: not RTC-capable — cannot be latched<br>25: safe |
| 3LB (ESP32) | 2 | download-mode entry strap | normal boot works; a latched HIGH can block entering serial-download mode until a power-cycle clears it |

nagyrobi:
Changed to draft until https://github.com/dalathegreat/Battery-Emulator/pull/2551 will be merged, will rebase after

nagyrobi:
Enabled this feature for the boards as below:
- Waveshare: GPIO6
- BECom: GPIO1
- LilyGo T-2CAN: GPIO3 and 2 (both the default and the option)
- LilyGo T-CAN485: GPIO25 (default 18 not supported)
- Stark CMR: GPIO25 (default 23 not supported)

nagyrobi:
Tested on Waveshare and T-CAN485, connecting to BMS_POWER GPIO a LED that showed that GPIO level remains untouched during reboot, and the Reset BMS command properly turns it off and back on in accordance with the settings.

nagyrobi:
@dalathegreat [Discord Discussion](https://discord.com/channels/775038294655762433/1506020692871741524/1522277159178342571)

dalathegreat (approved):
Great improvement. However, I forsee that we will need to add a way to actuate BMS_POWER on demand from the webserver, now that the reboot no longer cycles it on-off. I think it was used by some integration to powercycle the BMS remotely

(Example, Stark CMR, BMS power always on, except for when booting)

This can be done in a later PR once we figure out which component/board requires it

PR #2536, opened 2026-07-05
Double/Triple battery: Add hysteresis to out of sync voltage event

dalathegreat:
### What
This PR implements hysteresis to out of sync voltage event

### Why
User report on Discord, spamming of this event occasionally;
```
 591.001 Event: Too large voltage diff between the batteries. Third battery cannot join the DC-link
 624.001 Event: Too large voltage diff between the batteries. Second battery cannot join the DC-link
 644.003 Event: Too large voltage diff between the batteries. Second battery cannot join the DC-link
 900.003 Event: Too large voltage diff between the batteries. Third battery cannot join the DC-link
 902.003 Event: Too large voltage diff between the batteries. Third battery cannot join the DC-link
 909.003 Event: Too large voltage diff between the batteries. Third battery cannot join the DC-link
```

### How
Instead of raising the event if we are 1.5V out of sync for over 1s, we now need to be out of sync for 3seconds before it fires. This should remove any spam.

> [!TIP]

### DIFF
--- a/Software/src/devboard/safety/parallel_safety.cpp
+++ b/Software/src/devboard/safety/parallel_safety.cpp
@@ -22,12 +22,14 @@ void check_parallel_battery_safety(uint8_t batteryNumber) {
         datalayer.system.status.battery2_allowed_contactor_closing = true;
       }
     } else {  //Voltage between the two packs is too large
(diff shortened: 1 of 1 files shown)


dalathegreat:
Tested OK on the Discorod!

PR #2537, opened 2026-07-05
Security enhancement(wifi): option to disable the access point after connecting to the local network

### LINKED ISSUE #799: new feature request: AP turn off when successfully connected to Wi-Fi
### Expected Behavior

Turn off battery emulator AP when successfully connected to Wi-Fi
and turn on after x seconds when connection to Wi-Fi is lost.
(in such way by default behaves ESPHome code in HomeAsistant)

### Actual Behavior

 if such configuration in USER_SETTINGS.h enabled:
   #define WIFI
   #define WIFIAP 

 and if LILYGO is successfully connected to configured Wi-Fi, 
 then AP is not needed - risk for improper Wi-Fi work because AP uses network frequency

### Settings
 
 - Software version: `8.1`
 - Hardware used for Battery-Emulator: `HW_LILYGO`

nagyrobi:
### What
Add a settings-page checkbox "Disable access point after connecting to local network" (checked by default, persisted as NVS key APOFFONCONN). When AP broadcast is enabled and this option is on, the soft-AP stays up while the device cannot join the configured STA network, and is shut down (station-only mode) once STA obtains an IP.

### Why
Increase security at the sites where BE is deployed, by offering the option to disable AP after connection the local network. It is by default enabled, to increase user awareness.

This gives a captive-AP-style bootstrap experience: the AP stays available while the device can't reach the network (so you can configure it), and disappears once it's on the LAN.
Original behavior (keep both STA and AP modes) can be restored by clearing the checkbox and saving the settings+reboot.

Additional benefit is freeing the runtime resources of the seldomly used AP.

Fixes #799

### How
When "Broadcast Wifi access point" is enabled and this option is on, the device keeps its soft-AP running after boot for as long as it cannot join the configured station (STA) network. As soon as it successfully connects and obtains an IP, the AP is shut down and the device drops to station-only mode. While the device has not yet connected since boot, the AP stays available. Once STA obtains an IP the AP is shut down. 

- Boot, with AP broadcast option enabled: AP starts and stays up.
- While STA can't connect: no GOT_IP fires, so the AP remains available; the existing full-reconnect path also re-asserts the AP while the device has never connected.
- On first successful IP: onWifiGotIP runs once and, if enabled, tears the AP down and switches to station-only. ap_active prevents repeated toggling.
- Option unchecked: previous always-on-AP behavior is preserved.
- Option disabled / AP broadcast disabled: no change from current behavior.
- The unlikely event that the configured local networ SSID never comes back (eg. someone steals the AP or reconfigures the network to a different SSID) manual intervention is needed anyway, thus a BE reboot will restore access through the AP.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -195,6 +195,7 @@ void init_stored_settings() {
 
   // WIFI AP is enabled by default unless disabled in the settings
   wifiap_enabled = settings.getBool("WIFIAPENABLED", true);
+  wifi_ap_disable_after_connect = settings.getBool("APOFFONCONN", true);
   wifi_channel = settings.getUInt("WIFICHANNEL", 0);
   ssidAP = settings.getString("APNAME", "BatteryEmulator").c_str();
   passwordAP = settings.getString("APPASSWORD", "123456789").c_str();
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -521,6 +521,10 @@ String raw_settings_processor(const String& var, BatteryEmulatorSettingsStore& s
     return settings.getBool("WIFIAPENABLED", wifiap_enabled) ? "checked" : "";
   }
 
+  if (var == "APOFFONCONN") {
+    return settings.getBool("APOFFONCONN", true) ? "checked" : "";  // default: checked
+  }
+  
   if (var == "APPASSWORD") {
     return String("");
   }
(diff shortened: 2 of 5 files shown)


nagyrobi:
Successful connection to wifi network:
```
       0.214 init_Wifi enabled=1, ap=1, ssid=IOT_LOCAL
       0.215 transmitter registered, total: 1
       0.215 CAN receiver registered, total: 1
       0.217 Native Can ok
       0.217 Bit Rate prescaler: 4
       0.217 Time Segment 1:     13
       0.217 Time Segment 2:     6
       0.217 RJW:                4
       0.217 Triple Sampling:    no
       0.217 Actual bit rate:    500000 bit/s
       0.217 Exact bit rate ?    yes
       0.217 Sample point:       70%
       0.217 Event: The board was reset via software, webserver or OTA. Normal operation
       0.239 MQTT initialized successfully.
       0.239 Setup complete!
       0.319 Creating Access Point: BatteryEmulatorDog
       0.319 Access Point password is set (13 characters)
       0.319 Access Point created. IP address: 192.168.4.1
       0.319 start Wifi
       0.320 Connecting to Wi-Fi...
       0.320 Connecting to Wi-Fi SSID: IOT_LOCAL, Channel: 0
       0.321 init_Wifi complete
       0.385 Event: Wifi connected.
       0.386 Wi-Fi connected. status: 0, RSSI: 40 dBm, IP address: 0.0.0.0, SSID: IOT_LOCAL
       1.002 Event: Successfully communicating with battery. Battery detected!
       1.397 Wi-Fi Got IP. IP address: 192.168.5.219
       1.398 Connected to local network, disabling Wi-Fi access point.
       1.715 MQTT initialized
       1.768 Event: MQTT connected.
       1.768 Publishing buttons discovery
       1.779 MQTT connected
```

PR #2541, opened 2026-07-05
feat(wifi): long-press board BOOT button to start AP or factory reset

nagyrobi:
### What
Three runtime actions triggered by long-pressing the board button (typically the BOOT button on GPIO0, opt-in per board via the HAL). All actions are decided on button release, based on how long it was held, and the thresholds are checked highest-first so the tiers never overlap or double-fire. The color led blinks white at various speeeds to confirm the selected stage.

| Hold duration | Action | Visual feedback (white) |
| --------------|--------|-----------------|
| n≥ 5 s (and < 15 s) | Start the Wi-Fi access point | Slow blink (400ms) |
| ≥ 15 s (and < 30 s) | Wipe Wi-Fi client settings and reboot with AP on | Mid blink (200ms) |
| ≥ 30 s | Full factory reset and reboot | Fast blink (100ms) |

### Why
GPIO0/BOOT button has no other use at runtime. There are use cases which could benefit from having a way to physically interact with BE:

- The AP is the device's out-of-band access path: if the configured client network is unreachable or the user doesn't know the device's LAN address, a physical way to bring the AP up on demand lets them reach the web interface to reconfigure — without needing serial access or a reflash. This would also encourage people to turn the AP mode off for everyday use and only bring it up when they need it, easily.
- A full factory reset is too blunt when the only problem is wrong network credentials — a mistyped SSID/password or a static IP that doesn't fit the new network. This  is a targeted recovery: fix just the STA connection data and keep everything else (battery/inverter config, AP setup, MQTT, hostname, etc.) intact. After the wipe the device sits with no SSID, so the STA side goes idle cleanly (the connect/monitor code early-returns on an empty SSID — no reconnect churn), and the user reconfigures via the AP.
- Provides a physical, UI-independent recovery of last resort — usable when the device is misconfigured even to the point of being unreachable through the web interface.

### How
- Opt-in per board. GPIO0 is a strapping pin and only usually a button, so the pin is declared per board via AP_BUTTON_PIN(); boards that don't override it get none of these features.
- Fire-on-release + highest-threshold-first. This is what makes the three tiers safe and mutually exclusive, and prevents accidental destructive actions from a held/stuck pin.
- No hold feedback. Consistent with a single silent button, there's no on-device countdown; users can't see the 5/15/30 s boundaries. The two destructive tiers are 15 s apart to reduce mis-triggering.
- Reachability after a wipe/reset. Both reboot into a state reachable via the AP — at boot if AP broadcast is enabled, otherwise via the 5 s long-press.

### Which boards
These boards have the button for GPIO0 soldered from factory, thus this feature is activated:
- Waveshare ESP32‐S3‐RS485‐CAN
- LilyGo T-2CAN (and FD)
- LilyGo T-CAN485
- ESP32 Devkit V1
- Stark CMR
- BECom

> [!TIP]

### DIFF
--- a/Software/src/devboard/wifi/wifi.cpp
+++ b/Software/src/devboard/wifi/wifi.cpp
@@ -56,6 +56,11 @@ static uint16_t current_full_reconnect_interval = INIT_WIFI_FULL_RECONNECT_INTER
 static uint16_t current_check_interval = WIFI_CHECK_INTERVAL;
 static bool connected_once = false;
 
+static bool ap_button_inited = false;
+static bool ap_button_was_pressed = false;
+static unsigned long ap_button_press_start = 0;
+static const unsigned long AP_BUTTON_HOLD_MS = 5000;  // 5-second long-press to start AP
+
 void init_WiFi() {
   DEBUG_PRINTF("init_Wifi enabled=%d, ap=%d, ssid=%s\n", wifi_enabled, wifiap_enabled, ssid.c_str());
 
@@ -100,6 +105,8 @@ void init_WiFi() {
 
 // Task to monitor Wi-Fi status and handle reconnections
 void wifi_monitor() {
+  check_ap_button();
+
   if (ssid.empty() || password.empty()) {
     return;
   }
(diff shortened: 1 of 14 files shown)


nagyrobi:
On LilyGo T-CAN485:
```
       0.189 init_Wifi enabled=1, ap=0, ssid=MYNET
       0.189 Event: The board was reset via software, webserver or OTA. Normal operation
       0.190 Setup complete!
       0.279 start Wifi
       0.279 Connecting to Wi-Fi...
       0.280 Connecting to Wi-Fi SSID: MYNET, Channel: 0
       0.281 init_Wifi complete
       0.314 Event: Wifi connected.
       0.315 Wi-Fi connected. status: 0, RSSI: 83 dBm, IP address: 0.0.0.0, SSID: MYNET
       1.332 Wi-Fi Got IP. IP address: 192.168.5.219
      36.410 AP button long-press: starting Wi-Fi access point.
      36.419 Creating Access Point: BatteryEmulator
      36.419 Access Point password is set (9 characters)
      36.420 Access Point created. IP address: 192.168.4.1
```

The top banner also contains an extra row of info that the AP is left activated:

Vágólap_07-05-2026_01

Software/src/devboard/wifi/wifi.cpp
```diff
@@ -98,8 +109,69 @@ void init_WiFi() {
...
+      set_led_override(true, LED_COLOR_WHITE, 400);  // >=5 s
+    } else {
+      set_led_override(false, 0, 0);  // <5 s: no feedback yet
+    }
+  } else if (!pressed && ap_button_was_pressed) {
+    // Released: act based on how long it was held.
+    set_led_override(false, 0, 0);  // released: stop blink feedback
+    const unsigned long held = now - ap_button_press_start;
+    if (held >= AP_BUTTON_FACTORY_RESET_MS) {
+      logging.println("Button held >=30 s: performing factory reset and rebooting.");
```
dalathegreat: To keep codesize minimal, we can remove these logging prints (nobody will ever see the factory reset one other than now during development testing)

nagyrobi: Done

dalathegreat (approved):

PR #2551, opened 2026-07-06
Do a graceful restart for reboot/MQTT/OTA, keeping CAN active

jonny5532:
### What
Currently, reboots via OTA, WiFi and mqtt send pause_CAN=true, which can result in the inverter being paused before the power=0W restriction is sent.

Add a new `graceful_restart()` function which stops charge/discharge (but leaves CAN operating, equipment stop unasserted, and is not persistent).

Add a tick function to the main loop which will then restart the ESP after 2s once PAUSED (so the current has fallen), or after 5s regardless.

Note: the pause_CAN argument is no longer used anywhere - should it be, or can it be removed?

> [!TIP]

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -25,6 +25,7 @@ static bool battery_empty_event_fired = false;
 bool emulator_pause_request_ON = false;
 bool emulator_pause_CAN_send_ON = false;
 bool allowed_to_send_CAN = true;
+static uint32_t emulator_restart_request_millis = 0;
 
 //component detection
 bool battery_detected = false;
@@ -606,6 +607,32 @@ void setBatteryPause(bool pause_battery, bool pause_CAN, EquipmentStop equipment
 
   //immediate check if we can send CAN messages
   update_pause_state();
+}
+
+void graceful_restart() {
+  // Pause charge/discharge, and then restart the ESP32 within 5s (as soon as the power stops).
+
+  set_event(EVENT_RESTARTING, 0);
+
+  // Stop charge/discharge so we don't damage the contactors
+  setBatteryPause(true, false, EquipmentStop::UNCHANGED, false);
+
+  uint32_t now = millis();
+  emulator_restart_request_millis = now > 0 ? now : 1;
+}
+
+void update_restart_progress() {
+  // If is a restart has been requested, check the time and restart if the
+  // conditions are met.
+
+  if (emulator_restart_request_millis > 0) {
+    uint32_t now = millis();
+    uint32_t elapsed = now - emulator_restart_request_millis;
+    // Restart after 2s if the emulator has paused. Always restart after 5s.
+    if ((elapsed > INTERVAL_2_S && emulator_pause_status == PAUSED) || elapsed > INTERVAL_5_S) {
+      ESP.restart();
+    }
+  }
 }
 
 /// @brief handle emulator pause status and CAN sending allowed
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -628,7 +628,7 @@ void update_restart_progress() {
   if (emulator_restart_request_millis > 0) {
     uint32_t now = millis();
     uint32_t elapsed = now - emulator_restart_request_millis;
-    // Restart after 2s if the emulator has paused. Always restart after 5s.
+    // Restart after 5s if the emulator has paused. Always restart after 10s.
     if ((elapsed > INTERVAL_5_S && emulator_pause_status == PAUSED) || elapsed > INTERVAL_10_S) {
       ESP.restart();
     }
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -629,7 +629,7 @@ void update_restart_progress() {
     uint32_t now = millis();
     uint32_t elapsed = now - emulator_restart_request_millis;
     // Restart after 2s if the emulator has paused. Always restart after 5s.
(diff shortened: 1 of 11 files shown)


nagyrobi:
I'll definitely try this out!

Afaik pause_CAN needs to be used during Nissan's [SOH reset](https://github.com/dalathegreat/Battery-Emulator/wiki/Battery:-Nissan-LEAF---e%E2%80%90NV200#notes-on-30kwh-battery).

nagyrobi:
Tested. A reboot from the webpage still stops the inverter, while the OTA does not.

The other difference between the two that I see is that during OTA the status is being put to UPDATING (5 from system_status_enum), the LED is blue.

Apparently this is communicated towards the Fronius inverter too, as in its webpage battery status shows as "Updating", and goes back to "Active" afte BE boots normally - without stopping and starting.

We should probably fake this from within graceful_restart() towards Fronius.

jonny5532:
Now adds a new RESTARTING event which sets the same UPDATING state that OTA uses.

A bit strange perhaps, but interested to see if this fixes Fronius reboots.

There are some other new changes in here which i've also separated out as separate PRs (will rebase later).

nagyrobi:
Works! 

**Before and after OTA** (normal operation, also BE PAUSED state:
image

**With this patch**, during OTA and restart:
image

**Previously during restart**: (took 5 minutes, with no PV and battery generation):
image

Now it doesn't stop and start the inverter after rebooting the emulator.

Not strange at all, I was thinking similarly to set a global bool when graceful restart is in progress, and update Fronius-only based on that. Here I thought I could add just an if block to report updating instead of datalayer.system.status.system_status when the bool is true
https://github.com/dalathegreat/Battery-Emulator/blob/c612ce7adba65090e52d82cd443b0b91e7f06cc7/Software/src/inverter/BYD-MODBUS.cpp#L87

But your solution is better, could be beneficial for other inverters too.

Would be nice to print a log entry, did the current/power really go to 0 before reboot. I know for the moment that would be hard to catch but this is rather important, we can monitor through USB serial during reboot (and I think of adding support for a network syslog server).

Opened contactors from web interface (EQUIPMENT_STOP) after PAUSED:
image

Closed contactors from web interface:
image

And naturally after that, in RUNNING state:
image

This confirms that hard options still work when deliberately triggered from the web ui.

dalathegreat (approved):
LGTM 👍

Software/src/devboard/safety/safety.cpp
```diff
@@ -578,6 +579,32 @@ void setBatteryPause(bool pause_battery, bool pause_CAN, EquipmentStop equipment
...
+
+void update_restart_progress() {
+  // If is a restart has been requested, check the time and restart if the
+  // conditions are met.
+
+  if (emulator_restart_request_millis > 0) {
+    uint32_t now = millis();
+    uint32_t elapsed = now - emulator_restart_request_millis;
+    // Restart after 2s if the emulator has paused. Always restart after 5s.
+    if ((elapsed > 2000 && emulator_pause_status == PAUSED) || elapsed > 5000) {
```
dalathegreat: I like using INTERVAL_2_S & INTERVAL_5_S instead of raw values!

nagyrobi: @jonny5532

jonny5532: changed!

nagyrobi:
In happened once that inverter didn't go into UPDATING mode while I restarted BE.

Turns out that GEN24 [only polls every 5 seconds](https://gitlab.com/pelle8/inverter_resources/-/blob/main/byd_registers_modbus_rtu.md) for status, max charge/discharge power. So a grace of 5 seconds exactly is right on the edge, it may very well occur a race when the poll didn't make it on time. 

I'd say ideally the grace period should start from the moment we know that the inverted asked for these data. If we can't know that, we should be safer, let's extend this to 11 seconds (let 2 polls happen max) or at least 7 seconds. 

Small current shortcut is not enough unfortunately.

Software/src/devboard/safety/safety.cpp
```diff
@@ -608,6 +609,32 @@ void setBatteryPause(bool pause_battery, bool pause_CAN, EquipmentStop equipment
...
+
+void update_restart_progress() {
+  // If is a restart has been requested, check the time and restart if the
+  // conditions are met.
+
+  if (emulator_restart_request_millis > 0) {
+    uint32_t now = millis();
+    uint32_t elapsed = now - emulator_restart_request_millis;
+    // Restart after 2s if the emulator has paused. Always restart after 5s.
+    if ((elapsed > INTERVAL_2_S && emulator_pause_status == PAUSED) || elapsed > INTERVAL_5_S) {
```
dalathegreat:
```suggestion
    if ((elapsed > INTERVAL_5_S && emulator_pause_status == PAUSED) || elapsed > INTERVAL_10_S) {
```

Software/src/devboard/safety/safety.cpp
```diff
@@ -608,6 +609,32 @@ void setBatteryPause(bool pause_battery, bool pause_CAN, EquipmentStop equipment
...
+}
+
+void update_restart_progress() {
+  // If is a restart has been requested, check the time and restart if the
+  // conditions are met.
+
+  if (emulator_restart_request_millis > 0) {
+    uint32_t now = millis();
+    uint32_t elapsed = now - emulator_restart_request_millis;
+    // Restart after 2s if the emulator has paused. Always restart after 5s.
```
dalathegreat:
```suggestion
    // Restart after 5s if the emulator has paused. Always restart after 10s.
```

dalathegreat:
Updated the PR to now;
Reboot always after 5s if paused
Reboot always after 10s anyways

dalathegreat (approved):

PR #2555, opened 2026-07-07
feature: Add general logging to a syslog server

nagyrobi:
### What

Adds optional **remote syslog logging** as a new output sink for the existing general log stream. You can use your NAS as a syslog server, if you have one supporting such packages.

Under **Settings → Debug options** there's a new checkbox, *"Enable general logging to syslog server"*, which reveals three fields: the syslog **server IP**, **UDP port** (default 514), and **facility** (default 1/user). When enabled, every log line the firmware already produces is also sent to that server as an RFC 5424 UDP datagram.

Event messages are sent at their own severity (error → `err`, warning → `warning`, update → `notice`, info → `info`, debug → `debug`); all other log lines are sent as `debug`. Settings are persisted in NVS and survive reboot.

Since there's no wall-clock and no RTC in Battery Emulator, messages are sent with NILVALUE timestamp, which will make the syslog server stamp the messages on receipt with its own time.

### Why

Today the general log is only reachable in three ways, none of which suit a headless, always-on installation:

- **USB serial** — needs a physical cable and an attached terminal.
- **Web log** — a volatile in-RAM ring buffer that only fills while the debug page is open and is capped in size, so history is lost.
- **SD card** — only available on select hardware, and still requires pulling the card to read it (+SD flash wear?).

There's no way to centrally aggregate, retain, or monitor logs from a running unit. Syslog is the standard, lightweight answer: point the emulator at an existing rsyslog / syslog-ng / Graylog collector and get durable, searchable, filterable logs — and, because events map to real syslog severities, server-side alerting on warnings/errors becomes trivial. Good home NAS brands have free syslog server packages available which can be used for this task.

The implementation was kept deliberately minimal to add as little flash as possible.

### How

- **Reuses the existing sink pattern.** The central `Logging` class already fans out to USB / web / SD, each behind a runtime flag. This adds a fourth sink in the same place (`Logging::write()` and `Logging::printf()`), gated by a new `syslog_logging_active` flag — no changes to the ~626 existing log call sites.
- **No new library.** Sending is raw `WiFiUDP` over the lwIP stack that's already linked by WiFi and the async webserver, so the flash cost is negligible (a few hundred bytes plus one small static line buffer).
- **Line-framed, self-contained.** Output is buffered per line and one datagram is emitted per line, formatted as RFC 5424 with a NILVALUE (`-`) timestamp so the syslog server stamps on receipt — avoiding any NTP dependency. Sends are guarded on `WL_CONNECTED`.
- **Per-event severity.** `set_event()` passes the event's level to the logger via a lightweight `set_next_severity()` hint just before it prints; the syslog flush consumes it for that one line and otherwise falls back to the `debug(7)` default.
- **Settings & UI reuse existing plumbing.** The new keys hook into the generic NVS persistence arrays in `webserver.cpp`; the UI uses the existing template-placeholder and `data-*` conditional-display patterns (no new JavaScript).
- **Guard updates.** The `DEBUG_PRINTF/DEBUG_PRINTLN` macro guards and the internal early-return checks were extended with the syslog flag so logging works even when syslog is the *only* enabled sink.

**Flash impact:** negligible — reuses lwIP, no new dependencies.

**Testing:** verified datagrams are received and correctly severity-tagged against rsyslog (`imudp`); confirmed the checkbox reveals/hides the fields, that settings persist across reboot, and that disabling syslog leaves USB/web/SD behavior unchanged. Mock `Logging` used by the unit-test build gets a no-op `set_next_severity()` so the test target still compiles.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -187,10 +187,12 @@ void init_stored_settings() {
   datalayer.system.info.web_logging_active = settings.getBool("WEBENABLED", false);
   datalayer.system.info.CAN_SD_logging_active = settings.getBool("CANLOGSD", false);
   datalayer.system.info.SD_logging_active = settings.getBool("SDLOGENABLED", false);
+#ifndef SMALL_FLASH_DEVICE
   datalayer.system.info.syslog_logging_active = settings.getBool("SYSLOGEN", false);
   syslog_ip = settings.getString("SYSLOGIP").c_str();
   syslog_port = settings.getUInt("SYSLOGPORT", 514);
   syslog_facility = settings.getUInt("SYSLOGFAC", 1);
+#endif
   datalayer.battery.status.led_mode = (led_mode_enum)settings.getUInt("LEDMODE", false);
 
   //Some early integrations need manually set allowed charge/discharge power
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -188,8 +188,8 @@ void init_stored_settings() {
   datalayer.system.info.CAN_SD_logging_active = settings.getBool("CANLOGSD", false);
   datalayer.system.info.SD_logging_active = settings.getBool("SDLOGENABLED", false);
   datalayer.system.info.syslog_logging_active = settings.getBool("SYSLOGEN", false);
-  syslog_ip       = settings.getString("SYSLOGIP").c_str();
-  syslog_port     = settings.getUInt("SYSLOGPORT", 514);
+  syslog_ip = settings.getString("SYSLOGIP").c_str();
+  syslog_port = settings.getUInt("SYSLOGPORT", 514);
   syslog_facility = settings.getUInt("SYSLOGFAC", 1);
   datalayer.battery.status.led_mode = (led_mode_enum)settings.getUInt("LEDMODE", false);
 
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -187,6 +187,10 @@ void init_stored_settings() {
   datalayer.system.info.web_logging_active = settings.getBool("WEBENABLED", false);
   datalayer.system.info.CAN_SD_logging_active = settings.getBool("CANLOGSD", false);
   datalayer.system.info.SD_logging_active = settings.getBool("SDLOGENABLED", false);
+  datalayer.system.info.syslog_logging_active = settings.getBool("SYSLOGEN", false);
+  syslog_ip       = settings.getString("SYSLOGIP").c_str();
+  syslog_port     = settings.getUInt("SYSLOGPORT", 514);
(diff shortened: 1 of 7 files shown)


nagyrobi:
Tested on Software: 10.14.dev Hardware: LilyGo T-CAN485
image

Listening to syslog UDP packets is possible even as connected to the AP:

```
nc -l -u -p 514
<15>1 - batteryemulator BatteryEmulator - - - Battery pause begin 1 0 0 1
<12>1 - batteryemulator BatteryEmulator - - - Event: The emulator is trying to pause the battery.
<15>1 - batteryemulator BatteryEmulator - - - Battery pause begin 1 0 1 1
<11>1 - batteryemulator BatteryEmulator - - - Event: User requested stop, either via equipment stop circuit or webserver Open Contactor button
<15>1 - batteryemulator BatteryEmulator - - - Battery pause begin 0 0 0 1
<14>1 - batteryemulator BatteryEmulator - - - Event: The emulator is attempting to resume battery operation from pause.
<15>1 - batteryemulator BatteryEmulator - - - Battery pause begin 1 1 1 0
<11>1 - batteryemulator BatteryEmulator - - - Event: User requested stop, either via equipment stop circuit or webserver Open Contactor button
<12>1 - batteryemulator BatteryEmulator - - - Event: The emulator is trying to pause the battery.
<15>1 - batteryemulator BatteryEmulator - - - Safety: Pausing CAN sending
```

dalathegreat:
Nice, but this consumes 0.5% of all flash space. Is it really that useful? If so, consider wrapping it in ifdef -smallflashdevice to get consumption down!

nagyrobi:
My motivation is that it would be nice to be able to save the logs off the device. In case some catastrophic thing happens, one could look back what happened. 

For example the most feared issue is welded contacts. We do reboots etc. and we "try" to set currents to 0 before opening, but still, it's not really known if the demand was fulfilled by the inverter or the battery. After reboot, the web ui log is gone, no way to trace back. We don't have SD card slots on the majority of the boards, plus one can't see the contents of the card remotely.

It would be nice to really fill the code with as much logging as possible around the critical steps, especially the attempts to forcefully impose current/power limits and their results to be able to track possible issues and be certain that the commands were really respected.

We have 1000ms delay only after asking the inverter to reduce power to 0, but did we succeed? The new graceful reboot tries https://github.com/dalathegreat/Battery-Emulator/pull/2551 to wait at least 2 seconds, for this to happen, but after 5 seconds it cuts anyway. We should have a log entry with a warning in that case for people to start thinking this may lead to trouble later on when contactors will be worn out.

Mqtt is not really suitable to log events, because the purpose of a Home Automation system is different. There values get normalized, logs are automatically purged, etc. Syslog is the industry standard for troubleshooting based on what happened in the past.

Just my 2 cents

nagyrobi:
> wrapping it in ifdef -smallflashdevice to get consumption down! 

Done, LilyGo T-CAN485 build doesn't have it. But it does have the SD card slot so it's a fair trade-off.

dalathegreat (approved):
LGTM 👍

PR #2556, opened 2026-07-07
Improvement: Remove unreliable ESP32 CPU temperature code :thermometer: 

dalathegreat:
### What
This PR removes any trace of CPU temperature (Webserver, MQTT)

### Why
ESP32 temperature sensors are not calibrated from factory, and can swing 50 degrees from board to board. This is causing unnecessary stress and support load, so we remove it from the code. We previously removed the warning Events, but this PR removes it all.

### How
CPU temperature removed from both Webserver & MQTT.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -147,16 +147,6 @@ void logging_loop(void*) {
 }
 
 void update_calculated_values(unsigned long currentMillis) {
-  /* Update CPU temperature*/
-  union {
-    float temp;
-    uint32_t hex;
-  } temp = {.temp = temperatureRead()};
-  if (temp.hex != 0x42555555) {
-    // Ignoring erroneous temperature value that ESP32 sometimes returns
-    datalayer.system.info.CPU_temperature = temp.temp;
-  }
-
   /*Update free heap*/
   datalayer.system.info.CPU_free_heap = ESP.getFreeHeap();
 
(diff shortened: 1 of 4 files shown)


amarofarinha:
Hi!.
I understand the concern regarding the lack of factory calibration and the support overhead caused by absolute temperature inaccuracies across different boards. 

However, before completely removing this feature, could we consider keeping it as an optional/advanced metric (disabled by default), or clarify if there is a mid-ground approach? 

While the absolute value is indeed unreliable for ambient metrics, the sensor is highly effective for monitoring **relative temperature variations** on a single chip. For context, I currently use this telemetry in Home Assistant to track CPU temperature spikes, which is incredibly useful for:

* **Detecting abnormal loops:** Identifying hanging tasks or firmware loops that cause sudden CPU over-heating.
* **Hardware diagnostics:** Monitoring physical enclosure ventilation and thermal behavior over time.

Since the primary goal is to reduce support load from confused users, wouldn't it be possible to either:

1. Keep the MQTT topic but remove it from the Webserver UI?
2. Hide it behind an explicit configuration flag (e.g., `enable_cpu_temp_telemetry: true`) with a clear warning in the documentation stating that the values are uncalibrated and meant for delta-tracking only?

This would fully eliminate the support noise from general users while still preserving an essential diagnostic tool for advanced monitoring and automations.

image

image

nagyrobi:
It was me who suggested to remove it but the argument is valid.

I guess presenting a way to the user to enter a temperature offset would clarify for everyone that this is by design so imprecise, and the offset possibility would deliberately bring attention to that (with a nice tooltip with details). That would also reduce support stress.

fastbike:
So add it as an advanced option,  to enable the temp display and make the user set the offset. With a tool tip to explain  about the lack of temperature calibration?

nagyrobi:
Yeah something like that.

dalathegreat:
Closing, will open new PR later with the requested changes

PR #2557, opened 2026-07-08
Upgrade ESP-NOW to v2

nagyrobi:
### What
- transmit according to ESP-NOW v2
- send cell voltages at full 1 mV resolution for up to three batteries
- add event, system-status, and extended-vendor-metric frames
- consolidate the four send paths into one and give the TX buffer real backing storage.

### Why
ESP-NOW telemetry was missing a few generic information, and cell voltages were quantized to 20 mV — too coarse for cell-delta visualisation. v2's 1470-byte payload removes the size pressure that forced the quantization. The consolidation keeps the flash cost of the additions near zero and fixes a latent OOB write in the old flexible-array buffer.

### How
New `BAT_EVENTS`/`BAT_SYSTEM`/`BAT_EXTENDED` types on a 5 s cadence; count-first `uint16_t` cell frame trimmed to cell count; single `espnow_transmit()`/`espnow_send()` core; fixed 1466-byte payload buffer with `static_assert` bounds and an IDF-version error guard.

[Examples in the Wiki](https://github.com/dalathegreat/Battery-Emulator/wiki/Feature:-ESPNow) will have to be updated to show how it works after this change.

> [!TIP]


nagyrobi:
Updated the .h and .ino [wiki arduino examples](https://github.com/user-attachments/files/29792459/wiki_arduino_examples.zip) for anyone who wants to test. 
I'll do the [ESPHome LVGL display](https://github.com/user-attachments/files/29793107/be-monitor-guition.zip) version test myself.

jonny5532:
I'm glad you fixed the blatant memory issue with `ESPNOW_BATTERY_MESSAGE`, I raised it before the the original author didn't seem interested in fixing.

There seems to be a more general problem with the ESP-NOW stuff:
- I don't know if it has any users? I've seen some experimental ESP-screen projects floating about, I guess the authors are using these?
- If it does have users, then these changes will break the existing implementations.
- There doesn't seem to be any kind of versioning (and even if there was, the options would be either to keep old implementations in BE too, to support older displays (not desirable because of code-size), or for the device code to support multiple versions of the BE protocol, so will need maintaining).

My personal preference would be a different style of protocol, with key+size+value triples (or maybe use fixed-size 32-bit fields), so that new fields can be added in future without breaking existing implementations. But maybe that's overcomplicated, and just having fixed structs from now on, with new fields strictly added at the end, and the receivers being tolerant to larger frames than expected, would be enough.

More generally, I'm a bit concerned that BE is gaining features that only have one user - especially with the T-CAN485 so full. Even without the shackles of the 4mb T-CAN485, it would be easy to keep adding new features and before long we'd be hitting Stark's 8mb limit. There's another PR for a 'CYD' screen which appears to be another ESP-NOW based variant.

The ESP-NOW thing is useful I think though, it's an easy way of running cheap remote displays - but it probably also needs a maintained parallel project for the display, with instructions, so people can use it more easily...

nagyrobi:
> More generally, I'm a bit concerned that BE is gaining features that only have one user - especially with the T-CAN485 so full. Even without the shackles of the 4mb T-CAN485, it would be easy to keep adding new features and before long we'd be hitting Stark's 8mb limit. There's another PR for a 'CYD' screen which appears to be another ESP-NOW based variant.

I hear you.

No idea why he chose to develop a separate "CYD transport" on top of ESPNow instead of trying to improve the existing one.

I actually don't like leaving things unfinished so I thought the proper route is not to re-invent the wheel but instead try to improve what exists. Yes this is a breaking change but the main focus here is to keep flash usage small, apparently this change only introduces +184 bytes and brings up the data layer to the latest feature set. I think those few users who use it atm could manage to make the step forward, given that they are likely not many, and bringing up the feature set to be current is beneficial for them too.

First I've seen the oled screen in this project and after I saw the ESPNow I just thought let's try it. And honestly it's by far better than the oled so if it was me, I'd choose this one instead of the oled for a multitude of reasons:
- oled screens are too tiny
- no interactivity just a few static fields
- they will burn in in about 1 year of continuous operation
- they consume precious GPIO, require galvanic connection
- they only show data from one device

With ESPNow you can aggregate information from multiple Battery Emulators into one display, make it interactive (put them on tabs, pages etc). No physical contact, more safe to use and a plethora of screen sizes to choose from. Plus through ESPHome it becomes an additional way to integrate with Home Assistant if one dislikes MQTT.

Imho it's still not too late to remove certain features that could be overbloating, at first I thought let's drop ESPNow at least from the small flash units, but based on the above, I honestly would rather remove oled display support completely. If we had to find something useless, well that certainly is - apart from just using it to show off and make the install to look cool. A bigger colorful separate display would look cooler btw. 

I have similar thoughts about SD card support - afaik only the T-CAN485 has a SD card slot, but support for it goes in every other binary too. Again - what's the use of it? Store logs would be more appropriate on a syslog server (that you can run for free on your NAS with 2 clicks). CAN log/dump etc is a very special case worth a dedicated build (let's have say a T-CAN485 engineer's build which doesn't have the fancy features but has SD support).

As I mentioned on Discord I would also completely remove the options for "Customized MQTT topics", and always just set all 4 fields to the hostname. I would just rename "Custom Wifi hostname" to "Hostname (applies to MQTT topics, if enabled)". Using different values for these fields in MQTT is overkill, 99.99% of the use cases can be covered by simply setting them all to to the device hostname to uniquely idenitfy it. That could squeeze out some flash space too.

nagyrobi:
But to be very, very honest, if it comes to remove things to save flash storage space, I'd be happy to remove ESPNow all together in addition all the above. Battery Emulator is usually part of more complex systems (at least an inverter with a smart meter and some PV panels are involved), and these displays here are only able to show the BE side. A sane person wouldn't want a different display for each component of the system. The correct approach for a display would be to show information aggregated from all the components, being able to display relations between them. MQTT is already good enough to make that step and aggregate the data.

The RGB status led is already doing a great job in showing critical information about BE as a component, honestly, there's no real need for more at BE level.

jonny5532:
I think improving it is sensible, a better summary of my thoughts is:
- We should make it as future-proof as possible, so we don't need to break compatibility again (I would need check the code closely to see if there are any improvements here)
- Ideally, someone would make a nice remote-display that uses this implementation, with easy installation, so that it gets used more widely (any volunteers? ;)

nagyrobi:
> * Ideally, someone would make a nice remote-display that uses this implementation, with easy installation, so that it gets used more widely (any volunteers? ;)

Easy: https://github.com/nagyrobi/esp32evse_esphome-lvgl

dalathegreat:
Closing the CYD screen support PR. It should really use ESPNow directly, and prefferably the new V2 version

dalathegreat (approved):
Let's take the step to v2 (while there still aren't that many ESPnow screens in the wild), and update the examples. I sincerely hope we get better ESPnow options compared to hardwired screens!

nagyrobi:
Follow-up in https://github.com/dalathegreat/Battery-Emulator/pull/2700

PR #2558, opened 2026-07-08
Tesla Model 3/Y: Full CAN alert-matrix (DTC) decoding

hpetersen:
### What
- Completes Tesla Model 3/Y alert-matrix decoding for BMS (0x320), and adds
  PCS (0x3A4) and CP (0x31E) alert-matrix decoders to "More Battery Info" page.
- Fixes an incorrect bit position for BMS_a036_SW_HvpHvilFault.
- Adds a new "tesla_model3y_dtc.json" DTC JSON for offloading the ESP32 by storing it all on Github.

### Why
Previously only part of the BMS alert matrix was decoded, and PCS/CP faults
were not surfaced at all. This makes the full pack fault set visible in the UI.

### Testing
- Tested on the Stark board and flashed via OTA.
- Verified the Alert table renders all alerts in the bottom of the page.

> [!TIP]


hpetersen:
Skærmbillede 2026-07-08 kl  11 50 07

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -1626,6 +1632,10 @@ String processor(const String& var) {
     }
     content += "<button onclick='Cellmon()'>Cellmonitor</button> ";
     content += "<button onclick='Events()'>Events</button> ";
+    if (user_selected_battery_type == BatteryType::TeslaModel3Y ||
+        user_selected_battery_type == BatteryType::TeslaModelSX) {
+      content += "<button onclick='Faults()'>Faults</button> ";
```
dalathegreat: Faults should be under More Battery Info, no need to add a new page!

Software/src/battery/TESLA-ALERT-NAMES.cpp
```diff
@@ -0,0 +1,301 @@
+#include "TESLA-ALERT-NAMES.h"
+
+// See TESLA-ALERT-NAMES.h. Names from tesla-m3-pack-findings (firmware 2019.20.4.2).
+const char* const TESLA_BMS_ALERT_NAMES[100] = {
```
dalathegreat: Later on, we should add the fault descriptions to the DTC JSON file like other integrations do, which offloads the ESP32 by storing it all on Github. This PR takes up almost +1% of flash usage in this current state!

jonny5532:
Given how huge this is it would need to be non-SMALL_FLASH_DEVICE at least I think.

This could be an opportunity to try something new - my new frontend PR (#1544) has a mode for battery-specific diagnostic info which works differently, and this could follow that mechanism:

- The raw values from the CAN frames are stored directly in the extended datalayer, without being unpacked (which should take no flash and minimal RAM)
- None of the alert names or offsets are stored in the Flash
- A new webserver endpoint returns the extended datalayer as a big binary struct
- A Javascript file, loaded into the Tesla More battery info via a `<script>` tag, can then:
  - Query the webserver for this binary blob
  - Grab the CAN frame data from the binary blob (the offset can be hardcoded, as long as the layout remains stable)
  - Parse all of the flags from the CAN frames and render a HTML table with alert names/descriptions
- Either gzip and bundle the Javascript file (which will be smaller than this PR currently), or host it somewhere else like the DTC JSONs currently.

Here is how the new frontend JS does it:
https://github.com/jonny5532/Battery-Emulator/blob/3601fd6b7f9809cec450a2fe5a1bf781866b9c9b/frontend/src/ext/datalayer.ts (this is autogenerated from the ELF)
https://github.com/dalathegreat/Battery-Emulator/pull/1544/changes#diff-565989a520b976859a7da5d371075d07dd86f0b6efafadd4ce25166513f16105
https://github.com/jonny5532/Battery-Emulator/blob/3601fd6b7f9809cec450a2fe5a1bf781866b9c9b/Software/src/lib/TinyWebServer/webserver_api.cpp#L218

hpetersen:
Pushed a new commit that removes the "Faults" page and adds alerts to a table in the bottom of the "More Battery Info" page, and also did the DTC JSON file.

@jonny5532 I really like the idea, but I think it's part of a larger rebase of the whole repo.

hpetersen:
Skærmbillede 2026-07-09 kl  20 47 29

Software/src/battery/TESLA-BATTERY.cpp
```diff
@@ -2412,6 +2939,33 @@ void TeslaBattery::printFaultCodesIfActive() {
   printDebugIfActive(BMS_a176_SW_GracefulPowerOff, "ERROR: BMS_a176_SW_GracefulPowerOff");
```
dalathegreat, line 2939: To save a big deal of flash, we could remove all printDebugIfActive, much better to guide users to the new Webserver version of the faults section. Could you @hpetersen remove all printDebugIfActive ?

hpetersen: Sure, I can look into it, but today the log says ERROR: BMS_a145_SW_SOC_Change for example and after it would say ERROR: Tesla alert BMS45 active and you have to look up code 45 in the JSON? 

But yes, we could save a good chunk by removing the last big pile of hand-maintained BMS strings.

dalathegreat: We can merge this PR as-is, and focus on optimizing the strings in another PR. I personally think replacing the

ERROR: BMS_a145_SW_SOC_Change

With 

ERROR: BMS_a145

Would be enough, and maybe later on we even remove the USB printing entirely, when people are good with instead doing all debugging via webserver instead of USB printing

dalathegreat (approved):

PR #2559, opened 2026-07-08
Github: Make compile action use more caches for faster builds

jonny5532:
### What
Adds a second layer of caching (as used in the measure-size action), to cache the PlatformIO installation itself as well as just the downloaded files.

Cache key is based on a hash of the platformio.ini file, so if that changes it'll regenerate.

### Why
Should significantly speed up compiles, and reduce the amount of GitHub Action worker CPU time we're burning through...

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -40,12 +40,27 @@ jobs:
     - uses: actions/checkout@v7
       name: Checkout code
 
+    - name: Get platformio.ini hash
+      id: platformio_ini_hash
+      run: |
+        PLATFORMIO_INI_HASH=$(md5sum platformio.ini | awk '{ print $1 }')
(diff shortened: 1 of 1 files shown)


dalathegreat:
Did this really speed it up?

jonny5532:
I don't think there'll be a visible effect until it is merged (as PRs don't share caches, but they can use caches created by the main branch).

It should reduce build time by about 2 minutes, you can see it on these PR updates on the measure-size action:

https://github.com/dalathegreat/Battery-Emulator/actions/runs/29004464369/job/86072626493 (took 5m33)
https://github.com/dalathegreat/Battery-Emulator/actions/runs/29005114811/job/86074755565 (took 3m36)

It can be easily reverted if it causes problems or doesn't help (just revert the commit).

jonny5532:
Ah, having just rebased and pushed again, it has saved 1m on every build:

First time:
image

Second time:
image

dalathegreat (approved):

PR #2561, opened 2026-07-08
Improve binary size tracking

ngehrsitz:
### What
This PR refactors the binary size tracking to work for all boards

### Why
Help with the binary size reduction efforts

### How
It removes the separate rebuild and instead adds a `<board>size.json` build artifact. This makes it possible to calculate binary size differences without rebuilding which saves CI runtime.

> [!TIP]


ngehrsitz:
Example comment: https://github.com/ngehrsitz/Battery-Emulator/pull/6#issuecomment-4919354544

dalathegreat:
Thanks for the contribution. I just wonder, does it really help? As more boards gets added, the list gets longer and more confusing for contributors to understand. If the flash fits the 4MB board, you are golden, you will have so much headroom for the 8MB (and 16MB variants)

I personally think having just the 4MB OG ESP32 tracker to be enough. @jonny5532 ?

ngehrsitz:
Besides the performance improvement my reason for adding this was that there are some features that are only enabled for certain boards. Thus if you change something there you currently don´t see the impact.
But we could also shorten the summary to only show the worst boards or remove the comment entirely and only add a Github Check that fails if there is an increases by X%.

nagyrobi:
a Github Check that fails if there is an increase is misleading, as one may do that on purpose

dalathegreat:
Can you @ngehrsitz solve the merge conflict?

ngehrsitz:
Yes, I will also track the compiler versions. Otherwise it could happen that the artifacts are created by different compiler/pio versions and that could result in wrong numbers.

ngehrsitz:
I resolved the conflicts and added a few further improvements.

- We now track the toolchain to emit a warning if there is a mismatch
- The `compiler_warning_check` artifact is no longer uploaded
- I removed the `.bin` suffix for the artifacts since these are actually zip files

The test of my latest version is on https://github.com/ngehrsitz/Battery-Emulator/pull/12

dalathegreat (approved):
Awesome, builds faster and now with more info. Great work 🙌

dalathegreat:
@ngehrsitz , seems like it broke after being merged to main, see the recently opened https://github.com/dalathegreat/Battery-Emulator/pull/2579 which has no comment on code size

dalathegreat:
@ngehrsitz , another issue, The new measure size PR also broke the easyness of the build system. Now users downloading the artefact are trying to OTA update to the JSON file

image

dalathegreat:
I have reverted PR for now, started getting support requests for PRs where users could not find bin files

ngehrsitz:
> @ngehrsitz , another issue, The new measure size PR also broke the easyness of the build system. Now users downloading the artefact are trying to OTA update to the JSON file
> image

Sorry my changes exposed the `.pio` folder structure, I will fix that. https://github.com/dalathegreat/Battery-Emulator/actions/runs/29185504946/job/86630753527 is the run for #2579 I will investigate what went wrong there.

PR #2562, opened 2026-07-09
Improvement/MQTT and AP UI reductions, settings page UI

nagyrobi:
### What
This PR removes the "Customized MQTT topics" checkbox and its four dependent fields (MQTT topic name, Prefix for MQTT object ID, HA device name, HA device ID), the "Access point name:" field from the web interface, and removes the corresponding `MQTTTOPICS`, `MQTTTOPIC`, `MQTTOBJIDPREFIX`, `MQTTDEVICENAME`, `HADEVICEID`, `APNAME` settings from NVM. All four MQTT/HA identity values and the AP name are now derived automatically from the device's WiFi hostname instead of being separately configurable.

This still satisfies the vast majority of the use cases.

Alongside this, the WiFi hostname handling is changed so a hostname is always set at boot, regardless of whether MQTT is used: if the user has configured a custom hostname it's used as-is; otherwise the device now defaults to something like `battery-emulator-af34` (the last two bytes of its MAC address lowercase), giving every device a meaningful, distinct hostname out of the box. The existing "Custom Wifi hostname" field is relabelled to "Hostname (also Access Point SSID and MQTT topics):" with an updated tooltip explaining that this single value now drives the MQTT topic name, object-ID prefix, HA device name and HA device ID when MQTT is enabled. 

The hostname field in the settings shows the default one as a placeholder when cleared.

The Access Point SSID will be the same.

The mDNS hostname logic is simplified to reuse this same value rather than computing its own MAC-based suffix independently. mDNS was broken, this PR fixes it (as it consumes the hostname ).

Logging to syslog server will populate the logs with the same unified hostname.

The settings page has some further string optimizations to debloat the UI and reduce size.

### Why
This is driven by the need to reduce flash usage. Five separate settings (one bool, four strings) each carry their own NVS storage, their own HTML label/input markup, their own placeholder-resolver functions, and their own CSS show/hide rule tied to the removed checkbox — all of which take up flash space for a feature that, in practice, nearly everyone fills with the same value. Collapsing five independently-configurable values down to one (the hostname, which the device needs to expose anyway) removes that duplicated code and markup while keeping the one setting that's actually useful to change. It also removes a class of user-facing footgun where the topic name, prefix, device name and device ID could independently drift out of sync with each other or be left blank, which the "Customized MQTT topics" section made easy to do by accident.

AP SSID and MDNS name being consistent with the hostname, makes the device more easy to be identified, which is especially important when the user has multiple units in the network — one may forget to change one field from the defaults/others on one of the device and things can become confusing pretty quickly.

You can identify your device equally on every platform, no mistakes.

### How
- init_WiFi() now always calls WiFi.setHostname(). If custom_hostname is empty, a default of "battery-emulator-" + the last two bytes of WiFi.macAddress() (lowercased) is generated and used instead of leaving the hostname unset. init_mDNS() is simplified to just read back WiFi.getHostname() rather than recomputing its own MAC-suffix hostname again.
- mqtt.cpp: the mqtt_manual_topic_object_name branch that previously chose between NVS-stored custom values and compiled-in defaults is replaced with a single block that sets topic_name, default_entity_id_prefix, device_name, and device_id all from WiFi.getHostname() (with _ appended for the prefix). The five now-unused globals (mqtt_manual_topic_object_name, mqtt_topic_name, mqtt_default_entity_id_prefix, mqtt_device_name, ha_device_id) are removed.
- settings_html.cpp and webserver.cpp: removed the `MQTTTOPICS/MQTTTOPIC/MQTTOBJIDPREFIX/MQTTDEVICENAME/HADEVICEID` placeholder resolvers, the associated .if-topics CSS show/hide rule, and the checkbox + four input fields from the form. Relabelled the hostname field and expanded its title tooltip to explain the MQTT linkage. Shortened some labels and tooltips to improve clarity.
- mDNS starts only when there is an IP assigned (in AP-only mode no use for it), plus it correctly advertises the service as `http`. This advertises `_http._tcp.local` on port 80 with instance name derived from the hostname — which is exactly the standard way a "find web servers on the network".

### Compatibility note
Existing installations will see their MQTT topic name and HA device identity change from the previous values to the hostname after upgrading. Home Assistant will need its old retained discovery entries cleared (will simply create new entities under the new device). 

> [!TIP]


nagyrobi:
Software: 10.14.dev Hardware: LilyGo T-CAN485

image

image


image


image

nagyrobi:
Fixed the broken mDNS advertisment:

```bash
$ avahi-browse -r _http._tcp
+   eth0 IPv4 battery-emulator-6634                            Web Site             local
=   eth0 IPv4 battery-emulator-6634                            Web Site             local
   hostname = [battery-emulator-6634.local]
   address = [192.168.1.169]
   port = [80]
   txt = []

$ping battery-emulator-6634.local
PING battery-emulator-6634.local (192.168.1.169) 56(84) bytes of data.
64 bytes from battery-emulator-6634.local (192.168.1.169): icmp_seq=1 ttl=64 time=52.2 ms
64 bytes from battery-emulator-6634.local (192.168.1.169): icmp_seq=2 ttl=64 time=71.4 ms
```

syslog uses the same hostname too:
```bash
# nc -l -u -p 514
<15>1 - battery-emulator-6634 BatteryEmulator - - - Battery pause begin 1 0 0 1
<12>1 - battery-emulator-6634 BatteryEmulator - - - Event: The emulator is trying to pause the battery.
<15>1 - battery-emulator-6634 BatteryEmulator - - - Battery pause begin 0 0 0 1
<14>1 - battery-emulator-6634 BatteryEmulator - - - Event: The emulator is attempting to resume battery operation from pause.
<15>1 - battery-emulator-6634 BatteryEmulator - - - Battery pause begin 1 1 1 0
<11>1 - battery-emulator-6634 BatteryEmulator - - - Event: User requested stop, either via equipment stop circuit or webserver Open Contactor button
<12>1 - battery-emulator-6634 BatteryEmulator - - - Event: The emulator is trying to pause the battery.
<15>1 - battery-emulator-6634 BatteryEmulator - - - Safety: Pausing CAN sending
```

nagyrobi:
Tested on Software: 10.14.dev Hardware: LilyGo T-CAN485 and Waveshare ESP32-S3-RS485-CAN.

dalathegreat (approved):
Nice simplification. This change will require a new major release to be published (v11.0.0), along with info how to migrate to the new setup.

Especially the comment on Discord is helpful;
_I 99% bet whoever uses this, sets all 4 fields to the same string. Now all that has to be done after this patch is to set the hostname to that, and nothing will change on the other side._

Nice work :raised_hands:

PR #2566, opened 2026-07-09
security: Provisioning window for AP with default pass

nagyrobi:
## What

This PR adds a provisioning window for the Wi-Fi Access Point: when the AP is running with the factory-default password (`123456789`), and no client connects to it, it is automatically disabled after 5 minutes. If a custom AP password has been set, behavior is unchanged and the AP stays enabled indefinitely.

After the window expires, the AP can be re-enabled by:
- rebooting the device, or
- long-pressing the BOOT button for ≥5 s (existing behavior)

A new info event `EVENT_WIFI_AP_PROVISION_TIMEOUT` is raised when the window expires, with a message explaining how to re-enable the AP or make it permanent.

### Why

The AP ships with a well-known default password, and remains enabled by default. Leaving it permanently enabled means anyone within radio range can join the emulator's network and reach the webserver/settings of a device controlling a high-voltage battery. Limiting the default-password AP to a short provisioning window closes this exposure while keeping first-time setup and recovery access fully functional:

- New users still get 5 minutes at every boot to connect and start configuring the device
- Once connected, the AP doesn't shutdown until the user disconnects (or walks away + 5 minutes)
- Recovery access remains available at any time via the BOOT button long-press
- Users who set their own AP password keep today's always-on behavior

### How

- `wifi.cpp`: new `check_ap_provisioning_window()` called from `wifi_monitor()` (before the `ssid.empty()` early-return, so it also works on devices with no STA credentials configured). On expiry it calls `WiFi.softAPdisconnect(true)` — dropping only the AP bit from the Wi-Fi mode, STA/ESP-NOW stay up — clears `ap_active`, and latches `ap_provisioning_expired`.
- `init_WiFi_AP()` records the window start time **only on an off→on transition**, so the repeated re-inits from the STA reconnect fallback don't keep extending the window.
- The "no previous OK connection" rescue-AP fallback in `wifi_monitor()` is gated on `!ap_provisioning_expired`, so it can't resurrect the AP right after the window closed.
- The BOOT button ≥5 s handler clears the expired latch before starting the AP, opening a fresh window.
- `wifi.h` / `comm_nvm.cpp`: the `"123456789"` literal is promoted to a shared `DEFAULT_AP_PASSWORD` constant (single source of truth).
- `events.h` / `events.cpp`: new `EVENT_WIFI_AP_PROVISION_TIMEOUT` (level `INFO`) with a user-facing message.
- Timer math uses unsigned `millis()` subtraction (rollover-safe).

### Note
Changing the AP password via the web UI takes effect after reboot (as with other settings), so a password changed mid-window still results in AP shutdown at the 5-minute mark or after the user disconnects — the AP is still broadcasting the default password until reboot, so this is the safe choice.

Wiki needs to be updated after this is merged.

> [!TIP]

### DIFF
--- a/Software/src/devboard/utils/events.cpp
+++ b/Software/src/devboard/utils/events.cpp
@@ -411,8 +411,8 @@ String get_event_message_string(EVENTS_ENUM_TYPE event) {
     case EVENT_WIFI_DISCONNECT:
       return "Wifi disconnected.";
     case EVENT_WIFI_AP_PROVISION_TIMEOUT:
-      return "Wifi access point disabled: the provisioning window expired. Change password or "
-             "hold the BOOT button >=5 seconds or reboot to re-enable it.";
+      return "Wifi AP disabled: the provisioning window expired. Change password or "
+             "hold BOOT button 5 to 15 secs or reboot to re-enable.";
     case EVENT_MQTT_CONNECT:
       return "MQTT connected.";
     case EVENT_MQTT_DISCONNECT:
--- a/Software/src/devboard/utils/events.cpp
+++ b/Software/src/devboard/utils/events.cpp
@@ -411,8 +411,8 @@ String get_event_message_string(EVENTS_ENUM_TYPE event) {
     case EVENT_WIFI_DISCONNECT:
       return "Wifi disconnected.";
     case EVENT_WIFI_AP_PROVISION_TIMEOUT:
-      return "Wifi AP disabled: the provisioning window expired. Change password or "
-             "hold BOOT button 5 to 15 secs or reboot to re-enable.";
+      return "Wifi AP disabled due to cybersecurity concern. Change default password to keep AP "
+             "constantly on! Reboot/Hold BOOT button 5-15 seconds to re-enable AP temporarily.";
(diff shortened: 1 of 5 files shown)


nagyrobi:
image

Software/src/devboard/utils/events.cpp
```diff
@@ -409,6 +410,9 @@ String get_event_message_string(EVENTS_ENUM_TYPE event) {
       return "Wifi connected.";
     case EVENT_WIFI_DISCONNECT:
       return "Wifi disconnected.";
+    case EVENT_WIFI_AP_PROVISION_TIMEOUT:
+      return "Wifi AP disabled: the provisioning window expired. Change password or "
```
dalathegreat: Let's use a little bit simplified english to guide users towards the password change. I suggest;

"Wifi AP disabled due to cybersecurity concern. Change default password to keep AP constantly on! Reboot/Hold BOOT button 5-15 seconds to re-enable AP temporarily"

dalathegreat (approved):
Very good cybersecurity improvement 👍

PR #2568, opened 2026-07-09
Add CAN bus error detection and handling, and native bus-off recovery

jonny5532:
### What
This PR reports CAN bus errors (the warning state where REC/TEC exceed 96) on all buses, which usually indicate wiring problems (numerous failed tx/rx).

It also performs bus-recovery on native CAN, which previously would stay permanently off in the case of a bus-off, but now recovers.

(Curious to see how much code this takes up, the main hit is all the new events and related code.)

> [!TIP]


nagyrobi:
> (Curious to see how much code this takes up, the main hit is all the new events and related code.)

Well worth it imho

Software/src/devboard/utils/events.cpp
```diff
@@ -199,6 +205,11 @@ String get_event_message_string(EVENTS_ENUM_TYPE event) {
       return "High amount of corrupted CAN messages detected. Check CAN wire shielding!";
     case EVENT_CAN_NATIVE_TX_FAILURE:
       return "CAN_NATIVE failed to transmit, or no one on the bus to ACK the message!";
+    case EVENT_CAN_NATIVE_BUS_ERROR:
+    case EVENT_CANMCP2515_BUS_ERROR:
+    case EVENT_CANFD_BUS_ERROR:
+    case EVENT_CANFD_2_BUS_ERROR:
+      return "CAN bus has had multiple tx/rx errors. Check wiring!";
```
dalathegreat: Suggested text;

`Multiple CAN TX/RX errors. Check wiring!`

jonny5532: Have renamed!

I also realised that the NATIVE_TX_FAILED event was actually just a BUFFER_FULL like the others, so have renamed it, and combined them all under a generic "CAN failed to send. Buffer full or no one on the bus to ACK the message!" message, which saves some flash from avoiding repeated strings (it should be obvious from the event name which bus it is).

dalathegreat (approved):

PR #2570, opened 2026-07-10
Union datalayer entries, rejig constructors, to save Flash+RAM

jonny5532:
### What
Somewhat experimental PR to merge all the battery extended datalayer entries together into a big union (the zero initialized ones, anyway), since they're not used at the same time as each other.

This obviously saves some RAM, but also a decent chunk of code due to all the initializers.

The main downside is the struct entries now need to not have any default values, or have them put in the new DataLayerExtended constructor (which is somewhat ugly).

Ideally all the structs would be initially-zero anyway, as that avoids any extra code (and is a good practice anyway IMO, maybe I write too much Go).

In addition to the Flash saving, this saves 1360 bytes of static RAM.

(Don't merge it if you think it is too ugly, there might be a better way anyway...)

> [!TIP]

### DIFF
--- a/Software/src/datalayer/datalayer_extended.h
+++ b/Software/src/datalayer/datalayer_extended.h
@@ -2,108 +2,109 @@
 #define _DATALAYER_EXTENDED_H_
 
 #include <stdint.h>
+#include <string.h>
 
 struct DATALAYER_INFO_BOLTAMPERA {
   /** uint16_t */
   /** PID polling parameters */
-  uint16_t battery_5V_ref = 0;
-  uint16_t battery_capacity_my17_18 = 0;
-  uint16_t battery_capacity_my19plus = 0;
-  uint16_t battery_SOC_display = 0;
-  uint16_t battery_SOC_raw_highprec = 0;
-  uint16_t battery_max_temperature = 0;
-  uint16_t battery_min_temperature = 0;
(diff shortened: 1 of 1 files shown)


Software/src/datalayer/datalayer_extended.h
```diff
@@ -114,940 +115,977 @@ struct DATALAYER_INFO_BMWIX {
...
+      for (int i = 0; i < 13; i++)
+        data.battery_temperatures[i] = -40;
+      data.discharge_status = 14;
+      data.auto_calibrate_soc_enabled = true;
+      data.auto_calibrate_soc_drift_percent = 5;
+    };
+    initBydAtto3(bydAtto3);
+    initBydAtto3(bydAtto3_2);
+
+    rivian.pre_contactor_voltage = 3700;
```
dalathegreat, line 1081: I like it! Dealing with the extended_datalayer being initialized to 0 inside each battery integration is a good solution. We could even shave off these separately handled ones in another PR to get this function to only memset and nothing else!

dalathegreat (approved):

Software/src/datalayer/datalayer_extended.h
dalathegreat: Doing that in https://github.com/dalathegreat/Battery-Emulator/pull/2572

PR #2571, opened 2026-07-10
improvement: Flash-size 🪏🆓 reduction by sdkconfig override for small-flash targets

nagyrobi:
### What
Enables pioarduino **hybrid mode** on the two 4 MB `SMALL_FLASH_DEVICE` targets (`esp32devkit_330`, `lilygo_330`) and applies ESP-IDF sdkconfig size cuts, without moving off the Arduino framework or touching app code.

### Why
- `custom_sdkconfig` allows sdkconfig control (keeping Arduino stuff intact)
- Several arduino-esp32 managed components are dead weight (unused) and were also breaking the from-source build.

### How
- `custom_sdkconfig = file://sdkconfig.be_size.defaults` on the two envs.
- `custom_component_remove`: drop unused components (esp_insights,   esp_diagnostics/diag_data_store, esp_rainmaker/rmaker_common, network_provisioning, zigbee/zboss, esp-modbus) — reclaims flash *and* fixes the `https_server.crt` / esp-modbus build breaks. (firmware doesn't use esp-modbus at all — it uses the Arduino eModbus library. esp-modbus is only being dragged in transitively by arduino-esp32, so in the precompiled path it was invisible dead weight!)
- sdkconfig cuts: `-Os`, IDF+bootloader log level `NONE`, drop WPA-Enterprise, mbedTLS trims (TLS 1.3/DTLS off, `TLS_CLIENT_ONLY`, cert bundle `FULL→CMN`), assertions disabled.
- FMB timeout pins kept as a safety net for esp-modbus.

### Caveats
- TLS cuts are safe only if no on-device TLS server / secure MQTT (8883).
- Assertions-disabled turns clean aborts into undefined behaviour.
- Cert bundle can't be fully removed (unused `NetworkClientSecure` forces the header); `CMN` is the smallest self-contained option.
- This hybrid mode recompiles the IDF static libs from source, so builds for the two 4 MB `SMALL_FLASH_DEVICE` firmware targets (`esp32devkit_330`, `lilygo_330`) will be slower.

> [!TIP]


nagyrobi:
10.14.dev Hardware: LilyGo T-CAN485, fake battery

Performance improvement gained from a thinner wrap of Arduino around ESP-IDF is also significant:

Before:
Free heap: 116936, max alloc: 77812
Core task max load: 1875 us
Core task max load last 10 s: 224 us
Max load @ worst case execution of core task:
10ms function timing: 1796 us
Values function timing: 0 us


After:
Free heap: **157280**, max alloc: 86004
Core task max load: **1163** us
Core task max load last 10 s: 218 us
Max load @ worst case execution of core task:
10ms function timing: **1144** us
Values function timing: 0 us

jonny5532:
That is a remarkably big difference, especially since I would expect the linker to be able to remove the unused code already - but I guess things like TLS sneak in anywhere you might potentially connect to an https URL (or secure MQTT) etc. The RAM saving must be preallocated buffers for TLS etc (which is pretty memory heavy iirc).

I also don't understand the perf improvement, that must just be all the asserts/error-handling? Might well be flash-cache related, since the API functions will now be smaller and displace less flash each time they are loaded. Memory errors (stack and heap crashing etc) and watchdog timeouts aren't unusual (usually buggy code), so this may make it harder to isolate these (but hopefully they would also occur on the 2CAN etc build with the asserts still enabled).

The 7-minute builds might get boring though, but I suppose it is manageable. I wonder about using `--wrap` to stub things out to avoid full recomplication, although I guess that'll be a lot of work to do and keep updated (and more likely to encounter weird bugs).

jonny5532:
(I would love to remove the Arduino stuff completely, there is some pretty painful code in there (eg, the SPI code makes *so* many copies on the way into and out of the hardware). However it'd need changing a lot of code and replacing basically all of the libraries, including the OTA `Update` one (which I would very much like not to touch). We might be able to replace some of the subsystems though (eg, SPI, which hopefully won't get linked if we never call the Arduino SPI* stuff).

nagyrobi:
That was my original intention but I don't think it's worth it, at least not yet.

nagyrobi:
The performance improvement makes me go hmmm in respect with the other boards too...

Maybe we could switch to hybrid mode there too, but with less constaints (eg leave more debug strings in). To gain the performance side of things.

dalathegreat:
The pro's seem to outweigh the cons on this one. 

Very nice findings @nagyrobi :raised_hands: 

@jonny5532 , maybe the flash compare check could be built into the build, saving time not to have to build again?

ngehrsitz:
> The pro's seem to outweigh the cons on this one.
> 
> Very nice findings @nagyrobi 🙌
> 
> @jonny5532 , maybe the flash compare check could be built into the build, saving time not to have to build again?

#2561 does this and also eliminates the need to rebuild. Thus build time shouldn´t be a concern here.

nagyrobi:
Removing SPIFFS partition for the low flash ones

dalathegreat:
Could the SPIFFS change be in a separate PR, so that we can schedule that for the next mandatory USB flash update?

nagyrobi:
Reverted the SPIFFs change from this.
Got some different ideas about SPIFFS usage so yes it's not here anymore.

ngehrsitz:
Enabling LTO(11d866e7f7c81399e9aeec0bd38d24b26302fa69 ) gives even bigger savings: https://github.com/ngehrsitz/Battery-Emulator/pull/26#issuecomment-5014137541

nagyrobi:
> Enabling LTO([11d866e](https://github.com/dalathegreat/Battery-Emulator/commit/11d866e7f7c81399e9aeec0bd38d24b26302fa69) ) gives even bigger savings: [ngehrsitz#26 (comment)](https://github.com/ngehrsitz/Battery-Emulator/pull/26#issuecomment-5014137541)

Thanks for the tip, indeed!

nagyrobi:
On Waveshare ESP32-S3-RS485-CAN Free heap before: 128864, after 141456

dalathegreat (approved):
Tested OK on production units for 24h+:
- T-CAN485: BMWi3+Fronius
- Stark CMR: DoubleLEAF+Fronius

Insanely nice perf-gains 🙌

PR #2581, opened 2026-07-12
Improvement: 🛜 Static IP single-field IP entry, add missing DNS server setting

nagyrobi:
Replaces the twelve octet inputs with four validated text fields, adds the missing DNS setting (defaulting to the gateway), and migrates the legacy LOCALIP1-4/GATEWAY1-4/SUBNET1-4 NVS keys on first boot.

Placeholders show the addresses currently in use, so pinning an existing DHCP lease is just a matter of ticking the checkbox.

Moves the Wi-Fi interface related network settings up into the Network config box, and renames the title of  'Connectivity settings' box 'Integration settings' (as ESPNow and MQTT are features helping integration with other devices/systems).

### What
This PR replaces the twelve `<input type="number">` octet fields used for static IP configuration with four validated text fields (IP, gateway, subnet mask, DNS server), and adds a DNS server setting that was previously missing entirely. Legacy `LOCALIP1-4` / `GATEWAY1-4` / `SUBNET1-4` NVS keys are migrated automatically on first boot, so existing static IP users are unaffected.

### Why
Previously, with static IP, the user had no DNS set (0.0.0.0). `WiFi.config()` was called with three arguments only. In arduino-esp32 the STA branch stops the DHCP client and then unconditionally writes `ESP_NETIF_DNS_MAIN` from the (unset) `dns1` parameter, leaving the resolver at `0.0.0.0`. Anyone enabling static IP therefore lost all name resolution — eg. MQTT brokers addressed by hostname, and the release check all silently fail. 

Twelve inputs plus twelve NVS keys and twelve template-processor branches is a lot of flash and NVS for what is conceptually four addresses.

We had Wi-Fi related settings spread in several places on the page. Now they are in a single box as they belong together.

### How
`WiFi.config()` now receives a DNS argument, defaulting to the gateway when the field is left blank (correct on virtually every home network). The settings are stored as dotted-quad strings and parsed with `IPAddress::fromString()` at connect time; invalid or empty settings log a warning and fall back to DHCP rather than leaving the device unreachable. The IPv4 validation regex is injected once via a `%IPPATTERN%` placeholder and reused by all four fields plus the existing syslog field, instead of being duplicated in the HTML template. 

The migration function is temporary, and can be removed after a couple of major releases. (added to comment)

> [!TIP]


nagyrobi:
Tests on Software: 11.1.dev Hardware: Waveshare ESP32-S3-RS485-CAN 
image

dalathegreat (approved):
Good improvement. i remember when I hastily implemented the 4 different IP fields, thinking there must be some cleaner way to do this :joy: 

We can indeed drop the older keys after a while!

PR #2585, opened 2026-07-12
Improvement/Syslog 📜 pre-connectivity log backlog; task name as app name

nagyrobi:
### What
This PR adds a pre-connectivity log backlog.
Lines logged before the STA gets an IP are now buffered and replayed to the syslog server on connect, plus a one-shot "Bootup complete, running version" notice.

The name of the task where the messages came from appears in the app column.

### Why
Everything logged during boot was silently dropped by the `WL_CONNECTED` gate in `syslog_emit()` — exactly the lines that matter most when diagnosing a failed startup.

App name was always "BatteryEmulator" which is irrelevant since it's always the same.

### How
HOSTNAME now follows the same rule as `init_WiFi()` — custom hostname if set, otherwise `default_hostname()` (eFuse-derived, valid before WiFi starts) — cached to avoid a heap allocation per log line, and APP-NAME carries the FreeRTOS task that produced the line instead of a constant.
The connectivity gate is replaced by a queue plus a dedicated low-priority sender task: `syslog_emit()` only appends each assembled line to a lazily malloc'd 4 KB FIFO under a short-held mutex (no network I/O in the caller's task, so the core loop can't overrun), while `syslog_task()` — started from `onWifiGotIP()` — pops one record at a time and does all UDP sending outside the lock, which keeps packets in order and uncorrupted; lines queued before the link is up carry their boot uptime (`[boot +N.NNNs]`) since there is no wall clock at that point.

> [!TIP]

### DIFF
--- a/Software/src/devboard/utils/logging.cpp
+++ b/Software/src/devboard/utils/logging.cpp
@@ -67,7 +67,7 @@ static bool syslog_online(void) {
 }
 
 // Called ONLY from syslog_task, and never with the mutex held.
-static void syslog_send(uint8_t sev, const char* msg) {
+static void syslog_send(uint8_t sev, const char* proc, const char* msg) {
   IPAddress dst;
   if (!dst.fromString(syslog_ip.c_str())) {  // empty or invalid IP -> skip
     return;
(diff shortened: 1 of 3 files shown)


nagyrobi:
Software: 11.1.dev Hardware: Waveshare ESP32-S3-RS485-CAN 

Any warnings occurring during boot are caught-up to the syslog server now.

Could be already useful for https://github.com/dalathegreat/Battery-Emulator/discussions/2584 and when you're not close enough to examine via USB cable what's happening (accessing remotely).

nagyrobi:
One more subtle change: since this is an embedded platform, APP-NAME would be always the same, so instead much more useful to put task name in that column:

image

Now we can see at a glance which part of the firmware produced the log entry.

dalathegreat (approved):
LGTM 👍 Now I really wanna setup syslog at home!

PR #2588, opened 2026-07-12
Improvement: Nissan Leaf 🔋 balancing status

nagyrobi:
### WHAT
Noticed that though MQTT "Balancing Status" sensor reports `Unknown` for Nissan LEAF.

### WHY
`balancing_status` is an opt-in field: a handful of drivers (MEB, Zoe Gen2, the BMWs, RJXZS) set it, if silently inherits the `UNKNOWN` default the user has no way to tell "not supported" apart from "not working". Data is there though.

Users on Discord think there's an issue with the pack, generating potential support pressure.

### HOW
Inspired from Zoe, derive the aggregate state from the shunt bits the driver is already collecting — active if any resistor is on, ready if none are — and hold the field at `UNKNOWN` until group `0x06` has actually answered once, because polling is paused at boot.

#### Hysteresis
The LBC duty-cycles its bleed resistors and switches them off while it samples cell voltages, so a single snapshot showing no active shunt does not mean balancing has finished — it usually means we asked at the wrong instant.
Group `0x06` is one of seven PID groups on a 10 s rotation, so we get one snapshot per ~70 s and cannot distinguish a resistor that is on 90% of the time from one that is on 10%; taken naively, the status flaps between Active and Ready every poll and buries the event log in start/end pairs.
Requiring several consecutive all-clear polls before dropping back to `READY` makes the field mean "the pack is in a balancing phase", which is what a user actually wants on a dashboard, while the raw per-cell array and `balancing_active_cells` stay unsmoothed for anyone who wants the instantaneous truth.

> [!TIP]

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -203,9 +203,8 @@ void NissanLeafBattery::
       balancing_idle_polls++;
     }
 
-    balancing_status_enum new_status = (balancing_idle_polls < BALANCING_IDLE_POLLS_TO_END)
-                                           ? BALANCING_STATUS_ACTIVE
-                                           : BALANCING_STATUS_READY;
+    balancing_status_enum new_status =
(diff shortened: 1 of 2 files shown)


nagyrobi:
image
image

dalathegreat (approved):

PR #2594, opened 2026-07-13
Improve binary size tracking v3

ngehrsitz:
This PR is a repeat of #2561 while addressing the major issues:

1. The zip file structure is now the same as before with the only difference being that the artifacts have removed the `.bin` suffix so `battery-emulator-esp32devkit` instead of `battery-emulator-esp32devkit.bin`. This was done because tha artifact is actually a zip file containing the bin, so the previous naming was misleading.
2. The previous PR relied on an error prone Github lookup to identify what PR a comment should be posted on. The other downside of this design was that the measure step was no longer listed on the PR which made it harder to track failed runs. This PR calls the measure workflow from the compile workflow which solves both of these issues.
3. Previously we only searched for the last run of the compile workflow on the `main` branch. Instead we now search for the exact PR base SHA and if there are multiple runs we compare against the most recent run that has the most artifacts.

Tested on https://github.com/ngehrsitz/Battery-Emulator/pull/17

> [!TIP]

### DIFF
--- a/.github/scripts/parse_size.py
+++ b/.github/scripts/parse_size.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+"""Parse `pio run -t checkprogsize` output into a per-env size report.
+
+Consumed by .github/workflows/measure-firmware-size.yml to produce a
+firmware-size PR comment without rebuilding.
+
+checkprogsize output looks like:
+
+    RAM:   [==        ]  13.9% (used 45678 bytes from 327680 bytes)
+    Flash: [======    ]  62.8% (used 1234567 bytes from 1966080 bytes)
+"""
+import argparse
+import json
+import re
+import sys
+from dataclasses import asdict
+
+from size_report import BoardSize, MemoryBytes, Toolchain
+
+
+RAM_RE = re.compile(r"^RAM:.*used\s+(\d+)\s+bytes\s+from\s+(\d+)\s+bytes", re.MULTILINE)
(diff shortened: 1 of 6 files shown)


ngehrsitz:
This PR had to change the workflow filter in the `add-comment-to-pr.yml` workflow. That´s why this PR will not get a comment even though the measure-size step ran.

dalathegreat (approved):
Nice work, let's merge it and try it out 👍

jonny5532:
@dalathegreat - I note that this puts the json containing the size in the same artifactt zip as the firmware.bin, hopefully people won't be confused by this? Another option would be to create a second artifact with just the size JSON in.

It would be nice to link to the firmware artifacts from the rows of the table, so people can download them more easily (I can add post-merge if that's easier).

Also I'm tempted to tweak the table a bit to make it more readable, specifically:
- The SHAs should render as clickable links? maybe print in full and remove the code-parentheses and GitHub will do it automatically
- Only show one of 'Board' and 'Env', not both (takes up too much space)
- Not showing 'Base flash', you can figure it out from Max and delta, if you need to...
- Finding some way to indicate the % change as well (I think it is useful) whilst still making it fit (two rows?)
- Maybe putting flash after RAM as it's more interesting? (RAM is just the static allocations anyway)

(obviously subjective but IMO the new table is less readable than the old one)

dalathegreat:
I looked at the artefact .zip, and yes it is less confusing than it was before with the .pio folder that was invisible on MAC for some reason. Now it is a .bin and .JSON, and the .bin is thankfully named firmware, so I hope this will guide the user to pick the correct file.

Feel free to build further upon it!

PR #2598, opened 2026-07-13
Improvement: 🛜 Wipe RF PHY calibration data on factory reset and network reset

nagyrobi:
### What

This PR erases the RF PHY calibration data (the `phy` NVS namespace) during factory reset — both via the `/factoryReset` web route and the ≥30 s BOOT button hold — and also on the ≥15 s BOOT button WiFi-wipe, forcing a full RF calibration on the next boot. A log entry records the erase result.

### Why

Factory reset only clears the emulator's own settings namespace (`clearAll()`), leaving stale PHY calibration data behind — undesirable after board rework, antenna changes, drastic changes to wifi environment, or when RF misbehaves, which is exactly when these recovery gestures are used. The 15 s tier is the "WiFi is broken" gesture, so RF recalibration fits its intent.

This could otherwise be cleared only by fully erasing the board and reflashing a full image via serial.

### How

New `erase_phy_cal_data()` helper in `comm_nvm` wraps `esp_phy_erase_cal_data_in_nvs()` and logs success/failure; it is called from the three reset paths (webserver `/factoryReset`, wifi.cpp 30 s and 15 s button branches). A failed erase is logged but never blocks the reset; next boot performs full RF calibration (~100 ms extra RF init) and rewrites fresh cal data only once.

Log entries introduced as we can have a record on backends like syslog how this has been done.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -1,4 +1,5 @@
 #include "comm_nvm.h"
+#include <esp_phy_init.h>  // esp_phy_erase_cal_data_in_nvs()
 #include "../../battery/BATTERIES.h"
 #include "../../battery/Battery.h"
 #include "../../battery/Shunt.h"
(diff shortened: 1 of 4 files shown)


dalathegreat (approved):
Good! We had an earlier workaround for the calibration data clear that looked something like this in the Webserver Wiki;

I have removed this last week from the Wiki, and I presume this fix makes that totally obsolete!

_## Troubleshooting WiFi Connection / Back to factory settings on Lilygo
The following can help with instability with WiFi or slow performance found on newer boards of lilygo V1.1 boards from 2024-04-25, you might try to follow next steps to flash V1.0.0 version of original firmware, it is also bringing back to factory settings of lilygo so if you checked by mistake "Erase All Flash Before Sketch Upload" to enabled and your lilygo will be half bricked (wifi no connection at all, some strange errors in BE) you can back to factory and then again flash fresh BE. To do that follow next steps.

1. Download ESP32 Flash download tool https://dl.espressif.com/public/flash_download_tool.zip
2. Download V1.0.0 firmware https://github.com/Xinyuan-LilyGO/T-CAN485/tree/arduino-esp32-libs_v3.0.1/firmware
3. Run ESP32 Flash download Tool and set it like on picture (ESP32, Develop) and click OK

image

4. Load V1.0.0 firmware like on picture and write 0x00 next to it, then select COM of your lilygo (arduino IDE need to be closed) and then click START

image

5. Wait few seconds until Finish appear and Volia - your lilygo is like fresh new delivery and 100% working firmware!_

PR #2602, opened 2026-07-13
Add MQTT auto-discovery sensor for charging status

StarkJohan:
Adds a new charging_status MQTT auto-discovery sensor per battery, showing the same human-readable charge/discharge text already shown on the web page (e.g. "Battery charging! (Inverter limiting)", "Battery idle").
Extracts that value into a shared get_charging_status_text() helper (Software/src/devboard/utils/types.h/.cpp) so the web UI and MQTT sensor looks the same. Previously the text was constructed by the web page.
Web page rendering for the battery status now calls this helper instead.
Registers charging_status in batterySensorConfigTemplate[] in mqtt.cpp, so it's automatically cloned per battery (_2/_3) via the existing sensor-config machinery.
get_charging_status_text() returns const char* to avoid unnecessary allocation, consistent with the existing get_balancing_status_text() helper.


StarkJohan:
Closing in favor of a follow-up PR with a more structured charging_state/limiting_factor implementation.

PR #2604, opened 2026-07-13
Adds the "Battery charging! (Inverter limiting)" info via MQTT sensors

StarkJohan:
Adds the "Battery charging! (Inverter limiting)" info via MQTT sensors with two typed fields, charging_state (Idle/Charging/Discharging) and limiting_factor (Idle/Inverter/Emulator/Battery). This gives us access to more structured backend data instead of just generating info on the web page and a nice clear status via MQTT. The web UI looks the same by using get_charging_status_text() composed from the same two helpers. The limiting factor falls back to "Idle" when no limiting factor is set to better visualize the logic connection to the charging state.

image

### DIFF
--- a/Software/src/devboard/utils/types.cpp
+++ b/Software/src/devboard/utils/types.cpp
@@ -31,7 +31,7 @@ LimitingFactor get_limiting_factor(ChargingState state, bool inverter_limits_cha
       return LimitingFactor::Inverter;
     }
     if (user_settings_limit_discharge) {
-      return LimitingFactor::Emulator;
+      return LimitingFactor::UserSetting;
     }
     return LimitingFactor::Battery;
   }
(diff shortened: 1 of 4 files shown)


Software/src/devboard/utils/types.cpp
```diff
@@ -17,6 +17,88 @@ std::string getBMSStatus(system_status_enum status) {
...
+      return "Idle";
+  }
+}
+
+const char* limiting_factor_to_text(LimitingFactor factor) {
+  switch (factor) {
+    case LimitingFactor::Inverter:
+      return "Inverter";
+    case LimitingFactor::Emulator:
+      return "Emulator";
```
dalathegreat: Is "Emulator" self-explanatory enough, or should we consider "UserLimits"?

StarkJohan:
Maybe even UserSetting?

Software/src/devboard/utils/types.cpp
nagyrobi: It is: the user sees 3 components: battery, inverter, and a box between them, it's the "Emulator". It's clear like this, which box is the actual source of decision. UserLimits UserSetting is confusing

dalathegreat (approved):

PR #2607, opened 2026-07-13
Improvement: 🔋Charge power taper for Nissan Leaf

nagyrobi:
### What
This PR adds a CC-CV style charge power taper to the Nissan Leaf battery driver.

As the highest cell in the pack approaches a constant-voltage setpoint, `max_charge_power_W` is ramped down smoothly instead of being passed through from the BMS unmodified.

image

### Why
The Leaf BMS reports a near-binary charge limit on 0x1DC: effectively 2kW, or 0 W on SoH resetted pack, this produces a self-reinforcing oscillation near the top of charge:

1. The weakest cell group reaches the BMS's internal per-cell ceiling **under load**, before the pack is actually full.
2. The LBC cuts its reported charge limit to 0 W. The driver relays that, so `max_charge_power_W` (and therefore the inverter's charge ceiling) goes to 0.
3. Current stops. The cell relaxes back below the ceiling within seconds.
4. The LBC re-opens. The inverter ramps current back in and spikes the same cell immediately.
5. Repeat.

The result is a 0 <-> 2kW square wave on the allowed charge power, and repeated `EVENT_CHARGE_LIMIT_EXCEEDED`, since the inverter cannot unwind its ramp within one 1 s update tick and ends up more than 2 kW above the (now zero) limit.

image

Observed on a 30 kWh Leaf pack behind a Fronius GEN24 over BYD Modbus: the allowed charge power oscillates between 0 W and 2 kW bursts from roughly 97% SOC upwards, with `CHARGE_LIMIT_EXCEEDED` firing repeatedly. The pack effectively stops making useful progress in the top few percent while hammering the weakest cell group with repeated current surges.

image
image

This is a BMS-behaviour problem specific to the Leaf, which is why it is fixed in the Leaf driver rather than in shared safety code. Note that `MAX_CELL_VOLTAGE_MV` (4250) is an emergency-stop threshold, not a charge target, so the existing cell-overvoltage protection is nowhere near engaging in this region and cannot address it.

### How
A taper is appended to the end of the existing "extra safety functions" block in `update_values()`, after `battery_Relay_Cut_Request`.

Charge power is scaled linearly with the highest cell's remaining headroom to the setpoint. The inverter is therefore already down to a few hundred watts by the time the weak cell group nears the LBC's ceiling, so it is never driven hard enough to spike, the LBC never has cause to cut, and charging continues as a gentle trickle rather than a burst.

Result:
image

Curve with some defaults and a 2 kW LBC limit:

| max cell | published `max_charge_power_W` |
|---|---|
| <= 4000 mV | 2000 W (taper inert) |
| 4050 mV | 1333 W |
| 4100 mV | 666 W |
| 4140 mV | 200 W (floor) |
| >= 4150 mV | 0 W |

- **It can only ever lower the ceiling, never raise it.** The final comparison guarantees a BMS-commanded 0 W is never overridden.
- **Placement is deliberate.** It sits downstream of all three writes to `max_charge_power_W` (the 0x1DC limit, the `battery_Full_CHARGE_flag` stop, and the `battery_Relay_Cut_Request` stop). The `battery_Failsafe_Status` switch below it only raises events and does not touch charge power. Every existing stop path therefore still wins.
- **It gates on cell voltage, not SOC.** The LBC cuts on cell voltage, so tapering on the same variable tracks the actual cause and will not need re-tuning as the pack ages or as temperature shifts the SOC/voltage relationship.
- **It is inert until real cell data arrives.** `cell_max_voltage_mV` defaults to 3700 mV, well below the 4000 mV engage point, so the taper does nothing on boot or if group polling fails.
- `cell_max_voltage_mV` is populated in `handle_incoming_can_frame()` from the cell group polls, so it is current when `update_values()` reads it.

### Testing
- Bench/live validation on a 30 kWh Leaf + Fronius GEN24 in done; `CHARGE_TAPER_TARGET_MV` being tuned to sit just below the cell voltage at which the LBC is observed to cut.

> [!TIP]


dalathegreat:
Curious, did you try the inverter low pass filter option? It applies smoothing to the charge values to avoid oscillation

image

nagyrobi:
It has no effect.

Looking at it maybe because it acts on max_charge_current_dA, while BYD-MODBUS.cpp never reads max_charge_current_dA. Register 307 is derived from the power, and the current term comes from a setting.
This looks only useful for  (BYD-CAN, all the SMA variants, Pylon, Growatt, Kostal…) but not by BYD-Modbus?

It should apply the filter to max_charge_power_W too?

nagyrobi:
It doesn't look bad at all:
image

nagyrobi:
image

nagyrobi:
Only works nicely if the scaled SOC is set max to 100%. Can't rescale the soc to lower than 100.
image

nagyrobi:
Edit: this performs better than https://github.com/dalathegreat/Battery-Emulator/pull/2621. This is Leaf-specific, still worth it.

nagyrobi:
Should use https://github.com/dalathegreat/Battery-Emulator/pull/2628 instead of this.

PR #2608, opened 2026-07-13
MEB: Don't transmit CAN until BMS is detected

jonny5532:
### What
The new CAN-error-detection code complains during MEB startup, since BE sends messages to the BMS before the BMS is alive, which results in CAN errors.

Now wait until BMS is out of BMS_DISCONNECTED state, implying that it has received enough messages to be considered present.

**Needs testing from someone with MEB, might not be the correct solution (feel free to correct!)**

> [!TIP]


dalathegreat:
Closing this one, fixed in #2638

PR #2609, opened 2026-07-13
Reduce header dependencies by moving #includes about

jonny5532:
### What
Currently, if you change any battery header file, it recompiles all the others, due to them all being loaded via `BATTERIES.h`. Instead, move the includes so that only the files that actually reference the batteries (the `.cpp` ones) include the headers.

The same for inverters.

### Why
Makes recompilation faster (and probably shaves a few seconds off the bulk compile).

Not very important, was just bugging me slightly!

> [!TIP]


jonny5532:
(I was not expecting it to save 4kb! No idea why...)

dalathegreat (approved):
Awesome 🙌

PR #2610, opened 2026-07-13
GitHub: Remove extraneous + signs from measure action

jonny5532:
### What
It seems I didn't test it enough, increases on the firmware size measurement are showing as `++4` instead of `+4`, this fixes that.

> [!TIP]

### DIFF
--- a/.github/scripts/render_size_comment.py
+++ b/.github/scripts/render_size_comment.py
@@ -95,8 +95,6 @@ def fmt_bytes(n: int | None) -> str:
 def fmt_delta(delta: int | None) -> str:
     if delta is None:
         return ""
-    if delta == 0:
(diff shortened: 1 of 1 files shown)


ngehrsitz:
This shows `+0`, should probably be just `0`

jonny5532:
This was deliberate, IMO '+0' immediately reads as a delta of zero, whereas '0' doesn't - although I guess it is stylistic (what do other people think?)

ngehrsitz:
`+` usually indicates that it increased, but it didn´t. That´s why I find it a bit strange

dalathegreat (approved):
I am quite indifferent, but this PR removes a ton of lines, so less is more ;)

PR #2611, opened 2026-07-13
Unifiy build & release

ngehrsitz:
### What
This PR builds upon the changes in #2594 and now also reuses the compile step for the release process. This makes the compile step the single source of truth and removes the need to also adapt the release step when adding a new board.
It also removes the `esptool merge-bin` invocation and adds the factory images that the build already creates to the build artifact.

### Why

- Makes it easier to add new boards
- Increases release speed, since we reuse the matrix approach
- Prevents a mismatch between the `esptool merge-bin` parameters and `platformio.ini`
- Provide a directly flashable factory image for every build

### How
We now store both the factory and ota image in the build artifacts. The release step then only needs to replace the git shortref with the current release tag. A consequence of storing both images would be that this also duplicates the size of the archive since zip files can´t deduplicate over multiple files. To avoid this I have changed to artifact to be stored as a `.tar.xz` which does not have the same limitation and allows us to distribute both files without a size increase.

### Changes

- Release artifacts are now a `.tar.xz` instead of a `.zip`
- The filenames inside of the archive are now `BE_8828a0b_StarkCMR.factory.bin` and `BE_8828a0b_StarkCMR.ota.bin` instead of `firmware-pr1234-8828a0b.bin`. I removed the `pr` section for simplicity since it wouldn´t always be there.
- An artifact for `ESP32DevKitV1` will now also be created since adding the special handling to excluding it from the release did not seem worth it
- Added top level `permissions:` to restrict the default permissions that are available to a workflow
- `board_upload.flash_size`  for `stark_330` in `platformio.ini` has been aligned with the values from release-assets.yml, it probably makes sense to test this

Test release:
https://github.com/ngehrsitz/Battery-Emulator/releases/tag/v99.0.9
https://github.com/ngehrsitz/BE-Web-Installer/tree/main/images/v99.0.9

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
 
+permissions:
+  contents: read
+
 # This is the list of jobs that will be run concurrently.
 jobs:
   build:
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -103,13 +103,16 @@ jobs:
         SHORT_SHA="${SHA:0:7}"
         cp "$BUILD/firmware.factory.bin" "fw/BE_${SHORT_SHA}_${{ matrix.release_label }}.factory.bin"
         cp "$BUILD/firmware.bin"         "fw/BE_${SHORT_SHA}_${{ matrix.release_label }}.ota.bin"
+        cd fw
+        tar -cJf "../${{ matrix.artifact_name }}.tar.xz" *
 
     - name: Upload artifact for ${{ matrix.board_name }}
       if: matrix.release_label
       uses: actions/upload-artifact@v7
       with:
         name: ${{ matrix.artifact_name }}
-        path: fw
+        path: ${{ matrix.artifact_name }}.tar.xz
+        archive: false # No additional zip compression
         if-no-files-found: warn
 
   measure-size:
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
+  workflow_call:
 
 permissions:
   contents: read
@@ -20,21 +21,27 @@ jobs:
           - board_name: "Lilygo T-CAN485"
             pio_env: "lilygo_330"
             artifact_name: "battery-emulator-lilygo-t-can485"
+            release_label: "LilygoT-CAN485"
           - board_name: "Stark CMR"
             pio_env: "stark_330"
             artifact_name: "battery-emulator-stark-cmr"
+            release_label: "StarkCMR"
           - board_name: "Lilygo T-2CAN"
             pio_env: "lilygo_2CAN_330"
             artifact_name: "battery-emulator-lilygo-t-2can"
+            release_label: "LilygoT-2CAN"
           - board_name: "ESP32 DevKit V1"
             pio_env: "esp32devkit_330"
             artifact_name: "battery-emulator-esp32-devkit-v1"
+            release_label: "ESP32DevKitV1"
           - board_name: "BECom"
             pio_env: "BECom_330"
             artifact_name: "battery-emulator-becom"
+            release_label: "BECom_ES32S3"
           - board_name: "Waveshare RS485CAN"
             pio_env: "waveshare_330"
             artifact_name: "battery-emulator-waveshare-rs485can"
+            release_label: "WaveshareESP32S3RS485CAN"
           - board_name: "Compiler warning check"
             pio_env: "compiler_warning_check"
             artifact_name: "battery-dummy-check"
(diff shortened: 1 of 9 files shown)


.github/workflows/compile-common-image.yml
```diff
@@ -17,21 +21,27 @@ jobs:
...
             artifact_name: "battery-emulator-esp32-devkit-v1"
+            release_label: "ESP32DevKitV1"
           - board_name: "BECom"
             pio_env: "BECom_330"
             artifact_name: "battery-emulator-becom"
+            release_label: "BECom_ES32S3"
           - board_name: "Waveshare RS485CAN"
             pio_env: "waveshare_330"
             artifact_name: "battery-emulator-waveshare-rs485can"
+            release_label: "WaveshareESP32S3RS485CAN"
```
ngehrsitz, line 44: We can´t derive `release_label` from `board_name` because of `BECom` and `Waveshare RS485CAN`

ngehrsitz:
The measure step failed since `main` still uses `zip` archives. I did not implement backwards compatibility since this will only be the case for a very brief period during the changeover. If both have the new archives it works: https://github.com/ngehrsitz/Battery-Emulator/pull/20

jonny5532:
One thing to consider is that the artifact zips are used for distributing test firmware, thus:
- They should be easily accessible by newbies on windows machines (who probably won't have xz installed?)
- Ideally they'd only have one file in, so people don't try to OTA the wrong one.

Could we use separate artifacts, for the distributable OTA image vs the factory one, to avoid the potential confusion? I realise it will unavoidably use more storage, although I'm not sure there's a real limit anyway?

(and could then put the size JSON file in the factory one too, to simplify the OTA ones further).

I'm not sure that the extra compiles at release time are a huge issue, but it does introduce potential ambiguity I guess.

nagyrobi:
I actually like that we can have both the OTA and the factory version (which includes the bootloader) so we can test the whole thing. As long as the filenames differentiate clearly, we can expect from the users that they're educated enough to know which one to choose (+we can add to the docs that).

I miss that the filenames don't have the PR number, the goal was that one collects the various test binaries in a directory, simply viewing the files alphabetically groups the ones belonging to the same PR together (you can have more commits for a PR, test and compare). It was guarded against, if the build doesn't come from a PR, it wouldn't be there. A cosmetic observation, I would use an underscore instead of a dot before the "ota" and "factory" to emphasize the difference a bit.

And yes, a ZIP is more universal, despite many of us like the other.

ngehrsitz:
Github does impose a limit on the number of artifacts you can store. Maybe @dalathegreat can take a look at our current usage and if that actually is an issue: https://github.com/settings/billing
The limits I found are quite low: https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions
If we stick with zip then it doesn´t make a difference if it is one or two artifacts, it will duplicate the storage requirement either way.
Windows 11 does have support, for `tar.xz` but unfortunately not in Explorer you need to use powershell or tell the users to download 7Zip: https://superuser.com/a/1735194
But in terms of making it user friendly is it unreasonable to assume that someone downloading the artifact from a PR is capable of extracting a `.tar.xz`? After all we are not talking about the people just using the Webinstaller and downloading the uncompressed `ota.bin` from the releases section. This has to be someone who wants to install a development version...
I can also add the PR section back in if that´s the consensus.
I went with `.factory.bin` and `.ota.bin` because the ota variant was already present in our releases before. So it just made sense to mirror that. Tasmota also uses this style. The reason for keeping with the dot syntax is that it is a more specific subtype of the file type like with `.tar.xz`.

jonny5532:
Re: "This has to be someone who wants to install a development version...":

Most actual devs will compile it themselves, I think the primary audience for these PR firmwares is people that Dala or I or someone are trying to get to test a change on their battery/inverter. These users are often quite hardware-savvy but not fans of the software side, so I think it needs to be as low-friction as possible.

Re: factory images, I doubt many people will download them from the PR and install them directly (you need esptool etc, you might as well just build it yourself) - the vast majority will use the webinstaller for the initial install and then OTA (which is almost always sufficient).

ngehrsitz:
@jonny5532 Do these users have internet connectivity? Because if they do then the flow should be that you enter the pull request number on the OTA page and the rest happens automatically.
The way this could work without increasing the flash size required on the device would be that the user clicks a "I want to flash this PR in online mode" button. The page then downloads and executes some js/web assembly from the internet. This then takes care of fetching the archive from github, extracting and the uploading it.

nagyrobi:
Extra care must be taken with such solutions; should not open an attack vector for a maliciously infected browser to flash some "random" binary...

dalathegreat:
Yeah I do not like this change, too much risk it will increase support loads. Already very hard to ask people unzipping the .bin file from the .zip file, you would not believe the amount of people that try OTA updating to the .zip directly!

While this PR is nice that it saves storage costs, the current spending comes from running minutes. Last month we spent 0.89$ on storage, but 83.86$ on CPU time for all the runners. FYI @nagyrobi 

image

Honestly if this becomes a problem, I'd be happy to SPEND money on keeping it inefficient but simple for endusers.

nagyrobi:
Understood

ngehrsitz:
If we know that size is not a concern then we could skip compression alltogether and have both factory and ota image as separate direct `.bin` downloads. That would be the most user friendly option.
I will update this PR accordingly.

ngehrsitz:
Continuing in #2615

PR #2614, opened 2026-07-14
MEB Battery: Read SoH from BMS

malec1984:
### What
On newer firmware version SoH can be read from BMS. Use the read value instead of the calculated.

### Why
It might be closer to reality.

### How
Use UDS read data by ID and fetch the SoH from BMS.

> [!TIP]

### DIFF
--- a/Software/src/battery/MEB-BATTERY.cpp
+++ b/Software/src/battery/MEB-BATTERY.cpp
@@ -223,7 +223,7 @@ void MebBattery::
     // 0.935 and 0.9025 are the different conversions for different battery sizes to go from design capacity to
     // total_capacity_Wh calculated above.
 
-    if ( battery_soh_polled > 0) {
+    if (battery_soh_polled > 0) {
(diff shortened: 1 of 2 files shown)


Software/src/battery/MEB-BATTERY.cpp
```diff
@@ -1225,6 +1232,11 @@ void MebBattery::uds_response_handler(uint8_t* data, int len, enum isotp_tatype
             break;
           battery_soc_polled = data[3] * 4;  // 135*4 = 54.0%
           break;
+        case PID_SOH:
+          if (len < 5)
+            break;
+          battery_soh_polled = ((data[3] << 8) | data[4]);
```
dalathegreat, line 1238: Does the older bms report anything on this poll?

malec1984: If the DID is not available then it will respond with request out of range, so the variable battery_soh_polled will stay 0

dalathegreat (approved):

PR #2615, opened 2026-07-15
Unifiy build & release (uncompressed)

ngehrsitz:
Continuation of #2611
### What
This PR builds upon the changes in #2594 and now also reuses the compile step for the release process. This makes the compile step the single source of truth and removes the need to also adapt the release step when adding a new board.
It also removes the `esptool merge-bin` invocation and adds the factory images that the build already creates to the build artifact.

### Why

- Makes it easier to add new boards
- Increases release speed, since we reuse the matrix approach
- Prevents a mismatch between the `esptool merge-bin` parameters and `platformio.ini`
- Provide a directly downloadable and flashable factory and ota image for every build

### How
We now store both the factory and ota image as well as the size.json as separate uncompressed artifacts. The release step then only needs to fetch them. This increases the total storage required but avoids any confusion with the zip files.

### Changes

- Release artifacts are now uncompressed
- Artifact names are now only derived from `board_name` in `compile-common-image.yml`
- `BECom` was renamed to `BECom ES32S3` to stay similar to the existing artifact names. This changes the artifact name from `BE_v1.2.3_BECom_ES32S3.factory.bin` to `BE_v1.2.3_BEComES32S3.factory.bin` which drops the `_` and brings it in line with the other boards.
- `Waveshare RS485CAN` was renamed to `Waveshare ESP32S3RS485CAN`, this only affects the workflow and the size table
- The filenames for PR builds are now `BE_pr22-1fd4ebd_StarkCMR.factory.bin`
- The filenames for push builds are now `BE_main-4d2ee00_StarkCMR.factory.bin`
- An artifact for `ESP32DevKitV1` will now also be created since adding the special handling to excluding it from the release did not seem worth it
- Added top level `permissions:` to restrict the default permissions that are available to a workflow
- `board_upload.flash_size`  for `stark_330` in `platformio.ini` has been aligned with the values from release-assets.yml, it probably makes sense to test this

Tests:
https://github.com/ngehrsitz/Battery-Emulator/releases/tag/v90.0.0
https://github.com/ngehrsitz/BE-Web-Installer/tree/main/images/v90.0.0
https://github.com/ngehrsitz/Battery-Emulator/pull/22

> [!TIP]


ngehrsitz:
Notes from the other PR:

- The measure step on this PR will fail to calculate the delta from main since that still has zips
- It would be good to test the StarkCMR alignment @StarkJohan

ngehrsitz:
@dalathegreat @jonny5532 @nagyrobi Do you have any remaining concerns?

dalathegreat:
Change tested with Stark CMR;
board_upload.flash_size = 8MB
board_upload.maximum_size = 8388608

I think we can merge this yes! Nice work @ngehrsitz :raised_hands:

ngehrsitz:
Great, then let´s merge it 👍

dalathegreat (approved):

dalathegreat:
@ngehrsitz , After merging, this broke the build system. See https://github.com/dalathegreat/Battery-Emulator/actions/runs/29701182010/job/88230394250

ngehrsitz:
I will take a look

ngehrsitz:
Ah, I encountered this before, this is an issue with the merge, I will open a PR

PR #2619, opened 2026-07-16
Improvement: Nissan LEAF🔋perform CAN shut-down sequence before BMS reset power cut

nagyrobi:
Implements the shut-down sequence CAN transmissions from the Nissan battery control specification (CHG_STA_RQ=11b, BTONFN=0b, RLYP=0b, VCM_WakeUpSleepCommand=00b, CAN stop) before the BMS_POWER GPIO is pulled, for both the 24h periodic and MQTT-triggered BMS reset.

### What
Implements the Nissan LEAF "Shut down sequence" from the battery control specification as a graceful CAN shutdown, executed during BMS reset before the BMS_POWER GPIO is pulled. Applies to both the 24h periodic and MQTT-triggered reset, and only for LEAF-type batteries.

### Why
Cutting 12V power to the LBC mid-communication skips the shutdown handshake the BMS expects in the vehicle. Following the spec sequence lets the LBC stop cleanly and enter sleep before power removal, mirroring real vehicle IGN OFF behavior.

### How
Adds a generic `supports/request/completed_bms_shutdown_sequence()` hook on `Battery` and a new `BMS_RESET_SHUTDOWN_SEQUENCE` state that defers `bms_power_off()` until completion (90s timeout). The LEAF driver phases through CHG_STA_RQ=11b (0x1F2), BTONFN=0b and RLYP=0b (0x1D4, CRC recomputed), VCM_WakeUpSleepCommand=00b (0x50B), then stops CAN >1s after LBC silence and waits >1min per spec. Other battery types are unaffected; frame contents self-restore after the reset.

> [!TIP]


nagyrobi:
```
      90.165 MQTT message arrived: [battery-emulator-1/command/BMSRESET]
      90.166 Triggering BMS reset
      90.166 Battery pause begin 1 0 0 0
      90.166 Event: The emulator is trying to pause the battery.
      90.166 LEAF: Starting BMS shut-down sequence
     194.874 Battery pause begin 0 0 0 0
     194.874 Event: The emulator is attempting to resume battery operation from pause.
     194.874 Event: BMS reset event completed.
```

@dalathegreat question is, how do we know if this had any positive effect?

dalathegreat:
@nagyrobi,
I think the only way we can tell if it has had any positive effect, is to see if SOH% starts to move.

I fear if some DTC is active it could also halt SOH% calculation, as we also noticed in the Battery-Upgrade project. We really should get DTC reading into LEAF ASAP! ( I already did the JSON file ready!)

nagyrobi:
This now contains the DTC reading for Leaf and the BMS reset off time allowed to be longer, let's see what happens.

nagyrobi:
No difference in behavior from pack side, closing

PR #2621, opened 2026-07-16
Bugfix: 🐛Apply inverter limits low pass filter to power limits

nagyrobi:
### What
This PR makes the "Inverter limits low pass filter" setting effective for power-based inverters (e.g. BYD-Modbus) by filtering `max_charge/discharge_power_W` instead of the derived `_dA` currents.
Renames it in the web UI from "Inverter limits low pass filter:" to "Ramp up charge limits gradually:" to make it more self-explanatory for the end-users, adjusting the tooltip accordingly.

### Why
Previously the filter only touched the derived current fields, which BYD-Modbus never reads — so the setting was a no-op for it, and unfiltered limits (e.g. from a Nissan Leaf LBC bouncing near full SOC) cause inverter power oscillation.

### How
The filter is moved earlier in `update_calculated_values()` onto the power fields (state in local statics, since battery drivers overwrite the datalayer each cycle), and currents are then derived from the filtered power; initialization uses a first-run flag instead of `==0`, so ramp-up from a 0 limit is also smoothed while decreases stay instant.

#### Before, in `update_calculated_values()`:

1. Battery driver writes raw `max_charge/discharge_power_W` into the datalayer every 1 s.
2. Those are converted to target currents: `target = power_W * 100 / voltage_dV` (skipped if voltage ≤ 1.0 V).
3. If the checkbox on: per direction — if the stored `_dA` is 0, take the target immediately (init); if the target is *higher*, blend `new = (target*10 + old*90)/100` (~10 s time constant at the 1 s loop); if *lower or equal*, take it immediately (safety).
4. The filter's "memory" was the datalayer `_dA` field itself — which the user/remote limit clamps further down also modify, entangling filter state with clamping.
5. Consequence: only the `_dA` fields were smoothed; the power fields stayed raw, so power-based inverters (BYD-Modbus, VCU-CAN) bypassed the filter entirely — the checkbox was a no-op for them.
6. If the checkbox is off: targets are written straight to `max_charge/discharge_current_dA`. Done.

#### After, in `update_calculated_values()`:

1. Battery driver still writes raw `max_charge/discharge_power_W` into the datalayer every 1 s — drivers untouched. Filter moved to the power domain, before any conversion.
2. If the checkbox is on: on the very first loop pass, the raw values are taken as-is (one-time init flag). After that, per direction — if raw is *higher* than the filtered value, blend `new = (raw*10 + old*90)/100` (~10 s time constant); if *lower or equal*, take it immediately (safety drops still land within one cycle).
3. Filter state lives in local `static` variables, not the datalayer — so it survives the driver's overwrite each second and is no longer entangled with the user/remote clamping downstream.
4. The filtered values are written back into `max_charge/discharge_power_W`, so **everything** downstream sees smoothed limits: BYD-Modbus registers 306/307, VCU-CAN frames, MQTT, web UI.
5. Currents are then derived from the (already filtered) power: `_dA = power_W * 100 / conversion_voltage`, where conversion voltage is live pack voltage, falling back to `max_design_voltage_dV` if live reads ≤ 1.0 V, and skipping the update entirely if both are 0. Single filter stage — current-based inverters get equivalent smoothing to before, no double-filtering.
6. Checkbox off: raw power passes through untouched and currents derive exactly as on main — bit-for-bit legacy behavior.

Key behavioral difference vs. original: ramp-up from a 0 limit is now smoothed too (the old `==0` instant-init re-triggered on every dip to zero), which is what kills the GEN24 hunting when the Leaf bounces its limits near full SOC.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -201,7 +201,7 @@ void update_calculated_values(unsigned long currentMillis) {
     datalayer.battery.status.max_discharge_power_W = discharge_power_W_filtered;
   }
 
- /* Calculate allowed charge/discharge currents. Prefer live pack voltage for the conversion.
+  /* Calculate allowed charge/discharge currents. Prefer live pack voltage for the conversion.
      If unavailable (some drivers report 0 before battery comms are up), fall back to the design
      max voltage - conservative, since it under-estimates current for a given power. If that is
      also 0 (generic BMS drivers with no user-configured pack voltage, or BMS-reported cutoff
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -201,14 +201,21 @@ void update_calculated_values(unsigned long currentMillis) {
     datalayer.battery.status.max_discharge_power_W = discharge_power_W_filtered;
   }
 
-  /* Calculate allowed charge/discharge currents*/
-  if (datalayer.battery.status.voltage_dV > 10) {
-    // Only update value when we have voltage available to avoid div0. TODO: This should be based on nominal voltage
+ /* Calculate allowed charge/discharge currents. Prefer live pack voltage for the conversion.
+     If unavailable (some drivers report 0 before battery comms are up), fall back to the design
+     max voltage - conservative, since it under-estimates current for a given power. If that is
+     also 0 (generic BMS drivers with no user-configured pack voltage, or BMS-reported cutoff
+     values before first frame), keep the previous values to avoid div0. */
+  uint16_t conversion_voltage_dV = datalayer.battery.status.voltage_dV;
+  if (conversion_voltage_dV <= 10) {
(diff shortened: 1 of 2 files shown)


nagyrobi:
image
(red: no filter, green: with filter enabled)

nagyrobi:
Result, as approaching SOC (real) full:
image
Setting a scaled SOC max considerably lower (eg. 80%) will hardly mitigate, because the battery at that stage will still report a magnitude higher max charge power. So this is also useful hear the **real** SOC max.

Not as good as https://github.com/dalathegreat/Battery-Emulator/pull/2607#issuecomment-4968883751, but still usable.
This is a universal mitigation though (applies to any battery/inverter combo).

dalathegreat (approved):
I like universal mitigation 👍 Good solution!

PR #2622, opened 2026-07-16
Fix Pylon protocol compatibility with GCE and similar BMS

BatteryPlug:
## What

Three related fixes for BMS that speak the Pylontech HV CAN protocol but deviate slightly from the reference encoding. All three were found while commissioning a DIY 75S NMC pack with a **GCE HV BMS** driving a Solax X1 Hybrid G4.

1. **Max discharge power reported as ~4294959 kW.** The discharge current limit in frame `0x4211` (bytes 6-7) is decoded as `-(value - 30000)`. The GCE encodes this limit with a *positive* offset (`30000 + I`), like the charge limit, so the result was negative and underflowed the unsigned `max_discharge_current_dA` / `max_discharge_power_W` fields, sending garbage to the inverter. `abs(value - 30000)` yields the correct magnitude for both encodings.

2. **User-configured pack voltage limits had no effect.** `update_values()` wrote the BMS-announced charge/discharge cutoffs (frame `0x4220`) into `max/min_design_voltage_dV` on every cycle, overwriting the values set in `setup()` from the user's webserver settings. The cutoffs now only apply when the user has left the corresponding limit at `0`.

3. **Battery info page was sparse and mislabeled.** The frames already carry the highest/lowest cell index and hottest/coldest sensor index (`0x4231`/`0x4241` bytes 4-7) but they were discarded. The page now shows them, plus the BMS cutoffs in volts and the design voltage range actually in effect.

## Why

Without these, a GCE (and likely other slightly-off Pylontech-protocol BMS) reports an absurd discharge power and ignores user voltage limits, which is unsafe for a DIY pack.

## Testing

Tested on hardware (LilyGo T-2CAN, based on v11.0.0) with a GCE HV 75S BMS + Solax X1 Hybrid G4:
- Max discharge power now reads a sane value (8.5 kW) instead of 4294959 kW.
- Pack min/max design voltage set in the UI is now honored.
- Battery info page shows highest/lowest cell number and sensor indices.

No behavior change for BMS that already followed the reference encoding: `abs()` is a no-op for correctly-signed values, and the cutoff precedence only changes behavior when the user has explicitly set a non-zero limit.

### DIFF
--- a/Software/src/battery/PYLON-BATTERY.cpp
+++ b/Software/src/battery/PYLON-BATTERY.cpp
@@ -52,9 +52,17 @@ void PylonBattery::update_values() {
 
   datalayer_battery->status.temperature_max_dC = celltemperature_max_dC;
 
-  datalayer_battery->info.max_design_voltage_dV = charge_cutoff_voltage;
+  // Expose the BMS-announced cutoffs on the battery info page
+  extended_data.charge_cutoff_dV = charge_cutoff_voltage;
+  extended_data.discharge_cutoff_dV = discharge_cutoff_voltage;
 
-  datalayer_battery->info.min_design_voltage_dV = discharge_cutoff_voltage;
+  // BMS cutoffs only define the design voltage when the user has not set explicit limits
+  if (user_selected_max_pack_voltage_dV == 0) {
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Massive thanks for this patch! Nice work 👍

PR #2623, opened 2026-07-16
Add voltage-based SOC plausibility guard (NMC)

BatteryPlug:
## What

Adds an optional guard that clamps the SOC reported to the inverter into a plausibility band derived from the measured min/max cell voltage, using an NMC open-circuit-voltage table. It protects against BMS SOC drift on packs whose coulomb counter recalibrates poorly — common on DIY packs using a Pylontech-protocol BMS.

**How it works**
- 12-point NMC OCV lookup table (3.00 V = 0% … 4.15 V = 100%).
- Cell voltages are IR-compensated using pack current and a configurable per-cell internal resistance, so the estimate does not swing with load.
- Asymmetric band: the **weakest** cell caps the upper bound (discharge protection), the **strongest** cell floors the lower bound (overcharge protection) — correct for unbalanced/mixed-cell packs, where the extreme cell should govern, not the average.
- Operates on `real_soc` **before** SOC scaling, so `reported_soc` and the capacity figures stay consistent.
- Only active for **NMC** chemistry. Disabled by default. Bails out safely on missing/implausible cell data or dead CAN.
- New `EVENT_VOLTAGE_SOC_GUARD` (warning) fires whenever a clamp is applied, with the deviation (in %) as event data.

**Settings (webserver)**

| Setting | Meaning | Default |
|---|---|---|
| `VSOCGUARD` | Enable the guard | off |
| `VSOCIR` | Per-cell internal resistance, 0.1 mOhm units | 5 (0.5 mOhm) |
| `VSOCBAND` | Allowed deviation from the voltage estimate, 0.01% units | 1000 (±10%) |

## Why

On a DIY 75S NMC pack with a GCE HV BMS, the BMS reported **28% SOC while the weakest cell was already at 2.78 V**. The inverter kept discharging and drove the pack into deep over-discharge overnight (cell delta blew out to 450 mV). The cell voltage tells the truth in that situation; the coulomb-counted SOC does not. This guard cross-checks the two and clamps the SOC before it reaches the inverter.

## Testing

Tested on hardware (LilyGo T-2CAN) with a GCE HV 75S NMC BMS + Solax X1 Hybrid G4. With the guard enabled, a BMS SOC that diverges from the cell-voltage estimate is clamped and the event is raised; normal operation (BMS SOC plausible) passes through untouched.

## Notes

The OCV table is a typical NMC/graphite curve; the deviation band absorbs cell-to-cell curve variation. It could be extended to other chemistries with their own tables in future — kept to NMC here since that is what was tested. Builds on the Pylon/GCE fixes in #2622 but is independent of it.


jonny5532:
FYI there are already at least four NMC SoC-to-OCV mapping tables in BE, I'm not sure we need another...

Something like this is sensible, currently the BE cell voltage limits are set pretty wide. I don't think it needs to be this complicated though (synthesising a resistance-compensated SoC).

Simply setting max-power to 0 at cell voltage limits (either chemistry-hardcoded or maybe configurable), with a taper and hysteresis, and forcing SoC to 0%/100% where inverters require that, should be enough. This could probably be enforced across all of the batteries as an extra layer of safety.

(As it happens I am implementing something similar right now for MG, I know there's a another PR for current tapering based on SoC - I fear we might be duplicating each other's functionality somewhat...)

nagyrobi:
Battery (and inverter) independent taper based on SOC: https://github.com/dalathegreat/Battery-Emulator/pull/2628

PR #2624, opened 2026-07-17
Improvement: Make core-task CAN logging non-blocking to prevent task overruns

nagyrobi:
### What
This PR makes the CAN logging paths that run in the core task non-blocking:

- **SD CAN logging** (`add_can_frame_to_buffer`): the full log line (timestamp, direction, ID, DLC, all data bytes) is now built into a single buffer and pushed to the ring buffer with **one send and zero timeout**, instead of up to 9 separate sends with a 2 ms timeout each. Frames that don't fit because the SD writer is stalled are counted and reported as a `[N CAN frames dropped, SD buffer full]` marker in the log once the buffer drains.
- **`add_log_to_buffer`**: zero timeout, and the `logging.println()` call removed from the failure path.
- **USB CAN logging** (`print_can_frame`): the line is built with a single `snprintf` and written with **one `Serial.write`, only if `Serial.availableForWrite()` reports enough room**, instead of ~15 individual `Serial.print()` calls. Dropped frames are counted and reported as a `[N CAN frames not printed]` marker once the port drains.
- **`init_serial`**: `Serial.setTxBufferSize(1024)` before `Serial.begin()`, so bursts are absorbed by a TX ring buffer and `availableForWrite()` reports buffer space rather than raw UART FIFO space.

The log output format is byte-compatible with the current one (same spacing, uppercase hex, line endings), so existing parsers keep working. Two minor behavior improvements: log lines are now written atomically (a full buffer can no longer leave a header without its data bytes mid-line), and zero-DLC frames get a proper line termination.

### Why
With CAN logging enabled, the current implementation can block the core task long enough to delay the 10 ms tasks and trigger `EVENT_TASK_OVERRUN` ("Task took too long to complete"):

- **SD path**: 9 ring buffer sends × 2 ms timeout = up to **~18 ms of blocking per frame** whenever the ring buffer is full. SD cards routinely stall for tens to hundreds of ms during wear-leveling, so this happens in normal operation, and on a busy bus the delay compounds across queued frames.
- **USB path**: `Serial.print()` blocks when the TX buffer is full — i.e. whenever the host isn't draining the port or can't keep up with a busy 500 kbit/s bus.
- Blocking doesn't even save the data: if the sink is stalled, waiting in the core task just delays contactor/precharge handling without getting the frame out.

Additionally, `add_log_to_buffer` has a latent crash: its failure path calls `logging.println()`, which goes through `Logging::write()` back into `add_log_to_buffer()` while the ring buffer is still full — **unbounded recursion ending in a stack overflow** whenever the SD writer stalls long enough.

### How
- The philosophy follows what the logging subsystem already does for syslog: the producer side (core task) only ever does a memcpy into a queue and never waits on a sink. Sinks that can't keep up lose data, and the loss is made visible in the log instead of being silently absorbed as core-task latency.
- `add_can_frame_to_buffer` formats into a 320-byte static buffer (sized for a gap marker + CAN-FD 64-byte payload at 3 chars/byte) and issues one `xRingbufferSend(..., 0)`. The buffer is `RINGBUF_TYPE_BYTEBUF` and the SD writer task streams whatever chunk it receives, so single-send-per-line needs no consumer changes. A `static` counter tracks drops; the next successful send prepends the gap marker. `static` is safe because this is only called from the core task.
- The USB path mirrors the same pattern: build line → check `availableForWrite()` → single `Serial.write` or increment drop counter. The 1024-byte TX buffer matters on UART-console boards, where

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -67,6 +67,9 @@ void register_transmitter(Transmitter* transmitter) {
 // Initialization functions
 void init_serial() {
   // Init Serial monitor
+  // Buffered TX so CAN-log bursts are absorbed instead of dropped; also makes
+  // availableForWrite() report ring-buffer space rather than raw FIFO space.
+  Serial.setTxBufferSize(1024);
   Serial.begin(115200);
 #if (HW_LILYGO2CAN || HW_BECOM || HW_WAVESHARE)
   // Wait up to 100ms for Serial to be available. On the ESP32S3 Serial is
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -469,10 +469,9 @@ static void print_can_frame(CAN_frame frame, CAN_Interface interface, frameDirec
     static char usb_line[288];  // header + up to 64 CAN-FD data bytes at 3 chars each
     static uint32_t usb_frames_dropped = 0;
     unsigned long currentTime = millis();
-    size_t size =
-        snprintf(usb_line, sizeof(usb_line), "(%lu.%02lu) %s%d %lX [%u] ", currentTime / 1000,
-                 (currentTime % 1000) / 10, (msgDir == MSG_RX) ? "RX" : "TX",
-                 (msgDir == MSG_RX) ? (int)(interface * 2) : (int)(interface * 2) + 1, frame.ID, frame.DLC);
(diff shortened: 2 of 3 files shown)


dalathegreat (approved):
I knew there was some better way to do this! The initial implementation relied on raising the baud rate to as high as possible to avoid issues (Serial.begin(115200); was as high as we dared to go). Thanks for improving this 🙌

PR #2625, opened 2026-07-17
Improvement: Skip unchanged values in NVS settings store to reduce flash writes and prevent task overruns

nagyrobi:
### What
This PR makes the NVS settings store (`BatteryEmulatorSettingsStore` in `comm_nvm.h`) skip writes for values that haven't changed:

- `saveUInt()`, `saveBool()` and `saveString()` already read the old value, but only used it to update the `settingsUpdated` flag — the `put*()` call happened **unconditionally**. They now only write when the value actually differs, matching what `saveInt()` already did.
- All change checks (including `saveInt()`) now use an explicit `isKey()` guard instead of a sentinel default value.
- No API or call-site changes; `store_settings()` and everything else works as before. The `settingsUpdated` / `were_settings_updated()` semantics are unchanged (flag set only on real changes) — the write behavior now simply matches the flag.

### Why
Every `Preferences::put*()` is a synchronous NVS flash operation. During any SPI flash write the cache is suspended and code execution stalls on **both** cores — which shows up as `EVENT_TASK_OVERRUN` ("Task took too long to complete") correlated with saving settings from the web UI.

A typical web-UI save changes one or two values, but `store_settings()` writes **every** key, so each save produces a burst of flash writes instead of one or two. NVS is append-structured, so the unconditional rewrites also churn pages faster; a full page triggers NVS garbage collection, which performs a page *erase* — the longest flash stall of all.

Separately, the existing sentinel-based change detection in `saveInt()` has a latent bug: `getInt(name, INT32_MAX)` returns the sentinel for a missing key, so saving a value **equal to the sentinel** into a fresh key compared as "unchanged" and was silently never persisted. The naive way of extending change detection to the other types would have copied that flaw (a stored `false` or empty string is indistinguishable from a missing key when used as its own default).

### How
Each `save*()` method now uses the same pattern:

```cpp
if (!settings.isKey(name) || get*(name, <any default>) != value) {
  settings.put*(name, value);
  settingsUpdated = true;
}
```

- `isKey()` decides "missing key → always write"; only for existing keys is the stored value compared. This makes the default argument irrelevant, eliminating the sentinel-collision class of bugs for all four types (INT32_MAX/UINT32_MAX ints, `false` booleans, empty strings).
- The old value was already being read in every method, so there is no additional NVS read cost — the diff essentially moves the existing comparison from after the write to before it.
- Result: a typical settings save drops from dozens of NVS writes to one or two, and NVS page GC frequency falls proportionally.

1. **The `saveInt()` change is a bug fix, not just refactoring**: `saveInt("X", INT32_MAX)` on a fresh key was silently dropped before this PR. The other methods didn't have the bug only because they didn't skip writes at all.
2. **Skipped rewrites are not a corruption-defense regression**: with skip-on-unchanged, a key already holding the target value no longer gets its NVS entry refreshed. That is intended — NVS entries don't decay, and integrity is handled by NVS's own CRCs at the page level; rewriting identical values was never a corruption defense.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.h
+++ b/Software/src/communication/nvm/comm_nvm.h
@@ -62,8 +62,9 @@ class BatteryEmulatorSettingsStore {
   }
 
   void saveInt(const char* name, int32_t value) {
-    auto oldValue = getInt(name, std::numeric_limits<int32_t>::max());
-    if (value != oldValue) {
+    // isKey() check instead of a sentinel default: saving a value equal to the
+    // sentinel into a missing key must not be skipped.
+    if (!settings.isKey(name) || getInt(name, 0) != value) {
       settings.putInt(name, value);
       settingsUpdated = true;
     }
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
LGTM, nice find, and great for those that tweak settings often and inadvertently get perf-issues!

PR #2628, opened 2026-07-17
Feature: 🔨📉🔋SOC-based, universal charge power taper towards inverter 

nagyrobi:
### What
This PR implements an optional, battery-integration-independent charge power taper based on scaled SOC: the allowed charge power is linearly reduced from full at (100% − band) down to 0 W at 100% scaled SOC, with an optional configurable floor (minimum charge power held through the tail of the band). Configurable via a new checkbox next to the existing low pass filter setting, with taper band (percent) and floor (W) fields shown only when enabled. 

image

A log line is emitted when tapering engages:
`Charge power tapering based on SOC engaged at 91.8 (real) 95.1 (scaled)`

### Why
Today the charge limit stays at full BMS allowance until the battery reports full, then drops abruptly to 0 W, which can cause overshoot and oscillation on inverters with slow control loops (e.g. BYD-Modbus/Fronius GEN24 polling every 5 s). A gradual approach to full gives the inverter a smooth setpoint to follow and works for every battery integration centrally, unlike the per-driver `RAMPDOWNSOC` implementations which are keyed on real SOC and only exist in some drivers. The floor keeps inverters above their minimum stable charging power near full (avoiding standby cycling), provides a steady cell-balancing trickle, and guarantees the charge session terminates instead of asymptotically stalling just below full - since inverter and battery combos can vary a lot, this is a user setting rather than a constant. Recommended value for this is 5-10% of the inverter max power.

### How
A new `filter_charge_taper_soc()` runs after `update_machineryprotection()` and before `filter_inverter_limits()`, scaling `max_charge_power_W` in place (drivers rewrite it every cycle, so nothing compounds) and min-capping the derived `max_charge_current_dA`. The effective current cap (remote if active, else user) is applied to the taper input first, so the band acts on the limit the inverter actually sees rather than on BMS headroom; after the linear factor, the result is raised to the floor when configured (default 0 = disabled), dropping to 0 W only at exactly 100% scaled SOC where the existing battery-full cutoff takes over. Settings follow the existing `LOWPASSFILTER` pattern (`CHGTAPERSOC` bool, `CHGTAPERBAND` stored as whole percent loaded ×100 to pptt, `CHGTAPERFLOOR` in W); independent of the low pass filter — none, either, or both can be enabled.

The taper operates on the aggregate limit downstream of all multi-battery aggregation, so it works wherever the existing low pass filter is enabled or not, including the extremes-override behavior.

> [!TIP]


nagyrobi:
First test: 
image
I will add a configurable floor to eliminate the trickle noise when it flattens out.

nagyrobi:
This is perfect now!

- To be correct, set your **Max charge speed:**  and  **Max discharge speed:** in BE to the values supported by the inverter.
- You can Rescale SOC anywhere. 
- Taper reacts to the rescaled SOC. With band you select at what percentage minus from the top should it start tapering. Slows down charging until it reaches 100%.
- You can set a floor in W, under which it shouldn't oscillate anymore. When Max power goes under this, sticks constantly at this level until your SOC hits 100%. 
- After this, Max power will be 0 and charging stops.
- As cells will slowly drift down, SOC will decrease, Max power will slowly increase back, and it's up to the inverter to decide when to start charging. Fronius will only start charging again it when it goes under 99%.
- Rebooting the node in the process causes no hiccup, the taper will engage automatically.

Apparently this makes it work like Fronius intended.

I tested with SOC max rescaled to 95%, Band 5 and Floor 900W, then modified to 600W and then to 300W. It's recommended to be between 5-10% of the inverter rated power:
image

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -1745,6 +1762,22 @@ const char* getCANInterfaceName(CAN_Interface interface) {
         <input type='checkbox' name='LOWPASSFILTER' value='on' %LOWPASSFILTER% 
         title="Smooths sudden increases in the battery's charge power limits before sending them to the inverter to prevent oscillation, using a low pass filter." />
 
+        <label>Charge power tapering based on SOC:</label>
+        <input type='checkbox' name='CHGTAPERSOC' value='on' %CHGTAPERSOC%
+        title="Linearly reduces the allowed charge power over the top of the SOC window, from full power at (100pct - band) down to 0W at 100pct scaled SOC, for a smooth approach to full instead of an abrupt cutoff." />
+
+        <div class='if-chgtapersoc'>
+        <label>Charge taper band, percent: </label>
+        <input type='number' name='CHGTAPERBAND' value="%CHGTAPERBAND%"
```
dalathegreat: Most users are aware of RAMPDOWNSOC% setting, so I think the CHGTAPERBAND should be that instead
It is easier to think when should we start to ramp down power? 90%! Instead of a Band option (10%)

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -1745,6 +1762,22 @@ const char* getCANInterfaceName(CAN_Interface interface) {
...
+        min="1" max="50" step="1"
```
dalathegreat: Min value could be 97%, and lowest possible 50% (I assumed the change to RAMPDOWNSOC)

Software/src/devboard/webserver/settings_html.cpp
```diff
@@ -1745,6 +1762,22 @@ const char* getCANInterfaceName(CAN_Interface interface) {
...
+        title="Width of the taper band in percent of scaled SOC. 5 = tapering starts at 95pct and reaches 0W at 100pct" />
+
+        <label>Charge taper floor, W: </label>
```
dalathegreat: More undestandable name would be "float charge power"

dalathegreat (changes requested):

Software/src/devboard/webserver/settings_html.cpp
nagyrobi: Min value could be 97%, and lowest possible 50%  - a bit contradictory ????

Software/src/devboard/webserver/settings_html.cpp
nagyrobi: Agreed, changed to "Start tapering at SOC" with the same mental model as RAMPDOWNSOC — full power until the configured scaled SOC, then linear taper to 0W at 100%. Kept it in whole percent rather than pptt for readability. Internally it still derives the same band value, so the taper logic is untouched.

dalathegreat (approved):

PR #2630, opened 2026-07-18
CAN: Make sure we don't try to send to uninitialized interfaces

jonny5532:
### What
Currently, if you select an add-on interface that you don't have (or isn't wired correctly), it will fail to initialize, and then BE will crash when it later tries to send (and likely get stuck in a bootloop).

This deliberately nulls out the relevant pointers, and ensures that everywhere they are accessed, they are checked for non-null first.

Notably it doesn't try freeing the driver instances - `init_CAN` won't get called again, so the memory leak doesn't really matter, and we're in a fault state anyway. Trying to free the instances risks another whole category of bugs from trying to destruct a driver that has failed to fully initialize.

### Why
Avoids crashes/bootloops.

> [!TIP]


dalathegreat (approved):
Nice, this solves the crash I got when loading wrong hardware.bin onto my Stark. Very good improvement 🙌

PR #2635, opened 2026-07-18
Nissan LEAF: log detected battery generation (ZE0/AZE0/ZE1)

nagyrobi:
### What
This PR adds a log entry reporting the auto-detected Nissan LEAF battery generation once the battery is communicating.

### Why
The LEAF driver detects the generation from CAN frames, but it was only visible on the battery's advanced webserver page — a log line confirms it at startup for diagnostics and support requests.

### How
`update_values()` logs the generation via `logging.printf` once `battery_can_alive` is set, guarded by a one-shot member; since ZE0 is the default until AZE0/ZE1-specific frames arrive, an upgraded detection is logged again.

> [!TIP]

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -222,13 +222,13 @@ void NissanLeafBattery::
     reported_battery_Type = LEAF_battery_Type;
     switch (LEAF_battery_Type) {
       case AZE0_BATTERY:
-        logging.printf("Nissan LEAF battery generation identified: AZE0 (2013-2017)\n");
+        logging.printf("Battery generation: AZE0\n");
         break;
       case ZE1_BATTERY:
-        logging.printf("Nissan LEAF battery generation identified: ZE1 (2018-2023)\n");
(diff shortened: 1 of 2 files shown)


Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -216,6 +216,23 @@ void NissanLeafBattery::
     datalayer_battery->status.balancing_status = new_status;
   }
 
+  // Log the identified battery generation. ZE0 is the default until AZE0/ZE1
+  // specific CAN frames are seen, so an upgraded detection is logged again
+  if (battery_can_alive && (reported_battery_Type != LEAF_battery_Type)) {
+    reported_battery_Type = LEAF_battery_Type;
+    switch (LEAF_battery_Type) {
+      case AZE0_BATTERY:
+        logging.printf("Battery generation: AZE0\n");
```
dalathegreat, line 225: Better to before the switch case do;
logging.printf("Battery generation: );

Switch case
logging.printf("AZE0\n");

Should save some flash

nagyrobi:
Reconsidering later

Issue #2590, opened 2026-07-12
[BUG] ERROR 'CANFD_BUS_ERROR' on startup since V11.0.0


Webwanze:
**Describe the bug**
Error 'CANFD_BUS_ERROR' appears once in log directly after startup (powercycle).
But communication seems to be OK.
Using an add-on CAN-FD on WAVESHARE for VAG MEB Battery.

**To Reproduce**
restart BE

**Screenshots**

<img width="772" height="662" alt="Image" src="https://github.com/user-attachments/assets/b8c6c69b-fc0c-4ed6-9d2f-bb3f801129d4" />

**Version and settings (please complete the following information):**
- Software version: V11.0.0
- Battery used: VAG MEB
- Inverter communication protocol: none
- Hardware used for Battery-Emulator: WAVESHARE
- GPIO controlled contactors: no
- MQTT: yes

dalathegreat:
From Discord analysis on this issue;

_On MEB, if you do like me and power the BE and BMS in the same time then you always have the BE booting faster and there is no other ECU on the bus to ACK the messages sent, after a short time BMS is active so everything is fine_

Think we need to do the same as on LEAF; and to hold spamming messages towards the BMS before it has started.

jonny5532:
For clarity, you will now get a CAN_BUS_ERROR after about 15 failed transmits, or 96 corrupt receives - which as Dala says in this case is probably due to startup timing. It's a purely informationall message, everything will work as before.

We can either tweak the battery integrations to avoid this situation, or possibly just add a timer so it doesn't fire these CAN_BUS_ERROR messages in the first two seconds...

Webwanze:
I would prefer to have a timer to avoid such messages/warnings. Those are displayed on the display, even if they could be ignored.

Another question is, if warnings/errors should disappear somewhen automatically. e.g. when a (longer) period without a problem passed, some messages might be removed automatically. For example this one.

dalathegreat:
We should be able to remove this event entirely by first waiting on an incoming MEB message before blasting out the full vehicle emulation.

jonny5532:
The change has been made here, experimentally (see links on the PR for test firmwares): https://github.com/dalathegreat/Battery-Emulator/pull/2608

(Entirely untested, I hope it works!)

malec1984:
Should be covered by this BMS reset feature
https://github.com/dalathegreat/Battery-Emulator/pull/2638

dalathegreat:
[...]

Issue #2354, opened 2026-05-20
[BUG] No precharge after rebooting BE


Webwanze:
**Describe the bug**
When clicking 'reboot' or also after an OTA update, the precharge will not be executed.
Therefore the battery stays disconnected (HV), data is received and also displayed correctly.
After a powercycle everything starts up as expected - including precharging.

**To Reproduce**
Just click 'reboot' within WebIf

**Expected behavior**
Precharge should also be carried out when rebooting BE

**Screenshots**
- not needed -

**Version and settings (please complete the following information):**
- Software version: 10.9.0 (but did never work before)
- Battery used: VAG MEB 82 kWh
- Inverter communication protocol: SMA compatible BYD Battery-Box HVS
- Hardware used for Battery-Emulator: LilyGo T-CAN
- GPIO controlled contactors: no
- MQTT: yes

bradlrein:
In addition to this:
If the +12V ignition is switched by the "BMS Power" output with a relay, the precharge never happens, even after a cold reboot.
It seems to be a timing issue.

ekholm:
This looks like it was fixed since the report (which was on 10.9.0). Two relevant changes landed:

1. **v12.0.0** — the MEB BMS-reset state machine (`eeff4155`): on the first CAN message after boot, if the BMS is *not* in Init (i.e., it stayed awake because only the emulator rebooted — exactly the webUI-reboot/OTA case), the emulator now triggers a BMS reset so the pack walks through Init with KL_15 properly gated, and precharge proceeds.
2. **v11.1.0** — GPIO hold-across-reset (`3b289cbf`…`627d8c16`): outputs like BMS Power are held driven *through* a software reboot instead of glitching, which addresses the "+12V ignition switched by BMS Power relay" case from the comment.

Supporting data point: your log in #2691 (v12.0.0) shows `MEB: Start precharging` ~11 s after bootup, so the boot→precharge path is working on your setup on v12.

Could you retest the specific webUI-reboot and OTA paths on v12.0.0? If precharge now runs, this can be closed. @bradlrein — same ask for the BMS-Power-relay cold-boot case on ≥v11.1.0; if it still fails there, a fresh log in a new issue would help since it's a different mechanism (boot-time pin ordering rather than BMS state).

PR #2638, opened 2026-07-18
MEB Battery: Implement BMS Reset

malec1984:
### What
Many times the BMS goes in fault state. This PR is adding the BMS reset button in the More Battery Info page.
Fix the HVLM_13 message.
Update one naming from BMS_22 message.

### Why
Should help with getting the BMS in the right state.

### How
Allow the BMS to enter sleep by setting the KL15 signal to off and the network management to sleep.
Suspend TX from BE and wait for the BMS to enter sleep.
Resume TX from BE as normal.
Suppress the CAN bus error events which are set because the BMS is not setting ACK during wakeup.

> [!TIP]


Software/src/devboard/utils/events.cpp
```diff
@@ -194,6 +193,21 @@ void clear_event(EVENTS_ENUM_TYPE event) {
   }
 }
```
malec1984, line 199: @dalathegreat that's something I am not sure you want to have, but with this I can remove the entry for CAN errors which are not possible to avoid during BMS wakeup

Software/src/battery/MEB-BATTERY.cpp
```diff
@@ -1192,6 +1214,94 @@ void MebBattery::transmit_can(unsigned long currentMillis) {
...
+        bms_reset_ms = currentMillis;
+        //logging.println("MEB: BMS reset: resuming CAN communication");
+      }
+      break;
+
+    case BmsResetState::CLEAR_EVENTS:
+      if (currentMillis - bms_reset_ms >= BMS_RESET_CLEAR_EVENTS_MS) {
+        // Step 7: remove the CAN events that were set because the BMS did not ACK during
+        // the bus-wakeup phase. remove_event() zeroes occurences so
+        // the events actually disappear from the log.
```
malec1984: removing the events here, it's not the perfect way as I have created dependancy on the number of CAN interfaces. Maybe I should move this to a functions like remove_can_events and place it in the events.cpp file

jonny5532:
If it's not possible to avoid CAN errors during wakeup, I would suggest one of:

1. a global `ignore_can_errors(bool)` to globally disable the CAN error event firing (and reenable later)
2. a global `ignore_can_errors_until(ms)` to suppress until some time (safer)
3. adding a `suppress_errors` bool default argument to `CanBattery::transmit_can` which sets some kind of timeout (but can be limited per-interface)
4. passing CanBattery's `can_interface` along to a `ignore_can_errors` function for per-interface limiting

Unless you can think of something else? Option 1/2 are probably simplest!

malec1984:
@jonny5532 thanks for the suggestion, I decided for number 2, but with additional interface selection, so only the one used for battery is ignoring the errors.

Webwanze:
Tested with SMA SE @ MEB 82 kWh:

Fixed issue: 
[BUG] ERROR 'CANFD_BUS_ERROR' on startup since V11.0.0  #2590
[BUG] No precharge after rebooting BE  #2354

Did no tests with Kl15 etc.

Thank you @malec1984

Software/src/devboard/utils/events.cpp
```diff
@@ -524,12 +533,47 @@ const char* get_emulator_status_string(EMULATOR_STATUS status) {
...
+
+// Returns true while any interface has an open ignore window that 'event' belongs to.
+// Checked per interface so several windows can be active at once.
+static bool can_error_ignored(EVENTS_ENUM_TYPE event) {
+  uint32_t now = millis();
+  for (uint8_t i = 0; i < NO_CAN_INTERFACE; i++) {
+    // Signed difference keeps the "in the future" test correct across millis() wraparound.
+    if ((int32_t)(can_errors_ignore_until_ms[i] - now) > 0 && is_can_error_of_interface(event, (CAN_Interface)i)) {
+      return true;
+    }
```
jonny5532, line 560: I think we also need a:
```C
if(can_errors_ignore_until_ms[i] > 0) {
   can_errors_ignore_until_ms[i] = 0;
}
```
To reset it, so it doesn't become active again in 24 days? (that isn't quite right, as it should only reset if definitely past), maybe:

```C
if ((int32_t)(can_errors_ignore_until_ms[i] - now) > 0) {
   if(is_can_error_of_interface(event, (CAN_Interface)i)) {
      return true;
   }
} else if(can_errors_ignore_until_ms[i] > 0) {
   can_errors_ignore_until_ms[i] = 0;
}
```



(Maybe using the millis thing as both a flag and timeout is too hacky, but I like the reduced code...)

malec1984: I think better just use milli64() to be safe.

dalathegreat:
Okay now 3 people have used PR2638 successfully, are we happy with it now? :)

Webwanze:
@malec1984 

Tested again with the last small fix. Took bin from above, hope that's correct.

Finding: The new bin is a lot smaller than the previous one...

BE is starting up, but then runs into a kernel panic. Reboots. And so on ...

malec1984:
@Webwanze thanks for testing, it's probably related to some other PR which got merged, will check it anyway.

dalathegreat:
PR https://github.com/dalathegreat/Battery-Emulator/pull/2650 ongoing for fixing the FD issues. 

I am merging this PR now :+1: Nice work @malec1984 :raised_hands:

dalathegreat (approved):

PR #2640, opened 2026-07-18
Make USB debug logging non-blocking to prevent task overruns

nagyrobi:
### What
This PR makes the general USB debug-log path non-blocking: `Logging::write()`, `Logging::printf()` and `add_timestamp()` now route through a shared `usb_log_write()` helper that only writes when `Serial.availableForWrite()` reports room. Dropped output is reported with a `[USB log output dropped, host not draining port]` marker, always flushed before output resumes so the gap appears at the right place.

### Why
These three sites wrote to Serial unconditionally, so with USB logging active and a host that stops draining the port, every `DEBUG_PRINTF` from the core task blocks and raises `EVENT_TASK_OVERRUN`. This is the same bug class already fixed for CAN frame logging (#2624) — after this PR, no logging sink (SD, USB CAN frames, USB debug, web, syslog) can block the core task.

### How
One static helper mirrors the merged CAN-frame pattern: check room for marker + chunk, write both or drop and count — a stalled host now means drop-with-marker instead of block-until-delivered. `Serial.write` locks internally in both `HardwareSerial` and `HWCDC`, so multi-task callers are safe; the non-atomic drop counter can at worst miscount the marker, never corrupt data.

> [!TIP]

### DIFF
--- a/Software/src/devboard/utils/logging.cpp
+++ b/Software/src/devboard/utils/logging.cpp
@@ -13,6 +13,34 @@
 
 bool previous_message_was_newline = true;
 
+// ---- USB debug-log sink ----
+// Writes only when the TX ring buffer has room. This path runs in the core task
+// (and others): it must never block on a host that stops draining the port
+// (EVENT_TASK_OVERRUN). Lost output is reported with a marker once the port
+// recovers, and the marker is always flushed before output resumes so the gap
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Good improvement 👍

PR #2606, opened 2026-07-13
Make firmware size report clearer, add links

jonny5532:
### What
This modifies the new firmware size report, adding download links to the Firmware files, and hopefully making it a bit clearer:

image

(removes environment and previous-flash-size, and makes things bold instead of using warning icons)

Thoughts? I realise the layout is somewhat subjective. I removed the RAM (it's not very interesting/slightly misleading anyway).

> [!TIP]
> [You can help test this PR with this guide](https://github.com/dalathegreat/Battery-Emulator/blob/main/CONTRIBUTING.md#downloading-a-pull-request-build-to-test-locally-)

PR #2642, opened 2026-07-19
Fix measure-firmware-size artifact filter

ngehrsitz:
### What
This PR fixes an issue in the measure-firmware-size action introduced in #2606. The measure step currently fails when the base branch itself does not have artifacts that were created by a push and the artifact JSON map is thus empty. This is the case if you create a PR and then another PR that would merge into the first PR. On the second PR it will currently filter out the artifacts that exists on the first PR because they were not built by a `push` event.

### How

- Remove the push filter, filtering by SHA will still identify the correct artifacts
- Add more graceful handling for invalid JSONs

> [!TIP]

### DIFF
--- a/.github/scripts/render_size_comment.py
+++ b/.github/scripts/render_size_comment.py
@@ -312,7 +312,11 @@ def main() -> int:
             "> No size reports found in either the PR build or the base build.\n"
         )
     else:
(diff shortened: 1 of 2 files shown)


jonny5532:
Looks sensible!

dalathegreat (approved):

PR #2643, opened 2026-07-19
MQTT: reduce heap usage, especially on double/triple battery setups

nagyrobi:
### What

This PR implements six MQTT-side heap optimizations and fixes three bugs found along the way. It reduces both permanent and per-cycle heap usage of the MQTT module, with the largest gains on double/triple battery setups, where memory cost previously grew with battery count.

**⚠️ It contains one breaking change (multi-battery raw-topic consumers only):** battery _2 and _3 state values move from suffixed keys on `<name>/info` (`SOC_2`, `battery_voltage_3`, …) to **un-suffixed keys on per-battery topics** `<name>/info_2` and `<name>/info_3`.

Not affected:
- Single-battery setups — `<name>/info` topic, keys, and payload are unchanged.
- Home Assistant autodiscovery users - discovery configs are updated automatically (new `state_topic` + `value_template`); all `unique_id`s are preserved, so existing entities migrate in place with no duplicates.
- `spec_data_*` / `balancing_data_*` / `events` / `status` topics - unchanged.

Affected: very fey users who might be consuming battery 2/3 values from raw `<name>/info` JSON keys must point at `<name>/info_2` / `<name>/info_3` with un-suffixed keys. Needs a wiki + changelog note.

Bug fixes included:
- **UB in `publish_events()`**: `order_events.clear()` was called *inside* the range-for iterating `order_events`, invalidating the live iterators whenever more than one event was pending. The clear now happens after the loop.
- **Per-battery charged/discharged energy**: `set_battery_attributes()` always read battery _1's `total_charged/discharged_battery_Wh`, so batteries 2/3 published battery _1's energy counters. Each battery now reports its own totals.
- **Discovery completion flags** (`ha_common_info_published`, `ha_cell_voltages_published`, `ha_buttons_published`) were set on the *first* successful publish; a mid-loop failure then permanently skipped the remaining discovery configs. Flags are now set only after all configs went out, so failed discovery retries on the next cycle.

Additionally, cell voltage and balancing payloads are now published on a fixed 60 s cadence (independent of the configured publish interval), and `/status` `online` is published retained once on connect following the standard LWT pattern, instead of non-retained every cycle.

### Why

On 2–3 battery setups the MQTT module's heap footprint grew with battery count in several independent ways:

- Suffixed String keys (`doc["SOC" + suffix]`) allocated ~25 temporary heap Strings per battery on *every 5 s publish cycle*, and forced ArduinoJson to copy every key into the document pool (const `char*` keys are stored zero-copy by pointer; String keys are not). The combined `/info` document and payload also grew cumulatively with battery count.
- `sensorConfigs` was a permanent `std::list` of ~28 entries × battery count, each with 1-3 never-freed `strdup()` strings and a heap-allocating captured `std::function` - several KB of permanent heap serving one-shot discovery.
- Five separate `static JsonDocument`s each retained their own memory pool, and several early-return paths skipped `doc.clear()`, keeping a full payload allocated until the next successful cycle.
- The cell voltage / balancing state payloads (the largest recurring messages, 100+ elements per battery) built a full ArduinoJson document every cycle - the biggest transient allocation of the publish path.
- Hot paths rebuilt the same Strings repeatedly: six temporary topic Strings on every received MQTT message, temporary String wrappers around event strings.
- The esp-mqtt outbox had no size cap, so a stalled broker/Wi-Fi link let queued traffic grow on the heap until OOM.

Beyond peak usage, the constant allocation churn drives heap fragmentation over weeks of uptime - a real failure mode on long-running emulator installs.

Publishing cell voltage and balancing payloads at 5s cadence is completely overkil. This is slow-moving data, even the BMS refreshes it much slower.

### How

- **Per-battery state topics** (`/info`, `/info_2`, `/info_3`) with identical un-suffixed keys. All JSON keys become `const char*` literals (zero-copy in ArduinoJson), the temporary key Strings disappear, and document/payload size is constant per battery instead of cumulative. Battery _1 keeps the historical topic and keys, so single-battery users see no change.
- **On-the-fly discovery generation**: the per-battery name / entity id / value_template variants are generated into stack buffers at discovery time. The `sensorConfigs` list, its `strdup()`d strings, and the `std::function` conditions (now plain function pointers) are gone; nothing is retained on the heap for discovery.
- **One shared `JsonDocument`** across all publish paths (they run strictly sequentially), with an RAII `DocClearGuard` that clears it on scope entry and every exit path - including the early returns that previously leaked a  populated document.
- **Manual serialization of cell voltage / balancing state payloads** with `snprintf` straight into `mqtt_msg` (with buffer-overflow guards). The three copy-pasted per-battery blocks are deduplicated into shared helpers (resolves the in-code TODO). Discovery, which is one-shot and deeply structured, keeps using ArduinoJson.
- **Precomputed topics**: the per-battery info topics and the six command topics are built once in `init_mqtt()`; `mqtt_message_received()` compares against them instead of rebuilding six Strings per message; event enum/level strings are assigned as `const char*` (numeric event fields stay String-typed so the events payload is byte-identical to previous releases).
- **esp-mqtt outbox cap**: `mqtt_cfg.outbox.limit = 16 KB` bounds heap growth when the broker is unreachable, turning a slow OOM into a bounded, recoverable enqueue failure. Task stack is deliberately left at default pending a measured high-water mark.

### Notes:
- `mqtt.h` is unchanged. No symbols removed from `mqtt.cpp` are referenced anywhere else in the tree (verified by grep).
- Known cosmetic difference: manually serialized cell voltages print fixed 3 decimals (`3.700` where ArduinoJson printed `3.7`). Numerically identical; HA templates unaffected.
- Pre-existing quirk deliberately preserved: Tesla DC-DC discovery configs are published for batteries 2/3 but never receive state (same observable behavior as before - the sensors show as unavailable/unknown in HA either way).

> [!TIP]


nagyrobi:
On Lilygo T-CAN485:

Before:
Single batt, Send all cellvoltages via MQTT **disabled**, Free heap: 98736
Single batt, Send all cellvoltages via MQTT **enabled** , Free heap: 96664
**Double** batt, Send all cellvoltages via MQTT **enabled** , Free heap: 89812


After:
Single batt, Send all cellvoltages via MQTT **disabled**, Free heap: 104316
Single batt, Send all cellvoltages via MQTT **enabled** , Free heap: 100812
**Double** batt, Send all cellvoltages via MQTT enabled , Free heap: 98028

nagyrobi:
Send all cellvoltages via MQTT disabled, Waveshare ESP32-S3-RS485-CAN:

Before:
MQTT function (MQTT task) max load last 10 s: 4799 us
WIFI function (MQTT task) max load last 10 s: 3431 us

After:
MQTT function (MQTT task) max load last 10 s: 2548 us
WIFI function (MQTT task) max load last 10 s: 2022 us

dalathegreat:
Think this PR might require major version number bump to inform users about MQTT possisble breaking change

nagyrobi:
And a little surprise for Home Assistant autodiscovery users:

image

The **Visit** link appearing in HA will open BE web interface in the browser.

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -221,9 +236,9 @@ static String generateSensorDefaultEntityId(const String& object_id) {
 
 void set_common_discovery_attributes(JsonDocument& doc) {
   doc["device"]["identifiers"][0] = device_id;
-  doc["device"]["manufacturer"] = "DalaTech";
```
dalathegreat, line 224: Can we keep manufacturer, but have it listed as "[FOSS](https://en.wikipedia.org/wiki/Free_and_open-source_software)" instead?

nagyrobi: Done

nagyrobi:
Text suggestion for the release notes:

```
**MQTT: reduced memory usage and improved reliability**

This release significantly lowers the heap usage of the MQTT module, with the biggest gains on double and triple battery setups, and fixes a bug where batteries 2 and 3 reported battery 1's charged and discharged energy totals. Cell voltages and balancing states are now published every 60 seconds instead of every 5, combined into a single message, while all live values — power, current, battery voltage, SOC, and cell min/max/delta — continue to update at the normal publish interval. Availability on the `/status` topic is now published retained once on connect, following the standard last-will pattern.

Two changes affect existing setups. Battery 2 and 3 values have moved from suffixed keys on `<name>/info` (such as `SOC_2`) to their own topics, `<name>/info_2` and `<name>/info_3`, using plain un-suffixed keys. The separate `balancing_data*` topics have been removed, with that data now included in `spec_data*` under the `cell_balancing` key, ensuring perfect sync with `cell_voltages`. Home Assistant autodiscovery users need to take no action, as entities migrate automatically; only manually configured MQTT sensors, Node-RED flows, and similar integrations need updating.
```

dalathegreat (approved):
Extremely nice to see so great MQTT improvements 🙌 Will bump up to v12 with this!

PR #2646, opened 2026-07-19
Improvement: Remove "Use CanFD as classic CAN:" setting

dalathegreat:
### What
This PR removes the two checkboxes "Use CanFD as classic CAN:" & "Use CanFD2 as classic CAN:" from the webserver configuration page

### Why
The primary thing the setting did was make FD chip actively reject FD frames if ticked. This functionality was not useful in any type of integration, and it caused confusion when to use it

### How
We now always start out any MCP2518FD chip in FD mode, and allow sending of NormalCAN and CANFD messages on the interface. This simplifies configuration, with less chance for user mistakes

> [!TIP]


dalathegreat:
Tested OK on Stark CMR v2, with the CAN-FD port connected to a LEAF battery :heavy_check_mark:

PR #2649, opened 2026-07-19
Fix merge issue in Stark config

ngehrsitz:
### What
Fix the build issue from #2615

> [!TIP]


ngehrsitz:
The issue here was that #2571 added the same missing value, which is why we ended up with a duplicate after merging both PRs.
To prevent issues like this you could add the Github branch protection check `Require branches to be up to date before merging`. But that then means that you need to always merge in the latest changes before you can merge a PR.

dalathegreat (approved):
Thanks for the quick fix @ngehrsitz 🙌

PR #2650, opened 2026-07-19
Investigate broken CAN-FD on 11.2.dev

dalathegreat:
### What
This PR Investigates broken CAN-FD on 11.2.dev

### Why
On ESP32-S3's that have CAN-FD add-ons, the boards stop operating randomly or crash with the latest 10.12.dev changes. 

### How
It is suspected that PR #2571 is the reason, so let's unroll it one step at a time.

> [!TIP]

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -61,7 +61,6 @@ custom_component_remove =
     espressif/ksz8851snl
 build_flags = -I include -DHW_DEVKIT -DSMALL_FLASH_DEVICE -Wimplicit-fallthrough -Wextra -Wall
     -fno-exceptions -ffunction-sections -fdata-sections -Wl,--gc-sections
-build_unflags = -fno-lto
 ;build_cxxflags = -fno-rtti
 lib_deps = 
 
@@ -140,7 +139,6 @@ custom_component_remove =
     espressif/ksz8851snl
 build_flags = -I include -DHW_STARK -Wimplicit-fallthrough -Wextra -Wall
     -fno-exceptions -ffunction-sections -fdata-sections -Wl,--gc-sections
-build_unflags = -fno-lto
 ;build_cxxflags = -fno-rtti
 lib_deps = 
 
@@ -188,7 +186,6 @@ build_flags =
     -D ARDUINO_RUNNING_CORE=1       ; Arduino Runs On Core (setup, loop)
     -D ARDUINO_EVENT_RUNNING_CORE=1 ; Events Run On Core
     -fno-exceptions -ffunction-sections -fdata-sections -Wl,--gc-sections
-build_unflags = -fno-lto
 ;build_cxxflags = -fno-rtti
 lib_deps = 
 
@@ -233,7 +230,6 @@ build_flags =
     -D ARDUINO_RUNNING_CORE=1       ; Arduino Runs On Core (setup, loop)
     -D ARDUINO_EVENT_RUNNING_CORE=1 ; Events Run On Core
     -fno-exceptions -ffunction-sections -fdata-sections -Wl,--gc-sections
-build_unflags = -fno-lto
 ;build_cxxflags = -fno-rtti
 lib_deps =
 
@@ -286,7 +282,6 @@ build_flags =
     -ffunction-sections
     -fdata-sections
     -Wl,--gc-sections
-build_unflags = -fno-lto
 ;build_cxxflags = -fno-rtti
 lib_deps =
 
--- a/sdkconfig.be_size.defaults
+++ b/sdkconfig.be_size.defaults
@@ -88,7 +88,7 @@ CONFIG_MBEDTLS_TLS_CLIENT_ONLY=y
 # Remove assertion checks outright (not just the message strings). Larger flash
 # win than SILENT. Trade-off: a bug that would have cleanly asserted becomes
 # undefined behaviour instead. Switch back to _SILENT if you want the aborts.
-CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y
+# CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y
 # CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT is not set
 
 # -----------------------------------------------------------------------------


ngehrsitz:
I would suspect that it it something in either `custom_component_remove` or `custom_sdkconfig` because that is where we removed functionality.
The most suspicious option appears to be `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y` since the AI even says that it can incur undefined behavior.

dalathegreat:
This change fixed so BECom does not crash anymore after bootup with CAN-FD towards MEB :raised_hands: 

Unfortunately, T-2CAN still throws a communication lost after 120 seconds using CAN-FD :grimacing:

nagyrobi:
Fyi @ngehrsitz we have to remove this suggestion  “build_unflags = -fno-lto”

ngehrsitz:
Claude came up with a theory that sounds plausible, unfortunately I don´t have the hardware to investigate:
## 1. What PR #2571 changed

PR [#2571](https://github.com/dalathegreat/Battery-Emulator/pull/2571)
("Flash-size reduction by sdkconfig override for small-flash targets") applied a
batch of build-system changes across every hardware environment:

| Change | Detail |
| --- | --- |
| **LTO enabled** | `build_unflags = -fno-lto` on every env — i.e. link-time optimization is now **ON**. Added late (commit `11d866e`), lightly tested. |
| **pioarduino hybrid mode** | ESP-IDF static libs recompiled from source instead of using the precompiled Arduino libs. |
| **`custom_component_remove`** | Drops unused managed components, including SPI-adjacent MAC drivers: `w5500`, `dm9051`, `ksz8851snl`, `lan867x`. |
| **sdkconfig cuts** | `-Os`, IDF/bootloader log level `NONE`, WPA-Enterprise off, mbedTLS trims, **assertions disabled** (`CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE`). |

Assertions being disabled matters for diagnosis: a latent bug that used to `abort()`
cleanly now becomes undefined behaviour instead.

---

## 2. What "communication lost" actually is

"Communication lost" surfaces as the event **`EVENT_CAN_BATTERY_MISSING`**.

The liveness mechanism:

1. Battery CAN parsers refresh a countdown on every valid frame:
   `datalayer.battery.status.CAN_battery_still_alive = CAN_STILL_ALIVE;`
   where `CAN_STILL_ALIVE = 60` (`Software/src/devboard/utils/types.h:78`).
2. The safety loop decrements it once per second and raises the event at zero
   (`Software/src/devboard/safety/safety.cpp:296`):

   ```cpp
   // Check if the BMS is still sending CAN messages. If we go 60s without messages we raise an error
   if (!datalayer.battery.status.CAN_battery_still_alive) {
     set_event(EVENT_CAN_BATTERY_MISSING, can_config.battery);
   } else {
     datalayer.battery.status.CAN_battery_still_alive--;
     clear_event(EVENT_CAN_BATTERY_MISSING);
   }
   ```

   `update_machineryprotection()` runs on the 1-second cadence
   (`INTERVAL_1_S`) in `Software/Software.cpp:637,658`.

**Timing math:** the counter is 60 and drains at 1/second → the event fires
~60 s **after CAN-FD RX truly stops**. The observed ~120 s = frames flow for
~60 s after boot, RX then stalls, and a 60 s countdown follows → failure at ~120 s.

So the real fault is: **CAN-FD receive delivery stops after startup.**

---

## 3. Why RX stopping is plausible — the RX path is interrupt-driven

CAN-FD RX via the MCP2517/2518FD (`ACAN2517FD` library) on ESP32 is **fully
interrupt-driven, not polled**:

- Hardware INT → ISR lambda `[]{ canfd->isr(); }` registered in
  `Software/src/communication/can/comm_can.cpp` (`begin_canfd`, line ~283).
- `ACAN2517FD::isr()` (`ACAN2517FD.cpp:922`) only does:

  ```cpp
  void ACAN2517FD::isr (void) {
    BaseType_t xHigherPriorityTaskWoken = pdFALSE ;
    xSemaphoreGiveFromISR (mISRSemaphore, &xHigherPriorityTaskWoken) ;
    portYIELD_FROM_ISR () ;
  }
  ```

  **It is _not_ marked `IRAM_ATTR`.**
- A FreeRTOS task `myESP32Task` (`ACAN2517FD.cpp:51`) blocks on the semaphore and
  runs `isr_poll_core()`, which drains the chip FIFO into a software ring buffer
  `mDriverReceiveBuffer`.
- The application (`receive_frame_canfd_addon()` → `ACAN2517FD::available()` /
  `receive()`) only ever reads that software buffer.

**Consequence:** if the interrupt stops firing (or the task stops being woken),
the software buffer drains and no new frames are ever delivered — which is exactly
the observed symptom.

---

## 4. Board divergence: why T-2CAN and not BECom

The two boards wire the MCP2518FD interrupt(s) differently — the decisive difference:

| Board | Interrupt config | Source |
| --- | --- | --- |
| **BECom** | Single INT pin (`MCP2517_INT()` = GPIO10) | `hw_becom.h:23` |
| **T-2CAN** | Dual interrupt (`MCP2517_INT()` = NC; `INT0` = GPIO9, `INT1` = GPIO3) | `hw_lilygo2can.h:63-65` |

On the dual path, both pins attach to the **same ISR**
(`ACAN2517FD.cpp:588-589`). BECom being fixed while T-2CAN still fails is consistent
with the regression being sensitive to interrupt handling / timing rather than a
board-agnostic logic error.

---

## 5. Suspects, ranked

### Suspect 1 (prime): non-IRAM ISR + LTO

`ACAN2517FD::isr()` is not `IRAM_ATTR`. PR #2571 turned LTO on and switched to a
hybrid from-source IDF build. LTO reorders/inlines across translation units and
changes flash section placement — precisely what can flip a not-in-IRAM ISR from
"happens to work" to "faults or is silently missed while the flash cache is
briefly disabled" (e.g. during an NVS/SPIFFS write or wear-level flush). A single
missed RX interrupt at the wrong moment can then wedge RX permanently via Suspect 2.

### Suspect 2: buffer-full interrupt latch (permanent-stall amplifier)

`ACAN2517FD.cpp:1059-1067`: when the software ring fills, the driver disables the
RX interrupt in both software and hardware:

```cpp
//--- If mDriverReceiveBuffer is full, disable receive interrupt (added in release 2.17)
if (mDriverReceiveBuffer.isFull ()) {
  mRxInterruptEnabled = false ;
  if (mINT != 255 || mINT0 != 255 || mINT1 != 255) {
    uint8_t data8 = readRegister8Assume_SPI_transaction (INT_REGISTER + 2) ;
    data8 &= ~ (1 << 1) ; // Receive FIFO Interrupt disable
    writeRegister8Assume_SPI_transaction (INT_REGISTER + 2, data8) ;
  }
}
```

Re-arming happens **only** inside `receive()` (`ACAN2517FD.cpp:863-867`). If the
main-loop `receive_can()` pump ever races or stalls relative to a buffer-full
event, RX never re-arms and stays dead — turning a transient miss into a
permanent stall.

### Suspect 3 (low): `custom_component_remove` of SPI-adjacent components

Dropping `w5500` / `dm9051` / `ksz8851snl` / `lan867x` could, in the from-source
hybrid build, alter shared SPI/GPIO glue that the Arduino `SPIClass` /
`attachInterrupt` path relied on. Lower probability, same blast radius (SPI +
external interrupt).

### Suspect 4 (not a root cause, but a masker): assertions disabled

`CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE` means a latent driver bug that
previously asserted cleanly now runs into undefined behaviour, which could
*manifest* as "CAN-FD broken" without being the origin.

---

## 6. Recommended next steps

1. **Fastest bisect (hardware):** on the T-2CAN env, temporarily remove
   `build_unflags = -fno-lto` (LTO off) and re-test. If RX survives past 120 s,
   LTO is confirmed as the trigger.
2. **Durable fix (keeps flash savings):** mark the ISR path `IRAM_ATTR` and audit
   the ISR/task shared state (`mRxInterruptEnabled`, the ring buffer,
   `mHardwareTxFIFOFull`) for missing `volatile`/atomics. Note this repo already
   runs `-Werror=volatile`, so volatile-correctness gaps are exactly what LTO
   punishes.

dalathegreat:
After [6ba5440](https://github.com/dalathegreat/Battery-Emulator/pull/2650/commits/6ba54405159957e1b62e107c29fd1223ae57ac39) was in place, issues went away. I will merge this PR to make Main stable again, and we can re-visit the optimization once we can fix the FD library for S3

PR #66, opened 2023-10-05
Improvements to debug output for Tesla batteries.

matvey83:
I planned to make more improvements and also re-factor / clean up, but don't have the time right now.

PR #2652, opened 2026-07-19
Add IRAM_ATTR to MCP2518FD

ngehrsitz:
### What
This is one part of the suspected fix for the CAN FD issues from #2650. It purposefully does not fix the bug that a single missed interrupt can cause the 120s timeout. That should be fixed in a separate PR after we have verified that this fixes the root cause.

## Why

CAN-FD RX is **fully interrupt-driven**: a hardware INT wakes a FreeRTOS task via a semaphore, the task drains the chip FIFO into a software ring buffer, and the application polls that buffer. If the interrupt stops firing, the buffer drains and no frames arrive.

The registered handler chain was **not in IRAM**:

```
Arduino GPIO ISR dispatcher  →  []{ canfd->isr(); }  →  ACAN2517FD::isr()
```

When the flash cache is briefly disabled — which happens during **NVS writes and OTA** — a flash-resident ISR faults or is silently dropped. A single missed edge then trips the driver's buffer-full latch (`ACAN2517FD.cpp`), which disables the RX interrupt and only re-arms inside `receive()` — never reached once RX is wedged — so RX dies permanently. The battery liveness counter (`CAN_STILL_ALIVE = 60`, decremented 1/s in the safety loop) then drains and raises `EVENT_CAN_BATTERY_MISSING`, hence the ~120 s symptom.

This is a **pre-existing latent bug**, documented upstream in [pierremolinaro/acan2517FD#66](https://github.com/pierremolinaro/acan2517FD/issues/66) (same root cause, same recommended fix). It was exposed by the LTO + hybrid from-source build introduced in #2571: LTO changed code placement so the previously-lucky flash layout no longer held. **T-2CAN** is hit first because its dual-interrupt wiring (INT0 + INT1) registers the handler on two pins → ~2× the ISR pressure vs. single-INT boards like BECom.

## How

Three coordinated changes make the **entire** CAN-FD interrupt path cache-safe:

1. **Library handlers → IRAM.** `IRAM_ATTR` on `ACAN2517FD::isr()` and `poll()` (the two functions that run in true ISR context on ESP32). `isr_poll_core()` and the worker task are deliberately left in flash — they run in task context and do long SPI transfers.

2. **App trampolines → IRAM.** A raw lambda's `operator()` does not inherit `IRAM_ATTR`, so the `[]{ canfd->isr(); }` handlers are replaced with named `static void IRAM_ATTR` shims. (This is exactly the upstream #66 workaround.)

3. **Arduino GPIO dispatcher → IRAM.** `CONFIG_ARDUINO_ISR_IRAM=y` in `sdkconfig.be_size.defaults`. Without it, Arduino installs the shared GPIO ISR service with flag `0` and the dispatcher (`__onPinInterrupt`) itself runs from flash — so IRAM on our handlers alone would be insufficient. This is only actionable because #2571's hybrid mode recompiles the Arduino core from source.

### Safety of `CONFIG_ARDUINO_ISR_IRAM=y`

Enabling it forces **every** `attachInterrupt` handler on the shared GPIO service to be IRAM-safe. Audit of all interrupt users in the tree:

| User                         | Registration         | Shared GPIO service? | IRAM-safe?            |
|------------------------------|----------------------|----------------------|-----------------------|
| `ACAN2517FD` (CAN-FD)        | `attachInterrupt`    | Yes                  | ✅ made IRAM-safe here |
| `MCP2515_Lite` (classic CAN) | `attachInterruptArg` | Yes                  | ✅ already `IRAM_ATTR` |
| `ACAN_ESP32` (native TWAI)   | own `esp_intr_alloc` | No (separate source) | unaffected            |

No unaudited handler is affected.

### Hardware verification

Runtime behaviour requires physical hardware on a live CAN-FD bus and need to be verified @dalathegreat 

- [ ] Flash `lilygo_2CAN_330` to a T-2CAN with a CAN-FD battery; run **≥10 min** (pre-fix reliably dies at ~120 s). Pass = no `EVENT_CAN_BATTERY_MISSING`, steady RX, no `EVENT_CANMCP2518FD_INIT_FAILURE`.
- [ ] **Stress the flash-cache path** — trigger NVS/settings saves and an OTA update while CAN-FD traffic flows, and confirm RX survives. This is the exact condition (#66) that drops a non-IRAM ISR.
- [ ] Regression: flash `BECom_330`, confirm single-INT CAN-FD still runs ≥10 min.

### DIFF
--- a/sdkconfig.be_size.defaults
+++ b/sdkconfig.be_size.defaults
@@ -88,7 +88,7 @@ CONFIG_MBEDTLS_TLS_CLIENT_ONLY=y
 # Remove assertion checks outright (not just the message strings). Larger flash
 # win than SILENT. Trade-off: a bug that would have cleanly asserted becomes
 # undefined behaviour instead. Switch back to _SILENT if you want the aborts.
-# CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y
+CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y
 # CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT is not set
 
 # -----------------------------------------------------------------------------
--- a/sdkconfig.be_size.defaults
+++ b/sdkconfig.be_size.defaults
@@ -36,6 +36,14 @@ CONFIG_LOG_DEFAULT_LEVEL_NONE=y
 # No WPA2/WPA3-Enterprise (you use plain PSK). Drops the enterprise supplicant.
 # CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT is not set
 
+# Put the Arduino GPIO interrupt dispatcher in IRAM (installs the shared GPIO ISR
+# service with ESP_INTR_FLAG_IRAM and marks __onPinInterrupt IRAM_ATTR). This is a
+# CORRECTNESS fix, not a size cut: the CAN-FD add-on (ACAN2517FD via attachInterrupt)
+# must service its RX interrupt even while the flash cache is disabled, otherwise
+# edges are dropped and RX wedges. Only actionable because hybrid mode recompiles
+# the Arduino core from source. Requires all attachInterrupt handlers to be IRAM-safe
+CONFIG_ARDUINO_ISR_IRAM=y
+
 # Root-CA bundle: the FULL store is ~69 KB and this firmware does NO on-device
 # TLS verification (the GitHub release check runs in the browser). The feature
 # CANNOT be fully disabled: arduino-esp32's NetworkClientSecure library does an
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -59,6 +59,18 @@ static SPIClass* SPI2517_2;
 static ACAN2517FD* canfd_2 = nullptr;
 static ACAN2517FDSettings* settings2517_2;
 
+// IRAM_ATTR interrupt trampolines for the CAN-FD add-on(s). The registered
+// handler runs from the Arduino GPIO ISR dispatcher and must stay resident when
+// the flash cache is disabled (NVS writes, OTA), so it (and everything it calls)
+// lives in IRAM. A raw lambda's operator() would not inherit IRAM_ATTR, hence
+// these named shims. See pierremolinaro/acan2517FD#66.
+static void IRAM_ATTR canfd_isr_trampoline() {
+  canfd->isr();
+}
+static void IRAM_ATTR canfd_2_isr_trampoline() {
+  canfd_2->isr();
+}
+
 static bool native_can_initialized = false;
 //CAN logging filter settings
 uint16_t user_selected_CAN_ID_cutoff_filter = 0;  //Messages below this ID will not be logged in webserver
@@ -280,7 +292,7 @@ bool init_CAN() {
 }
 
 static bool begin_canfd() {
-  const uint32_t errorCode2517 = canfd->begin(*settings2517, [] { canfd->isr(); });
+  const uint32_t errorCode2517 = canfd->begin(*settings2517, canfd_isr_trampoline);
   canfd->poll();
   if (errorCode2517 != 0) {
     logging.print("CAN-FD Configuration error 0x");
@@ -294,7 +306,7 @@ static bool begin_canfd() {
 }
 
 static bool begin_canfd_2() {
-  const uint32_t errorCode2517_2 = canfd_2->begin(*settings2517_2, [] { canfd_2->isr(); });
+  const uint32_t errorCode2517_2 = canfd_2->begin(*settings2517_2, canfd_2_isr_trampoline);
   canfd_2->poll();
   if (errorCode2517_2 != 0) {
     logging.print("CAN-FD 2 Configuration error 0x");
--- a/Software/src/lib/pierremolinaro-ACAN2517FD/ACAN2517FD.cpp
+++ b/Software/src/lib/pierremolinaro-ACAN2517FD/ACAN2517FD.cpp
@@ -894,7 +894,7 @@ bool ACAN2517FD::dispatchReceivedMessage (const tFilterMatchCallBack inFilterMat
 //----------------------------------------------------------------------------------------------------------------------
 
 #ifdef ARDUINO_ARCH_ESP32
-  void ACAN2517FD::poll (void) {
+  void IRAM_ATTR ACAN2517FD::poll (void) {
     BaseType_t xHigherPriorityTaskWoken = pdFALSE ;
     xSemaphoreGiveFromISR (mISRSemaphore, &xHigherPriorityTaskWoken) ;
     portYIELD_FROM_ISR () ;
@@ -919,7 +919,7 @@ bool ACAN2517FD::dispatchReceivedMessage (const tFilterMatchCallBack inFilterMat
 //----------------------------------------------------------------------------------------------------------------------
 
 #ifdef ARDUINO_ARCH_ESP32
-  void ACAN2517FD::isr (void) {
+  void IRAM_ATTR ACAN2517FD::isr (void) {
     BaseType_t xHigherPriorityTaskWoken = pdFALSE ;
     xSemaphoreGiveFromISR (mISRSemaphore, &xHigherPriorityTaskWoken) ;
     portYIELD_FROM_ISR () ;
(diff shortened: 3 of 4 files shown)


nagyrobi:
Nice catch

sdkconfig.be_size.defaults
```diff
@@ -36,6 +36,17 @@ CONFIG_LOG_DEFAULT_LEVEL_NONE=y
...
 # CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT is not set
 
+# Put the Arduino GPIO interrupt dispatcher in IRAM (installs the shared GPIO ISR
+# service with ESP_INTR_FLAG_IRAM and marks __onPinInterrupt IRAM_ATTR). This is a
+# CORRECTNESS fix, not a size cut: the CAN-FD add-on (ACAN2517FD via attachInterrupt)
+# must service its RX interrupt even while the flash cache is disabled, otherwise
+# edges are dropped and RX wedges (observed as EVENT_CAN_BATTERY_MISSING ~120s on
+# T-2CAN). Only actionable because hybrid mode recompiles the Arduino core from
+# source. Requires every attachInterrupt handler to be IRAM-safe: the only two
+# users of the shared GPIO service are ACAN2517FD (made IRAM-safe alongside this)
```
jonny5532: I'm not sure listing the users of GPIO interrupts here is useful, since it'll get out of date if anyone adds any others (and clearly won't update this comment).

jonny5532: (specifically the "the only two users ..." bit, which GitHub's commenting doesn't make entirely clear)

ngehrsitz:
I rebased the branch, would be great if someone with the hardware could test it

sdkconfig.be_size.defaults
ngehrsitz: I dropped that part

jonny5532:
I've tried this on my non-FD 2CAN, it doesn't stop it crashing - it is still tripping a stack canary in the ACAN2517FD process.

I'll go through and selectively comment things until it starts working to try and work out what the cause is. It feels like it is crashing the stack, but I can't imagine that disabling assertions somehow makes code use *more* stack than it used to...

(It seems a sensible improvement anyway, but not sure it is the actual issue here...)

ekholm:
A lead on the stack canary @jonny5532 is hitting: the driver task that services the chip is created with

```cpp
xTaskCreate (myESP32Task, "ACAN2517Handler", 1024, this, TASK_ACAN2517FD_PRIORITY, &mESP32TaskHandle) ;
```

(ACAN2517FD.cpp, in `begin()`). On ESP-IDF — and therefore Arduino-ESP32 — `xTaskCreate` takes the stack size in **bytes**, unlike vanilla FreeRTOS where the same parameter is in words. So this task runs on a 1 KiB stack, not the 4 KiB that `1024` would mean on a standard FreeRTOS port. 1 KiB is very thin for a task that executes the whole `isr_poll_core()` path (SPI transaction begin/end with mutex, register reads, FIFO drain into the ring buffer), and the canary/watchpoint sits at the end of exactly that stack. A trip there is textbook stack overflow, and it would be load-dependent — deeper call nesting when FIFOs are full or error paths run — which fits "crashes under load" better than the interrupt placement did, and also fits it reproducing on the non-FD 2CAN.

Two cheap experiments: (1) log `uxTaskGetStackHighWaterMark(mESP32TaskHandle)` after some bus traffic — headroom near zero confirms it; (2) bump the create call to 4096 and re-run the failing case. Cost is ~3 KiB RAM per CAN channel, which is affordable. If that stabilizes it, the IRAM_ATTR change is still worth keeping for the flash-operation windows — but the stack size would be the actual crash.

ngehrsitz:
I bought an MCP2518FD board, let´s see if I can narrow down the root cause once I get it.

ekholm:
I ran this on hardware — the mechanism your PR targets reproduces, and the change removes it.

Setup: Stark CMR (ESP32/LX6) with an MCP2518FD add-on in internal loopback at ~470 self-TX/s plus a MEB battery's cyclic traffic, ~750 INT/s total. Flash-cache windows forced by repeated settings saves through the stock `/saveSettings` (one NVS write + commit each), `CONFIG_ARDUINO_ISR_IRAM=y`.

**Before (stock lambda chain, flag on): 10 saves → 9 Guru Meditations**, all `Cache disabled but cached memory region accessed`, board crash-looping. Panic PCs symbolized against the exact ELF: 8 of the 9 land *inside the `begin_canfd()` lambda itself*, the 9th in ACAN2517FD flash text.

**After (named `IRAM_ATTR` top-halves): 100/100 saves, zero panics**, and the ISR counter tracks RX 1:1 all the way through the storm — interrupts are being serviced during the flash windows, not deferred past them.

Worth noting what that implies: the sdkconfig flag on its own is actively harmful without the handler changes, because the dispatcher then runs with the cache disabled and jumps straight into flash-resident handler code. Your three changes are load-bearing as a unit.

Caveats so nobody over-reads this: it's a Stark, not the T-2CAN in your checklist; `ISR_IRAM=y` isn't the shipping default (with it off the failure mode is silent FIFO loss rather than a crash); and it says nothing about the stack canary @jonny5532 hit — I still think that's a separate defect, the 1 KiB `xTaskCreate` stack from my earlier comment.

One variation if it's useful: I got the same result by putting the IRAM top-halves in `comm_can.cpp` and waking the driver's task through its public `mISRSemaphore`, which leaves the vendored library untouched. Happy to share that.

ekholm:
I put a T-2CAN on the bench to run your checklist item 1, and hit something that changes the
picture for that board specifically. Measured, not read from the HAL:

| MCP2518FD output | actually reaches | `hw_lilygo2can.h` declares |
|---|---|---|
| **nINT** (dedicated) | **GPIO8** | `MCP2517_INT()` = `GPIO_NUM_NC` |
| nINT0 | GPIO9 | `MCP2517_INT0()` = GPIO9 |
| nINT1 | nothing on the scanned set | `MCP2517_INT1()` = GPIO3 |

LilyGO's own `pin_config.h` corroborates the first two rows — it declares `MCP2518_INT 8`
(the pin the HAL marks NC), and it is visibly where the HAL's `INT_0=9 / INT_1=3` pair came
from, minus the `INT 8` line above it. The third row is explained by the schematic
(`T-2Can-Fd_V1.0.pdf`): **INT1# reaches GPIO3 only through a DIP switch**, open as shipped on
both of our units — while nINT# is hard-wired to GPIO8 with a 10 K pull-up. So the stock
declaration's RX-interrupt path depends on an undocumented switch position, and the dedicated
nINT works on every board as shipped.

Method: put each chip output into a known state over SPI and watch every candidate GPIO as an
input — GPIO0/GPIO1 driven directly via `IOCON` in GPIO mode (pure continuity), the dedicated
nINT via the mode-change interrupt, with a negative control that disables the interrupt while
keeping the identical register writes. Reproduced on both of our FD boards, and the classic
T-2CAN's MCP2515 interrupt measures to GPIO8 as well, so GPIO8 is the CAN interrupt net on
that PCB in both build variants.

Per DS20005688B, nINT0 is the transmit interrupt and nINT1 the receive interrupt. So on a
T-2CAN the driver is attached to a pin carrying only TX, plus a pin that carries nothing —
**receive interrupts are never delivered at all**. What actually drains the RX FIFO is the
driver task's 500 ms semaphore timeout, and the arithmetic closes: the controller RX FIFO is
27 deep, 27 / 0.5 s = 54 frames/s predicted against 55 measured.

Two consequences for this PR:

**Your checklist item 1 cannot pass on a T-2CAN from this PR alone.** "Steady RX for ≥10 min"
needs an RX interrupt to exist; no amount of the handler chain being in IRAM creates one. On
our rig, receive-only traffic from a peer at ~425 frames/s gives:

- stock pin declaration: **55–56 frames/s**, ISR count **0**
- `MCP2517_INT()` = GPIO8, INT0/INT1 `NC`: **292 frames/s**, ISR count tracking RX 1:1

That is a one-file change in `hw_lilygo2can.h`, disjoint from everything this PR touches — the
two compose, and I think item 1 passes with both. The change is public and runnable:
https://github.com/ekholm/Battery-Emulator/commit/a5f8a2303308ce80b92fd5ae30baa1557bae3065 (and
the freed-GPIO3 revert: https://github.com/ekholm/Battery-Emulator/commit/3fd7e1367d5153e47d1d4d9fcc6dcc2da8fa4131).

**I also suspect it explains your "T-2CAN is hit first" observation better than the ~2× ISR
pressure reading.** It isn't double the interrupt pressure, it's none: with RX served only by
a 500 ms poll, that 27-deep FIFO is permanently overflowing under any real bus load, which
should trip the buffer-full latch you describe far more readily than on a board whose RX
interrupt works. It says nothing about the stack canary @jonny5532 hit on a **non-FD** 2CAN,
where this defect does not apply — I still think that one is the 1 KiB `xTaskCreate` stack.

One knock-on if that HAL change is wanted: GPIO3 becomes free again, so the BMS_POWER move to
GPIO45 can be reverted. The pin conflict that motivated it was real, but it was between
BMS_POWER and an INT1 declaration that no trace backs.

Happy to open the HAL fix as its own small PR, or to leave it and just hand over the numbers —
whichever is less work for you.

PR #2653, opened 2026-07-20
Remove checkprogsize rebuild

ngehrsitz:
### What
After #2571 the `checkprogsize` invocations are now unable to reuse the artifacts from the previous compile step, thus causing a unnecessary 2min rebuild.

### How
By capturing the size in the first bild we can avoid the rebuild.

> [!TIP]


dalathegreat (approved):

dalathegreat:
Nice, those 10min builds were brutal :sweat_smile:

PR #2656, opened 2026-07-20
Improvement: BYD - Change source pack voltage & current from CAN / add insulation resistance

demon1300:
### What
* Pack voltage from 0x438 (deci-volts) instead of polled DID 0x0008
* Pack current from 0x444 (deci-amps, offset 5000) instead of polled DID 0x0009
* Insulation resistance from 0x43A added to the battery info page
* Poll rotation drops from 22 to 20 DIDs to improve timings.

### Why
The BMS already broadcasts both values, faster and better:

* 0x438 gives 10x the voltage resolution of the whole-volt DID
* 0x444 arrives at ~101 ms vs the ~4.4 s poll cycle — this also gives the contactor open-interlock a fresh current sample instead of a stale one
* Insulation resistance was not being used (ideally would confirm with scan tool to ensure correct scaling).

### How
All three decodes are checksum-gated; the existing computeBydChecksum() applies unchanged to received frames.

Validated against vehicle captures before implementation.

* 0x438 tracks 0x444 within ±0.7 V
* 0x444 current sign confirmed on two charging logs, same encoding as the retired DID
Transmitted 0x441 link voltage deliberately left on 0x444, so the precharge handshake is untouched. New state is per-instance, so dual-battery setups are covered.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -424,6 +424,8 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
       if (rx_frame.data.u8[7] == computeBydChecksum(rx_frame.data.u8)) {
         battery_voltage_dV = (rx_frame.data.u8[6] << 8) | rx_frame.data.u8[5];
+      } else {
+        datalayer_battery->status.CAN_error_counter++;
       }
       break;
     case 0x43A:
@@ -431,6 +433,8 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       if (rx_frame.data.u8[7] == computeBydChecksum(rx_frame.data.u8)) {
         battery_insulation_ohm_per_volt = (rx_frame.data.u8[3] << 8) | rx_frame.data.u8[2];
         battery_insulation_valid = true;
+      } else {
+        datalayer_battery->status.CAN_error_counter++;
       }
       break;
     case 0x43B:
@@ -477,6 +481,8 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
         lastCurrentSampleMillis = millis();
         BMS_SOH = rx_frame.data.u8[4];
         BMS_voltage_available = true;
+      } else {
+        datalayer_battery->status.CAN_error_counter++;
       }
       break;
     case 0x445:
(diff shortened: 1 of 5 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -421,9 +422,16 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
+      if (rx_frame.data.u8[7] == computeBydChecksum(rx_frame.data.u8)) {
+        battery_voltage_dV = (rx_frame.data.u8[6] << 8) | rx_frame.data.u8[5];
+      }
       break;
     case 0x43A:
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
+      if (rx_frame.data.u8[7] == computeBydChecksum(rx_frame.data.u8)) {
+        battery_insulation_ohm_per_volt = (rx_frame.data.u8[3] << 8) | rx_frame.data.u8[2];
+        battery_insulation_valid = true;
+      }
```
dalathegreat, line 438: If the checksum does not match,  we should do as other integrations and increase the CAN error counter

```
case 0x0E9:
  datalayer.battery.status.CAN_battery_still_alive = CAN_STILL_ALIVE;
  if (rx_frame.data.u8[0] == expected_CRC(&rx_frame, 0x05)) {  //If CRC matches expected
    pack_voltage = (rx_frame.data.u8[6] << 8) | rx_frame.data.u8[7];
  } else {  //If CRC does not match expected, increment error counter
    datalayer.battery.status.CAN_error_counter++;
  }
  break;
```

demon1300: Good catch — done. I applied it to 0x43A and 0x444 as well since they use the same checksum gate. Used datalayer_battery->status.CAN_error_counter++ rather than datalayer.battery... so it attributes correctly on dual-battery setups.

dalathegreat (approved):
Nice improvement for BYD 👍

btoonen:
Thanks Demon and Dale,
Getting more and more professional every update. 
I get a scary low 14kohm while in use. What are values you are getting?  
Greetings Ben

demon1300:
> Thanks Demon and Dale, Getting more and more professional every update. I get a scary low 14kohm while in use. What are values you are getting? Greetings Ben

Hi Ben,
Yes, the pack will have an IR fault with anything less than 500 ohms per volt.  This will impact the discharge rate of the pack and the upcoming native charge termination & BMS SOC calibration (I have the battery behaving like it does in the car and will calbirate itself).  Stay tuned for future updates.

PR #2659, opened 2026-07-21
Fake battery: support triple setup, mirror pack voltage, fake balancing

nagyrobi:
### What
Add TestFake to the triple-battery switch in setup_battery() so battery3 is instantiated and MQTT publishes the third set of entities.

Batteries 2 and 3 mirror battery 1's user-set fake voltage, and the primary now defaults to 370.0V at boot instead of 0V.

Each cycle, any cell above the pack average gets its resistor flagged on. Since the voltages are re-randomised ±20 mV every pass, roughly half the cells light up and the set churns between cycles — good for exercising a dashboard, and it applies to all three instances since it's written through datalayer_battery.

### Why
Upon test-benching https://github.com/dalathegreat/Battery-Emulator/pull/2643, I noticed that enabling the triple option on the Fake battery just produces a log entry `[boot +0.124s] User tried enabling triple battery on non-supported integration!`.

Additionally I noticed the `Event: Too large voltage diff between the batteries. Second battery cannot join the DC-link`, when I tried to adjust the voltage in the settings. Fixing that so the voltage setting propagates correctly to the double and triple options.

Couldn't test to see behavior of the balancing data, all fields were `false` through MQTT.

> [!TIP]

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -431,6 +431,9 @@ void setup_battery() {
         battery3 = new RelionBattery(&datalayer.battery3, can_config.battery_triple,
                                      &datalayer.system.status.battery3_allowed_contactor_closing);
         break;
+      case BatteryType::TestFake:
+        battery3 = new TestFakeBattery(&datalayer.battery3, can_config.battery_triple);
+        break;
       default:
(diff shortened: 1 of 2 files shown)


nagyrobi:
image

dalathegreat (approved):

PR #2661, opened 2026-07-21
Web UI: only offer double/triple battery for integrations that support it

nagyrobi:
### What
This PR hides the "Double battery" and "Triple battery" checkboxes on the settings page for battery integrations that don't implement parallel batteries.
The double/triple contactor GPIO options and the Battery 2/3 interface selectors follow the same gating.
A previously stored double/triple setting is also cleared when the user switches to an integration that can't use it.

### Why
Currently every integration offers both checkboxes, even though only 17 support a second battery and 4 support a third.
Enabling them on an unsupported integration silently does nothing except print "User tried enabling double battery on non-supported integration!" to the debug log.
Worse, the setting stays in NVS after switching battery type, so an invisible enabled flag can linger and confuse later configuration.

### How
The capability lists are extracted out of the two switch statements in `setup_battery()` into `battery_supports_double()` / `battery_supports_triple()`, which now also gate object creation, so there is a single source of truth.
`settings_html.cpp` generates the `.if-dblcapable` / `.if-tricapable` CSS rules from those predicates at runtime, reusing the existing `form[data-battery=N]` + `display: contents` mechanism, so adding a new integration needs no UI edit.
`/saveSettings` clears `DBLBTR` / `TRIBTR` when the saved battery type doesn't support them, since a hidden checkbox still submits its stored state.

> [!TIP]

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -336,6 +336,47 @@ Battery* create_battery(BatteryType type) {
   }
 }
 
+// The integrations that can be instantiated a second time on a separate
+// interface. Must match the switch in setup_battery() below.
+bool battery_supports_double(BatteryType type) {
+  switch (type) {
+    case BatteryType::BoltAmpera:
+    case BatteryType::BydAtto3:
+    case BatteryType::NissanLeaf:
+    case BatteryType::BmwI3:
+    case BatteryType::CmfaEv:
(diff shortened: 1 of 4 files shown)


dalathegreat (approved):
Yes! I had been thinking about this earlier, thanks for making it a reality!

nagyrobi:
This came in the way while testing Fake with MQTT...

PR #2662, opened 2026-07-21
Leaf: Balancing status detection based on reported shunts

nagyrobi:
### What
This PR makes the Nissan LEAF report a meaningful balancing status: **Ready/Pending/Balancing**. The driver decodes the 96 balancing shunt bits from LBC poll group `0x06` into the per-cell array, but never rolled them up into the aggregate `balancing_status` field. It now classifies that field as `READY`, `ACTIVE` or `BLOCKED` from how the shunt set behaves over successive polls, renames the user-facing `BLOCKED` wording to "Pending" in both the MQTT string and the cell monitor legend. It leaves the raw per-cell array and `balancing_active_cells` count untouched for anyone who wants the instantaneous snapshot.

### Why
A populated shunt bitmap on its own does not mean the pack is balancing. On a 30 kWh pack the LBC keeps a small bleed engaged much of the time — at high, low and even mid SoC — and holds a static set for hours while it waits for the pack to settle, long before it does the vigorous top-of-charge balancing a user actually cares about. Reporting `Active` for that entire time makes the sensor useless for the one thing it is good for: knowing when a real balance is running and when it has finished. So the status has to separate genuine balancing from the BMS measurement period, using the only balancing telemetry the bus exposes.

### How
Analyzed data: [history_balancing_export.csv](https://github.com/user-attachments/files/30301123/history_balancing_export.csv) (exported from Home Assistant's SQLite database, where they were collected through MQTT cell_balancing json)

The design is constrained by what the BMS transmits. The balancing shunts are switched internally by the BMS - group `0x06` is a read-only window onto shunt on/off state, and there is no separate "measuring" versus "bleeding" flag anywhere in the protocol. LeafSpy, which reads far more of the bus than this driver, uses exactly these same bits and nothing richer. So "is this real balancing?" cannot be read from a single frame; it can only be inferred from how the shunt set behaves over time.

The distinguishing observation is *rate of change*. During genuine balancing the BMS duty-cycles the shunts - it bleeds one set of cells for a few minutes, then swaps to the next — so the bitmap keeps moving, though only every few polls rather than on every read. During the idle measurement phase the same set of flagged cells sits mostly unchanged for hours. The classifier is therefore edge-based with hysteresis, comparing each fresh group `0x06` read against the previous one:

| observation | status |
|---|---|
| no cells active | `READY` |
| bitmap changed since the last read | `ACTIVE` |
| bitmap unchanged for `BALANCING_STATIC_POLLS_FOR_BLOCKED` reads | `BLOCKED` |
| first read of a new phase | `BLOCKED` until a change confirms `ACTIVE` |

Any change flips the status to `ACTIVE` and it *stays* there across the static gaps between swaps; the pack is only demoted to `BLOCKED` after the configured number of consecutive no-change reads. That threshold (default 6, about seven minutes at the ~70 s group `0x06` cadence) is the single tuning knob and must be set longer than the LBC's gap between swaps — measured at a few minutes on the test pack. This replaces an earlier "all N reads in a sliding window must be distinct" rule, which demanded a change on nearly every poll and so flapped between `ACTIVE` and `BLOCKED` throughout a real balance, since the bitmap genuinely holds still for a poll or two between swaps.

`ACTIVE` is treated as the balancing edge for the event log, so `EVENT_BALANCING_START` / `EVENT_BALANCING_END` fire only on entry to and exit from real balancing, not when the pack merely stops being flagged. The stored bitmap and the static-poll counter are cleared on a zero-cell read (→ `READY`) and whenever a BMS reset is in progress (→ `UNKNOWN`), since neither the classification nor the history survives an LBC power cycle; this covers both the 24 h periodic reset and the MQTT-triggered one, as both drive `bms_reset_status` out of `BMS_RESET_IDLE`. The whole thing costs one stored bitmap and two counters — a few bytes of RAM.

**Observed on:** Two 2017 30 kWh (AZE0) packs. The 24 / 40 / 62 kWh generations are untested. If a generation's LBC holds each cell set longer than ~7 minutes between swaps, raise `BALANCING_STATIC_POLLS_FOR_BLOCKED`; if it swaps faster, it can be lowered. The `balancing_data` MQTT topic exposes the raw bitmap, so the swap interval can be measured directly from one logged balancing episode before changing the constant.

image
image
image

> [!TIP]


nagyrobi:
image
image


image
image

This will be beneficial for BMW-PHEV and BMW-IX users who also use this flag.

nagyrobi:
Further data analyzed in https://github.com/dalathegreat/Battery-Emulator/pull/2662/commits/131087b9d6ce9cf09745da8df06c109656342f4c: [export_balancing_data_aze0_v2.csv](https://github.com/user-attachments/files/30419383/export_balancing_data_aze0_v2.csv)

nagyrobi:
And even a bigger recording of cell behavior in https://github.com/dalathegreat/Battery-Emulator/pull/2662/commits/3718f2ac34e2a5c32f0514f0581bdcb5be2826ef: [export_balancing_data_aze0_v3.zip](https://github.com/user-attachments/files/30631641/export_balancing_data_aze0_v3.zip)
232 hours - 12,819 reads.

dalathegreat (approved):

PR #2666, opened 2026-07-22
Improvement: BYD-Modbus - report STANDBY until DC bus is live

nagyrobi:
### What:
Add datalayer.system.status.dc_bus_live as the single source of truth for "DC actually energized to the inverter", set by each contactor topology (GPIO state machine, BMW-I3 self-managed switch, BMW-SBOX). BYD-Modbus now remaps register 300 from ACTIVE to STANDBY while the bus is not yet live, and gates the voltage register (308) on the reported status. FAULT,
UPDATING and STANDBY pass through unchanged.

### Why:
Fronius GEN24 event log is filled with error message `GEN24-1187 No bat­tery voltage meas­ured but bat­tery is act­ive` each time Battery Emulator is being rebooted. This causes restarting the inverter, stopping PV production, and entering a long "Starting" procedure of several minutes (as required by the grid operator).

image

BydModbusInverter doesn't override controls_contactor(), so the emulator closes contactors on its own schedule (10s boot gate + precharge) with no inverter handshake. Meanwhile system_status defaults to ACTIVE and is served from boot, so a Fronius polling during that window sees an ACTIVE battery against a dead DC input and complains. STANDBY is the protocol-native "present but not delivering" state that resolves the mismatch.

### How:
dc_bus_live defaults true, so direct-wired setups keep today's behaviour; only topologies that actually gate the bus pull it low. The GPIO path sets it from contactorStatus == COMPLETED, BMW-I3 from its own contactors_engaged, BMW-SBOX from shunt.contactors_engaged. BYD-Modbus is the only consumer; the field is designed to generalise to other inverters later. No timing constants changed; reg 300 still first written on the 1s update loop.

BMW-I3 and BMW-SBOX need their own edits precisely because they're the exceptions that bypass handle_contactors(), everyone else uses the state machine in comm_contactorcontrol.cpp.

> [!TIP]

### DIFF
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -55,6 +55,7 @@ void BmwI3Battery::update_values() {  //This function maps all the values fetche
     datalayer.system.status.system_status = STANDBY;
     // During balancing sleep, report contactors as open so an old engaged state is not latched.
     datalayer.system.status.contactors_engaged = 0;
+    datalayer.system.status.dc_bus_live = false;
   }
 
   // Map internal balancing state to datalayer balancing_status
@@ -141,6 +142,8 @@ void BmwI3Battery::update_values() {  //This function maps all the values fetche
       datalayer.system.status.contactors_engaged = 0;
       break;
   }
+  // I3 drives its own DC switch, so DC is live once its contactors report engaged.
+  datalayer.system.status.dc_bus_live = (datalayer.system.status.contactors_engaged == 1);
 }
 
 void BmwI3Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
--- a/Software/src/battery/BMW-I3-BATTERY.cpp
+++ b/Software/src/battery/BMW-I3-BATTERY.cpp
@@ -1,6 +1,7 @@
 #include "BMW-I3-BATTERY.h"
 #include <Arduino.h>
 #include "../communication/can/comm_can.h"
+#include "../communication/contactorcontrol/comm_contactorcontrol.h"
 #include "../datalayer/datalayer.h"
 #include "../datalayer/datalayer_extended.h"
 #include "../devboard/utils/common_functions.h"  //For CRC table
@@ -55,7 +56,9 @@ void BmwI3Battery::update_values() {  //This function maps all the values fetche
     datalayer.system.status.system_status = STANDBY;
     // During balancing sleep, report contactors as open so an old engaged state is not latched.
     datalayer.system.status.contactors_engaged = 0;
-    datalayer.system.status.dc_bus_live = false;
+    if (!contactor_control_enabled) {
+      datalayer.system.status.dc_bus_live = false;
+    }
   }
 
   // Map internal balancing state to datalayer balancing_status
@@ -143,7 +146,10 @@ void BmwI3Battery::update_values() {  //This function maps all the values fetche
       break;
   }
   // I3 drives its own DC switch, so DC is live once its contactors report engaged.
-  datalayer.system.status.dc_bus_live = (datalayer.system.status.contactors_engaged == 1);
+  // Guarded so the GPIO contactor state machine stays authoritative when enabled.
+  if (!contactor_control_enabled) {
+    datalayer.system.status.dc_bus_live = (datalayer.system.status.contactors_engaged == 1);
+  }
 }
 
 void BmwI3Battery::handle_incoming_can_frame(CAN_frame rx_frame) {
--- a/Software/src/battery/BMW-IX-BATTERY.cpp
+++ b/Software/src/battery/BMW-IX-BATTERY.cpp
@@ -840,6 +840,10 @@ void BmwIXBattery::transmit_can(unsigned long currentMillis) {
     if (!contactor_control_enabled) {
       HandleBmwIxCloseContactorsRequest(counter_10ms);
       HandleBmwIxOpenContactorsRequest(counter_10ms);
+      // CAN mode: ContactorState tracks completion of the close/open command sequences.
+      // Note this is command-completion, not pack feedback - if the pack ignored the sequence
+      // we report live anyway, which matches pre-dc_bus_live behaviour (never STANDBY-locks).
+      datalayer.system.status.dc_bus_live = (ContactorState.closed && !ContactorState.open);
     }
     counter_10ms++;
 
--- a/Software/src/battery/BMW-SBOX.cpp
+++ b/Software/src/battery/BMW-SBOX.cpp
@@ -95,12 +95,14 @@ void BmwSbox::transmit_can(unsigned long currentMillis) {
 
     if (contactorStatus == SHUTDOWN_REQUESTED) {
       datalayer.shunt.contactors_engaged = false;
+      datalayer.system.status.dc_bus_live = false;
       return;  // A fault scenario latches the contactor control. It is not possible to recover without a powercycle (and investigation why fault occured)
     }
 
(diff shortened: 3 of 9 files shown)


jonny5532:
When you say `everyone else uses the state machine in comm_contactorcontrol.cpp`, what about all the battery integrations that control the contactors themselves via CAN? (such as MG, Atto3, CMP, etc)

Presumably they'll need to update this flag as they detect the contactors closing/opening. If they don't, presumably it'll break Fronius for these users?

nagyrobi:
#### BE Rebooting:
image

#### BE After boot, before contactors closed:
image

#### BE After contactors closed:
image


No more restarting the inverter and stopping PV!! And no more spam in the inverter logs.

nagyrobi:
> Presumably they'll need to update this flag as they detect the contactors closing/opening. If they don't, presumably it'll break Fronius for these users?

With the default of `true`, they don't break - the patch is simply a no-op for them. If an MG/Atto3/CMP integration never touches `dc_bus_live`, it stays true, and BYD-Modbus reports `system_status` straight through exactly as it does today.

But you're right, we might want to extend this functionality for them too

nagyrobi:
OK, so here's how it goes:

The guarantee is structural, not hopeful, and rests on three facts:

**1. The flag defaults `true`, and only an explicit writer can make it `false`.** For any battery that never touches `dc_bus_live`, the field sits at `true` for the lifetime of the process. There is no timer, no timeout, no inference anywhere that lowers it on their behalf.

**2. There is exactly one reader, and it can only act when the flag is `false`.** The BYD-Modbus remap is `if (reported_status == ACTIVE && !dc_bus_live) → STANDBY`. With the flag pinned `true`, the condition is unreachable, and register 300/308 handling reduces to byte-for-byte the pre-patch code. Every other inverter protocol (SMA, Solax, Pylon, Kostal, BYD-CAN, …) never reads the flag at all, so for them the patch is inert regardless of battery. It can be decided later if this could be useful for them.

**3. The one writer that runs regardless of battery type is correct for every battery.** The GPIO state machine writes the flag only when `contactor_control_enabled` — and in that wiring, the emulator's external contactors *are* the physical gate between any pack and the inverter, so "COMPLETED = DC live" holds no matter whose cells sit behind them. When GPIO control is disabled, that writer never executes. So an un-instrumented, self-managing battery in CAN mode has *zero* writers → flag stays `true` → today's behavior exactly.

So batteries with no self-managed contactors get the fix via the GPIO path; instrumented self-managers get it via their own signal; un-instrumented self-managers are **unchanged** - they keep the original boot-window "active but no DC" quirk (not fixed), but nothing regresses, and no inverter interaction changes.

#### Battery integrations that don't change:

The deciding principle is risk asymmetry. A wrongly-`true` flag merely reproduces today's behavior — the transient warning. A wrongly-`false` flag reports STANDBY forever, so the Fronius **never engages** and the installation is dead until a firmware fix. Skipping is free; guessing wrong bricks someone's system. So the bar for wiring a battery was: *an in-file, documented or operationally-trusted signal that means "pack HV output is connected" — not permission, not intent, not an undocumented raw value.* Each skip fails that bar for a specific, verifiable reason:

- **CMP Smart, eCMP, Rivian, FoxESS, Ford Mach-E, Tesla Legacy** — a contactor state *is* decoded, but as raw values with no value↔meaning mapping anywhere in the tree; today they're display-only telemetry. Tesla Legacy is the instructive case: it looks like modern Tesla, but its 4-bit field has no in-file enum and legacy DBCs are known to differ, so `== 4` would be a guess. Any of these can be promoted later by someone with the pack who can confirm the closed value on real hardware — each is then a one-line change following the established pattern.
- **BMW PHEV** — has a state signal (0x53A), but its own comments say bench testing showed it lags reality ("contactors close even while 0x53A still…"). A documented-unreliable signal is worse than none.
- **Geely Geometry C** — contactor closing doesn't work at all yet per the file's TODO; nothing to report on.
- **Nissan LEAF, MG-HS** — the available signal is *permission*, not closure. LEAF's `battery_MainRelayOn_flag` is literally commented "No-Permission=0, Main Relay On Permission=1" in the header; MG-HS only manages `battery_allows_contactor_closing`. Permission precedes closure, so wiring it would claim DC-live early — reintroducing the very race this patch fixes. (Standard Fronius+LEAF wiring uses GPIO contactors and is covered by the state machine anyway.)
- **Everyone else** (Kia/Hyundai, Renaults, Volvo, LV BMSes, …) — no self-managed pack contactors toward the inverter; the physical DC gate in those installs is the emulator's GPIO contactors or fixed wiring, both already handled.

Software/src/battery/BMW-I3-BATTERY.cpp
```diff
@@ -55,6 +56,9 @@ void BmwI3Battery::update_values() {  //This function maps all the values fetche
     datalayer.system.status.system_status = STANDBY;
     // During balancing sleep, report contactors as open so an old engaged state is not latched.
     datalayer.system.status.contactors_engaged = 0;
+    if (!contactor_control_enabled) {
+      datalayer.system.status.dc_bus_live = false;
```
dalathegreat, line 60: Question, there are some users that use permanently on contactors, (Early LEAF builds before we had GPIO control). What happens on these, when contactor_control_enabled is false, I assume the dc_bus_live gets forced to false, and Fronius will never get the state?

So only fix on these setups would be to enable GPIO contactor control (even though it is not wired)

nagyrobi: Well I guess that kind of manual control should definitely be deprecated, (also because it exposes safety risks). Contactor control works super-reliable, it's something that must be a requirement now that we're mainstream.

It is probably nostalgic to think of the good old times of manual control, but I believe strongly that there is a time to let things go. It's a part of the Wiki that I would't even migrate to the new site.

Various types of SSRs are available, there are even more choices to do the move.

Another reason to bump the major release number. Downgrade is possible, so anyone who made the upgrade and runs into this, can revert to the earlier version, until they install the 3 or 4 SSRs. 

Otherwise just enabling contactor control checkbox doesn't enforce any feedback about their physical existence so I think it can be used as a trick to let the system think they are there...

malec1984:
What about using the system_status properly? Maybe that should be set to STANDBY when battery contactors are open.
Or what could be the meaning of system_status is STANDBY?

nagyrobi:
> what could be the meaning of system_status is STANDBY?

That depends on the inverter, does it have/need a STANDBY status at all? I only own GEN24, and since all this knowledge was gained by reverse engineering, so far it turns out by experimenting that STANDBY and UPDATING have to be sent during boot and reboot/ota towards the inverter to avoid it restarting PV production, re-syncing to grid and logging these events with errors. 

Pretty sure other inverters behave differently, they may support fewer or more statuses. Whoever owns such inverters may observe unintended behaviours and suggest corrections, what we know so far that they likely work as they are now, and if there are no complaints, probably introducing an unexpected STANDBY there may break things.

This PR targets BYD-Modbus (Fronius) only, and politely asks the batteries to hand over this information. Other inverters may benefit from this in other ways (maybe not through system_status).

dalathegreat:
I will add this note when doing a release;

NOTE :warning: New hardware requirement for Fronius :warning: This release changes when battery voltage is reported towards Fronius inverters. It will only update voltage when contactors are engaged. **This means that old legacy installls using manual A/B/C switches for turning on battery contactors will no longer function with Fronius inverters.** Only GPIO controlled contactors and CAN controlled contactors will work. This is a new stricter safety requirement to get the Fronius inverter to startup faster and with less errors. The bonus is that GPIO controlled contactors is inherently safer than manual A/B/C triggering.

dalathegreat (approved):

PR #2668, opened 2026-07-23
Improvement: Add guards to periodic BMS reset

nagyrobi:
### What
Rename the setting to 'Periodic BMS reset:' and add a configurable interval (24h default, 48h) plus two optional guards, shown in the web UI only when the periodic reset is enabled.

Low SOC defers: while either the real or the scaled SOC is below 15%, the reset stays due and runs as soon as SOC recovers, with the following period anchored to that point. Balancing skips a single period instead, so a pack that reports balancing more or less permanently cannot suppress the reset indefinitely. The log names whichever condition applied. Remote MQTT resets are unaffected.

image

Installations upgrading from a build that only stored the enable flag default to the previous 24h interval.

### Why
Two problems to fix: 
- A BMS power cycle at low SOC may return with a different, lower SOC than it was before reset:
image

If it's too low, it may trigger the inverter to forcefully start charging from grid if this happens over night. We defer the reset until the battery charges back to 15% and perfom the reset then. And anchor following period to that point, likely next day won't run into the same situation.

- Resetting mid-balance can be unhelpful: the BMS loses an in-progress balancing run, interrupting the balancing procedure and forcing the BMS to re-start the several hours long cell measurements (AZE0):
image

Some setups may turn out that cell measurements + balancing procedures require more than 24 hours to complete, resetting as often as once a day may end up in never finishing the balance properly. However we only allow one skip - if the balancing ACTIVE status is longer than a day (for batteries that report this) that is likely stuck. Extending the period increases the chances for a graceful finish.

image

Why do we need the setting? It's observed that for example AZE0 likes to finish balancing. If BMS rebooted during balance, it won't restart. ZE1 can overbalance, so it may be wise to intrerrupt it daily. The user has to have a choice.

Without the checkboxes enabled, it will behave as before.

### How
The periodic branch of `handle_BMSpower()` consults the guards before calling `start_bms_reset()`; a guarded occurrence restarts the interval so it is retried a full period later, and the log names the condition. Remote MQTT resets are unaffected. Three new NVS keys are added, with `PERBMSRESETH` falling back to 24h so existing installs keep their current behaviour. Five unit tests cover the interval and both guards.

> [!TIP]


nagyrobi:
Real world test on Software: 11.2.dev Hardware: LilyGo T-CAN485 

- Balancing active at the 24h mark, so it skips it BMS reset
- Balancing active at the 48h mark, performs the BMS reset (as it's only allwed to skip one period)
- No balancing at the 72h mark, but SOC under 15%, BMS reset deferred
- A few minutes after the 72h mark, SOC went above 15%, BMS reset performed


```
   86400.006 BMS reset: Periodic reset skipped for one period, balancing active on battery.
  172800.006 Battery pause begin 1 0 0 0
  172800.006 Event: The emulator is trying to pause the battery.
  172833.018 Battery pause begin 0 0 0 0
  172833.018 Event: The emulator is attempting to resume battery operation from pause.
  172833.018 Event: BMS reset event completed.
  259200.016 BMS reset: Periodic reset deferred, real SOC below 15 percent.
  259430.752 BMS reset: Running previously deferred periodic reset.
  259430.752 Battery pause begin 1 0 0 0
  259430.752 Event: The emulator is trying to pause the battery.
  259463.768 Battery pause begin 0 0 0 0
  259463.768 Event: The emulator is attempting to resume battery operation from pause.
  259463.768 Event: BMS reset event completed.
```

dalathegreat (approved):

Issue #1840, opened 2025-12-15
Double battery more battery info incorrect


MartijnAir:
With a double battery setup on the „more battery info” page the info shown is incorrect. The info of 1 battery will be shown twice. 

Double battery setup and „more battery info” page will show this anytime

Expected fix is; both batteries will show independant information 

Version 9.3.0
Battery used Nissan Leaf 24kw 2x
BYD Battery-Box Premium HVS over CAN Bus GOODWE
Stark CMR
GPIO controlled contactors; yes
MQTT; no

MartijnAir:
<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/30cef143-60b9-4e6e-942e-ec3e1ef2bc34" />
<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/fd0e0357-123e-4f55-8ef1-5bf78244b73e" />

fastbike:
I have also observed this with double battery setup of 40kWh Leaf batteries

dalathegreat:
Fix coming to v12.0.0, main now fixed!

Issue #899, opened 2025-02-18
Double Battery - More Battery Info


malcolmputer:
I'm running a double battery setup with two 30kWh Leaf packs.  When I select the "More Battery Info" page I only get more information on one pack.  In addition to missing on the information, this prevents the clicking of the reset SoH for the 2nd battery as well.

dalathegreat:
Fix coming to v12.0.0, main now fixed!

PR #2670, opened 2026-07-23
Bugfix: Leaf per-battery data on advanced page for double/triple setups

nagyrobi:
### What
This PR fixes issues https://github.com/dalathegreat/Battery-Emulator/issues/1840 and https://github.com/dalathegreat/Battery-Emulator/issues/899, the "More battery info" page for double/triple Nissan Leaf setups: battery 2 previously displayed battery 1's values, and battery 3 showed no data at all. Each Leaf now shows its own real data (Insulation, GIDS, HX, temperatures, serial number, etc.).

### Why
The Leaf HTML renderer read the hardcoded global `datalayer_extended.nissanleaf`, and batteries 2/3 were constructed with a `nullptr` extended pointer, so their driver writes were silently discarded. This made the page misleading for multi-battery users and left no per-battery Insulation reading.

### How
Adds `nissanleaf_2`/`nissanleaf_3` structs to the `DataLayerExtended` union (zero RAM growth — 3×72 B fits under the 752 B Tesla member), passes them in `BATTERIES.cpp`, and constructs `NissanLeafHtmlRenderer` with the per-instance pointer. A new `renders_own_battery_data()` flag on `BatteryHtmlRenderer` gates battery 2/3 rendering: reworked renderers (Leaf, BMW-i3, Bolt, BYD Atto 3, KIA-64) show real data, legacy ones show the "limited to Main Battery" notice instead of wrong values. Future integrations follow the same three steps and flip the flag.

> [!TIP]

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -407,7 +407,8 @@ void setup_battery() {
           battery2 = new BydAttoBattery(&datalayer.battery2, &datalayer_extended.bydAtto3_2, can_config.battery_double);
           break;
         case BatteryType::NissanLeaf:
-          battery2 = new NissanLeafBattery(&datalayer.battery2, nullptr, can_config.battery_double);
+          battery2 = new NissanLeafBattery(&datalayer.battery2, &datalayer_extended.nissanleaf_2,
+                                           can_config.battery_double);
           break;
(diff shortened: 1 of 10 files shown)


dalathegreat (approved):
Massive thanks for fixing this long-standing bug! 🙌

PR #2675, opened 2026-07-24
Sungrow: SBR cleanup and updates

aferrara:
### What
This PR updates the Sungrow SBRXXX battery emulator to the **`.30` BMS firmware** frame set, corrects mislabelled and undecoded frame comments, and moves each frame's constant/config bytes into the header initialisers so `update_values()` carries only live telemetry.

### Why
Real Sungrow SBR batteries have moved to BMS firmware `.30` (`SBRBCU-S_22011.01.30`), which changes the CAN frame set: `0x400` (and the irregular `0x1F0` firmware-update channel) stop, and a static `0x402` appears in their place. An inverter that has taken the firmware update expects the `.30` frames, so the emulator has to match. While in the file, several comments were vague ("Magic number") or actively wrong — the `0x706`/`0x714` min/max cell-voltage labels were reversed — and many constant/config bytes were rewritten every cycle, obscuring which bytes are actually dynamic.

### How
- **`.30` frame set:** the `0x707` firmware-patch byte is now `0x30`; a static `0x402` (`00 00 55 …`) replaces `0x400`, which is removed entirely. SoC still reaches the inverter via `0x500`/`0x702`. `0x1F0` is documented only (not emulated).
- **Comment / label fixes:** vague comments are replaced with the decoded field meaning, and the reversed min/max labels in `0x706`/`0x714` are corrected. `0x714` is genuinely max-first (bytes 2–3 > bytes 6–7), which the emulator already emitted correctly — only the comments were wrong. Confirmed against a real `.30` SBR capture.
- **Frame-data reorganisation:** constant, model-derived and static-but-unknown bytes are moved from `update_values()` into the header `.data{}` initialisers (`0x500`, `0x705`, `0x707`, `0x70F`, `0x713`/`0x714`, `0x719`); `update_values()` now writes only per-cycle telemetry. Payloads are byte-for-byte unchanged (transmit-identical, verified on-bus against a real `.30` battery). A short convention note in `setup()` records where each frame's bytes come from (header / `setup()` / `update_values()` / Modbus reply).

> [!TIP]

### DIFF
--- a/Software/src/inverter/SUNGROW-CAN.cpp
+++ b/Software/src/inverter/SUNGROW-CAN.cpp
@@ -85,6 +85,17 @@ bool SungrowInverter::setup() {
   SUNGROW_707.data.u8[5] = (battery_config.nameplate_wh >> 8);      // Nameplate capacity (high)
   SUNGROW_707.data.u8[6] = battery_config.module_count;             // Num of modules
 
+  // 0x70F mux 0 b4-7 - max continuous power (W) = nominal voltage x 30 A. Per the SBR datasheet each
+  //                    module is 64 V nominal and the pack is 30 A continuous, so 1920 W per module
+  //                    (SBR096 = 5760, SBR160 = 9600, SBR224 = 13440). On a real battery this is a
+  //                    static rating - invariant across SoC and season - so it belongs here, not in
+  //                    update_values(). The live limits are reported separately in 0x701.
+  uint16_t max_power_w = 1920 * battery_config.module_count;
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):
Nice continued reverse engineering on this protocol 👍 Approved, shall we merge right away?

aferrara:
It's in good shape to merge straight away. There's still some remaining signals that I haven't figured out yet, but they are rapidly shrinking. I'll open a new PR as I learn more.

PR #2677, opened 2026-07-26
Fix inverter CAN alive-event handling in update_machineryprotection

ekholm:
Two related fixes to the CAN aliveness logic in `safety.cpp`, plus regression tests.

### 1. `EVENT_CAN_INVERTER_MISSING` never cleared with the long CAN timeout option

When the "long inverter CAN timeout" option was added (e5aa5b94), the original decrement + `clear_event()` pair was moved into the normal-timeout `else` branch, and the new long-timeout branch re-implemented only the decrement — so with the option **enabled**, the missing-inverter event is never cleared once raised.

Since the event is ERROR level, this is severe: `system_status` goes to FAULT, power limits are zeroed every cycle, and after 10 s of FAULT the contactor control latches `SHUTDOWN_REQUESTED` — so a single inverter CAN dropout longer than the timeout permanently stops the system until power cycle, **even after the inverter recovers**. This inverts the intent of the option, which exists to be *more* tolerant of slow-starting inverters.

**Fix:** hoist `clear_event(EVENT_CAN_INVERTER_MISSING)` above the timeout-mode branch, mirroring the battery path a few lines up. The half-rate decrement behavior of the long-timeout mode is unchanged (covered by a test).

### 2. Detected-events never fire for drivers that refresh the counter above `CAN_STILL_ALIVE`

The "component detected" checks compare with `== CAN_STILL_ALIVE`, but some inverter drivers deliberately refresh the counter above that value to get a longer timeout: the SMA family uses `CAN_STILL_ALIVE * 3`, Sofar uses `* 2`. (Those drivers predate the detection events: bdc8c986 vs 35314c7d.)

For those inverters:
- `EVENT_CAN_INVERTER_DETECTED` never fires in normal operation, and
- it instead fires **mid-outage**, when the decrementing counter happens to pass exactly 60 — logging "Successfully communicating with inverter!" while the inverter has been silent for two minutes.

**Fix:** compare with `>=` instead of `==` at all five detection sites (battery 1/2/3, inverter, charger). This is correct in all paths: detection latches on the first refresh, and the boot value is deliberately `CAN_STILL_ALIVE - 1` so a never-seen component stays undetected. Battery and charger drivers have no multipliers today, so those three changes are prophylactic — they protect against the same drift if a battery driver ever gets a longer timeout the same way.

### Tests

`test/inverter_alive_tests.cpp` (new, registered in the test CMakeLists):

- `DetectionFiresWithMultipliedRefreshValue` — fails without fix 2
- `MissingEventClearsOnRecoveryWithLongTimeout` — fails without fix 1
- `MissingEventClearsOnRecoveryWithNormalTimeout` — control, passes before and after
- `LongTimeoutStillDecrementsAtHalfRate` — guards the unchanged half-rate decrement

Note: the detection test must run before the others in the suite because `inverter_detected` is a file-static latch with no reset (documented in the test file; default definition order handles it).

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -286,7 +286,7 @@ void update_machineryprotection() {
 
     //Check if we have ever seen the Battery
     if (!battery_detected) {
-      if (datalayer.battery.status.CAN_battery_still_alive == CAN_STILL_ALIVE) {
+      if (datalayer.battery.status.CAN_battery_still_alive >= CAN_STILL_ALIVE) {
         battery_detected = true;
         set_event(EVENT_CAN_BATTERY_DETECTED, 1);
       }
@@ -312,7 +312,9 @@ void update_machineryprotection() {
 
     //Check if we have ever seen the inverter
     if (!inverter_detected) {
-      if (datalayer.system.status.CAN_inverter_still_alive == CAN_STILL_ALIVE) {
+      // >= not ==: some drivers refresh the counter above CAN_STILL_ALIVE for a
+      // longer timeout (SMA x3, Sofar x2), which would never match equality
+      if (datalayer.system.status.CAN_inverter_still_alive >= CAN_STILL_ALIVE) {
         inverter_detected = true;
         set_event(EVENT_CAN_INVERTER_DETECTED, 1);
       }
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Thanks for the contribution 🙌

PR #2679, opened 2026-07-26
Dyness Pylon: Corrected calculation of cell votages

tekampe-oss:
### What
This PR corrects the cell voltage reading for dyness batteries

### Why
Current cell voltage is wrong.

### How
Calculating correctly based on a comparison with app data.

### DIFF
--- a/Software/src/battery/PYLON-BATTERY.cpp
+++ b/Software/src/battery/PYLON-BATTERY.cpp
@@ -188,21 +188,19 @@ void PylonBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       // Handle EMUS individual cell voltage
       uint8_t cell_start_voltage = rx_frame.data.u8[0] - 1;
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;  // Keep battery alive on cell data
-
+      //three cellls per message
       for (uint8_t i = 0; i < 3; i++) {
         uint8_t cell_index_voltage = cell_start_voltage + i;
         if (cell_index_voltage < MAX_CELLS && (actual_cell_count == 0 || cell_index_voltage < actual_cell_count)) {
-          // Cell voltage: 3100mV base + (byte value × 4mV)
-          // Calaculation looks a little bit off.
-          uint16_t cell_voltage = 3100 + (rx_frame.data.u8[(i * 2) + 1] * 4);
-          // Only update if voltage is in valid range (2500-4200mV)
-          if (cell_voltage >= 2500 && cell_voltage <= 4200) {
-            uint16_t current_voltage = datalayer_battery->status.cell_voltages_mV[cell_index_voltage];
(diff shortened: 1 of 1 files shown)


Software/src/battery/PYLON-BATTERY.cpp
```diff
@@ -188,21 +188,19 @@ void PylonBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
-          // Only update if voltage is in valid range (2500-4200mV)
-          if (cell_voltage >= 2500 && cell_voltage <= 4200) {
-            uint16_t current_voltage = datalayer_battery->status.cell_voltages_mV[cell_index_voltage];
-            // Reject sudden large changes (>1000mV)
-            if (current_voltage == 0 || abs((int)cell_voltage - (int)current_voltage) <= 1000) {
-              datalayer_battery->status.cell_voltages_mV[cell_index_voltage] = cell_voltage;
-            }
+          //start with the second byte and advance by 2 bytes for each cell (3 cells per message)
+          int16_t cell_voltage = (int8_t)rx_frame.data.u8[(i * 2) + 1];
+          if (cellvoltage_min_mV > 3200) {
```
dalathegreat, line 199: This  check does nothing for actual cell being processed, it just checks if the minimum cell voltage in the system is over 3200mV, then we get cellvoltages into the array.

I suggest removing the check entirely!

Software/src/battery/PYLON-BATTERY.cpp
```diff
@@ -188,21 +188,19 @@ void PylonBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       // Handle EMUS individual cell voltage
       uint8_t cell_start_voltage = rx_frame.data.u8[0] - 1;
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;  // Keep battery alive on cell data
-
+      //three cellls per message
```
dalathegreat: Spelling error

dalathegreat (changes requested):

Software/src/battery/PYLON-BATTERY.cpp
tekampe-oss: It does.

Unfortunately, I wasn't able to find a better solution. The calculation is correct when using a signed int8 for voltages between 2880 mV and 3520 mV. (3200 + value* 2.5) However, for voltages above 3520 mV, the value overflows the signed byte.
It appears that the next higher byte in the message does not contain any useful information. Therefore, if the minimum cell voltage is above 3200 mV, I reinterpret the value as an unsigned byte instead.

The min cell voltage is provides as a 16 bit value, so it is correct for the stack.

tekampe-oss: I add a comment, to make it more clear.

dalathegreat (approved):
Nice work making life better for Dyness owners! 🙌

PR #2680, opened 2026-07-26
Bugfix: missing newline on truncated log lines

nagyrobi:
### What
This PR terminates truncated log lines with a newline in `Logging::printf`, and returns early when `vsnprintf` produces nothing usable.

### Why
A line longer than `MAX_LINE_LENGTH_PRINTF` loses its trailing newline, so the next message - possibly logged hours later - is appended to it without its own timestamp. Affects all logging platforms.

### How
`vsnprintf` returns the length the line would have had, so a value at or above the buffer size means truncation; in that case the last character is replaced with a newline. A non-positive result now returns early instead of reading `message_buffer[-1]`, matching the `size == 0` check the `write()` path already has.

> [!TIP]

### DIFF
--- a/Software/src/devboard/utils/logging.cpp
+++ b/Software/src/devboard/utils/logging.cpp
@@ -382,8 +382,24 @@ void Logging::printf(const char* fmt, ...) {
 
   va_list args;
   va_start(args, fmt);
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
👍

PR #2681, opened 2026-07-26
Nissan LEAF: read and decode battery DTCs

nagyrobi:
### What
Adds DTC readout and decoding for the Nissan LEAF battery. The LEAF already supported clearing DTCs, but had no way to read them.

### Why
`web_data/dtc/nissan_leaf_dtc.json` (204 entries) was already in the repo thanks @dalathegreat .

### How
- `supports_read_DTC()` / `read_DTC()` wire the existing WebUI button to the LBC.
- Request is UDS ReadDTCInformation `19 02 0E` on 0x79B; the reply arrives on   0x7BB as `59 02 <mask>` plus 4 bytes per DTC (3-byte code + status),  multi-frame with flow control.
- The reply is intercepted before the periodic group-poll handling, because a DTC first frame carries `0x02` in the byte that handler reads as a group number (it would be mistaken for the cell-voltage group). The `0x59` vs `0x61` service byte separates them.
- `NISSAN-LEAF-HTML.h` renders the table itself and calls the existing `get_dtc_json_loader_html()`, same as the BMW iX / PHEV / Tesla renderers. Codes are matched on Nissan's 5-character short form (`P33D7`, `U1000`), which is how the JSON file is keyed; a non-zero failure-type byte is shown as `P33D7-2F` but kept out of the lookup key.
- No files outside the LEAF integration are touched.

8 unit tests added, driven by a real logged LBC capture (U1000, P33D7, P33D9, P33DD): single-frame, multi-frame, no-DTCs, negative response, timeout, group-reply-not-consumed, and the two rendering cases.

**Protocol provenance**

Taken from a logged LBC capture on the openinverter forum (<https://openinverter.org/forum/viewtopic.php?t=370&start=50>). The same capture's clear-DTC frame is byte-identical to the `LEAF_CLEAR_DTC` already shipping in this repo, which is a useful cross-check that the rest of it is sound. The `19 02 0E` request appears three times in that trace, including once after an erase returning zero codes.

> [!TIP]


nagyrobi:
image

dalathegreat:
Tested successfully on Double LEAF 30kWh setup :raised_hands:

dalathegreat (approved):

PR #2683, opened 2026-07-27
Validate SOC window consistently and stop silently reverting Min-SOC on reboot

ekholm:
The SOC window (min/max percentage) is guarded by three different rules at three entry points:

| Entry point | Rule enforced today |
|---|---|
| `/updateSocMin`, `/updateSocMax` live routes | none — any value accepted |
| `store_settings()` | none — persists whatever is live |
| Boot load in `init_stored_settings()` | stored minimum must be ≤ 50.0% |

Two user-visible consequences:

1. **Min-SOC above 50% silently reverts on reboot.** Set 60% → accepted, works, persisted → on reboot the boot guard rejects it and falls back to the compiled default with no trace. (The settings page's own prompt caps input at 50, but the route itself doesn't — anything driving the HTTP API directly hits this.)
2. **Nothing enforces min < max.** With an inverted window, `CONSTRAIN()` pins the scaled SOC to exactly 100% toward the inverter while the scaled total capacity wraps negative into a huge unsigned Wh value.

### Fix

One validator, used by every entry point (`settings_validation.h/cpp`), values in pptt: `-10.00% ≤ min`, `max ≤ 100.00%`, and min at least 1.00% below max — which enforces `min < max` and also keeps the SOC-scaling divisor from getting arbitrarily small (the scaling code only guards against zero).

- **Live routes** validate the new value against the other half of the pair and reject invalid combinations with HTTP 400. The settings page now surfaces the rejection (previously a non-200 response was silently ignored) and its prompts match the server rules.
- **Boot load** validates the stored pair; when invalid it keeps the compiled defaults for both values (silently, per review — no new event).

### Deliberate choices

- The old 50% boot cap on the minimum is dropped: large reserve floors (60–70%) are legitimate for backup-power installations, and the cap had no counterpart on the live path or the max side. The −10% floor the settings UI has always offered is kept.
- NVS encoding (`percent*10`) and keys (`MINPERCENTAGE`/`MAXPERCENTAGE`) are unchanged — no migration. A minimum >50% stored by this version and then downgraded hits the old guard and reverts.
- A missing `MINPERCENTAGE` key keeps its long-standing effective behavior (applied as 0.0%).

### Tests

`test/soc_window_tests.cpp` (new, registered in the test CMakeLists): validator rule coverage (typical pairs incl. negative min and >50% min, min==max, min>max, ceiling, floor, exact-gap boundary) and apply/reject semantics (valid pair applied to the datalayer, invalid pair leaves it untouched). Full suite: 90/90 pass. Firmware compiles clean under the `-Werror` warning-check environment.


dalathegreat:
Thanks for the improvement! I like more validation on user input, but I do think that raising an event is overkill. We are at quite high flash usage, and I atleast personally don't think wasting the flash on more events for a scenario that is very unlikely is worth it. Silently reverting to a valid value incase the stored value is invalid is OK for our usecase!

ekholm:
Agreed — event dropped in 5f136d10. The boot path now silently keeps the compiled defaults when the stored pair is invalid (`set_soc_window()` writes nothing on an invalid pair, so it can never apply half of one). The validation itself is unchanged: the live routes still reject bad values with HTTP 400 and an alert in the UI. PR description updated to match.

Also pushed c8ae79a0: the maximum now has a 1.00% floor. A stored max of 0 collides with the NVS never-stored sentinel and would be silently swapped for the compiled default at boot — same accept-then-lose class this PR removes, so the validator now makes it unrepresentable.

ekholm:
Related: #2629 approaches the same subsystem from the other side — it fixes what capacity scaling does when MinSoc is negative, while this PR fixes which min/max pairs can reach the scaling code at all. For the record: the validator here deliberately keeps the −10% minimum floor (`SOC_WINDOW_MIN_FLOOR_PPTT = -1000`), precisely because a negative MinSoc to fully discharge batteries behind inverters with fixed cutoffs is a supported use case — so the two changes are independent and don't step on each other. The only shared file is webserver.cpp in different regions; happy to rebase whichever lands second.

PR #2685, opened 2026-07-27
Allow BMS reset off time up to 600 seconds

nagyrobi:
### What
This PR raises the accepted range of the "Periodic BMS reset off time:" setting from 1-59 to 1-600 seconds, and keeps the CAN liveness watchdog from tripping during the longer off times.

### Why
To test the statement in Leaf "GEN4_e_Battery_control_spec_ver1.0.pdf" page 4, "IGN to be OFF for more than 6 min 30 seconds every day. "
Off times above 59 s would previously have latched `EVENT_CAN_BATTERY_MISSING`, since the safety layer faults after `CAN_STILL_ALIVE` seconds of silence.

### How
Widened the web UI validation bound and `user_set_bms_reset_duration_ms` to `uint32_t` (600000 ms no longer fits a `uint16_t`).
For off times above 59 s only, `handle_BMSpower()` refreshes `CAN_battery_still_alive` on all configured batteries every 59 s while the reset runs, plus once at power-on so the BMS gets a full window to rejoin the bus; shorter off times keep the original unmasked behaviour.
Since this is experimental only, the UI will not change, meaning users will still be instructed to choose between 1-59s.

> [!TIP]


nagyrobi:
Software: 12.0.0 Hardware: LilyGo T-CAN485

With Periodic BMS reset off time: 400 s
```
      24.746 MQTT message arrived: [battery-emulator-6634/command/BMSRESET]
      24.746 Triggering BMS reset
      24.746 Battery pause begin 1 0 0 0
      24.746 Event: The emulator is trying to pause the battery.
     427.756 Battery pause begin 0 0 0 0
     427.756 Event: The emulator is attempting to resume battery operation from pause.
     427.756 Event: BMS reset event completed.
```
image

dalathegreat (approved):

PR #2688, opened 2026-07-27
Deduplicate CAN aliveness checks and fix shared corrupted-CAN warning being cleared by healthy batteries

ekholm:
### The bug: one battery's corruption warning is masked by another battery's health

`EVENT_CAN_CORRUPTED_WARNING` is a single event shared by all batteries, but each battery block in `update_machineryprotection()` set **or cleared** it independently — so the last block to run wins. With a double/triple battery setup, a healthy battery 2 (or 3) silently clears battery 1's active corruption warning every cycle: battery 1 can be receiving a stream of malformed CAN frames and the event log shows nothing.

Fixed by evaluating all configured batteries together after the battery blocks: the warning is active while **any** battery's `CAN_error_counter` exceeds `MAX_CAN_FAILURES`, and clears only when all are healthy. Event data is now the first offending channel (previously: last writer's channel — a strictly-better tiebreak, listed here as the only intentional semantic change).

The new regression test `CorruptedWarningNotMaskedByHealthySecondBattery` fails on current main and passes with this change.

### The cleanup the fix falls out of

The CAN aliveness handling (detected-event latch on first counter refresh, missing-event at counter zero, decrement + clear otherwise) was copy-pasted four times: battery 1, battery 2, battery 3, charger. This is the same copy-paste drift that produced the bugs fixed in #2677 — and this corruption-warning bug is another instance of it.

The four blocks are replaced by one helper:

```cpp
static void check_can_component_alive(uint8_t& still_alive_counter, bool& detected, EVENTS_ENUM_TYPE detected_event,
                                      EVENTS_ENUM_TYPE missing_event, uint8_t missing_event_data);
```

**The inverter block is deliberately not converted** — it has the long-timeout option and startup-grace special cases the others don't, and stays exactly as #2677 left it.

Everything else is behavior-preserving: same detection latch semantics (`>=`, matching #2677), same event data arguments (charger keeps `charger->interface()`), same guards around each call site.

### Tests

`test/battery_alive_tests.cpp` (new, registered in the test CMakeLists):

- `CorruptedWarningNotMaskedByHealthySecondBattery` — the masking regression, fails on current main
- `SecondBatteryCorruptionRaisesWarning`, `CorruptedWarningClearsWhenAllBatteriesHealthy` — aggregation behavior
- Detection at multiplied refresh values, missing-at-zero / clear-and-decrement-on-refresh for battery 1, battery 2, and charger paths — all pass before and after (behavior preservation)

Full suite: 99/99 pass. Firmware compiles clean under the `-Werror` warning-check environment.

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -38,6 +38,26 @@ bool inverter_detected = false;
 battery_pause_status emulator_pause_status = NORMAL;
 //battery pause status end
 
+// Shared CAN-aliveness handling for battery 1/2/3 and the charger: latch the
+// detected-event on the first counter refresh, raise the missing-event when the
+// counter runs out, decrement and clear it otherwise. The inverter has its own
+// logic (long-timeout option, startup grace) and is handled separately.
+static void check_can_component_alive(uint8_t& still_alive_counter, bool& detected, EVENTS_ENUM_TYPE detected_event,
+                                      EVENTS_ENUM_TYPE missing_event, uint8_t missing_event_data) {
+  if (!detected) {
+    if (still_alive_counter >= CAN_STILL_ALIVE) {
+      detected = true;
+      set_event(detected_event, 1);
+    }
+  }
+  if (!still_alive_counter) {
+    set_event(missing_event, missing_event_data);
+  } else {
+    --still_alive_counter;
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Good improvement, thanks for the fix! 👍

PR #2693, opened 2026-07-28
Remove dead build_cxxflags lines: RTTI is already disabled by the IDF default

ekholm:
Removes the seven `build_cxxflags = -fno-rtti` lines (six commented, one active on `lilygo_330`) and documents once in the `[platformio]` section why no RTTI flag is needed. Two findings behind this:

### 1. `build_cxxflags` is not a PlatformIO option — the flag has never been active

Every `lilygo_330` build prints:

```
Warning! Ignore unknown configuration option `build_cxxflags` in section [env:lilygo_330]
```

so the `-fno-rtti` intended by 07442e42 ("Make fno-rtti a cxxflag only") has been silently ignored since it was added. Presumably it started life in `build_flags`, broke C compilation, and was moved to an option name that doesn't exist (the real PlatformIO spellings are `build_unflags`/`build_src_flags` — neither is what's wanted here anyway).

### 2. It doesn't matter: RTTI is already disabled everywhere

The generated sdkconfig has `CONFIG_COMPILER_CXX_RTTI` unset (the ESP-IDF default), and the pioarduino hybrid build applies that to all app C++. Verified in the ELF: zero app-class typeinfo symbols out of ~78 polymorphic classes — the only `_ZTI`/`_ZTS` symbols are libstdc++ exception classes from the prebuilt toolchain library. Also grep-verified: nothing in `Software/` uses `dynamic_cast` or `typeid`, so nothing could ever have needed RTTI.

### Verification

- A/B builds of `lilygo_330` (this change vs. current main): **identical size (1,909,136 B)**; the only differing bytes are the embedded compile time/date + ELF hash in the ESP-IDF app descriptor header. Same result independently measured on `esp32devkit_330` (byte-identical at 1,908,207 B with reproducible timestamps).
- The "unknown configuration option" warning is gone from the build output.
- No functional change; the size CI will show a zero delta.

Context: with `esp32devkit_330` at 97.1% of its OTA slot, flash work is ongoing — this removes a false "already handled via -fno-rtti" belief from the config so future size investigations start from accurate ground truth.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,6 +10,10 @@
 
 [platformio]
 src_dir = ./Software
+; C++ RTTI is already disabled in every environment by the ESP-IDF default
+; (CONFIG_COMPILER_CXX_RTTI unset in the generated sdkconfig) - no compiler
+; flag is needed. Note that 'build_cxxflags' is not a PlatformIO option and
+; is silently ignored ("Warning! Ignore unknown configuration option").
 
 [env:compiler_warning_check]
 platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.39/platform-espressif32.zip
@@ -23,7 +27,6 @@ board_build.partitions = min_spiffs.csv
 framework = arduino
 build_flags = -I include -DHW_LILYGO -DSMALL_FLASH_DEVICE -Wimplicit-fallthrough -Wextra -Wall -Werror
     -fno-exceptions -ffunction-sections -fdata-sections -Wl,--gc-sections
-;build_cxxflags = -fno-rtti
 lib_deps = 
 
 [env:esp32devkit_330] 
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -10,6 +10,10 @@
 
 [platformio]
 src_dir = ./Software
+; C++ RTTI is already disabled in every environment by the ESP-IDF default
```
dalathegreat: Can you remove this large comment? IMO not needed

dalathegreat (changes requested):

platformio.ini
ekholm: Removed in d6b1bb3f — the PR is now a pure deletion of the seven lines. (The full explanation stays in the PR description / commit message for the archaeology.)

dalathegreat (approved):

PR #2697, opened 2026-07-28
Sync settings-page defaults with boot defaults (6 mismatches)

ekholm:
For a setting whose NVS key was never saved, the settings page renders one default while the firmware boots with another: the operator believes X is active while the firmware runs Y — and saving the form as-displayed silently **persists** the wrong value. Concrete case: a device with never-saved `CHGPOWER` runs a 1000 W charger override while the page shows 0; one save of the settings form turns the working 1000 W into 0 without the user ever touching that field.

Six keys are affected (verified as the complete set by two independent passes over `comm_nvm.cpp` vs the settings processors):

| Key | Boot default | Page displayed | Fix |
|---|---|---|---|
| `CHGPOWER` | 1000 W | 0 | render → **1000** |
| `DCHGPOWER` | 1000 W | 0 | render → **1000** |
| `BATTCHEM` | NCA (1) | Autodetect (0) | render → **NCA** |
| `CTATTEN` | ADC_11db (3) | ADC_0db (0) | render → **ADC_11db** |
| `INVSUNTYPE` | SBR064 (0) | SBR096 (1) | render → **SBR064** |
| `MQTTPORT` | 0 | 1883 | **boot → 1883** |

**Five are display fixes** — the render literal changes to what boot has always applied, so device behavior is untouched; the page just stops lying. For `CHGPOWER`/`DCHGPOWER` the direction was verified against the consumers: `override_charge_power_W` feeds `max_charge_power_W` directly (ENNOID, iMiEV, …), so 0 means *zero watts*, not "no override" — aligning boot to the displayed 0 instead would have disabled charging on never-saved devices.

**One is a boot fix**, where the stored default was the broken half: `MQTTPORT` 0 → 1883. Port 0 is never valid, 1883 is the MQTT standard and what the page has always shown. This is benign: MQTT is disabled by default, and anyone who enabled it via the form saved the pre-filled 1883 along with it.

**One question left open deliberately:** `BATTCHEM` arguably *should* default to Autodetect — but that changes chemistry resolution (and with it cell-voltage limits) for never-saved packs, which is safety-adjacent. This PR keeps the historical boot value (NCA) on both surfaces; if Autodetect is the intended default, that's a deliberate two-literal follow-up.

After this change, render default == boot default for every key, so the accept-then-lose scenario is closed for all six. Longer term, single-sourcing each default (per-setting descriptor, fits naturally with the settings work in #1544) makes this bug class structurally impossible — this PR is just the sync, kept frontend-agnostic on purpose.

No test coverage exists for these paths (webserver isn't in the host test build); verification is the green `-Werror` build, a ~zero size-CI delta (literal-for-literal), and the full host suite still passing (99/99).

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -264,7 +264,7 @@ void init_stored_settings() {
   static_dns = settings.getString("DNS").c_str();
 
   mqtt_server = settings.getString("MQTTSERVER").c_str();
-  mqtt_port = settings.getUInt("MQTTPORT", 0);
+  mqtt_port = settings.getUInt("MQTTPORT", 1883);
   mqtt_user = settings.getString("MQTTUSER").c_str();
   mqtt_password = settings.getString("MQTTPASSWORD").c_str();
 
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -276,9 +276,8 @@ String settings_processor(const String& var, BatteryEmulatorSettingsStore& setti
            capability_css("if-tricapable", battery_supports_triple);
   }
   if (var == "BATTCHEM") {
-    return options_for_enum(
-        (battery_chemistry_enum)settings.getUInt("BATTCHEM", (int)battery_chemistry_enum::Autodetect),
-        name_for_chemistry);
+    return options_for_enum((battery_chemistry_enum)settings.getUInt("BATTCHEM", (int)battery_chemistry_enum::NCA),
+                            name_for_chemistry);
   }
   if (var == "INVTYPE") {
     return options_for_enum_with_none(
@@ -310,8 +309,8 @@ String settings_processor(const String& var, BatteryEmulatorSettingsStore& setti
 
   if (var == "CTATTEN") {
     return options_for_enum_with_none(
-        (adc_attenuation_enum)settings.getUInt("CTATTEN", (int)adc_attenuation_enum::ADC_0db), name_for_adc_attenuation,
-        adc_attenuation_enum::ADC_0db);
+        (adc_attenuation_enum)settings.getUInt("CTATTEN", (int)adc_attenuation_enum::ADC_11db),
+        name_for_adc_attenuation, adc_attenuation_enum::ADC_0db);
   }
 
   if (var == "EQSTOP") {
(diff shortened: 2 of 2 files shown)


jonny5532:
This is better than the old behaviour, but it does commit the defaults to flash on first save, meaning if BE ever changes them in future, people who have already saved won't get the newer defaults.

I think it would be better to only save to flash if people have changed the defaults away from the default-at-the-time values (which is what my new frontend does IIRC) - thus, if a default proves unsuitable, BE can change it without them having to tell everyone (who never deliberately did anything) to update their values too.

ekholm:
Agreed — persisting only user-changed values is the better model, and it also keeps NVS smaller and lets defaults evolve. The obstacle in the current codebase is that the save handler would need to know every key's default, and right now the defaults live as duplicated literals spread across the boot readers and the page processors (which is exactly how this PR's six mismatches happened — this PR aligns two of the three copies).

Your new frontend's settings tables in #1544 are the right shape for this: one row per setting with its default, consulted at save time. That's also where [my comment there](https://github.com/dalathegreat/Battery-Emulator/pull/1544#issuecomment-5102304846) connects — the same table needs to carry the on-flash NVS type per key (the MINPERCENTAGE getInt/getUInt mismatch and the missing i32 table), because "value == default" comparisons and key removal only work reliably when the read back uses the type the key was written with.

I'm preparing a similar backend-side move: all settings (key, default, range, on-flash type) in a single descriptor table, so the boot reader, the settings page, and the save handler pull from one source. On top of either table, your policy becomes trivial: on save, skip the write when the submitted value equals the current default, and remove the key if it was previously stored. One semantic trade-off to agree on: a user who deliberately selects the current default becomes indistinguishable from one who never touched the setting and will silently follow future default changes — I think that's acceptable, but it's a policy choice.

Suggest keeping this PR as the minimal read-side alignment so the form stops lying about what boot assumes, with don't-materialize-defaults implemented on a descriptor table where it can be done once instead of ~140 times. Does that ordering work for you?

ekholm:
Rebased onto current main (it was 129 behind); the change replays byte-identical.

While rebasing I re-ran the audit from scratch against current main rather than just re-checking the six — comparing every non-bool key's default on both surfaces with the enum casts resolved. 64 keys appear on both; after this PR none of them mismatch, and none of the 129 commits since filing introduced a new one. Build is clean and the host suite is 160/160.

One thing the re-audit made clearer, which is really an argument for @jonny5532's point rather than mine: the boolean settings avoid this bug class by luck, not by design. `getBool()` defaults its parameter to `false`, which happens to match every boot reader that passes `false` — all of them except `WIFIAPENABLED`, which boots `true` and only stays consistent because the page renders it from the live variable instead of a literal. Add one bool with a `true` boot default and the six become seven, silently.

@jonny5532 — the question from my last comment is still open: does the ordering work for you? Keep this as the minimal read-side alignment so the form stops lying about what boot assumes, and implement don't-materialize-defaults on a descriptor table where it's done once rather than ~140 times. Happy to go the other way if you'd rather see the save-side policy first, but it needs the per-key defaults to exist in one place before it can be written at all.

dalathegreat (approved):

PR #2700, opened 2026-07-28
Upgrade ESP-NOW to v2 (v2)

nagyrobi:
### What
Upgrades the ESPNow integration to v2.

### Why
The v1 implementation broadcast raw datalayer structs byte-for-byte. Any field added, reordered or resized broke every receiver, and cell voltages had to be quantized to 20 mV steps to fit the 250 byte ESP-NOW v1 frame.

### How
Replace it with a self-describing key/length/value stream:

* The length lives in the tag byte, encoded independently of the type class, so a receiver can skip a record whose *type* it has never seen, not just an unknown key. New keys and new types can therefore both be added without another compatibility break.
* Cell voltages are sent as raw uint16 millivolts, unquantized, for all three batteries, alongside a balancing bitset.
* Everything currently published over MQTT is exposed, plus the emulator-level status, link state and event stream.
* ESP-NOW v2 (1470 byte) frames are used when the IDF supports them, detected at runtime via esp_now_get_version(). On a v1-only build the cell array is split into index-tagged chunks, so the same firmware works either way.
* New ESPNOWMACS setting: a list of receiver MAC addresses shown on the settings page only while ESPNow is enabled. Empty keeps the previous broadcast behaviour.

Frames are paced one per tick instead of bursting a dozen packets at once, and serialization uses a single static buffer with no String, printf or allocation in the send path.

We will add very easy examples to build displays with working code to receive and show the BE information. The plan is to keep these in a separate repo.

> [!TIP]


nagyrobi:
A test ESPHome configuration for [Guition ESP32-S3-4848S040](https://www.aliexpress.com/item/1005008214872438.html) 480*480 smart screen: [be-monitor-guitionn.zip](https://github.com/user-attachments/files/30507478/be-monitor-guitionn.zip)


In ESPHome config you need to adjust the `emulator_mac:` yaml key with the MAC address of your running Battery Emulator, which is displayed in the index page.

In Battery Emulator turn on "Enable ESPNow:" and optionally fill in "ESPNow receiver MACs:" addresses of displays to show the data (multiple possible, comma separated):
image


> For those unfamiliar with ESPHome, check out [their Getting Started](https://esphome.io/#getting-started) doc. It doesn't require Home Assistant, despite the project is affiliated with it.

battery_emulator_espnow_esphome_lvgl_1
battery_emulator_espnow_esphome_lvgl_2
battery_emulator_espnow_esphome_lvgl_3
battery_emulator_espnow_esphome_lvgl_4
battery_emulator_espnow_esphome_lvgl_5
battery_emulator_espnow_esphome_lvgl_6
battery_emulator_espnow_esphome_lvgl_7

Plan is to make the config nicer with wigdets etc and multiple screens/resolutions would be supported in a separate repo.

nagyrobi:
@jonny5532 what do you think?

jonny5532:
Looks good! Would be nice if it were a bit smaller than 3kb on the T485 but I can't see how you could do that, the code is pretty tight. Seems fairly future proof, with spare slots in the key byte. You could also reserve a key for 'extended key' which is followed by a uint16_t or uint32_t, to make it even more extensible. Wouldn't affect this code, but might be worth building into the demo receive code so that they are more tolerant to future BEs.

I probably wouldn't worry about support ESP-NOW v1 at all anymore? The only chip that doesn't support it is the ESP8826 which won't be in any modern screens, and ESP-IDF 5.4 is old now, and will be EOL in Jan 2027...

(Removing the detection and const-ing `max_payload` saves 300 bytes!)

jonny5532:
Oh, hah, just read the comment:
`0xFF        reserved as an escape for a future 16 bit key space`
so you had already thought of that...

(although I think your 'Wire format' doc comment might need a `if(key==0xFF) i+=2;` or similar for future compatibility?)

nagyrobi:
Thanks! Looking

nagyrobi:
Dropped v1 so it's indeed about 2.6k on T485. Addressed wire format too.

nagyrobi:
@jonny5532 what do you think?
I'd start designing UIs.

dalathegreat (approved):

PR #2703, opened 2026-07-29
Add MEB ODIS bridge mode for car-less pack diagnostics

ekholm:
Adds a runtime-selectable **"MEB ODIS Bridge"** battery type for car-less MEB pack diagnostics: connect ODIS (via a VAS interface) on the emulator's native CAN side and the MEB battery on the CAN-FD side, and the emulator bridges frames both ways. No emulation runs in this mode — the bridge is a passive gateway that keeps the aliveness counter fed. Demand for this came up in recent community discussion around diagnosing packs outside the vehicle.

### What this contains

- The crude bridge mode from `feature/ODIS-BRIDGE` (5683e715 + df90e71b), merged up to current main — including adapting the FD→native forwarding into the dual-MCP2518 RX path (`_receive_frame_canfd`, first controller only) and re-slotting the `MebOdisBridge` enum after the battery types added on main since (now 55).
- **One behavior change on top: oversize FD frames are dropped instead of truncated.** The previous code clamped `DLC > 8` to 8 before forwarding. ODIS speaks UDS over ISO-TP, and a truncated ISO-TP frame has an inconsistent length nibble — the tool receives garbage that parses as a malformed message and can corrupt the diagnostic session. Dropping fails cleanly instead: the request times out, and the tool retries or reports a communication error rather than chasing wrong data.

### Known limitation (deliberate, documented at the code site)

Only ≤8-byte frames bridge in the FD→native direction. The MEB BMS broadcasts 64-byte FD frames and answers UDS on FD, so responses longer than 7 payload bytes won't pass yet. Proper ISO-TP re-segmentation (FD single-frame → classic first/consecutive frames, likely reusing the in-tree `uds_isotp`) is the planned follow-up; this PR ships the honest subset rather than a lying superset. The native→FD direction forwards as-is (classic frames are always ≤8 bytes).

### Supersedes `feature/ODIS-relay`

The April 2025 `feature/ODIS-relay` branch is fully contained here in modern form: its compile-time `MEB_GATEWAY_MODE` define became the runtime-selectable battery type, its passive gateway pseudo-driver became `MebOdisBridge`, and its DLC-clamp bug (`rx_frame.DLC == 8;` — a comparison, so the clamp never executed) is moot under the drop semantics. The stale branch can be deleted.

### Testing

Host suite 101/101 green; firmware compiles clean under the `-Werror` warning-check environment. The bridge path itself isn't host-testable (it lives in the CAN driver RX path), and no hardware smoke test with a real ODIS setup has been run on this machine — flagging that explicitly for anyone able to try the artifact from the size-CI build.


dalathegreat:
Think we can close this now since https://github.com/dalathegreat/Battery-Emulator/pull/2730 is available and reduces ODIS need

PR #2704, opened 2026-07-29
Actions: Cache custom_sdkconfig image for faster builds

jonny5532:
### What
The addition of `custom_sdkconfig` for smaller firmware also extended the build time by several minutes due to the extra compilation step.

This PR tweaks the caching, as well as some silly hacks to convince PlatformIO that the `custom_sdkconfig` doesn't need rebuilding.

### Why
Screenshot_20260729_082246

### How
After some trial and error and looking through the PlatformIO internals it seems the following is necessary for a build to reuse an existing custom platform library build:
- The `sdkconfig.defaults` in the project folder, generated during the initial compile, must be preserved and reused
- The custom sdkconfig itself (`sdkconfig.be_size.defaults`) must have the same mtime as the initial compile, and because git doesn't preserve mtimes, it requires an explicit `touch` with a hardcoded time

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -42,27 +42,54 @@ jobs:
     - uses: actions/checkout@v7
       name: Checkout code
 
-    - name: Get platformio.ini hash
-      id: platformio_ini_hash
-      run: |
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Now this is amazing 🙌 Nice work @jonny5532 , you are the king of the build-system 👑

PR #2706, opened 2026-07-29
Improvement: BYD - source cell min/max from 0x446 broadcast, drop redundant UDS polls

demon1300:
### What
Source BYD Atto 3 min/max cell voltage and cell numbers from the 0x446 broadcast instead of UDS polling, and drop the now-redundant 0x002A–0x002D polls. Adds MQTT min/max cell number sensors.

### Why
0x446 already broadcasts the cell extrema at ~1Hz; polling the same values was ~5× slower and used four rotation slots. Broadcast-sourcing refreshes cell min/max faster and speeds up every other polled value too. Exposing the cell numbers over MQTT lets a weak cell be tracked over time.

### How
Decode 0x446 (b0/b4 = 1-based min/max cell number, min/max mV 12-bit LE in b1:b2 / b5:b6) into the existing cell min/max fields, so all readers are fed unchanged, just fresher. Remove the four UDS poll cases, re-thread the poll rotation over them, drop the unused constants. Add the two MQTT sensors.

Tested on my own Atto 3 pack: 0x446 min/max voltages and cell numbers matched the 0x43D full sweep.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -505,6 +505,12 @@ void BydAttoBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
       break;
     case 0x446:
       datalayer_battery->status.CAN_battery_still_alive = CAN_STILL_ALIVE;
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Great improvement 👍

btoonen:
Thnx

Issue #2390, opened 2026-05-31
[BUG] BMW i3 faulting after 50 days of operation


dalathegreat:
**Describe the bug**
The BMW i3 battery goes into a FAULT state due to overtemperature, plus SOC% goes to 0% after 50 days of operation

**To Reproduce**
Run the Battery-Emulator for 50 days nonstop with a BMW i3 60AH battery

**Expected behavior**
Nothing bad happens after 50 days

**Screenshots**

<img width="326" height="185" alt="Image" src="https://github.com/user-attachments/assets/6c9132bd-944a-43bd-9b36-5ab285934f37" />

<img width="1016" height="577" alt="Image" src="https://github.com/user-attachments/assets/60e0ebd4-6569-4127-b01b-f9e6781a4fff" />

**Version and settings (please complete the following information):**
- Please attach screenshot of configuration for easier debugging
- Software version: 10.5.0
- Battery used: BMW i3
- Inverter communication protocol: BYD Modbus
- Hardware used for Battery-Emulator: LilyGo T-CAN
- GPIO controlled contactors: no
- MQTT: no

ekholm:
The 50-day mark is the fingerprint: **2³² ms = 49.71 days — `millis()` wraps.** Swept the tree for wrap-unsafe time comparisons; five sites, and one belongs to the i3:

`BMW-I3-BATTERY.cpp` `update_values()` gates the wakeup pin with `else if (millis() > INTERVAL_1_S)` — an absolute comparison meant as "after boot." When `millis()` wraps at 49.7 days, that condition goes false for one second and **the wakeup line drops mid-operation**. An SME held awake for 50 days getting a 1 s wakeup-line drop is a solid candidate for the observed corruption: poll multiplexer desync or SME re-init → implausible temperature (overtemp fault) + SOC 0%.

Same pattern exists in `CMP-SMART-CAR-BATTERY.cpp` (identical wakeup gating). Lesser cousins: `safety.cpp` inverter startup grace (`millis() > INVERTER_STARTUP_GRACE_MS` — SMA missing-event gets re-suppressed for 5 min every 49.7 days, cosmetic), `KIA-E-GMP` one-shot init guard (benign, latch already set), and `Software.cpp`'s remote-limit expiry (`currentMillis > timestamp + timeout` — MQTT remote limits can mis-expire at wrap).

[...]

PR #2707, opened 2026-07-29
Make absolute millis() comparisons wrap-safe (49.7-day rollover)

ekholm:
`millis()` wraps at 2³² ms ≈ 49.7 days. Five sites compare it as an absolute since-boot value instead of using wrap-safe arithmetic. Per-site effect at the wrap, honestly graded:

| Site | Effect at the 49.7-day wrap |
|---|---|
| `BMW-I3-BATTERY.cpp` wakeup gating | None in practice — the GPIO output latch holds HIGH through the 1 s post-wrap window; worst case is a ≤1 s delayed HIGH-restore if equipment-stop/balancing ends inside exactly that window. Hygiene fix. |
| `CMP-SMART-CAR-BATTERY.cpp` wakeup gating | Same pattern, same (non-)effect. Hygiene fix. |
| `KIA-E-GMP-BATTERY.cpp` voltage-limit one-shot | Benign — the one-shot latch was set long before the wrap. Hygiene fix. |
| `safety.cpp` inverter startup grace | **Real, modest** — for inverters using the startup grace window (SMA family), a *new* inverter-comms loss beginning right after the wrap is detected up to 5 minutes late, because the grace window re-arms. An already-raised event is not cleared by this. |
| `Software.cpp` remote-limit expiry | **Real, both directions** — `timestamp + timeout` overflows near the wrap, so freshly set remote limits expire immediately; and once `currentMillis` wraps, stale limits can persist up to another 49.7 days. |

### Fix

- Since-boot gates use the in-tree monotonic `millis64()` (the event system already uses it) — one-line changes, identical semantics before the wrap.
- The remote-limit expiry uses wrap-safe unsigned subtraction (`currentMillis - timestamp > timeout`) — no 64-bit math needed, and boot state (`timestamp = 0`) behaves as today.

### Relation to #2390

Found while investigating #2390 — but deliberately **not** claimed as a fix for it. As noted in the issue discussion, the emulator holds the pack awake far beyond an ECU's designed ignition-cycle uptime, and a 32-bit millisecond counter wrapping *inside the SME itself* matches the ~50-day fingerprint and the power-cycle recovery exactly — which no emulator-side change can prevent. This PR removes the emulator's own instances of the same bug class.

### Testing

Full host suite green; firmware compiles clean under the `-Werror` warning-check environment.

A 32-bit wrap regression test isn't possible in the host suite, and not just because the emulated `millis()` returns 64-bit `unsigned long`: truncating the emulated clock to `uint32_t` was tried and makes things worse. Production code stores timestamps in `unsigned long` variables and computes deltas at that width — uniformly 32-bit on the target, uniformly 64-bit on the host. A 32-bit clock feeding 64-bit delta arithmetic breaks code that is provably correct on the target (`bms_reset_tests` fails falsely, since `20000 - 0xFFFFFFF5` no longer reduces mod 2^32). Notably, that test already starts its clock at 2^64−10 on purpose — the suite exercises the same wrap-safe-subtraction idiom at the host's own wrap boundary, which is the consistent analog of what the target does at 2^32. The suite therefore serves as a no-regression check for the subtraction idiom and normal semantics; the 32-bit-specific sites rely on review.

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -321,8 +321,12 @@ void update_calculated_values(unsigned long currentMillis) {
   /*Update free heap*/
   datalayer.system.info.CPU_free_heap = ESP.getFreeHeap();
 
-  /* Check is remote set limits have timed out */
-  if (currentMillis > datalayer.battery.settings.remote_set_timestamp + datalayer.battery.settings.remote_set_timeout) {
+  /* Check is remote set limits have timed out. Wrap-safe subtraction: the naive
+     "currentMillis > timestamp + timeout" comparison both overflows near the 49.7-day
+     millis() wrap (fresh limits expire immediately) and inverts after it (stale limits
+     persist up to another 49.7 days) */
+  if ((currentMillis - datalayer.battery.settings.remote_set_timestamp) >
+      datalayer.battery.settings.remote_set_timeout) {
     datalayer.battery.settings.remote_settings_limit_charge = false;
     datalayer.battery.settings.remote_settings_limit_discharge = false;
     datalayer.battery.settings.max_remote_set_charge_dA = 0;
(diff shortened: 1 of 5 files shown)


dalathegreat (approved):
Massive thanks for the fix! Merging to get experience asap on especially BMW integration

PR #2709, opened 2026-07-30
Make core time arithmetic width-deterministic (uint32_t)

ekholm:
`millis()` is 32 bits wide on the target: the wrap at 2^32 ms IS the semantics of every timestamp comparison. Declaring time values as `unsigned long` makes that width platform-dependent — correct on the ESP32 where `unsigned long` is 32-bit, silently different on the 64-bit host test build, where the wrap never occurs. Consequence (discussed on #2707): wrap behavior was untestable, and the host suite exercised different arithmetic than the target runs.

This narrows the **core** time-carrying variables to `uint32_t`: `Software.cpp` interval timers and the `update_calculated_values()` parameter, the precharge/BMS-reset state-machine timestamps in `comm_contactorcontrol.cpp`, the datalayer `remote_set_timestamp`/`remote_set_timeout`, and `MyTimer`. The remote-limit expiry check moves into `update_remote_limit_expiry()` in safety.cpp so the host suite can reach it (behavior unchanged).

With production widths deterministic, the emulated `millis()` now returns a truncating `uint32_t` — the host runs exactly the arithmetic the target runs, wrap included. The BMS-reset overflow tests move from the host's 2^64 boundary to the target's real 2^32 boundary, and new wrap regression tests cover the two real bugs fixed in #2707 (SMA startup-grace re-arm; remote-limit expiry, both overflow directions) — each verified to fail against the pre-#2707 logic.

Driver-level `unsigned long` (including the `transmit_can(unsigned long)` signatures, 83 sites) is deliberately untouched — that mechanical sweep is proposed separately to keep this reviewable.

Testing: full host suite green (103 tests), `-Werror` warning-check build green, size CI expected ~neutral.

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -41,10 +41,12 @@
 // The current software version, shown on webserver
 const char* version_number = "12.1.dev";
 
-// Interval timers
-volatile unsigned long currentMillis = 0;
-unsigned long previousMillis10ms = 0;
-unsigned long previousMillisUpdateVal = 0;
+// Interval timers. Time values are uint32_t on purpose: that is the width
+// millis() actually has on the target, so the wrap arithmetic is identical on
+// every platform (including the 64-bit host test build).
+volatile uint32_t currentMillis = 0;
+uint32_t previousMillis10ms = 0;
+uint32_t previousMillisUpdateVal = 0;
 // Task time measurement for debugging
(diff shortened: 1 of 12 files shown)


jonny5532:
Looks sensible! I hate the vagueness of the traditional C types...

dalathegreat (approved):
👍

Issue #1664, opened 2025-10-23
[BUG] FoxESS EP11 using FoxESS HV2600 Protocol, Contactor not closing


vgyda:
**Describe the bug**
Using the FoxESS EP11 Battery (FoxESS HV2600 Battery Protocol) with a LilyGo T-CAN485 connected to a Fronius Symo Gen24 Plus 10.0 (BYD 11kwH HVM Battery over Modbus Protocol). Fronius Side works correctly to my knowledge and the problem lies between the Battery and Emulator. 

The Emulator receives all the Details about the Battery, SOC, Voltages, Cell Voltages and initial communication works. Pre-Charge also working? (as the Fronius and Fluke MM is showing around 100-110V on the Battery Input on the Fronius Inv.) 

Emulator errors out after initialisation with two errors: (Battery_Overvoltage) and then later with (CAN_Battery_Missing). No more communication with the Battery after first contact. Contactors never close. (Voltage stays at 100V). Fronius waits for the Battery Voltage to show up. FoxESS Battery shows "Standby" and "SOC 50%" (got delivered with 50% SOC)

**About the two Errors:**
"Battery_Overvoltage" is being shown, but the emulator correctly gets the Voltage of around 398V and shows it on the status screen.  (FoxESS HV2600 Batterys work on a voltage between 102.4 and 409.6, depending on configuration. EP11 is similar voltage wise to the already supported HV2600 HS20.8)
"CAN_Battery_Missing" stays on Error/Event screen, but CAN Communication is still flowing. (Multiple Logs can be provided, TX and RX both flowing and battery is resending Cell Voltages) Checked out the excellent work from https://github.com/FozzieUK/FoxESS-Canbus-Protocol and Frames look as expected.
The FoxESS EP11 is made for the H-Series Inverters (H1, H3, H3-Pro, KH) which the emulator supports on the inverter side. 

**What I suspect:**
Battery Voltage could be hardcoded in the FoxESS HV2600 Battery Protocol somehow and the Emulator is hoping for a different voltage? 
FoxESS updated the CAN Protocol and Battery waits for some kind of ACK? As the CAN Protocol shows the battery resending all the time the Main CAN-Bus Frame with Cell-Voltages and additional data. FozzieUK wrote in his repository "The BMS responds to the inverter polls, but only after it has sent the inverter one copy of the BMS data and all pack data will it close it's contactor and become operational." Maybe somewhere in this step there is a problem.

**Expected behavior**
[...]

PR #2710, opened 2026-07-30
FoxESS battery: refresh CAN keep-alive and stop overwriting BMS-reported voltage limits

ekholm:
Two defects in the FoxESS battery driver, found investigating #1664. Both affect every FoxESS battery user, not just EP-series packs.

### 1. The driver never refreshes the CAN keep-alive counter

`FOXESS-BATTERY.cpp` was the only battery driver with no `CAN_battery_still_alive` refresh anywhere in its receive handler. Consequence: `EVENT_CAN_BATTERY_MISSING` (ERROR level) fires 60 seconds after boot regardless of live traffic → `system_status` goes to FAULT → charge/discharge limits are forced to 0 → the contactor never closes and the inverter sees a dead battery.

Fixed by refreshing the counter in the eight periodic `0x187x` status cases — recognized traffic only, so the existing still-alive test contract (bogus frames must not renew the counter) holds.

### 2. `update_values()` overwrites the BMS-reported voltage limits every cycle

The `0x1872` (BMS_Limits) handler correctly writes the BMS-reported `max/min_design_voltage_dV` — and then `update_values()` re-applies the hardcoded HS-series pack table on the next cycle, clobbering them. The table always wins. Any pack outside the HS table trips a false `EVENT_BATTERY_OVERVOLTAGE` that zeroes charge power: the reporter's EP11 sits at ~398 V against the HS15.6 table's 350.4 V max.

Fixed by making the `0x1872` limits authoritative once received. The table remains the fallback before the first `0x1872` arrives and stays the only source for `number_of_cells`. The `0x1872` power-limit fields are untouched (working today).

### Relation to #1664

These two account for the reporter's "impossible" `CAN_Battery_Missing` and `Battery_Overvoltage` errors. This fixes the BE-side defects in #1664; whether the EP11 then closes contactors on the standard poll sequence remains open (may need EP-specific work), but these fixes are prerequisite to finding out and benefit all FoxESS users.

### Testing

New regression tests (`test/battery/foxess_tests.cpp`): the keep-alive test and the 0x1872-authority test both **fail against the unfixed driver** (verified) and pass with the fix; the limits test also pins the pre-0x1872 table fallback. Full suite 102/102 green; firmware compiles clean under the `-Werror` warning-check environment.

### DIFF
--- a/Software/src/battery/FOXESS-BATTERY.cpp
+++ b/Software/src/battery/FOXESS-BATTERY.cpp
@@ -23,64 +23,81 @@ void FoxessBattery::
 
   datalayer.battery.status.max_charge_power_W = ((datalayer.battery.status.voltage_dV * max_charge_power_dA) / 100);
 
+  // The HS-series pack table. Sole source for number_of_cells; the design
+  // voltage limits from it apply only until the BMS reports its own limits in
+  // 0x1872 (BMS_Limits) - packs outside this table (e.g. EP-series) would
+  // otherwise trip false BATTERY_OVERVOLTAGE against table values.
+  uint8_t cells = 0;
+  uint16_t table_max_design_voltage_dV = 0;
+  uint16_t table_min_design_voltage_dV = 0;
   switch (NUMBER_OF_PACKS) {
     case 1:  //HS2.6 (48V invalid combo for most HV inverters)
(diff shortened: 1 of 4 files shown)


dalathegreat (approved):
👍

PR #2713, opened 2026-07-30
Equipment stop: wait for zero current before opening contactors (bounded)

### LINKED ISSUE #1750: [Feature request] Current drop to 0 before contactor opening
When using the equipment stop function GPIO contactor control opens the contactors immediately. That behaviour can cause contactor welding and is more suitable for emergency stop.

A more graceful stop is suggested to not risk damage of the contactors. This could be achieved by requesting charging/discharging stop and opening contactors when current drop to zero.

A maximum time delay could be added to not risk unexpected behavior. If current doesn’t drop to zero before contactor opening, an error should be entered in log and emulator enter fault state.

ekholm:
Today, equipment stop zeroes the power limits and drops all contactor outputs **in the same 10 ms tick** — faster than any inverter can ramp down, so the contactors open under load (arc/weld risk, the exact scenario described in #1750).

The graceful machinery already exists — v11 applied it everywhere except this path: the pause state machine only reports PAUSED below 1.8 A, the BMS reset waits for zero current with a 10 s give-up ("so we don't weld the contactors"), and graceful restart waits for PAUSED. This PR applies the same pattern to the e-stop path.

### Behavior

- **Equipment stop** now holds the contactors in COMPLETED until the pause state machine reports PAUSED (current < 1.8 A) **or** a 7 s timeout (10 s originally; lowered per review - equipment stop doubles as an emergency stop in practice). Every e-stop entry point (button, MQTT, web) already issues the battery pause, so the PAUSED condition is reached without extra wiring in normal cases.
- **On timeout** it opens anyway and raises `EVENT_ERROR_OPEN_CONTACTOR` (data = 1), so a forced open under load is visible in the event log.
- **Inverter-commanded opening** keeps today's immediate behavior — the inverter has already stopped power transfer before revoking its permission.
- **The latched FAULT shutdown path is untouched** — genuine faults still open immediately.

Fixes #1750 (offer made in the issue).

### Testing

Four new host tests: hold-under-load until the bounded timeout (+ event raised), prompt open when already paused, open as soon as the pause completes mid-wait, and a regression guard that the inverter path stays immediate. The two hold-behavior tests fail against the unfixed code (verified). Full suite 104/104 green; firmware compiles clean under the `-Werror` warning-check environment.

### DIFF
--- a/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
+++ b/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
@@ -35,7 +35,7 @@ const uint8_t OFF = 0;
 #define PRECHARGE_COMPLETED_TIME_MS \
   1000  // After successful precharge, resistor is turned off after this delay (and contactors are economized if PWM enabled)
 #define ESTOP_OPEN_TIMEOUT_MS \
-  10000  // Equipment stop: max time to wait for the pause to reach zero current before opening contactors anyway
+  7000  // Equipment stop: max time to wait for the pause to reach zero current before opening contactors anyway
 uint16_t pwm_frequency = 20000;
 uint16_t pwm_hold_duty = 250;
 #define PWM_ON_DUTY 1023
--- a/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
+++ b/Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
@@ -34,6 +34,8 @@ const uint8_t OFF = 0;
   500  // Time after negative contactor is turned on, to start precharge (not actual precharge time!)
 #define PRECHARGE_COMPLETED_TIME_MS \
   1000  // After successful precharge, resistor is turned off after this delay (and contactors are economized if PWM enabled)
+#define ESTOP_OPEN_TIMEOUT_MS \
+  10000  // Equipment stop: max time to wait for the pause to reach zero current before opening contactors anyway
 uint16_t pwm_frequency = 20000;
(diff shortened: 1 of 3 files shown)


Software/src/communication/contactorcontrol/comm_contactorcontrol.cpp
```diff
@@ -34,6 +34,8 @@ const uint8_t OFF = 0;
   500  // Time after negative contactor is turned on, to start precharge (not actual precharge time!)
 #define PRECHARGE_COMPLETED_TIME_MS \
   1000  // After successful precharge, resistor is turned off after this delay (and contactors are economized if PWM enabled)
+#define ESTOP_OPEN_TIMEOUT_MS \
+  10000  // Equipment stop: max time to wait for the pause to reach zero current before opening contactors anyway
```
dalathegreat: I know equipment stop gets kinda mis-used as emergency stop (even though it is not). Due to this, I would recommend we lower the timeout from 10s -> 7s, just to make it more responsive and still allow for hopefully non-welded contactors

ekholm:
Agreed — 7 s is the better balance for the misuse-as-emergency-stop case. Lowering ESTOP_OPEN_TIMEOUT_MS to 7000 and updating the timeout test; push coming shortly.

dalathegreat (approved):
Good improvement 👍 Thanks!

ekholm:
Done in 4afb0e6d — timeout is 7 s, test asserts still-held at 5 s and opened at 8 s, PR description updated.

PR #2714, opened 2026-07-31
Add virtual destructors, per-driver TYPE statics, and Tesla variant subclasses

ekholm:
Lifecycle hygiene and preparation for a later factory-switch cleanup. **No behavior change anywhere in this PR** — everything is additive or a mechanical rename.

### 1. Virtual destructors on the polymorphic bases

`Battery`, `InverterProtocol`, `Transmitter`, `CanReceiver` are factory-created and held through base pointers, but none had a virtual destructor. This isn't hypothetical: the host test suite **already deletes batteries polymorphically** (`still_alive_tests.cpp` does `delete battery;` through the base pointer in SetUp/TearDown), which is undefined behavior today. The destructors legalize existing practice rather than enable new practice.

### 2. `static constexpr TYPE` on every driver class

Each battery, inverter, charger and shunt class gains a `TYPE` constant alongside its existing `Name` — one line per class, read by nothing yet; it's groundwork for replacing the hand-synced factory/name switches (which currently need 3-4 coordinated edits per new driver) with a single compile-time table in a follow-up. The switches themselves are deliberately untouched here.

### 3. Tesla variant split

Two enum values (`TeslaModel3Y`/`TeslaModelSX`) mapped to one class that read `user_selected_battery_type` deep inside the driver to resolve its own variant. Now `TeslaModel3YBattery` / `TeslaModelSXBattery` subclass `TeslaBattery`, and the variant-dependent tail of `setup()` (pack design limits, reported protocol name) is a pure virtual `apply_variant_config()` implemented per subclass and called at the same point in the sequence — the constructed type *is* the variant, and the global read is gone. The `Name3Y`/`NameSX` special case disappears (each subclass carries its own `Name` and `TYPE`), the factory constructs the right subclass, and the MQTT Tesla type-sniffing becomes a capability query on the instance.

### 4. Tesla contactor alias cleanup (fold-in while the file was open)

`battery_contactor` was a historical alias of `BMS_contactorState`: its own 0x20A parse has been commented out for a long time, and it was assigned verbatim from the 0x212 parse. Deleted, with the five readers switched to `BMS_contactorState` directly. Value 4 meant "closed" and 0 meant SNA under both the old and current encoding — a pure rename.

### Testing

New tests construct both Tesla variants with `user_selected_battery_type` deliberately set to `None` and assert the variant-dependent pack limits and reported protocol name resolve from the constructed type alone. Full host suite 111/111 green; firmware builds clean on the `-Werror` warning-check env plus Stark (ESP32) and LilyGo T-2CAN (ESP32-S3) targets; a sweep confirms every `Name`-carrying driver class has its `TYPE` static.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -188,9 +188,9 @@ const char* name_for_battery_type(BatteryType type) {
     case BatteryType::SimpBms:
       return SimpBmsBattery::Name;
     case BatteryType::TeslaModel3Y:
-      return TeslaBattery::Name3Y;
+      return TeslaModel3YBattery::Name;
     case BatteryType::TeslaModelSX:
-      return TeslaBattery::NameSX;
+      return TeslaModelSXBattery::Name;
     case BatteryType::TeslaLegacy:
       return TeslaLegacyBattery::Name;
     case BatteryType::TestFake:
@@ -315,8 +315,9 @@ Battery* create_battery(BatteryType type) {
     case BatteryType::StellantisSmallWide4x4:
       return new StellantisSmallWide4x4Battery();
     case BatteryType::TeslaModel3Y:
+      return new TeslaModel3YBattery();
     case BatteryType::TeslaModelSX:
-      return new TeslaBattery();
+      return new TeslaModelSXBattery();
     case BatteryType::TeslaLegacy:
       return new TeslaLegacyBattery();
     case BatteryType::TestFake:
@@ -456,8 +457,10 @@ void setup_battery() {
           battery2 = new TestFakeBattery(&datalayer.battery2, can_config.battery_double);
           break;
         case BatteryType::TeslaModel3Y:
+          battery2 = new TeslaModel3YBattery(&datalayer.battery2, can_config.battery_double);
+          break;
         case BatteryType::TeslaModelSX:
-          battery2 = new TeslaBattery(&datalayer.battery2, can_config.battery_double);
+          battery2 = new TeslaModelSXBattery(&datalayer.battery2, can_config.battery_double);
           break;
         default:
           break;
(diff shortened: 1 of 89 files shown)


jonny5532:
9kb on T485 is pretty painful for no obvious benefit - as you note, the destructors are never called on the hardware, we might be better off changing the tests.

What is your plan for "groundwork for replacing the hand-synced factory/name switches"?

ekholm:
Fair challenge — the 9 KB was real and I measured where it went: each of the ~80 polymorphic driver classes gained two emitted destructor bodies (complete + deleting) plus vtable slots, and the multiple-inheritance CAN drivers added per-base vtables and thunks. Agreed it wasn't worth it.

Reworked, and I think you'll like this better: the base classes now declare `protected: ~X() = default;` — deleting through a base pointer becomes a **compile error** instead of UB, which is the same correctness guarantee pointed the other way. Tests updated to delete concrete types. **The PR now costs under 1 KB on every board** (the remainder is the one new capability virtual) — see the refreshed size report above.

On the groundwork question: adding a battery today touches half a dozen hand-synced lists — the factory switch, name maps, settings-page id lists, NVM key wiring. The per-driver TYPE statics here are the first step toward drivers describing themselves, so those parallel switches collapse into one table the factory and webserver iterate. That direction is unchanged by the destructor rework.

ekholm:
Nudging this one since it's been quiet since the rework: the virtual destructors you flagged are gone, replaced with protected non-virtual ones — deleting through a base is now a compile error rather than UB, and the cost dropped from +9.5 KB to **+620 B** on the warning-check env (the remainder is the one new capability virtual). The refreshed size report above has the per-board numbers.

Rebased onto current main and green. Happy to split the Tesla-variant subclasses out into their own PR if that makes this easier to review — the TYPE statics and the destructor change are the part the rest of the driver work builds on.

PR #2715, opened 2026-07-31
Add multi-language support: catalog storage, endpoints, and Languages UI

ekholm:
## What

Adds the storage and UI plumbing for language support — no strings are translated by this PR, and a stock device behaves exactly as today.

- **Catalog storage on the existing `spiffs` partition** — the partition every board already has, mounted by nothing today. Storage is a small raw slot store (~4 KB of code) directly on `esp_partition_*`: a double-buffered, CRC'd, sequence-numbered directory plus data extents. Every mutation commits by flipping to the older directory block, so power loss at any point — including mid-upload — leaves the previous state fully intact. No filesystem driver: linking LittleFS costs ~80 KB of flash, which the 4 MB boards' OTA slot cannot absorb; this feature needs at most a handful of rarely-written blobs.
- **`LANGUAGE` setting** (NVS-backed, empty = English) with a live `/updateLanguage` route.
- **`/api/i18n` endpoints**: list (with active language), serve (gzip catalogs and raw packs, with ETag/If-None-Match revalidation), validated atomic upload (128 KB cap, atomic via staging + directory flip), delete, and an explicit format action (the store is never auto-formatted).
- **Languages panel** on the settings page (between Hardware config and Integration settings): selector, drag-drop upload, delete, format.

## Why the partition is safe to take

Nothing in the tree mounts or writes the `spiffs` partition today, on any board. No partition-table change, so it is OTA-safe fleet-wide; an unformatted partition simply means the feature is off until a user explicitly formats it.

## Flash cost

Under 10 KB once (measured on the warning-check env: +9.2 KB). The companion PR (string conversion) more than pays it back — the pair nets **about −20 KB** vs current main; numbers there.

## Testing

- Host suite: slot-store power-loss-mid-upload, replace-then-power-loss, corrupted-directory recovery, compaction, full-store rejection, filename validation.
- Hardware-verified on an 8 MB board: format on a dirty partition, upload/list/serve/delete round-trips, invalid-name and oversize rejection, language selection persisting across hard resets, catalogs surviving app reflash, ETag 304 revalidation, and graceful boot with the feature off on an untouched partition.

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -18,6 +18,7 @@
 #include "src/datalayer/datalayer.h"
 #include "src/devboard/display/display.h"
 #include "src/devboard/espnow/espnow.h"
+#include "src/devboard/i18n/i18n.h"
 #include "src/devboard/mqtt/mqtt.h"
 #include "src/devboard/safety/parallel_safety.h"
 #include "src/devboard/sdcard/sdcard.h"
@@ -730,6 +731,8 @@ void setup() {
   init_events();
 
   init_stored_settings();
+
+  init_i18n_storage();  // Language catalogs; mount failure just disables the feature
 
   // AP-button recovery must always run
   xTaskCreatePinnedToCore((TaskFunction_t)&connectivity_loop, "connectivity_loop", 4096, NULL, TASK_CONNECTIVITY_PRIO,
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -18,6 +18,7 @@
 #include "src/datalayer/datalayer.h"
 #include "src/devboard/display/display.h"
 #include "src/devboard/espnow/espnow.h"
+#include "src/devboard/i18n/i18n.h"
 #include "src/devboard/mqtt/mqtt.h"
 #include "src/devboard/safety/parallel_safety.h"
 #include "src/devboard/sdcard/sdcard.h"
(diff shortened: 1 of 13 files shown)


ngehrsitz:
What exactly is the advantage of storing the strings separate from the code? Is it that you can then compress them?
Because if that´s the advantage then I think we are better off with one of the two frontend rewrites(#1544/#2215) that also move all the strings out of the backend.

ekholm:
Compression is a side benefit — the point is languages without firmware builds. The catalogs are plain keyed JSON served from `/api/i18n/<lang>.json`; a language is a data file a user installs on the device, not something baked per-release, and the storage is the currently idle spiffs partition, so the 4 MB boards get languages they could never fit in flash. That's also why this isn't in competition with #1544/#2215: the rewrites relocate the ENGLISH strings to the frontend, but they don't give you 8 translated languages, per-device packs, or storage for them — a keyed catalog stack is exactly what an SPA would consume natively (fetch the JSON, look up by key). If/when a rewrite lands, the server-rendered page shims die, and the catalogs, endpoint, and upload path carry over unchanged.

ngehrsitz:
I think the approach here for uploadable language packs is the correct one, but I am unsure if it´s worth the complexity as long as we don´t have the new frontend architecture. What do others think?

PR #2716, opened 2026-07-31
Convert firmware strings to language catalogs: 8 languages, net flash reduction

ekholm:
## What

Converts the firmware's user-visible strings to language-catalog lookups, building on the storage stack PR #2715 (stacked on it — its commits appear here until it merges). With this pair, BE speaks **eight languages** at launch: English, German, French, Dutch, Hungarian, Swedish, Finnish, Ukrainian — and adding another is one spreadsheet column, no code.

- **Generator** (`tools/i18n_gen.py`): one spreadsheet (`tools/translations.fods`) is the single source of truth; the generator emits the key enum, the compiled bootstrap table, and per-language pack files. CI regenerates and diffs, and enforces the append-only ID sidecar. Generated artifacts are not committed.
- **`TR()` runtime**: index cached at boot (~8 KB), strings seek-read from the store. Dense packs resolve by direct index, holed packs by binary search. Fallback chain: active pack → English pack → compiled bootstrap → visible `[T####]` marker.
- **Converted**: all event messages, the web UI (status page, settings, cellmonitor, events, CAN log/replay, advanced pages, shared DTC section), all battery-driver info pages, and MQTT discovery **display names only**.
- **Bootstrap** (26 keys, ~1.4 KB compiled in): WiFi provisioning, the Languages panel, and save/reboot — a device with an empty store can always be provisioned and given a language.
- **Factory image builder**: devices ship with English pre-installed (plus a preferred-language hint the installer can set); OTA upgraders from older firmware do a one-time pack upload from the Languages panel.

## Safety rules (enforced, not conventions)

- Translated text never passes through printf-style formatting: substitution is literal token replacement (`{0}`/`{1}`), and CI rejects format specifiers in any translation.
- MQTT **entity identity is untouched**: only the discovery `name` field is translated; `unique_id`, `object_id`, topics, `device_class`, `state_class`, and state payload values stay machine vocabulary. (Home Assistant shows updated names on the next discovery publish.)
- Logs stay English. CAN interface names ("CAN-FD Native", "MCP2515", …) are treated as technical designations and deliberately untranslated, consistent with the selector options.

## Size (warning-check env)

| | bytes | delta |
|---|---|---|
| main (pre-i18n baseline) | 1,951,184 | — |
| + #2715 | 1,960,404 | +9.2 KB |
| + this PR | **1,916,228** | **−34.1 KB vs baseline** |

The conversion alone frees ~43 KB (string removal plus collapsing per-site render chains into shared helpers). Beyond the one-time win: **string growth stops costing flash** — new drivers, events, and texts land in pack files on the partition, not in the binary.

## Relation to #1544

The mechanism is frontend-agnostic: keyed catalogs with numeric IDs. The web UI rework can consume the same packs; nothing here assumes the current page structure.

## Testing

137 host checks green (store, TR/expand, generator CI rules, ETag source). Hardware-verified on an 8 MB board: translated rendering in all eight languages including Cyrillic, language switching, missing-pack fallback to English, wiped-store recovery via bootstrap + markers, upload restore, 304 revalidation.


ekholm:
Regarding maintainability, 
I have been running the same pattern for years in a production environment, that is; a master sheet of keys and strings, parsed into generated language files consumed by different front ends (desktop and web).
A translator only needs to update one cell in the sheet. This is the low-maintenance route, not the high one.

nagyrobi:
This has to be considered also from support point of view. 

Members here do this work in free time. Users generating support issues posting questions and UI screenshots in various languages will increase support load, and decrease success rate from the user's perspective (will receive help likely slower if not asking in English).

I highly appreciate the effort put into this, but I still don't see the real benefit from the entire project's point of view.

nagyrobi:
UI changes require adjusting translations constantly. Proper regionalized strings need active maintainers with native-speaking engineering languages, as AI-hallucinated technical terms look unserious. AI translation from English to other languages is pretty good for casual talk and flowing context. Technical literature is a different kind of story, and context needs to be derived from functionality, rather than just guesswork.

All this generates us more and more work in the future.

ekholm:
You are absolutely right to rise the concern.

It would be a marginal problem to have a TR described in one language long with a screen shot in the same language. The TR can easialy be translated with for instance google, and a screen shot will have the structure regardless of what language the user has selected.
It is easy to set up a automated job that translate TRs in non english languages into english. I really think that a poor automated translation beats a TR written in uncomprehensible english. But I think it boils down to what users we want to include/exclude into the spirit of BE. 

Yes, blindly trusting AI will bite back, trust me ;)
But those language that I used AI translate into have active community member as their native language (I hope there are some native speaking finnish guys here, I bet @dalathegreat it native swedish speaking :)
The current traslation list is 1300+ strings, it's a one time sweep to proof read them. and I suggest that we wait for the new UI to land and proof read those strings. This PoC might anyway die sooner than later, but show cases how to save flash and provide multi language support. 
In the projects I have been involded in that uses the same pattern, and with a constant evolving UI, the delta over changed strings that need translations are suprisingly small. In other words is it a small set of strings that would need corrections (after an initial automated translation). 

Disclaimer: This text written all by me...

PR #2718, opened 2026-08-01
Improvement: Source BYD power limits and temperatures from CAN broadcasts

demon1300:
### What
Sources allowed charge/discharge power and pack temperatures from the BMS broadcasts 0x345 and 0x447 instead of UDS polling. Drops 8 poll slots (SOC, voltage, current, both power limits, three temperature/sensor DIDs), so the rotation goes from 16 targets to 8 — a full cycle every 1.6 s instead of 3.2 s.

### Why
The data was never the problem, the poll rate was. Power limits arrived every 3.2 s over UDS; 0x345 carries the same values every ~100 ms.

I validated this against live captures from my own pack. The discharge word matches DID 0x000E exactly, 28/28 samples, and I caught it mid-ramp going 2053 → 2111 over 600 ms — movement the poll cycle would have flattened entirely. Charge power matches DID 0x000A, r = 1.0000 across 898 pairs.

At 3.2 s the inverter is acting on an allowance the BMS has already pulled. That's exactly what I see at top of charge when the limit collapses.

### How
Both frames are checksum-gated on the existing BYD inverted-sum byte before anything is accepted. A bad checksum bumps CAN_error_counter and leaves the previous values in place.

Otherwise this is a straight swap of the source: the values land in the same variables the polls used to fill, and everything downstream is untouched. The power gate and the temperature guard in update_values() are unchanged from main, so behaviour on signal loss is exactly as it was on the polled path.

3 unit tests: decode and checksum rejection for each of 0x345, 0x447 and 0x444.

**Not included**
real_soc is unchanged, still on 0x447. The two SOC bases do differ by a few percent mid-pack and I'm not yet confident which is the better source, so I've left it alone rather than change it blind.

Testing: the decodes are validated against live captures from my own Atto 3, and the driver builds and runs on hardware.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -174,9 +174,16 @@ void BydAttoBattery::
   }
   // End taper
 
-  // Hold power at zero until the pack confirms closed (0x344 bit7), and while opening/idle
+  // Requires a closed pack and both sourcing broadcasts fresh; a lost signal must not leave an
+  // allowance standing.
+  const uint32_t now_millis = millis();
+  const bool power_limits_stale =
+      !powerLimitFrameReceived || (now_millis - lastPowerLimitFrameMillis > POWER_LIMIT_STALE_MS);
+  const bool temperatures_stale =
+      !temperatureFrameReceived || (now_millis - lastTemperatureFrameMillis > TEMPERATURE_STALE_MS);
   if (!(contactor_feedback & BMS_FEEDBACK_MAIN_CLOSED) ||
-      (contactorState != CONTACTORS_CLOSING && contactorState != CONTACTORS_ACTIVE)) {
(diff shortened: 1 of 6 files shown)


Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -174,9 +174,16 @@ void BydAttoBattery::
   }
   // End taper
 
-  // Hold power at zero until the pack confirms closed (0x344 bit7), and while opening/idle
+  // Requires a closed pack and both sourcing broadcasts fresh; a lost signal must not leave an
```
dalathegreat: Do we really need a stale check? No other battery checks for this (except BMW iX for some reason)

demon1300: Yeah looking at it again, it was excessive.  check_can_component_alive already covers total loss...  Removed it now!

dalathegreat (approved):

PR #2722, opened 2026-08-01
Add DFRobot Edge101 support

ngehrsitz:
### What
This adds support for the [DFRobot Edge101](https://www.dfrobot.com/product-2934.html) board that I am running for two weeks now.

Schematic: https://dfimg.dfrobot.com/wiki/23038/DFR0886_edge101-esp32_schematics_1.0.pdf

### Why

- isolated CAN
- isolated RS485
- Ethernet (Will be included in a follow-up PR for easier review)
- External antenna
- PCIe expansion slot with UART
- Case with DIN Rail mount (that fits perfectly inside of a Fronius Verto) dfrobot101_side
- External IO header (where you could attach an external MCP2518FD)
- SD Card slot
- 16MB Flash
- RTC (not configured since this project only uses runtime)
- Good availability: https://octopart.com/de/part/dfrobot/DFR0886
- Affordable (I bought two from Digikey for 80€ including shipping)

This is one the best options with ethernet that I could find. The other notable alternatives are the LILYGO T-Connect Pro(#2234) and the [Waveshare ESP32-S3-POE-ETH-8DI-8DO](https://www.waveshare.com/ESP32-S3-POE-ETH-8DI-8DO.htm). But the Waveshare would only barely fit inside of the Verto even if you cut off the sides of the case. The LILYGO is more expensive and has even more features but is not as widely available. The main downsides of this board is that it´s not an S3.

### How
~~Unfortunately adding SPI SD support also increases the binary size of the other boards until we can restore LTO from #2650~~

> [!TIP]

### DIFF
--- a/.github/workflows/compile-common-image.yml
+++ b/.github/workflows/compile-common-image.yml
@@ -35,6 +35,8 @@ jobs:
             pio_env: "BECom_330"
           - board_name: "Waveshare ESP32S3RS485CAN"
             pio_env: "waveshare_330"
+          - board_name: "DFRobot Edge101"
+            pio_env: "dfrobot_edge101_330"
           - board_name: "Compiler warning check"
             pio_env: "compiler_warning_check"
 
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -34,7 +34,7 @@
 #include "src/inverter/INVERTERS.h"
 
 #if !defined(HW_LILYGO) && !defined(HW_LILYGO2CAN) && !defined(HW_STARK) && !defined(HW_3LB) && !defined(HW_BECOM) && \
-    !defined(HW_WAVESHARE) && !defined(HW_DEVKIT)
+    !defined(HW_WAVESHARE) && !defined(HW_DEVKIT) && !defined(HW_DFROBOT_EDGE101)
 #error You must select a target hardware!
 #endif
 
--- a/Software/src/devboard/hal/hal.cpp
+++ b/Software/src/devboard/hal/hal.cpp
@@ -26,6 +26,9 @@ void init_hal() {
 #elif defined(HW_DEVKIT)
 #include "hw_devkit.h"
   esp32hal = new DevKitHal();
+#elif defined(HW_DFROBOT_EDGE101)
+#include "hw_dfrobot_edge101.h"
+  esp32hal = new DFRobotEdge101Hal();
 #else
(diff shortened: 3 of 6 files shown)


jonny5532:
Interesting! Surprisingly cheap, I can see why they went with the ESP32 and not the ESP32S3 (since the original has a built in ethernet MAC which avoids the fiddlier SPI-based ones). Has 16mb flash though so no big problem.

CAN/RS485 is isolated, which is nice, although it shares the same isolation domain like the Waveshare (would rather have the CAN isolated only...). CAN inverter folk would need a MCP2518FD module and some duponts (but no soldering required).

I'm a bit surprised that it isn't possible to keep the SPI SD code out of the T485? even with `#define` hackery?

nagyrobi:
Very nice!
Is Ethernet port also isolated?

Another reason to consider a unified build…

jonny5532:
Although if you went unified-build, this (along with Stark) wouldn't use it:

- ESP32 4mb (T485, devkit)
- ESP32 8mb (Stark only)
- ESP32 16mb (DFRobot only)
- ESP32S3 16mb (T-2CANs, Becom, Waveshare)

(although there is scope for rearranging things so they build the same `.elf` but then split off the firmware binaries after that... but we'd probably still keep the flash-constrained 4mb boards separate)

nagyrobi:
I was just thinking. Agree that a unified build could only be realistic for ESP32S3 atm.

ngehrsitz:
#2746 should be merged first. I also have an SD refactoring PR prepared based on top of that which allows us to reduce flash size instead of adding it like in the current version.

ngehrsitz:
I rebased and now the flash increase should be gone. The reason it is missing here is that due to the Github Actions outage the build on `main` did not run.
I will also test with an external MCP2518 once that board arrives.

dalathegreat:
Nice work @ngehrsitz , let's merge this into main :raised_hands: 

I have created a Wiki page for t his hardware, could you add some pics/info to it when you have time? You are now the DFRobot expert :joy: https://github.com/dalathegreat/Battery-Emulator/wiki/Hardware:-DFRobot-Edge101

dalathegreat (approved):

ngehrsitz:
I made some edits. I am also working on the Integration of the MCP2518FD board that has arrived and finalizing the ethernet support.

PR #2723, opened 2026-08-02
Fix Stop V2X button performing a restart instead of a stop

ekholm:
During a code review this surfaced: the `chademoStop` entry in `battery_commands` uses `supports_chademo_restart()` as its condition and `chademo_restart()` as its action — both halves point at restart, so the "Stop" button (confirm text "stop V2X?") restarts the session instead. `supports_chademo_stop()` / `chademo_stop()` exist and are implemented in the CHAdeMO driver, but are referenced nowhere.

This is more than a mislabeled button: restart jumps the state machine straight to `CHADEMO_IDLE`, while the unreachable `CHADEMO_STOP` state is the orderly teardown — stop-control flag to the vehicle, permissions revoked, and pin 10 held until current ≤5 A / voltage ≤10 V, per the IEEE A.7.2.9 EV-contactor-protection note in that code. The mis-wire makes the protective wind-down unreachable from the UI.

Fix: wire the command to the stop predicate and action.


dalathegreat:
Chademo does not function, the people running successful branches are on v8, and these have not been merged to main. not needed to fix this!

PR #2724, opened 2026-08-02
Stop overwriting Tesla GTW driver defaults with zeros on fresh installs

ekholm:
During a code review this surfaced: `init_stored_settings()` loads the five GTW keys with fallbacks `0/false/0/0/0`, unconditionally overwriting the driver's intended defaults (`BATTERIES.cpp`: country `17477` = ASCII "DE", right-hand-drive `true`, mapRegion `2`, chassisType `2`, packEnergy `1` = 74 kWh). The values are written raw into the emulated gateway's 0x7FF frame with no 0-means-skip semantics, so any install that never saved these settings broadcasts country 0 / 50 kWh / chassis 0 — and the driver-side initializers are effectively dead code.

Fix: use the current global value as the `getUInt`/`getBool` fallback, so the driver's static initializer remains the single source of defaults. Installs with stored settings are unaffected (stored keys win, as before), and the settings page now renders the intended defaults instead of zeros.

Also checked the neighbouring driver-setting loads for the same pattern (Daly derating, CT clamp calibration, LEAF interlock, BYD auto-cal): their core fallbacks already match the driver defaults, so nothing else was touched.

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -141,11 +141,17 @@ void init_stored_settings() {
   user_selected_daly_power_at_0_degree_C = settings.getUInt("DALYPWR0C", 800);
   user_selected_use_estimated_SOC = settings.getBool("SOCESTIMATED", false);
   user_selected_tesla_digital_HVIL = settings.getBool("DIGITALHVIL", false);
-  user_selected_tesla_GTW_country = settings.getUInt("GTWCOUNTRY", 0);
-  user_selected_tesla_GTW_rightHandDrive = settings.getBool("GTWRHD", false);
-  user_selected_tesla_GTW_mapRegion = settings.getUInt("GTWMAPREG", 0);
-  user_selected_tesla_GTW_chassisType = settings.getUInt("GTWCHASSIS", 0);
-  user_selected_tesla_GTW_packEnergy = settings.getUInt("GTWPACK", 0);
+  // Fall back to the value the global already holds, which is the driver's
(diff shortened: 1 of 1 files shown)


Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -141,11 +141,17 @@ void init_stored_settings() {
   user_selected_daly_power_at_0_degree_C = settings.getUInt("DALYPWR0C", 800);
   user_selected_use_estimated_SOC = settings.getBool("SOCESTIMATED", false);
   user_selected_tesla_digital_HVIL = settings.getBool("DIGITALHVIL", false);
-  user_selected_tesla_GTW_country = settings.getUInt("GTWCOUNTRY", 0);
-  user_selected_tesla_GTW_rightHandDrive = settings.getBool("GTWRHD", false);
-  user_selected_tesla_GTW_mapRegion = settings.getUInt("GTWMAPREG", 0);
-  user_selected_tesla_GTW_chassisType = settings.getUInt("GTWCHASSIS", 0);
-  user_selected_tesla_GTW_packEnergy = settings.getUInt("GTWPACK", 0);
+  // Fall back to the value the global already holds, which is the driver's
```
dalathegreat: Can you remove the comment? Bit unnecessary 6 row beheemoth 😅

dalathegreat (changes requested):

Software/src/communication/nvm/comm_nvm.cpp
ekholm: done👍

dalathegreat (approved):

PR #2725, opened 2026-08-02
Fix parallel-battery checks being skipped while a pack reads exactly 370.0 V

ekholm:
During a code review this surfaced: `check_parallel_battery_safety()` returns early when either pack reads exactly 3700 dV, because that is the datalayer init default. But 370.0 V is a legitimate operating voltage, and the early return skips the code that manages `battery2/3_allowed_contactor_closing` — in both directions: a main pack genuinely sitting at 370.0 V blocks the second battery from ever being granted closing permission, and packs already joined that drift apart while one crosses exactly 3700 never reach the 10-second disengage. The SOH-deviation check has the same pattern at exactly 9900 pptt (a pack whose real SOH is 99.00 % silently loses the check).

Fix: treat the sentinel as a startup grace rather than a continuous condition. Each check latches once its pack pair has been seen with both values off-sentinel; from then on the check runs permanently, including through later sentinel crossings. A pack that never leaves the init value behaves exactly as today (fail-closed, no permission granted).

Regression test added: `VoltageSyncTest.Battery2DisengagesWhenMainSitsAtTheSentinelVoltage` — latch on a valid pair, drift >1.5 V apart while the main pack reads exactly 3700, assert the 10 s disengage still fires. Verified to fail against the previous code.


nagyrobi:
This appears with fake triple

ekholm:
Thanks for testing — I couldn't reproduce it, so rather than guess I wrote the configuration into the test suite. Rebased onto current main at the same time.

The new cases reproduce a fake triple the way the firmware actually runs it: the loop order from `Software.cpp`, where each secondary instance's `update_values()` is followed immediately by its own `check_parallel_battery_safety()`, and the mirror in `TEST-FAKE-BATTERY.cpp`, where instances 2 and 3 take battery 1's pack voltage on every update. That mirror lands in the same tick as the check that reads it, so the difference the check sees is zero at every step.

Five cases: resting at the default 370.0 V; a full sweep of the user-set voltage from 245 V to 404 V and back; passing through exactly 370.0 V after the latch has engaged, which is the case this PR changes; one tick of lag between an instance and battery 1; and an instance that stops tracking altogether.

The useful part is the comparison rather than the cases themselves — I ran all five against the previous continuous-skip behaviour as well, and they behave identically. For a fake triple this PR is a no-op. The PR's own regression case does fail against the old behaviour, as it should.

The fifth case is the discriminating one: an instance frozen 20 V away from battery 1 stays silent for three seconds and raises the event on the fourth. So if a difference event is appearing on a fake triple, it points at instance 2 or 3 not mirroring battery 1 the way the fake driver intends — which I'd want to understand before this merges, because that would be a bug in the fake setup rather than in this check. Could you tell me the exact event text from the events page, and what the three pack voltages read when it appears?

One thing I noticed while writing these that is easy to mistake for this PR: at the default fake voltage all three instances read exactly 370.0 V, so the startup grace never lifts and batteries 2 and 3 are never permitted to close their contactors. That is equally true before this change — the old code took the same early return — but it is the kind of thing you'd notice while testing a fake triple.

The branch also carries a small commit moving the latch and the drift counters from function-local statics to file scope, with a reset behind `UNIT_TEST`. Firmware behaviour is unchanged; without it no test can start from the module's boot state, because the first case to latch it latches it for the whole run.

PR #2726, opened 2026-08-02
Add "Inverter is offgrid" setting that downgrades grid-presence faults to warnings

### LINKED ISSUE #2691: [BUG] MEB + SMA: Manual battery start via ENABLE not possible when CAN_INVERTER_MISSING
**Describe the bug**
My setup is also build to be functional in backup scenarios.

use case:
- Grid fails at night (no solar power)
- - Inverter is without power
- - -> no communication SMA SE <-> BE
- - -> SMA ENABLE disabled
- - BE still alive (Backup battery on 12V)
- - MEB Battery off, as ENABLE disabled

In this case, I need to do a 'black start'. 
- Switch House from Grid to Backup
- Set ENABLE manually to 12V ('backup mode switch')
- - Battery PreCharge will start and contactors will close
- - SMA SE is now powered by battery
- - SMA SE will set ENABLE after a while
- - House is running on backup.

This is working as long as CAN_INVERTER_MISSING is not set.
When CAN_INVERTER_MISSING is set, PreCharge contactor will rattle. No PreCharge is happening. Cannot startup system.

**Expected behavior**
Ignore CAN_INVERTER_MISSING

- Software version: V12.0.0; Same with V11.1.0
- Battery used: MEB
- Inverter communication protocol: SMA - BYD HVS
- Hardware used for Battery-Emulator: Waveshare
- MQTT: yes

```
       0.395 Wi-Fi Got IP. IP address: 192.168.1.97
       0.395 Bootup complete, running version 12.0.0
       0.397 mDNS responder started.
       0.867 MQTT client started, connecting to broker...
       0.881 Event: MQTT connected.
       1.002 Event: Successfully communicating with battery. Battery detected!
      11.101 Battery pause begin 0 0 0 0
      11.101 Event: The emulator is attempting to resume battery operation from pause.
      11.102 MEB: BMS status 1 -> 0
      11.102 MEB: Voltage dV 3700 -> 3827
      11.308 MEB: First CAN msg received
      11.317 MEB: BMS status 0 -> 1
      11.401 MEB: Start precharging 0 -> 1
      12.000 MEB: Voltage dV 3827 -> 3830
      14.000 MEB: Voltage dV 3830 -> 3827
     ...
	 ...
     296.000 MEB: Voltage dV 3830 -> 3827
     298.000 MEB: Voltage dV 3827 -> 3830
     300.001 MEB: Voltage dV 3830 -> 3827
     300.263 AP provisioning window expired (factory-default AP password), disabling access point.
     300.267 Event: Wifi AP disabled due to cybersecurity concern. Change default password to keep AP constantly on! Reboot/Hold BOOT butto
	 
CAN_INVERTER_MISSING event
	 
     301.001 Event: Inverter not sending messages via CAN for the last 60 seconds. Check wiring!
     303.001 MEB: Voltage dV 3827 -> 3830
     305.001 MEB: Voltage dV 3830 -> 3827
     ...
	 ...
     414.008 MEB: Voltage dV 3827 -> 3830
     416.008 MEB: Voltage dV 3830 -> 3827
	 
Backup-Mode switch turned on -> ENABLE = 12V
	 
     416.995 Precharge: Starting sequence
     417.005 Precharge: Disabling Precharge bms not standby/active or equipment stop
     417.008 MEB: Voltage dV 3827 -> 3830
     417.025 Precharge: Starting sequence
     417.035 Precharge: Disabling Precharge bms not standby/active or equipment stop
     417.055 Precharge: Starting sequence
     417.065 Precharge: Disabling Precharge bms not standby/active or equipment stop
     ...
	 ...
     418.535 Precharge: Disabling Precharge bms not standby/active or equipment stop
     418.555 Precharge: Starting sequence
     418.565 Precharge: Disabling Precharge bms not standby/active or equipment stop
     418.585 Precharge: Starting sequence
     418.595 Precharge: Disabling Precharge bms not standby/active or equipment stop
	 
Backup-Mode switch turned off -> ENABLE = 0V

     419.008 MEB: BMS Voltage intermediate dV 910 -> 0
     421.008 MEB: Voltage dV 3830 -> 3827
     423.008 MEB: Voltage dV 3827 -> 3830
     ...
```

ekholm:
Implementation of the offgrid checkbox proposed in #2691, kept minimal so it can be merged as-is or adjusted freely:

- New setting "Inverter is offgrid (downgrade grid-presence faults to warnings)" on the settings page (`INVOFFGRID`), default off, shown for every inverter.
- When set, events on a small allowlist — initially just `EVENT_CAN_INVERTER_MISSING` — are recorded at WARNING instead of ERROR. The downgrade goes through one `effective_level()` helper at every point the level is read (aggregation, level string, syslog severity, recompute), so storage, aggregation and display all agree; the static severity table in `init_events()` is untouched, and extending the allowlist is one line per event.
- Effect: with the checkbox set and an unpowered inverter, inverter absence no longer drives `system_status` to FAULT, so battery-side precharge can proceed and black start works (this also covers the UPS / night-outage scenarios from the thread). With the checkbox off, behavior is exactly today's.
- Tests: grid-tied → ERROR; offgrid → WARNING and the aggregate stays below FAULT (the property that actually unblocks precharge); an unlisted event is not downgraded. The offgrid case is verified to fail without the fix.

Fixes #2691


Webwanze:
@ekholm 

Did some tests:
When new mode is activated, the error is shown as a warning and the battery will start when ENABLE is activated.
Therefore does what it should!

**First finding:**
Checkbox on configuration page is missaligned:
image

**Second finding / additional wish**
When in normal mode, BE behaves like before.
It starts precharging - stopps it again. Contactor will start to rattle.
We should prevent precharge from starting in this case. It harms the contactor and might also generate high voltage (peaks).

nagyrobi:
I think this should not be generic, just for the SMA-type, which exhibits this behavior.

Webwanze:
Independent from the battery/inverter:
Why should we start precharging to end it a few milliseconds after?

nagyrobi:
The principle is also a bit problematic imho. I have to enter the sertings web interface, a enable a checkbox, during a full power outage, to be able to black start. I may not have the WiFi up, nor a charged phone, or only my wife's at home.

It should be either automatic, or accomplished with a physical button or switch, and not need any kind of playing with the settings in such a stressful situation.

ekholm:
@Webwanze, I create a separate RP for the contactor rattle, it a bug in the contactor handling logic. 
And the spotted layout issues, fix come soon.

@nagyrobi, I believe that the use case was for a off-grid system and not a measure to handle a sudden black out.

ekholm:
Thanks for putting it on real hardware — good to know the ENABLE path does what it should.

**Alignment:** my fault and easy to fix. The rows in that settings section are label/control pairs dropped straight into a CSS grid; the conditional rows wrap in a `<div class="if-...">` and only survive because those classes are styled `display: contents`. Mine copied the wrapper without the class, so the whole row landed in the first column. I will drop the wrapper — the row is unconditional and does not need one — and rename the label to "Inverter run entirely offgrid" as suggested in #2691.

**Precharge starting and stopping:** you are right that it should not do that, and it turns out to be independent of this PR. In `handle_precharge_control()` the idle state starts a sequence when the battery requests precharge and the inverter permits it, but the precharging state aborts back to idle when the system is not ACTIVE, when an equipment stop is active, or when the BMS is not standby/active. The start gate does not test the first two, so whenever one of those is the blocker the sequence starts, is aborted on the next tick, and starts again — at the 10 ms loop cadence, which is the rattle you hear. The same asymmetry applies while an equipment stop is active. Since it is not caused by this PR I put it up separately as #2756.

@nagyrobi on your two points: the setting is persistent and meant to be enabled once when commissioning an offgrid installation, not toggled during the outage — the web UI is needed at setup time, not in the stressful moment. And it follows the shape proposed in #2691 (a checkbox that downgrades grid-presence faults), which is why it is protocol-agnostic rather than SMA-only; if it should be narrowed to specific inverters, that is worth settling in that issue. A physical switch or automatic detection is a good direction and does not conflict with this — the same downgrade could later be driven by a GPIO input the way the equipment stop is, without changing the event handling.

ekholm:
The alignment fix and label rename are pushed (46702bb1) — the wrapper `<div>` is gone, so the label and checkbox occupy their own grid columns, and the label now reads "Inverter run entirely offgrid". Verified on a Stark CMR: the row renders correctly and the setting persists as expected.

dalathegreat (approved):
Good base to build further on when dealing with offgrid systems 👍

PR #2727, opened 2026-08-02
Add a favicon to the web UI

ekholm:
The web UI serves no favicon, so browsers fall back to a blank tab icon and request `/favicon.ico` on every page load.

This embeds a small battery glyph as an inline SVG data-URI `<link>` in the shared page header (`INDEX_HTML_HEADER`), so it:

- applies to every page, including the login screen,
- needs no extra route or auth handling,
- works offline (no external request),
- costs about 1 KB of flash.

The data URI is base64-encoded rather than percent-encoded on purpose: these pages are rendered through the `%VAR%` template processor, which would consume a percent-encoded (`%3C...`) data URI.

Verified on hardware (`stark_330`): the favicon link is served intact and renders in the browser tab.

### DIFF
--- a/Software/src/devboard/webserver/favicon.h
+++ b/Software/src/devboard/webserver/favicon.h
@@ -1,0 +1,9 @@
+#ifndef FAVICON_H
+#define FAVICON_H
+
+// Battery glyph served from /favicon.svg. Stored once as raw SVG; the pages
+// only carry a link tag to it (see INDEX_HTML_FAVICON_LINK in index_html.h).
+#define FAVICON_SVG \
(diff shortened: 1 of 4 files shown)


dalathegreat:
Can you feature gate this so small flash devices don't get it?

ngehrsitz:
Would it be smaller if we just use the 🔋 emoji: https://stackoverflow.com/a/61238584?

ekholm:
I know, but I wanted something unique. The emoij trick is also different depending on which platform renders it. 
There are two possibilities, one that about halves the size using the the same icon, and the you opted for, using the emoij, which about quarter the size.
 I can fix either one, but we really don't have a flash size problem, stay tuned ;)

ekholm:
Done, both points in one rework: the icon is now served once as `/favicon.svg` (long cache lifetime, so browsers fetch it once) with only a link tag in the page header — and it's gated off `SMALL_FLASH_DEVICE` entirely, so the small-flash boards pay zero. The boards that keep the icon drop from ~2.6 KB to ~1 KB.

dalathegreat (approved):
Good compromise 👍

PR #2734, opened 2026-08-02
Nissan LEAF: serialise UDS requests and use BS=0 flow control for DTC reads

nagyrobi:
### What
Three fixes to the LEAF DTC readout added in #2681: it could be lost entirely when it collided with the periodic group polling, an oversized reply was abandoned part-way leaving the LBC stranded mid-transfer, and a truncated list was shown as if complete. Adds one generic datalayer field so any battery can report a truncated DTC list.

### Why
Captured on hardware: `03 19 02 0E` went out 1 ms after a flow control, in the middle of a group 0x90 transfer, and the LBC silently discarded it — no `59 02` ever came back. A later capture announced 599 bytes (149 codes); we filled our 32-code buffer at frame 18 of 86 and stopped acknowledging, leaving the LBC waiting on a flow control that never arrived.

### How
`uds_busy` tracks our outstanding request on 0x79B/0x7BB and no new one goes out until the previous answer is complete, error or good; a quiet-channel check and an SOH-clear guard cover the rest. Oversized replies are now drained to the end, storing the first 32 and reporting the true count ("32 codes shown of 149 reported"). Status mask stays at `0x0E`.

> [!TIP]

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -383,7 +383,8 @@ void NissanLeafBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
           if (uds_rx_remaining == 0) {
             uds_busy = false;
           }
-        } else if (rx_pci == 0x20) {  //Consecutive frame: up to seven more payload bytes
+        } else if (rx_pci == 0x20) {      //Consecutive frame: up to seven more payload bytes
+          uds_request_millis = millis();  //Still answering, so restart the no-answer timer
           uds_rx_remaining = (uds_rx_remaining > 7) ? (uds_rx_remaining - 7) : 0;
           if (uds_rx_remaining == 0) {
             uds_busy = false;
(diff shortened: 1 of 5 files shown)


nagyrobi:
On AZE0 with 0xFF:
kép

malec1984:
You must not use the mask FF, many ECUs are returning all the possible DTC list...

nagyrobi:
Hmm. That would actually be a good feature - to know what are the DTCs that would be possible at all, versus the ones really triggered.

Maybe put a checkbox on the page with something like "List all the possible DTCs" which would run it witx 0xFF?

malec1984:
That should normally be available in the respective JSON file if one really want to check.

nagyrobi:
The JSON is much wider afaik. There are 204 entries in the JSON, and only 149 returned with 0xFF. I guess the JSON covers all the possible DTCs for all Nissan models, while apparently on AZE0 only 149 match.
Plus, it would be nice to reveal any undocumented ones which may be missing from the JSON.
Maybe this would also help confirming that readout actually works...?

The confusion is because the page says "32 codes" and "Loaded 204 entries, 32/32 DTCs matched" the user would think only 32 are exitsting. 
Not listing all of them but I'll change these notifications.

malec1984:
There is no confusion.... The 32 are listed because that's the MAX buffer size allocated for DTCs and 32/32 matched means how many are actually matching a description in JSON. Normally on a good integration you should see less than let's say 10 active DTCs

nagyrobi:
Understood. 
This is now on AZE0 with 0x0E:

image

Also the page would say like `32 codes shown of 149 reported — read 3s ago` in case of more than 32 DTCs, and shows as before `1 codes — read 3s ago` if there's just a few. Didn't want to mess with a new checkbox and list all of them because that would have been costly on RAM. But at least now we would know if there are more than 32.

nagyrobi:
So we can say that my AZE0 only has the U1000 active, and nothing else, right? The tests above confirm that DTC reading works, if there would be more, we would see them, right?

dalathegreat (approved):
Thanks for looking into this @nagyrobi & @malec1984 🙌

PR #2738, opened 2026-08-03
UdsCanBattery superclass

jonny5532:
### What
A superclass that batteries can extend to provide basic UDS functionality. Part of the MG HS/ZS/MarvelR/MG5 refactor, but separated out as a separate commit for feedback.

Functionality:
- Uses the new ISOTP lib for flow control
- Subclass can set a list of PIDs to scan, and it will scan them, one every 200ms.
- PIDs are retried up to 10 times (sometimes they timeout during BMS startup)
- PID results are passed back to `handle_pid` callback for batteries to consume
- The callback can return a PID to allow out-of-turn scanning (eg, one PID more often than the main sequence)
- UDS 'sequences' can be triggered, with a battery callback to send the frames in a sequence (interleaved with PID scanning).
- Some sequences are provided internally (reset DTC, read DTC).
- A basic HTML renderer is included that includes battery-generated string, with the standard DTC readout below.
- A `pause_uds` function allows different degrees of UDS traffic to be paused with a specified timeout
- Any UDS traffic from elsewhere (eg, an external scan tool) will pause our traffic for 5s

**Note: includes ISOTP tweaks which will hopefully be merged first**

Original code was me, refactor was me + DS4F-0731, comments are 90% mine, bugs entirely mine ;)

### Why
Tries to reduce a lot of the common code between integrations, to do with PID scanning and DTC readout.

Handles the sequencing between PID scans and other UDS operations (to avoid confusing the BMS by sending two things at once).

Experimentally rebasing some existing integrations on this superclass has saved about 3kb a time (as well as providing extra features like the DTC read).

> [!TIP]

### DIFF
--- a/Software/src/battery/UdsCanBattery.cpp
+++ b/Software/src/battery/UdsCanBattery.cpp
@@ -1,0 +1,502 @@
+#include "UdsCanBattery.h"
+
+#include <Arduino.h>
+#include "../devboard/utils/events.h"
+#include "../devboard/utils/logging.h"
+
+// Timeouts (to wait for a UDS response) in 100ms ticks
+constexpr uint16_t UDS_TIMEOUT_CLEAR_DTC = 25;
+constexpr uint16_t UDS_TIMEOUT_READ_DTC = 20;
+constexpr uint16_t UDS_TIMEOUT_READ_DID = 2;
+// How many times to retry a particular PID read before giving up.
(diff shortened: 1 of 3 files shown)


jonny5532:
(Size is uninteresting because no batteries use this class yet - including the MG refactor it is **+3,336** on T485, probably about 50% UdsCanBattery vs MG code).

dalathegreat (approved):

PR #2739, opened 2026-08-03
Safety :warning: Apply battery2/battery3 power limits to main battery

jonny5532:
### What
(Sorry, yet another PR!)

In multi-battery setups, apply the power limits of battery 2/3 to the first battery if they are lower (since the first battery limits are the ones that are followed).

(This code has been tested on a double battery setup, works as expected)

### Why
Ensures that batteries are not overcharged/discharged - a SoC discrepancy due to drift, plus a significant delta (a not unreasonable scenario) could cause the 2nd/3rd battery to want to shut off charge before the first one is full. This ensures that this happens.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -333,6 +333,24 @@ void update_calculated_values(unsigned long currentMillis) {
     datalayer.battery.settings.max_remote_set_discharge_dA = 0;
   }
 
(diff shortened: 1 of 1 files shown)


nagyrobi:
Yeah was thinking about this the other day

dalathegreat (approved):

PR #2740, opened 2026-08-03
Give every environment the same managed-component set

ekholm:
`custom_component_remove` and `custom_sdkconfig` were repeated per-environment in `platformio.ini` — seven copies that had to be kept identical by hand, and `compiler_warning_check` never got them. That has two consequences:

1. **Build-order-dependent failures.** `custom_component_remove` mutates the shared PlatformIO package store, so environments with different remove-lists corrupt each other's framework state. Which error you get depends on interleaving — the "missing" component was different every time (`network_provisioning`, `libsodium`, `esp-libhelix-mp3`), which is what made it look transient and survive retries.
2. **The warning check measured the wrong image.** Building without the removals and without `custom_sdkconfig`, it came out ~39 KB larger than the firmware users actually get, sat at ~100% of the partition, and failed changes for reasons unrelated to their content.

Fix: move both options into the global `[env]` section, which every environment inherits. One copy means nothing to keep in sync; a future environment inherits the correct set by default, and overriding has to be written deliberately.

Verified:
- The failing sequence (compiler warning check, then BECom immediately after) is green.
- Board firmware is byte-identical: BECom builds to 1,925,200 with and without the change.
- Warning check drops 1,950,828 → 1,911,799 (−39,029) — it now measures the image that ships.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,6 +11,28 @@
 [platformio]
 src_dir = ./Software
 
+[env]
+custom_sdkconfig = file://sdkconfig.be_size.defaults
+custom_component_remove =
+    espressif/esp_insights
+    espressif/esp_diag_data_store
+    espressif/esp_diagnostics
+    espressif/esp_rainmaker
+    espressif/rmaker_common
+    espressif/network_provisioning
+    espressif/esp-zboss-lib
+    espressif/esp-zigbee-lib
+    espressif/esp-modbus
+    espressif/esp_modem
+    espressif/esp-sr
+    espressif/esp-dsp
+    espressif/qrcode
+    chmorgan/esp-libhelix-mp3
+    espressif/lan867x
+    espressif/w5500
+    espressif/dm9051
+    espressif/ksz8851snl
+
 [env:compiler_warning_check]
 platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.39/platform-espressif32.zip
 board = esp32dev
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -11,6 +11,34 @@
 [platformio]
 src_dir = ./Software
 
+[env]
+; Applied to every environment. These two must be identical everywhere: the
```
dalathegreat: Nice improvement. Drop the excessive comment, and we can merge this! 👍

dalathegreat (changes requested):

jonny5532:
Just to confirm, I don't think this will have affected the online builds or size reporting? Since they have a fresh framework build from zero, cached for each environment.

I can see it being annoying locally though when switching between envs.

ekholm:
Correct for the framework corruption — fresh per-env builds mean CI never sees that; it's a shared-local-machine problem only. The one online thing that was affected is the compiler_warning_check job: it built without the removals and without custom_sdkconfig, so it measured an image ~39 KB larger than any board ships, at ~100% of the partition — that's the 1,950,828 → 1,911,799 change in the body. The per-board size bot was fine (those envs already had the options).

ekholm:
I'm using worktrees, and to make that work, I on top of the this env fix I need to have a mutex on pio builds, as everybuild fiddles with the global pio config (located in user home folder). This prevent building different worktrees parallelm that otherwise would not work

dalathegreat:
This PR made the builds take 7 minutes , insane performance regression compared to the 2minute it normally takes

ekholm:
That's the CI cache being invalidated, not the builds getting slower. The env cache key hashes `platformio.ini` (the "Try env cache" step in compile-common-image.yml), so any PR touching that file makes every job fall back to a cold framework build — the workflow's own comment describes exactly this fallback. It's one-time: after the merge, the first run on `main` saves caches under the new hash and PR builds are back to the normal ~2 minutes. Any future `platformio.ini` edit would cost the same 7-minute round, with or without this change.

The one env that genuinely does more work is `compiler_warning_check`: it now builds the same slimmed from-source framework the boards do, instead of the stock prebuilt one — cached the same way, so it's only slow on the same cold runs. That's the cost of it measuring the image that actually ships (it was ~39 KB off before).

dalathegreat (approved):

PR #2744, opened 2026-08-03
Optimize CRC8 calculation for MEB CAN frames.

malec1984:
### What
Use the existing lookup table from utils common.
This replaces the bit loop calculation which wasted more cpu.

### Why
optimize cpu usage.

### How
Improve the CRC calculation.

> [!TIP]


dalathegreat (approved):
Nice, would be good to have unit tests on CRC calculation also!

dalathegreat (approved):

PR #2746, opened 2026-08-03
Add shared ESP32 and ESP32S3 targets

ngehrsitz:
### What
building on #2740 we can deduplicate even further. I was a bit too late to comment on that PR.

> [!TIP]

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,7 +11,7 @@
 [platformio]
 src_dir = ./Software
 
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
LGTM 👍 Only concern is that we  sometimes had to have compiler warning run on different platform of the esp32, due to some versions being released having compiler warnings stemming from the platform itself. But incase we encounter that again, we just have to split out platform from the [env]

Issue #2732, opened 2026-08-02
[Feature request] BYD ATTO 3: isolation-monitor override for battery balancing


TheOfficialBatteryMan:
## Problem

When a BYD ATTO 3 Blade battery is connected to a solar inverter, the two isolation-monitoring systems can interfere. My BMS then reports about `3 ohm/V` and sets confirmed DTC `P1A0000` (Serious Electric Leakage Fault).

We expect this serious DTC to inhibit cell balancing. The inverter already performs its own isolation test, so Battery Emulator needs an experimental option to inhibit the BYD test/fault while balancing the pack.

## Findings

Requests use CAN ID `0x7E7`; replies use `0x7EF`.
This is behind lvl 1 security
- Activate override: `04 31 01 20 08 00 00 00`
- Release override: `04 31 02 20 08 00 00 00`
- Check state: `04 31 03 20 08 00 00 00`

Positive replies begin `71 01`, `71 02`, and `71 03` respectively.

After activating the override, the reported isolation value rises to approximately `2900-3000 ohm/V` in BE or 7000 in scanner. After clearing DTCs, `P1A0000` no longer returns. The value still fluctuates, so the routine may suppress the BYD test or fault evaluation rather than completely stopping measurement.

## Requested implementation

Add guarded, experimental BYD controls to:

1. Activate the isolation override.
2. Release the override.
3. Read and display its state.
4. Optionally clear DTCs after activation.

Integrate this with the existing BYD diagnostic state machine so it does not conflict with DTC, SOC-calibration, or balance-time requests. Keep it manual, show failures/timeouts, and warn that another suitable isolation monitor must remain active.

The result can be verified by checking whether the accumulated per-cell balance hours increase without `P1A0000` returning.

dalathegreat:
Implemented in v12.2.0 release

PR #2755, opened 2026-08-05
Feature: BYD - add isolation resistance control

demon1300:
### What
A control on the BYD more battery info page to turn the pack's own insulation-resistance monitor off (and back on), with a live status showing whether it's on, and a "keep it off after every restart" tickbox.  This is enabled by default.

### Why
A repurposed pack often runs with its case earthed to the frame, which the pack reads as a leakage fault — it then refuses to charge in the correct manner and drops rated power by raising a DTC. In these installs the inverter already does its own insulation test and is responsible for it, so the pack's duplicate check is just getting in the way. Turning it off lets the pack run normally.

### How
Battery Emulator sends the same diagnostic command a dealer scan tool uses (over the pack's UDS channel, unlocked with the security key we already had for SOC reset etc). The monitor forgets the setting every time the battery is power-cycled, so the tickbox re-sends the "off" command a few seconds after each startup. Status is read from a broadcast the pack already sends to check the status of the monitor.

> [!TIP]

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.cpp
+++ b/Software/src/communication/nvm/comm_nvm.cpp
@@ -275,11 +275,12 @@ void init_stored_settings() {
   datalayer_extended.bydAtto3.auto_calibrate_soc_drift_percent =
       constrain(settings.getUInt("BYDAUTOCALDRIFT", 5), 1u, 20u);
   datalayer_extended.bydAtto3.auto_calibrate_soc_enabled = settings.getBool("BYDAUTOCALEN", true);
-  datalayer_extended.bydAtto3.keep_iso_disabled = settings.getBool("BYDKEEPISOOFF", true);
   datalayer_extended.bydAtto3_2.auto_calibrate_soc_drift_percent =
       constrain(settings.getUInt("BYDAUTOCALDRFT2", 5), 1u, 20u);
   datalayer_extended.bydAtto3_2.auto_calibrate_soc_enabled = settings.getBool("BYDAUTOCALEN2", true);
-  datalayer_extended.bydAtto3_2.keep_iso_disabled = settings.getBool("BYDKEEPISOOFF2", true);
+  // One isolation-monitor setting for both batteries
+  datalayer_extended.bydAtto3.keep_iso_disabled = settings.getBool("BYDKEEPISOOFF", true);
+  datalayer_extended.bydAtto3_2.keep_iso_disabled = datalayer_extended.bydAtto3.keep_iso_disabled;
 }
 
 void clear_wifi_sta_settings() {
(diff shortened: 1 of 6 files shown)


TheOfficialBatteryMan:
If you're happy, you have this set up. I'll delete my issue now.
https://github.com/dalathegreat/Battery-Emulator/issues/2732

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -706,6 +706,46 @@ void init_webserver() {
...
+  def_route_with_auth("/bydAtto3IsoEnable2", server, HTTP_GET, [](AsyncWebServerRequest* request) {
+    datalayer_extended.bydAtto3_2.UserRequestIsoRoutineEnable = true;
+    request->send(200, "text/plain", "OK");
+  });
+  def_route_with_auth("/bydAtto3KeepIsoDisabled2", server, HTTP_GET, [](AsyncWebServerRequest* request) {
+    if (request->hasParam("value")) {
+      bool enabled = request->getParam("value")->value().toInt() != 0;
+      datalayer_extended.bydAtto3_2.keep_iso_disabled = enabled;
+      Preferences prefs;
+      prefs.begin("batterySettings", false);
```
dalathegreat, line 726: doing prefs calls usually results in the entire application hitching/stuttering, since it blocks core operation entirely. I just saw that we already do that on other BYD settings, so I guess it is fine, but just good info nonetheless

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -275,9 +275,11 @@ void init_stored_settings() {
   datalayer_extended.bydAtto3.auto_calibrate_soc_drift_percent =
       constrain(settings.getUInt("BYDAUTOCALDRIFT", 5), 1u, 20u);
   datalayer_extended.bydAtto3.auto_calibrate_soc_enabled = settings.getBool("BYDAUTOCALEN", true);
+  datalayer_extended.bydAtto3.keep_iso_disabled = settings.getBool("BYDKEEPISOOFF", true);
```
dalathegreat: Is there a need that we need two separate parameters for Bat1 and Bat2? Is there going to be some benefit to be able to disable only one BYD battery's isolation monitoring, but not the other?

demon1300: Not really, it was just about having more granular control; what's exposed for one battery is also exposed for the second.  Probably over complicating what will be left alone (default IR monitor off) in 99% of installs though. 
Just on that, I wonder if I should even remove the ability to turn it on or off so as to make it just an automated function and very simple for the user to understand? What are your thoughts here?

Software/src/devboard/webserver/webserver.cpp
demon1300: Noted!

Software/src/communication/nvm/comm_nvm.cpp
demon1300: I've removed the controls for the second battery, just made it follow the primary.

TheOfficialBatteryMan: I think it should be in settings as a set up step. Then if someone wants it off they change settings which requires a restart. Labelling to explain and default on

demon1300: > I think it should be in settings as a set up step. Then if someone wants it off they change settings which requires a restart. Labelling to explain and default on

We could, however it would need to be built.  Currently it looks like the settings are global, with the few per battery settings being just 'hidden'.  Not great for long term per battery settings...  Can look at doing this if @dalathegreat agrees that's the right place to put it.

dalathegreat (approved):

dalathegreat:
I still don't think we need to allow users to enable iso monitoring at all, but we can remove it in a future PR once this is confirmed working OK

PR #2756, opened 2026-08-05
Stop starting an automatic precharge that is aborted on the next tick

ekholm:
`handle_precharge_control()` leaves `AUTO_PRECHARGE_IDLE` when the battery requests precharge and the inverter permits it, but the `AUTO_PRECHARGE_PRECHARGING` state aborts back to idle when the system is not ACTIVE, when an equipment stop is active, or when the BMS is not standby/active. The start gate tests neither of the first two, so whenever one of them is the blocking condition the sequence starts, is aborted on the following tick, and starts again immediately.

Entering the sequence drives the precharge PWM and switches the inverter-disconnect contactor, and the state machine runs on the 10 ms core-loop cadence, so this is a continuous ~20-30 ms switching cycle rather than a one-off. It was reported as an audible contactor rattle on a MEB + SMA setup where the battery requests precharge while the system is in FAULT; the same asymmetry applies while an equipment stop is active.

The fix gives the start gate and the abort check one shared predicate, so a sequence is only started under the conditions it needs in order to keep running. Sequences that used to start and abort within a tick no longer start; nothing that completes today is affected, since these are the conditions the sequence already had to satisfy to survive its first tick. The timeout / BMS-fault path to `AUTO_PRECHARGE_FAILURE` and the restart-after-BMS-reset path in `AUTO_PRECHARGE_COMPLETED` are unchanged.

Tests: `precharge_control.cpp` was not in the test build, so this adds it along with two test-only pieces it needed — `ledcWriteTone` in the Arduino emulation, and a recorder for pin writes in the same spirit as the CAN frame recorder, so the tests can assert that the inverter-disconnect contactor is never switched rather than only that a state variable stayed put. Six tests cover the two blocked cases, the restart loop itself, and three cases that must keep working. Checked against the unfixed file: the three that describe the bug fail there and pass here.

### DIFF
--- a/Software/src/communication/precharge_control/precharge_control.cpp
+++ b/Software/src/communication/precharge_control/precharge_control.cpp
@@ -12,8 +12,6 @@ uint16_t precharge_max_precharge_time_before_fault = 15000;
 uint16_t Precharge_max_PWM_Freq = 34000;
 
 // Hardcoded parameters
-#define Precharge_default_PWM_Freq 11000
-#define Precharge_min_PWM_Freq 5000
 #define Precharge_PWM_Res 8
 #define PWM_Freq 20000  // 20 kHz frequency, beyond audible range
 #define PWM_Precharge_Channel 0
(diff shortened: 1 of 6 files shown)


Webwanze:
@ekholm 

Good Morning,
tested this PR in V12.2.0. Works like expected!

Thank you @ekholm !

dalathegreat (approved):

dalathegreat:
After merging this, the build system fails on the newly added tests: https://github.com/dalathegreat/Battery-Emulator/actions/runs/32851326846/job/97812695117

Wanna have a go at it @ekholm ?

PR #2757, opened 2026-08-05
MEB HTML optimize strings

malec1984:
### What
Optimize the MEB HTML code.
Interim voltage has also status codes.

### Why
Better performance and reduce code size.

### How
Mainly use lookup tables.
Reserve buffer space to avoid heap fragmentation.

> [!TIP]


dalathegreat (approved):
This looks way harder to maintain, but the integration is already at the point where more stuff might not need adding to this page, so the performance gain is very nice. Approved!

PR #2761, opened 2026-08-05
Refactor SD support to use SPI

ngehrsitz:
### What
This PR is another prerequisite for #2722. The existing Lilygo board used the dedicated SDIO controller of the ESP32 for SD card access. This is generally the recommended option since it is slightly faster. The main drawback here is that you can only use it on dedicated pins. The alternative is using one of the SPI controllers which is possible on most pins.
The DFRobot board unfortunately does not use the correct pins for SDIO, thus forcing the use of SPI. Even though I placed the additional SPI SDCard implementation behind `ifdef`s the build system was not smart enough to only conditionally compile it in which resulted in the significant flash size increase in the current version of the PR. Neither setting `lib_ldf_mode = chain+` or `lib_ignore` could be used as a workaround.
Therefore this PR also moves the Lilygo to use SPI. Fortunately this also yields a significant flash size reduction most likely since most the SPI code gets compiled in anyways. I also gated the feature so that it does not impact the other boards unlike before.
One of the things I noticed is that the SPI bus assignments are currently somewhat arbitrary which can lead to overlaps. In a follow-up PR I will add some logic that assigns the same bus for matching pins.

> [!TIP]

### DIFF
--- a/Software/src/devboard/hal/hw_3LB.h
+++ b/Software/src/devboard/hal/hw_3LB.h
@@ -56,12 +56,6 @@ class ThreeLBHal : public Esp32Hal {
   virtual gpio_num_t INVERTER_CONTACTOR_ENABLE_PIN() { return GPIO_NUM_36; }
   virtual gpio_num_t INVERTER_CONTACTOR_ENABLE_LED_PIN() { return GPIO_NUM_NC; }
 
-  // SD card
-  virtual gpio_num_t SD_MISO_PIN() { return GPIO_NUM_2; }
-  virtual gpio_num_t SD_MOSI_PIN() { return GPIO_NUM_15; }
-  virtual gpio_num_t SD_SCLK_PIN() { return GPIO_NUM_14; }
-  virtual gpio_num_t SD_CS_PIN() { return GPIO_NUM_13; }
-
   // LED
   virtual gpio_num_t LED_PIN() { return GPIO_NUM_4; }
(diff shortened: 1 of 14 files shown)


Software/src/devboard/hal/hw_3LB.h
```diff
@@ -56,12 +56,6 @@ class ThreeLBHal : public Esp32Hal {
   virtual gpio_num_t INVERTER_CONTACTOR_ENABLE_PIN() { return GPIO_NUM_36; }
   virtual gpio_num_t INVERTER_CONTACTOR_ENABLE_LED_PIN() { return GPIO_NUM_NC; }
 
-  // SD card
-  virtual gpio_num_t SD_MISO_PIN() { return GPIO_NUM_2; }
-  virtual gpio_num_t SD_MOSI_PIN() { return GPIO_NUM_15; }
-  virtual gpio_num_t SD_SCLK_PIN() { return GPIO_NUM_14; }
-  virtual gpio_num_t SD_CS_PIN() { return GPIO_NUM_13; }
```
ngehrsitz, line 63: This board also had SD pins defined, but these are used for the ethernet addon: https://github.com/malcolmputer/3lb/blob/main/exports/pdfs/3lb.pdf

dalathegreat (approved):
T485 schematic it looks like 1-bit SDIO (And not 4-bit), so changing to 1-but SDI mode has no performance drop 👍 

Extremely nice optimization 🙌

PR #2769, opened 2026-08-07
BETA: Streaming can logging over WiFi

jonny5532:
### What
Adds a `/dump_can` endpoint which retrieves endless streamed CAN logs. 

Can either consume in a browser window or use something like:
`while true; do curl http://192.168.4.1/dump_can >> can_log.txt; done`
to log forever.

Also makes the existing CAN logger and SD logger use the new CAN formatter (which is considerably faster). Note that I haven't tested the SD CAN logging (although it should now use the new bus numbering unlike before). One small addition is that FD frames are now distinguished: `rx0`/`tx1` for non-FD, `RX0`/`TX1` for FD.

One annoyance is that the stream is infinite, so a browser won't actually save it as a download - cancelling deletes the partial file. You can save it by starting a new stream in another tab, which will stop the current one and finish downloading. If we add a user interface, we could add a 'Stop' button which just starts/stops a stream to trigger this.

### Why
Was originally implemented in the TinyWebServer branch, then backported to the new-frontend-on-AsyncWebServer branch, and now backported to the current frontend.

Might be useful if it isn't too big.

Implementation mostly guided AI, there's some rather subtle locking going on (but I've checked through it a few times and also tested). Requires a tweak to the vendored AsyncWebServer but I don't think we're in a position to update that anyway.

> [!TIP]

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -114,7 +114,7 @@ void connectivity_loop(void*) {
       update_espnow();
     }
 
-    ota_monitor();
+    webserver_tick();
 
     END_TIME_MEASUREMENT_MAX(wifi, datalayer.system.status.wifi_task_10s_max_us);
 
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -10,6 +10,7 @@
 #include "src/devboard/sdcard/sdcard.h"
 #include "src/devboard/utils/events.h"
 #include "src/devboard/utils/logging.h"
+#include "src/devboard/webserver/webserver_can_streaming.h"
 #include "utils.h"
 
 #include <esp_private/periph_ctrl.h>
@@ -317,7 +318,7 @@ void transmit_can_frame_to_interface(const CAN_frame* tx_frame, CAN_Interface in
 
 #ifdef SDCARD
   if (datalayer.system.info.CAN_SD_logging_active) {
-    add_can_frame_to_buffer(*tx_frame, frameDirection(MSG_TX));
+    add_can_frame_to_buffer(*tx_frame, interface, frameDirection(MSG_TX));
   }
 #endif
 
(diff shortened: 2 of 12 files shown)


nagyrobi:
We might consider deprecating old methods of getting CAN logs.

jonny5532:
I have noticed that every 30 seconds or so you get a corrupt entry like this (some examples):

```
(6663.352) Rx0 295 [8] 45 ac 00 03 70 4e 37 (6e3
...
(6698.734) Rx0 172 [8](6698.834) Rx0 173 [8] 00 00 c7 c8 0f bf 0f b4
...
(6726.981)(6727.077) Rx0 295 [8] 65 ac 00 03 70 4e 37 c3
...
(6731(6731.655) Rx0 173 [8] 00 00 c7 c8 0f bf 0f b4
...
(6746.379) Rx0 173 [8] (6746.450) Rx0 295 [8] f5 ac 00 03 70 4e 37 53
```

After much unsuccessful investigation, this seems to be a AsyncTCP bug. I tried upgrading AsyncTCP but that made it work worse - seems to be something to do with memory fragmentation and the handoff to lwIP.

If we do switch to the TinyWebServer thing with preallocated ringbufs this shouldn't be an issue. As it stands it isn't really an issue, CAN software will ignore the malformed frame.

ngehrsitz:
What exactly did you try when upgrading AsyncTCP? I also had upgrading the webserver on my todo list, but when I looked into it AI flagged some performance issues exactly related to CAN streaming.

jonny5532:
IIRC I just swapped in these files:

https://github.com/ESP32Async/AsyncTCP/tree/main/src

and hacked it about a bit so the existing core/stack `#define`s still worked.

It compiled fine, but the streaming connection dropped after 10-30s, I didn't look any further. Perhaps it is now aborting where previously it just silently failed.

dalathegreat (approved):
Merging to gather more feedback, extremely nice and useful feature!

PR #2771, opened 2026-08-07
Calculate uptime strings without ctime

ngehrsitz:
### What
While looking into RTC support for #2722 I noticed that there was a `ctime` dependency even though the time is never synced. This meant the logs were implicitly named after uptime already. I replaced it with an smaller uptime based calculation.

> [!TIP]

### DIFF
--- a/Software/src/battery/BMW-IX-HTML.cpp
+++ b/Software/src/battery/BMW-IX-HTML.cpp
@@ -1,4 +1,5 @@
 #include "BMW-IX-HTML.h"
+#include "../devboard/utils/time_format.h"
 #include "BMW-IX-BATTERY.h"
 
 String BmwIXHtmlRenderer::get_status_html() {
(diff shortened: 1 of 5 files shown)


nagyrobi:
> While looking into RTC support

+1 for RTC support for Waveshare too! It too has RTC battery input socket on board.

ngehrsitz:
> > While looking into RTC support
> 
> +1 for RTC support for Waveshare too! It too has RTC battery input socket on board.

The question is what do we use it for? Currently the only thing I can think of would be giving all logs an explicit timestamp. But I am not sure if that is worth the binary size impact since we would need to add SNTP and RTC code.

nagyrobi:
Periodic BMS reset could use a proper scheduling/timing instead of just counting 24/48 hours from boot.

malec1984:
What about not using any SNTP server? Just a simple button in web ui to set time from browsing device.

nagyrobi:
And a timezone delta selector.

ngehrsitz:
> What about not using any SNTP server? Just a simple button in web ui to set time from browsing device.

If you don´t include SNTP then only the two boards with RTC will survive a reboot. Also from a user perspective I would expect that it automatically syncs if I have network access.
Therefore for me the main question is whether it is worth spending the flash to get logs with a global time?

nagyrobi:
Logging to Syslog uses the time the message enters the server so no issue there.

I would find beneficial a kind of scheduler on long term, like for the periodic BMS reset, or sending some CAN commands on demand etc.

dalathegreat (approved):
Another insanely nice performance gain, thanks 🙌

PR #2773, opened 2026-08-07
Feature/BYD native charging session

demon1300:
### What
Lets the BYD Atto 3 BMS terminate the charge itself and recalibrate its own SOC and SOH, instead of Battery-Emulator stopping the charge at its own cell clamp.

I've also modified the more battery info page and broken it down into more logical sections and the AUTO-cal option will be greyed out when native termination is enabled.

### Why
The BMS only recalibrates at a genuine top-of-charge that it reaches under its own control, so a pack that is always stopped short drifts and never gets a proper full-charge event. It is also the prerequisite for the BMS running its own cell balancing, which only arms after a native termination.

### How
BE runs a real charge session on the pack bus, in place on an already-closed pack (0x86 → 0x85 → 0x84, no contactor movement), using only 0x24A/0x47E/0x36A/0x36D.

Details:

* The session auto-arms on sustained charge current and re-arms only after a real discharge; a refused grant backs off for 15 minutes.
* It owns the taper from cell max 3500 mV up to the observed 3742–3753 mV termination band. Below 3500 mV the BMS advisory still governs power.
* The outcome is reported through EVENT_BYD_CHARGE_TERMINATED, the More Battery Info page, and eight new charge_* MQTT sensors.
* On by default.
* Native termination is primary battery only at this stage.
* Requires the pack case to be isolated from earth: the BMS refuses a charge session while it reports an insulation fault. Packs using the isolation-monitor disable from #2755, with the re-assert fix from #2880, satisfy this automatically.
The minimal frame set was established by subtractive bisection — entry, charge and natural termination all succeed with only those four frames transmitting, so the other companion frames seen in vehicle logs are passengers rather than requirements.

A few other bits:
* Auto-cal's SOC write & taper is disabled while this is on — the battery does the calibration itself.
* Enable toggle persists across reboots. A web UI panel shows session state, live grant, and the last termination's cell spread.

> [!TIP]


dalathegreat (approved):

btoonen:
image
Seems to work perfect, thanks

PR #2775, opened 2026-08-08
Cover the machinery-protection suite, and stop a healthy pack clearing two shared warnings

ekholm:
`update_machineryprotection()` raises the protection events for each configured battery in sequence. Two of those events are shared by every pack, and both were set *and cleared* from inside the per-battery blocks — so the last block to run decided the outcome for all of them.

- `EVENT_CELL_DEVIATION_HIGH`: a healthy pack cleared a cell-deviation warning raised by another pack, on every cycle.
- `EVENT_SOH_DIFFERENCE`: the battery 1 vs battery 3 comparison cleared the warning raised by the battery 1 vs battery 2 comparison. This one only shows with three packs configured, which is presumably why it hasn't been reported.

Both are the same shape as the shared corrupt-flag masking fixed in #2688, so both are fixed the same way and in the same place: evaluated once after the per-battery blocks, raised if any configured pack is offending, and cleared only when none is. The SOH check keeps its existing behaviour of leaving the event untouched when no pack pair has a real reading to compare. Single-battery behaviour is unchanged.

The rest is tests. The protection suite had no direct coverage: temperature, pack voltage, cell limits with their latching and hysteresis behaviour, the user-set voltage window and its release hysteresis, battery full/empty, the inverter overdraw alarms with their tolerance margin and failure count, the current-limit clamp, the pause and fault paths, emergency recovery charge, and how the shared events behave across two and three packs. 34 cases added; four fail against the unfixed file and the rest pass either way as regression guards.

`CELL_CRITICAL_MV`, `CELL_HYSTERESIS_MV`, `HYSTERESIS_OFFSET_DV` and the two recovery-charge limits move to `safety.h` so the tests derive from the thresholds rather than hardcoding them; their sibling limits already live there and no value changes.

Also removes an ordering constraint in `inverter_alive_tests.cpp`: the fixture now clears the detection latch directly, the way `battery_alive_tests.cpp` already clears its latches, so that file no longer needs its detection test kept first and the suite can be run shuffled.

### DIFF
--- a/Software/src/devboard/safety/safety.cpp
+++ b/Software/src/devboard/safety/safety.cpp
@@ -6,7 +6,6 @@
 #include "../../inverter/INVERTERS.h"
 #include "../utils/events.h"
 
-static uint16_t cell_deviation_mV = 0;
 static uint8_t charge_limit_failures = 0;
 static uint8_t discharge_limit_failures = 0;
 static bool battery_full_event_fired = false;
@@ -16,11 +15,6 @@ static bool battery_empty_event_fired = false;
 // Some inverters take a while to boot and start sending CAN. Suppress the
 // inverter-missing error during this startup window (measured from power-on).
 #define INVERTER_STARTUP_GRACE_MS 300000  // 300 s
-#define CELL_CRITICAL_MV 100              // If cells go this much outside design voltage, shut battery down!
-#define LOWEST_ALLOWED_CELLVOLTAGE_RECOVERY_CHARGE_MV 2000  //If cells are below this, recovery charge not allowed
-#define MAX_CHARGEPOWER_RECOVERY_CHARGE_DA 50
-#define HYSTERESIS_OFFSET_DV 20
-#define CELL_HYSTERESIS_MV 20  // Re-allow charge only once max cell drops this far below limit (avoids chatter at knee)
 
 //battery pause status begin
 bool emulator_pause_request_ON = false;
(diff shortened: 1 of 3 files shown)


ekholm:
#2832 merged while this PR was open and changes the same function — it gates the charge/discharge limit comparison off while paused or faulted, and this PR's `ChargeLimitExceededOnlyAfterRepeatedOvershoot` / `DischargeLimitExceededOnlyAfterRepeatedOvershoot` cases pin exactly those events. So the interaction was checked by building the composition rather than by reading the diffs: this branch is now rebased onto current main (which includes #2832), `safety.cpp` merged clean, and both cases pass unchanged. The two changes are complementary — the pause gate decides *when* the comparison runs, these tests pin the repeated-overshoot debounce *when it runs*.

One fixture line was added on the rebase: `emulator_pause_status = NORMAL;` in `SetUp()`. After #2832 that global gates both limit checks, and it lives outside the datalayer, so without the reset a test ending mid-pause would fail these cases for an unrelated reason.

The new pause/FAULT gate itself has no test yet. If wanted, I'd extend this PR with two cases covering it: pausing mid-overshoot clears both events and resets the failure counters, and an idle inverter at 0 W takes the clearing branch. A few lines each on this file's existing fixture — say the word and I'll push them.

ekholm:
Refreshed onto current main, narrowed to the test-coverage commit. The per-pack cell-deviation/SOH events that were the second commit predate the per-pack event refactor that has since merged — rather than carry a competing convention here, they'll come separately on top of the new `EVENT_BATTERY2_*` shape. Two test pins also got corrected in the rebase: the event data field widened to `int16_t` since this was written (the overheat assertion was truncating through the old `uint8_t` cast), and the temperature-deviation event is now deliberately non-latching upstream, which the suite now pins instead of fighting.

Issue #2758, opened 2026-08-05
RFC: Treat CAN controllers as an array of devices, not five hand-written interfaces


ekholm:
**The problem.** `comm_can.cpp` handles its interfaces by writing each one out by hand. Every operation — init, transmit, receive, stop, restart, speed change, error reporting — is a per-interface branch or a copy of the previous one. The two MCP2518FD paths are the clearest case: `begin_canfd()` and `begin_canfd_2()` are identical except for the pointer and the log text, the two transmit cases are byte-identical fifteen-line blocks, and the `bool first` parameter threaded through the receive path exists only because the state is in named globals instead of an indexed structure.

That has cost more than tidiness. Both FD chips raise the same `EVENT_CANMCP2518FD_INIT_FAILURE`, so the event log cannot say which controller failed to start. And because `CANFD_NATIVE` and `CANFD_ADDON_MCP2518` are two names for one physical chip on the boards that have it, every received FD frame is mapped twice and then de-duplicated by two `TODO`-marked guards in `map_can_frame_to_variable()`.

**What I'd like to propose.** One small class per physical controller — native TWAI, MCP2515, MCP2518FD — and a table mapping the logical `CAN_Interface` values onto those devices. Every operation becomes a lookup or a loop over the table. Frame conversion exists once per controller type instead of once per interface. Received frames are dispatched once per physical frame and fanned out to whichever logical interfaces map to that device, so the double-mapping and both guards disappear rather than being worked around.

Health state moves the same way: the eight named booleans in the datalayer (`can_native_send_fail`, `can_2515_bus_error`, …) become one array indexed by device, each device writing only its own entry, and the safety monitor loops instead of repeating the same eight-line block eight times.

**What it makes possible.** Today the shape of the code assumes one native controller, one classic add-on and at most two FD add-ons, each with hardcoded slots and its own enum entries. With devices in an array, a board can have several controllers of any kind, indices are assigned at registration instead of written down, and the frame capability (classic or FD) becomes a property of the device rather than something encoded in its name — which matters for chips whose native controller supports FD. Adding a controller becomes one registration rather than roughly ten scattered edits.

[...]

PR #2780, opened 2026-08-08
Refactor CAN interfaces into a device table - one code path for all controllers

ekholm:
`comm_can.cpp` handles its five interfaces by writing each one out by hand. Every operation — init, transmit, receive, stop, restart, speed change, error reporting — is a per-interface branch or a copy of the previous one. The two MCP2518FD paths are the clearest case: `begin_canfd()` and `begin_canfd_2()` are identical except for the pointer and the log text, the two transmit cases are byte-identical fifteen-line blocks, and the `bool first` parameter threaded through the receive path exists only because the state is in named globals rather than an indexed structure.

That has cost more than tidiness. Both FD chips raised the same `EVENT_CANMCP2518FD_INIT_FAILURE`, so the event log could not say which controller failed to start. And because `CANFD_NATIVE` and `CANFD_ADDON_MCP2518` are two names for one physical chip on the boards that have it, every received FD frame was mapped twice and then de-duplicated by two `TODO`-marked guards.

This introduces one small class per physical controller — native TWAI, MCP2515, MCP2518FD instantiated twice — and a table mapping the logical `CAN_Interface` values onto those devices. Every operation becomes a lookup or a loop. Frame conversion exists once per controller type instead of once per interface.

Behaviour-preserving, with two deliberate exceptions: the second MCP2518 now raises its own init-failure event, and frames on the shared FD controller are dispatched once per physical frame rather than mapped twice and de-duplicated. Health flags move into a datalayer array indexed by device, and the events shared by several devices are decided once per event from the union of the devices sharing them, with a bitmask payload naming which — a per-device decision would let a healthy chip clear a faulty sibling's event.

No changes to the `CAN_Interface` values, the registration API, settings, or the bus numbering in the CAN log. Adding a controller becomes one table entry rather than roughly ten scattered edits.

This is not a code-size reduction: the duplication is replaced by structure. It costs flash — the size bot measured **+1,064 to +1,344 bytes** depending on the board, so a little over a kilobyte everywhere, not under it. (An earlier revision of this description said "within a kilobyte per board"; that was wrong, and the bot's figures are the ones to go by.) The two boards where that matters are Lilygo T-CAN485 at 94.3% of flash and ESP32 DevKit V1 at 92.0%; the rest sit under 56%.

Stacked on #2774, whose 13 dispatch tests were written against the behaviour *before* this refactor and pass unchanged after it — that is the evidence this preserves the registration and fan-out contract, rather than it being asserted from a reading of the diff. 36 further cases cover the health-event aggregation, transmit routing through the device table, and the bus-off recovery decision.

The transmit path moved out of `comm_can.cpp` for the same reason the registry did: that file also drives the TWAI controller, the MCP2515 and both MCP2518FDs, so nothing in it links on a host, and the test build had been replacing `transmit_can_frame_to_interface` wholesale. The gate, the frame logging and the table lookup need no hardware, so they now live beside the routing; the emulation boundary is the controller. That is pure code motion — `esp32devkit_330` `firmware.bin` is byte-identical before and after it — and it means the existing driver tests reach their recorded frames through the real transmit path rather than around it.


Software/src/communication/can/comm_can.cpp
```diff
@@ -17,60 +18,28 @@
...
-//CAN logging filter settings
-uint16_t user_selected_CAN_ID_cutoff_filter = 0;  //Messages below this ID will not be logged in webserver
+static void dispatch_frame(CAN_frame* rx_frame, class CanDevice* dev);
 
-bool init_CAN() {
-  // Native CAN (onboard the ESP32)
-
-  auto nativeIt = can_receivers.find(CAN_NATIVE);
+// The native TWAI controller on the ESP32 itself.
+class NativeTwaiDevice : public CanDevice {
```
ngehrsitz, line 33: I would move every controller into it´s own file

Software/src/communication/can/comm_can.cpp
```diff
@@ -17,60 +18,28 @@
...
+    log_interface = CAN_NATIVE;
+    buffer_full_event = EVENT_CAN_NATIVE_BUFFER_FULL;
+    bus_error_event = EVENT_CAN_NATIVE_BUS_ERROR;
+  }
 
-  if (nativeIt != can_receivers.end()) {
+  bool init(CAN_Speed speed) override {
     auto se_pin = esp32hal->CAN_SE_PIN();
     auto tx_pin = esp32hal->CAN_TX_PIN();
     auto rx_pin = esp32hal->CAN_RX_PIN();
```
ngehrsitz: Can we also remove the dependency from hal by passing in the pins at construction?

Software/src/communication/can/comm_can.cpp
```diff
@@ -143,33 +215,271 @@ bool init_CAN() {
...
+
+  Config cfg_;
+  ACAN2517FD* can_ = nullptr;
+  ACAN2517FDSettings* settings_ = nullptr;
+  EVENTS_ENUM_TYPE init_fail_event_ = EVENT_NOF_EVENTS;
+};
+
+//CAN logging filter settings
+
+bool init_CAN() {
```
ngehrsitz, line 489: Here you still have to go trough every chip which goes against the idea of abstracting chip specifics away.

Software/src/communication/can/comm_can.cpp
```diff
@@ -143,33 +215,271 @@ bool init_CAN() {
...
+// trampolines.
+static Mcp2518Device* fd_instances[MAX_CAN_FD_DEVICES] = {};  // value-initialized: all null
+
+// Logical interface -> physical device. Entries may repeat: on boards where
+// CANFD_NATIVE and CANFD_ADDON_MCP2518 are the same chip, both slots point at
+// the same device, and dispatch_frame delivers each physical frame to the
+// receivers of every logical interface mapped to it.
+// MCP2518FD add-on controller (CAN-FD over SPI). Instantiated twice on boards
+// with two FD chips; the SPI bus wiring (including bus sharing between the
+// two instances) is board topology and stays in init_CAN().
```
ngehrsitz, line 297: This is the exact issue that I faced when I tried to share an SPI bus between SD and CAN. The best idea I had so far is moving SPI init to the hal.

ngehrsitz (commented):
I only took a brief look at this, but this is definitely the right direction. I think we need to go even further.
The way I envison peripherals setup at the moment is something along these lines:

1. Read the settings to get the user intent/different possible pin configurations(In the future this could even be extended so the pins without hardwired peripherals can be freely selected)
2. hal starts initializing the SPI buses on their fixed pins
3. (Now hal could even loop trough the available drivers and try an auto-detection step for a certain bus if it there is no explicit user choice. This would especially be useful when you share an SPI with an sdcard and you want to make sure none is present)
4. hal+settings hands off the spi bus and pin configuration to the drivers which initialize

That way the existing pattern where can needs to read the settings and hal would be inverted.

ekholm:
> I only took a brief look at this, but this is definitely the right direction. I think we need to go even further. The way I envison peripherals setup at the moment is something along these lines:
> 
> 1. Read the settings to get the user intent/different possible pin configurations(In the future this could even be extended so the pins without hardwired peripherals can be freely selected)
> 2. hal starts initializing the SPI buses on their fixed pins
> 3. (Now hal could even loop trough the available drivers and try an auto-detection step for a certain bus if it there is no explicit user choice. This would especially be useful when you share an SPI with an sdcard and you want to make sure none is present)
> 4. hal+settings hands off the spi bus and pin configuration to the drivers which initialize
> 
> That way the existing pattern where can needs to read the settings and hal would be inverted.

Thanks for taking the time, I appreciate the feed back!

In other projects I have done along the lines you suggest, HAL configures periherals, and drivers get the bus (I2C, SPI etc) along with address or CS and other per instances dependent signals (like the int pin for CAN). For multicore/threaded environemnt also a per bus mutex. 
It is quite "easy" to get where you want. My feeling though is that moving towards such a solution need a series of bite size PRs. I'm trying to navigate the way of working of the battery emulator, this is my first open source project I have decided to spend my spare time on.  

I agree, the hardware ought to be descirbed parametrically, what pins are available and what peripherals are on the board. Take the Stark for instance, it has GPIO pins available on a header, as well as a set of outputs along with CAN and CAN FD, etc. 
That's its base capabilities, then as the GPIO pins are exposed, there are a set of additional "add-ons" that can be attached and requires addtional drivers.
So, along with the board hardware, and available drivers, the user could at the first run configure what add-on are hooked up. Then the software configures the mapping from drivers to pins. 

My co-programmer (=AI) will post a more elaborate reply :)

ekholm:
This is a better frame than the one I had, and the inversion is the part that matters: today CAN reads the settings and asks the HAL for pins, where in your sketch the HAL owns the buses and hands drivers what they need. This PR moves one step along that line and stops short of the rest, so it's worth being explicit about which is which.

What's in it now: all three device classes take a `Config` at construction and don't touch the HAL at all. `Mcp2515Device` no longer takes bus pins — SCK/MISO/MOSI belong to the bus rather than to a device hanging off it, so `init_CAN()` claims them once, brings up a single `SPIClass`, and hands the device that bus plus the only two lines it actually owns, its CS and its INT. That's your step 4 for CAN, with `init_CAN()` standing in for the HAL. What's *not* in it is steps 2 and 3: bus bring-up still happens inside `init_CAN()`, which is exactly the wrong place by your argument, and I agree it belongs in the HAL.

On sharing — I went looking after your comment and I think there's a live instance on the T-CAN485 rather than just a hazard in principle. `SD_SPI_BUS()` is VSPI and `MCP2515_BUS()` defaults to VSPI, but the two use different pins: SD on 14/2/15, MCP2515 on 12/34/5. Each subsystem constructs its own `SPIClass` and calls `begin()` with its own pins, and because those reach the peripheral through the GPIO matrix, the second `begin()` re-points VSPI and the first device goes quiet. `lilygo_330` builds with `-D SDCARD`, so the combination is reachable. I haven't reproduced it on hardware — I don't have that board with a card in it — so treat it as a code reading rather than a bug report, but it matches what you describe running into.

That also changed my mind about the fix. My first instinct was an `alloc_pins()`-style claim with conflict detection, and it doesn't work: pin-level exclusion is too strict for a bus, since two devices correctly sharing SCK/MISO/MOSI would be rejected, and too lax for this bug, since the pin sets don't overlap so nothing fires. What has to be exclusive is the chip select. What the bus needs is a single owner that decides its pins once — which is your step 2.

Concretely, that would also delete something rather than add to it: this PR still carries `if (MCP2517_BUS() == MCP2517_BUS2())` → reuse the first bus, which is the shared-bus case solved once by hand for one pair of chips. A HAL-owned bus would make that special case disappear instead of being generalised. If it's useful I'm happy to write the idea up as a design note rather than build it into this PR — you have the SD case and I have the CAN one, which is a better starting point than either alone.

On free pin selection, from your step 1: agreed, and it's the piece I'd most like to see exist. It needs the board to state which pins are hardwired and which are spare, and with what capability — an input-only pin can't serve as a chip select — or the UI ends up offering something the silicon won't do. I've been working on that description separately; happy to share it if it's useful, but it isn't a dependency of this PR.

Autodetection, your step 3, is the one I'd be careful with: probing an unknown device on a shared bus can disturb whatever else is on it, and a wrong guess is worse than an unanswered question. As a fallback when the user hasn't chosen, with an explicit choice always winning, it seems right.

ekholm:
Unrelated to the review, but worth flagging before it becomes a merge surprise: #2769 touches `comm_can.cpp` and `comm_can.h`, which this PR rewrites. Whichever of the two lands first is fine by me — if #2769 goes first I'll rebase and re-express the streaming hook on the device table, where it should actually be tidier, since frame dispatch is a single point here instead of four.

Software/src/communication/can/comm_can.cpp
ekholm: Agreed, and I'd like to do it as the immediate follow-up rather than here — the classes are new in this diff, so splitting them out now would move the same code twice inside one review. Once the shape is settled, one file per controller is the obvious next step.

Software/src/communication/can/comm_can.cpp
ekholm: Good catch, and it was an inconsistency in my own diff rather than a disagreement: `Mcp2518Device` already took a bus and its pins in a config struct while the other two still reached for the HAL. All three now take a `Config` at construction and reference the HAL nowhere, which also makes them constructible in a host test.

It grew slightly past what you asked, along the line your step 4 implies. `Mcp2515Device` doesn't take bus pins at all now — SCK/MISO/MOSI belong to the bus, not to a device hanging off it, so `init_CAN()` claims them once, brings up a single `SPIClass`, and hands the device that bus plus the only two lines it actually owns, its CS and its INT. That's what `Mcp2518Device` was already doing; the other two just hadn't caught up.

`alloc_pins()` stays in `init_CAN()` — it's the runtime conflict detector and belongs with the wiring rather than the device. Though as the top-level reply says, pin-level exclusion isn't really the right tool for a bus.

Software/src/communication/can/comm_can.cpp
ekholm: Fair — the per-chip branching is gone from every operation, but not from construction, which is what you're looking at. Driving construction from a declaration instead of code is the direction (I sketched it in #2758), but it needs the board declarations to exist first, so I'd rather not fake it here.

Software/src/communication/can/comm_can.cpp
ekholm: That's the most useful of the four, because you hit it from the SD side and I hit it from the CAN side. I've answered at the top rather than here, since it turned into the main thread — short version: I think there's a live instance of it on the T-CAN485 today, and working through it changed my mind about what the fix should look like.

ekholm:
Closing the loop on the conflict flagged above: #2769 landed first, and this PR is now rebased over it (head `c173136b`, mergeable, suite 239/239).

The resolution turned out to be an argument for the device table rather than a cost: the streaming and CAN-log calls now sit in the single dispatch funnel all controllers share, so #2769's streaming works unchanged on every channel — including SPI add-on controllers — through one call site instead of one per path.

PR #2781, opened 2026-08-08
Fronius: Add total energy amount & BMS temperature to BYD MODBUS registers

malec1984:
### What
Report active power.
Report total charged/discharged counters.
Device temperature, perhaps this can be enhanced to BMS internal temperature.
Fake battery add some numbers for total charged/discharged.
Verified only on Fronius GEN24.

### Why
Better integration.

> [!TIP]

### DIFF
--- a/Software/src/inverter/BYD-MODBUS.cpp
+++ b/Software/src/inverter/BYD-MODBUS.cpp
@@ -118,9 +118,9 @@ void BydModbusInverter::handle_update_data_modbusp301_byd() {
     mbPV[305] = std::min(datalayer.battery.status.reported_remaining_capacity_Wh,
                          static_cast<uint32_t>(57960u));  //Cap to 58kWh
   }
-  mbPV[306] = std::min(max_discharge_W, static_cast<uint32_t>(30000u));  //Cap to 30000 if exceeding
-  mbPV[307] = std::min(max_charge_W, static_cast<uint32_t>(30000u));     //Cap to 30000 if exceeding
-  mbPV[310] = datalayer.battery.status.voltage_dV;  // DC inner voltage.
+  mbPV[306] = std::min(max_discharge_W, static_cast<uint32_t>(30000u));       //Cap to 30000 if exceeding
+  mbPV[307] = std::min(max_charge_W, static_cast<uint32_t>(30000u));          //Cap to 30000 if exceeding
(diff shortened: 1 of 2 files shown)


Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -115,11 +118,18 @@ void BydModbusInverter::handle_update_data_modbusp301_byd() {
...
+  mbPV[310] = datalayer.battery.status.voltage_dV;                            // DC inner voltage.
+  mbPV[311] = static_cast<int16_t>(datalayer.battery.status.active_power_W);  // DC inner power (before contactors).
   mbPV[312] = datalayer.battery.status.temperature_min_dC;
   mbPV[313] = datalayer.battery.status.temperature_max_dC;
+  // U64 for total charged/discharged Wh (314-317 and 318-321), but datalayer uses only 32-bit.
+  mbPV[316] = datalayer.battery.status.total_charged_battery_Wh >> 16;
+  mbPV[317] = datalayer.battery.status.total_charged_battery_Wh & 0xFFFF;
+  mbPV[320] = datalayer.battery.status.total_discharged_battery_Wh >> 16;
+  mbPV[321] = datalayer.battery.status.total_discharged_battery_Wh & 0xFFFF;
+  mbPV[322] = datalayer.battery.status.temperature_max_dC;  // Fill device temperature, perhaps BMS temperature.
```
dalathegreat, line 132: The temperature_max_dC is a uint16_t value, so think this needs to be cast as <int8_t>

Also since it is dC, in an int8 range it will go  between -12.7 -> 12.7 *C. so I guess it needs to be sent as C instead of dC?

dalathegreat (changes requested):

Software/src/inverter/BYD-MODBUS.cpp
malec1984: I believe it is correct because the mbPV array for modbus registers uses uint16_t
Actually I see a different issue with total discharged being a negative value and that is not correct,
there are 4 modbus registers used for that, so 318 319 320 321 ....

dalathegreat: Doh, you're right! Nice work!

dalathegreat (approved):

PR #2782, opened 2026-08-08
Move battery HTML renderers into their own translation units

ekholm:
Every renderer defined inline in a `*-HTML.h` header is emitted in whichever TU instantiates it — in practice `BATTERIES.cpp`, which makes the factory object absorb ~127 KB of renderer code. Two costs:

1. **Per-driver flash attribution is wrong.** The size bot and reviewers see renderer cost charged to `BATTERIES.o`, not to the driver that owns it — a driver PR's true flash cost is systematically understated, and `BATTERIES.o` looks mysteriously huge.
2. **`BATTERIES.cpp` is a mega-TU** that recompiles on every driver-HTML touch.

This moves the renderer method bodies into per-renderer `.cpp` files, following the pattern BMW-I3, BMW-IX, IONIQ-28 and E-GMP already use (21 converted; all 25 now uniform). Headers keep the class with declarations; trivial one-liners stay inline. Battery headers that relied on transitive datalayer includes through their HTML header now include what they use directly.

Measured on `esp32devkit_330`:

- `BATTERIES.cpp.o`: 184,211 → 56,958 B of flash sections (**−69%**)
- renderer cost now lands per driver (`TESLA-HTML.o` 27,550 B, `ECMP-HTML.o` 14,229 B, `MEB-HTML.o` 9,858 B)
- **firmware total −6,472 B** (1,807,019 → 1,800,547) — the mega-TU was defeating `--gc-sections`

No behaviour change: pure code motion. Each relocated method body was compared statement-for-statement against the inline original across all 25 renderers. Unit tests green.


jonny5532:
Nice saving - yes the current `.h` thing is a bit weird, it'll be nice to able to tell which integrations actually use the most flash.

Semi-relatedly, the new `UdsCanBattery` thing that is currently in use in MG-GEN1-BATTERY and CMFA-BATTERY moves the HTML rendering to a method on the battery, which avoids needing an extra class, and also gives the renderer access to `private` variables inside the battery to avoid needing to copy them to the datalayer thing, and avoiding extra static slots for multiple batteries. It is opt-in, and won't be useful for the more complex batteries anyway, but I can't see a downside. You could even move the HTML method to its own TU in a separate `.cpp` if desired.

ekholm:
Thanks — and the visibility point is the one I hadn't expected to get for free. Once each renderer is its own translation unit, a driver's HTML cost lands in the map file under that driver instead of being smeared across whatever happened to include the header.

On `UdsCanBattery`: I think the strongest argument for it is the one that's easiest to overlook from the call site, which is the static slots. The current renderers read the driver's `datalayer_extended` struct, and that struct is where multi-pack support gets paid for: `nissanleaf`, `nissanleaf_2` and `nissanleaf_3` are three copies inside the union arm, `KiaHyundai64`/`_2` are two more, and `bydAtto3`/`bydAtto3_2` sit outside the union entirely, in the non-zero-defaults section, so that pair is paid whether or not an Atto 3 is configured. A renderer that reads its own battery object needs none of it — the second pack is a second object, and the third costs nothing new. The private access is a pleasant secondary effect; the duplication is the part that scales badly.

I don't think the two ideas compete, which is why I'd rather not fold one into the other here. This PR only changes *where* renderer code is compiled, not *who owns the renderer*. A battery that renders from a method still wants that method in its own `.cpp` rather than in a header every consumer pulls in — which is exactly your last sentence, and it's what I'd do. So the split applies to the renderers that exist today, and a driver converted to the `UdsCanBattery` shape later simply keeps its `.cpp`.

Where I'd be careful is the conversion itself. Moving an existing driver to a method-based renderer means untangling its `datalayer_extended` entry at the same time, since that is what the renderer currently reads and what the `_2`/`_3` copies exist for. That is per-driver work with a real blast radius, and the sort of thing that wants to happen deliberately rather than opportunistically alongside a mechanical move. Landing the TU split first keeps those two questions separable.

jonny5532:
> why I'd rather not fold one into the other here

Ah I wasn't suggesting you should, I was just thinking aloud and wondered if you saw any great problem with it, but sounds like you don't.

Indeed switching to `UdsCanBattery` (or some other in-battery-class HTML rendering) is quite intrusive so is sensible to do slowly.

ekholm:
Rebased onto current main. Two renderers left the PR along the way — upstream's UdsCanBattery migration replaced the Zoe Gen1 and CMP Smart renderers with the generic UdsBatteryHtmlRenderer, so their decomposition files are dropped and 19 of the original 21 conversions remain (all 23 renderers in the tree now uniform). The Nissan LEAF renderer body was re-extracted from current main so none of the recent LEAF display improvements are disturbed. Savings re-measured on esp32devkit: −5,760 B against today's main.

Issue #2711, opened 2026-07-30
Parallel battery safety: voltage-diff gate is one-directional — battery 1 can (re-)close onto a live link unchecked


ekholm:
The 1.5 V parallel-join rule in `check_parallel_battery_safety()` only gates battery 2/3 toward battery 1. Battery 1's own closing is never voltage-checked, which breaks the invariant in the reverse direction: if battery 1 drops out (BMS reset, WUP glitch, fault recovery) while battery 2 keeps its contactors closed, battery 1 later re-closes through its normal path with no check against the live link — and the resulting surge is limited only by the two packs' internal resistances.

The lower layers don't cover it: the GPIO precharge is time-based, so the resistor only equalizes dead-link capacitance — against a link held stiff by another closed pack, POS still closes after the timer with whatever ΔV remains. And pack-internal windows (e.g. MEB's ~20 V) are far looser than 1.5 V.

Proposed fix (happy to PR): make the gate symmetric. `check_parallel_battery_safety()` additionally computes a main-battery permission (blocked while battery 2/3 is engaged and ΔV > 1.5 V); `handle_contactors()` consumes it in the START_PRECHARGE transition, and the main battery instances of CAN-commanded packs get the same gating pointer their battery-2 instances already take. Healthy startups are unaffected — battery 2 is never engaged before battery 1 in the normal flow — so this only bites in the recovery scenarios where the surge risk actually exists.

The design question I'd like input on before building: how to determine "another pack is closed". For the GPIO second-battery output the state is known; for CAN-commanded packs I'd add a Battery API function in the style of the existing capability virtuals: `virtual ContactorState reported_contactor_state() { return ContactorState::Unknown; }` with a tri-state enum (Unknown/Open/Closed). Drivers that parse contactor state (i3, Tesla, MEB, ...) override it; the gate keys on REPORTED Closed (a commanded-but-failed close must not deadlock the main pack), and Unknown falls back to BE-commanded state. Per-instance for free on double/triple setups, and the web UI could reuse the same call for a contactor-status display later. Reasonable?

ekholm:
Following up on the API question above — the implementation exists now, and a shape is easier to judge than a description.

[...]

PR #2783, opened 2026-08-08
Parallel battery safety: make the 1.5 V join gate symmetric

ekholm:
Fixes the one-directional gate reported in #2711.

The 1.5 V parallel-join rule in `check_parallel_battery_safety()` only gates battery 2/3 toward battery 1. Battery 1's own closing is never voltage-checked, which breaks the invariant in the reverse direction: if battery 1 drops out (BMS reset, WUP glitch, fault recovery) while battery 2 keeps its contactors closed, battery 1 later re-closes through its normal path with no check against the live link — and the resulting surge is limited only by the two packs' internal resistances.

The lower layers don't cover it: the GPIO precharge is time-based, so the resistor only equalizes dead-link capacitance — against a link held stiff by another closed pack, POS still closes after the timer with whatever ΔV remains. And pack-internal windows (e.g. MEB's ~20 V) are far looser than 1.5 V.

**What this does**

`check_parallel_battery_safety()` additionally computes a main-battery permission, false while another pack holds the link and the packs differ by more than 1.5 V. `handle_contactors()` consumes it in the `START_PRECHARGE` transition, and the main instances of the CAN-commanded packs that already carry a gating pointer (i3, KiaHyundai64, Pylon) take it.

Healthy startups are unaffected — battery 2 is never engaged before battery 1 in the normal flow — so this only bites in the recovery scenarios where the surge risk actually exists. The gate is close-only: it never opens an already-closed main pack, because opening the main under load is its own hazard.

**"Another pack is closed" keys on what the pack reports**

Per the API discussion in #2711: `enum class ContactorState : uint8_t { Unknown, Open, Closed };` with `virtual ContactorState reported_contactor_state()` on `Battery`. Drivers that parse contactor state override it — i3 (precharge counts as Closed: the link is already energizing), Tesla (CLOSING/CLOSED/WELDED → Closed, SNA stays Unknown), MEB (the `dc_bus_live` mode set → Closed, Error → Unknown). Unknown falls back to the emulator-commanded state, so a commanded-but-failed close cannot deadlock the main pack. Kia E-GMP has a `contactor_status` parse whose semantics I could not verify against a log, so it is deliberately left un-overridden rather than guessed.

The `batteryNumber == 2` and `== 3` blocks were verbatim copy-paste (the battery-3 block's comment still said "second battery"). They are extracted into one parameterised helper with the new main-battery gate as its third consumer — same shape as the `check_can_component_alive()` helper from #2688.

**Open item for discussion**

Tesla's main instance is **not** wired to the gate. Tesla has no `contactor_closing_allowed` member or consumption path, and adding one means touching its TX command logic — that felt like a separate change rather than something to smuggle into a safety fix. Tesla still *reports* its contactor state, so it correctly blocks other packs; it just isn't itself blocked. Happy to add it here or follow up, whichever you prefer.

**Testing**

Six host tests: main close blocked at ΔV > 1.5 V against an engaged pack; allowed within the window; allowed regardless of ΔV when no pack holds the link; the existing battery-2 gating unchanged (regression guard); Unknown reported state falling back to commanded; and the gate holding the FSM in DISCONNECTED. Verified negatively — the three gate tests fail on unfixed code. Full suite green, `-Werror` green.


jonny5532:
This is an important improvement and should definitely improve the safety of double/triple battery. 

Some random notes:

1. I found the code very tricky to understand, it took careful working through, with all the pass-by-reference return values and statics. I do wonder if it could be expressed more simply - the old code, whilst lacking, was at least easy to understand.
2. It does look largely correct as well as my meat-brain can determine. I then fed your AI into my AI and the main observation was that zero voltage (and the 370V sentinel) values bypass the checks, but I can't actually conceive a scenario where that wouldn't trigger bigger alarms from the other safety checks.
3. This will need the `reported_contactor_state` method adding to MG-GEN1, which I can do, as well as making it acknowledge the new battery1 flag. Probably easiest to do as a follow-up before the next release.
4. One scenario this doesn't address is batteries that are both CAN controlled, and are using the GPIO interconnect contactor (although neither did the original code). The additional contactor is highly recommended, although not strictly necessary, and introduces some sequencing requirements (ie, it should close the pack contactors first, and then gate the interconnect contactor on the voltage diff AND the pack contactor closure). Should probably be a separate PR anyway.

ekholm:
Thanks — that's a more careful read than a PR this size usually gets, and point 2 turned out to be the important one.

**The sentinel (2).** Your AI was right, and the case is sharper than "bypasses the checks". In this PR the skip sits *above* the line that computes `main_blocked_by_joiner`, so it doesn't only skip the old joiner-side check — it disables the gate this PR adds. `battery1_allowed_contactor_closing` defaults to true on purpose, so single-battery systems are unaffected, which means a system that boots with either pack reading exactly 3700 leaves the main battery free to close onto a live link at any voltage difference. The same early return also makes the gate sticky: while a pack reads the sentinel, `main_blocked_*` keeps whatever value it last held.

As for why something bigger doesn't catch it — I don't think anything does, and that's the problem. 3700 dV is the datalayer's "not decoded yet" default, but it is also an ordinary reading for a pack sitting at 370.0 V, and nothing treats that as a fault. The existing test in #2725 shows the joiner-side consequence directly: at exactly 370.0 V the 10-second disengage never happens and no other event is raised.

So I've rebased this onto **#2725**, which turns that skip into a startup grace that latches once both packs have been seen off the sentinel. Because your refactor collapses the two duplicated blocks into one helper, the fix now lands in a single place rather than twice — the ordering works out in the refactor's favour. Two tests fail against this PR as filed and pass after the rebase: the existing `Battery2DisengagesWhenMainSitsAtTheSentinelVoltage`, and a new `MainGateStillEngagesWhenAPackSitsAtTheSentinelVoltage` that latches, drops the main pack to exactly 370.0 V with the engaged pack 25 V away, and requires the gate to engage.

Zero voltage I've left alone — that one is a genuine absence of data rather than a value the hardware can legitimately produce.

**Readability (1).** Fair, and I'd rather fix it than defend it. The state is now a single data-only `ParallelJoinState` in the header — the two startup-grace latches, the two drift timers and the two main-gate blocks in one declaration, with the behaviour staying in the free functions, following the shape the safety layer uses. Reinitialising is one assignment of a fresh instance rather than a list of fields that the next one added would quietly escape, and a test can now set a latch directly instead of driving the check through N ticks to reach a precondition.

That leaves the out-parameters, which are the part you actually had to work through. There are four, and the tidier shape is probably to group the per-joiner fields into a nested struct — `{voltages_seen, seconds_out_of_sync, main_blocked}` — so the helper takes one reference to a joiner's state and returns its permission, instead of four separate by-reference arguments. That also happens to be the shape this wants if it ever becomes an array over N packs. I'd rather agree it with you than guess: happy to do it here before merge, or as a follow-up.

**MG-GEN1 (3).** Please do, and thank you — a follow-up before the next release sounds right. The default `reported_contactor_state()` returns `Unknown`, which falls back to the BE-commanded state, so MG-GEN1 is no worse off than today until it lands; it just doesn't get the benefit of the BMS's own view.

**Interconnect contactor (4).** Agreed, separate PR. Your sequencing description is the part I'd want written down before anyone implements it — close the pack contactors first, then gate the interconnect on voltage difference *and* pack-contactor closure. That's a different state machine rather than another condition on this one, and expressing it here would make an already-tricky function worse.

Two practical notes.

This PR now carries #2725's four commits underneath its own two — GitHub won't let a fork PR use a fork branch as its base, so it still shows `main` and the diff includes both, the same way #2716 sits on #2715. #2725 wants reviewing first; this PR's own change is `9a721144` and `e04d4963`.

And if you check the host suite with `ctest`, it won't show state leaking between cases — `gtest_discover_tests` gives each case its own process. Running `./tests` directly puts them all in one process, which is where an unreset latch actually bites. That is worth knowing independently of this PR: on current main, `MillisWrapTest.InverterStartupGraceDoesNotRearmAfterWrap` passes under `ctest` and in isolation, but fails when run after `InverterAliveTests` in the same process. Nothing to do with this change — I hit it while rebasing and it reproduces on an unmodified `main` checkout — but it is the same class of problem, and CI cannot see it.

nagyrobi:
Claude's opinion:

I pulled the branch (`ekholm:fix/parallel-join-symmetry`) and read the merged state, not just the PR-as-filed diff, since it's stacked on #2725.

**Short version:** the bug it identifies is real and worth fixing, the `ContactorState` abstraction is a genuinely good addition, and the tests are real tests. But as filed it introduces a worse hazard than the one it fixes, in exactly the drivers it wires up. I would not merge it yet.

**The main problem — the "close-only" claim is false for the CAN path**

The PR states the gate never opens an already-closed main pack. That holds for `handle_contactors()`, which only reads the flag at the `DISCONNECTED → START_PRECHARGE` transition. It does not hold for the drivers it points at `battery1_allowed_contactor_closing`:

- `BMW-I3-BATTERY.cpp:427` — flag false → `BMW_10B.data.u8[1] = 0x00`, "keep contactors open", sent continuously. On an already-closed pack that is an open command.
- `KIA-HYUNDAI-64-BATTERY.cpp:227/239` — flag false → stops transmitting the frames that hold it closed.

And `main_blocked_by_joiner` has **no debounce at all**, unlike the joiner side's 3 s alert / 10 s disengage. One bad voltage decode with battery 2 engaged → main i3 pack commanded open, under load, for up to a second. That's the exact hazard the e-stop path in that same file goes to real lengths to avoid (waiting for `PAUSED`, current below 1.8 A, arcing/welding comment). Two paralleled packs with independent sensors disagreeing by >1.5 V transiently is not exotic.

Fix is cheap: only assert the block when the main is not already closed (`contactors_engaged == 0` / `contactorStatus == DISCONNECTED`), and debounce it if it's going to feed a continuous enable.

**Second: the gate isn't checked at the moment that matters**

It's sampled once leaving `DISCONNECTED`, then the FSM runs on timers through `NEGATIVE_CONTACTOR_TIME_MS` + `precharge_time_ms` to POS closure with no re-check — and the flag only updates at 1 Hz. The surge happens when POS closes. Re-checking in the `POSITIVE` case is a two-line change.

**Smaller, but checkable**

- Pylon is listed as a consumer. `PYLON-BATTERY.cpp` never reads `contactor_closing_allowed` — the wiring is inert.
- MEB's `reported_contactor_state()` maps `default:` to `Open`, so `Init`, `Balancing` and any out-of-range `BMS_mode` fail open. Tesla and i3 correctly map their default to `Unknown`. Inconsistent, and the wrong direction for a safety path.
- Main blocking raises no dedicated event. If a pack reports Closed permanently (welded), the main is locked out forever with only `EVENT_VOLTAGE_DIFFERENCE_BAT2` to explain it.
- The test re-declares `enum State` and `extern State contactorStatus` to match the FSM's file-scope definition. Works today, ODR violation, silently breaks if anyone reorders that enum.
- The branch is behind main (missing `safety_current_range_dA`), and the diff carries #2725's commits, so the thing under review is 14 files rather than the two commits that matter.

**The meta point**

jonny5532's review was good and the author's response to it was substantive — that's a healthy exchange, and the AI-assistance disclosure is the right call. But the description is long, confident, and two of its load-bearing claims don't survive a grep. That's the failure mode worth naming: a reviewer reads "the gate is close-only" and doesn't go check `BMW-I3-BATTERY.cpp:427`. For a change that decides when contactors close onto a live pack, the description needs to be treated as a claim to verify, not a summary to trust.

PR #2784, opened 2026-08-09
Renault Zoe Gen1: DTC Read feature (UDS broadcast)

Dibblah1:
### What
* **Jonny (@jonny5532)**: (Thanks for this MASSIVE piece of work - I wouldn't have even known where to start!) Migrated `RenaultZoeGen1Battery` from legacy `CanBattery` to `UdsCanBattery`, replacing custom ISO-TP state machine code with automated `handle_pid()` UDS polling.
* **Allan (@Dibblah1)**: 
  * Decoded fast telemetry from broadcast CAN frames (`0x155`, `0x424`, `0x425`, `0x42E`, `0x445`, `0x654`) to eliminate active polling overhead.
  * Added KWP2000 DTC header validation (`0x59 <mask >`) and payload offset handling to `UdsCanBattery` when `pid_scan_id_bytes == 1`.
  * Implemented Extended Session `0xC0` DTC readout sequence and raw 6-digit hex DTC code formatting (`0463F1`).
  * Added 2-bit diagnostic monitor fault thresholds (`>= 2 ? "FAULT" : "OK"`) and optimized HTML string allocation (`content.reserve(270)`).
  * Aligned synthetic CAN test logs with authentic vehicle CAN bytes.
  * Battery still alive checks now explicitly look for the heartbeat

### Why
* **Standardisation**: Aligns Renault Zoe Gen1 with the repository's `UdsCanBattery` framework.
* **Efficiency**: Removes unnecessary PID polling/calculation for telemetry already broadcast by the BMS; reduces memory footprint. Only one trigger is required for CAN_STILL_ALIVE - the heartbeat tells us there's a battery there.
* **Functionality**: Adds DTC readout / clear, human-readable diagnostic status formatting, and raw hex DTC display.

### How
* Inherited `RenaultZoeGen1Battery` from `UdsCanBattery` and configured `set_pid_scan_mode(PidScanMode::OneByteLocalId)`.
* Decoded broadcast telemetry in `handle_incoming_can_frame()` and restricted active UDS PID scan list.
* Updated `UdsCanBattery::handle_dtc_response()` to accept KWP2000 2-byte response headers when `pid_scan_id_bytes == 1`.
* Overrode `read_DTC()`, `on_uds_sequence_step()`, and `get_dtc_standard_code_string()` in `RenaultZoeGen1Battery`.
* Aligned synthetic test capture logs with authentic vehicle CAN bytes.
* Formatted 2-bit diagnostic monitors with `>= 2` fault evaluation and buffer reservation.
* Removed CAN_STILL_ALIVE triggers from other pathways apart from heartbeat.
.

> [!TIP]


Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
```diff
@@ -631,3 +228,16 @@ void RenaultZoeGen1Battery::setup(void) {  // Performs one time setup at startup
...
+static const uint16_t ZOE_STATE_OPEN_SESSION = 1;
+
+void RenaultZoeGen1Battery::read_DTC() {
+  send_sequence_message(ZOE_STATE_OPEN_SESSION, SID::DiagnosticSessionControl, (const uint8_t*)"\xC0", 1, 20, 2);
+}
+
+void RenaultZoeGen1Battery::on_uds_sequence_step(uint16_t state, uint8_t sid, const uint8_t* data, uint16_t len) {
+  if (state == ZOE_STATE_OPEN_SESSION && sid == UDS_RESPONSE_SID_OF(SID::DiagnosticSessionControl)) {
+    // Session 0xC0 granted! Transmit UDS ReadDTCInformation with status mask 0xFF
+    send_sequence_message(UDS_STATE_READ_DTC, SID::ReadDTCInformation, (const uint8_t*)"\x02\xFF", 2, 20, 2);
```
malec1984: Reading DTC with mask FF is usually not a good idea because you will get all the DTCs list, normally you query for active and pending DTCs. On the other hand I don't know how Renault has implemented the DTC read routine.

Dibblah1: Yep, agreed and it's a choice which I've defaulted to "all the things". In this case, this is only triggered when a user presses the "Read DTC" button - it's not run on a schedule. Completely agree that if this *was* run on a schedule that it should probably filter better, but this just makes the button functional / available for now.

In my case, this brought up 1 current and one historic DTC. Clearing the historic DTC worked and removed it from the list. So the code does work, it's just not the right thing to run automated.

Was that what you meant?

malec1984: I know it's on the user button, but with FF mask the ECU should return a very long list of DTCs, if it's not the case here then Renault Implementation is probably not according to the autosar standard

Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
```diff
@@ -631,3 +228,16 @@ void RenaultZoeGen1Battery::setup(void) {  // Performs one time setup at startup
   datalayer_battery->info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
   datalayer_battery->info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
 }
+
+static const uint16_t ZOE_STATE_OPEN_SESSION = 1;
+
+void RenaultZoeGen1Battery::read_DTC() {
+  send_sequence_message(ZOE_STATE_OPEN_SESSION, SID::DiagnosticSessionControl, (const uint8_t*)"\xC0", 1, 20, 2);
```
jonny5532: I expect it works mostly fine, but strictly this shouldn't do the `send_sequence_message` here, since it'll be called from the webserver thread and the inner logic isn't thread-safe.

The pattern that it uses internally is:

```C++
void UdsCanBattery::reset_DTC() {
  start_sequence(ZOE_STATE_RESET_DTC_START);
}

void RenaultZoeGen1Battery::on_uds_sequence_step(uint16_t state, uint8_t sid, const uint8_t* data, uint16_t len) {
  if (state == ZOE_STATE_RESET_DTC_START) {
     ... do the send_sequence_message here ...
  } else if ... etc etc
```

ie, it just calls 'start_sequence' in the `reset_DTC`, with a new state `..._START`, and then the `on_uds_sequence_step` is called with that new state (from another thread) and actually does the send.

Also avoids issues like trying to reset_DTC whilst something else (like a PID scan) is occurring.

Software/src/battery/RENAULT-ZOE-GEN1-BATTERY.cpp
jonny5532: Yeah we only store 32 DTCs, so if there are more than that they'll be truncated (and you might miss active ones).

It is handy to see the lot when debugging, but I'd recommend reducing the number shown to users in the release builds, or you'll just get endless questions about why there are some many DTCs...

jonny5532:
LGTM!

The other batteries use an enum and switch in the on_uds_sequence_step for clarity, although there's only two states so the way you've done it will work fine.

dalathegreat (approved):
Nice work! ZoeGen1 users rejoice! 🙌

PR #2788, opened 2026-08-09
Autodetect build version

ngehrsitz:
### Why
When switching between multiple debug builds the current versioning scheme with a static `12.3.dev` does not tell you what version you are actually running. This means it is possible to accidentially flash the wrong version to a board without a way to notice it.

### What
Therefore with this PR we now read the build information from the environment and pass that information along into our code. We can then use `ifdef`s to emit a build version. There are three main types of versions:
1. Release builds that are directly built from a tag -> we just use that
release_build

2. PR builds where we reference the last tag in the history to determine the version. This is arguably more correct than the current statically incremented version since this tells you which changes are already included, whereas `12.3.dev` does not tell you anything about what changes after `12.2` are included. We also reference the bash sha instead of the default merge sha. This is less precise but in my opinion more useful since you can directly compare against the PR.
pr_build

3. All other builds like a push use a fallback pattern where we use whatever information is still avialable. As long as a full checkout is present `GIT_ANCESTOR_TAG "dev-" GIT_SHORT_SHA` should always work.

I also added a link so that you can directly go to the Release and PR builds, otherwise this change would even be flash negative.

### How
By adding a small python script that calls git and reads the environment vairables set in the Github workflow.

> [!TIP]

### DIFF
--- a/Software/src/devboard/webserver/webserver.cpp
+++ b/Software/src/devboard/webserver/webserver.cpp
@@ -990,7 +990,18 @@ String processor(const String& var) {
     // Start content block
     content += "<div style='background-color: #303E47; padding: 10px; margin-bottom: 10px; border-radius: 50px'>";
     content += "<div id='bxUpd' style='text-align:center'></div>";
-    content += "<h4>Software: " + String(version_number);
+    content += "<h4>Software: ";
+#if defined(GIT_TAG) && defined(GITHUB_ORG) && defined(GITHUB_REPO)
+    content += "<a href='https://github.com/" GITHUB_ORG "/" GITHUB_REPO "/releases/tag/" GIT_TAG
+               "' target='_blank' style='color:#fff'>" +
+               String(version_number) + "</a>";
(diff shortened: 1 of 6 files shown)


Software/src/devboard/webserver/webserver.cpp
```diff
@@ -1685,7 +1696,11 @@ String processor(const String& var) {
     // In-UI update notification (browser-side; skips dev builds, 6h cached) - issue #1660
     content += "<script>";
     content += "(function(){var cur='" + String(version_number) + "';";
-    content += "if(cur.indexOf('dev')>=0)return;";
+#ifdef GIT_TAG
+    content += "if(false)return;";
+#else
+    content += "if(true)return;";
+#endif
```
ngehrsitz, line 1703: The reason for not removing the script for PRs is that release builds would then be larger than PR builds which we obviously don´t want.

nagyrobi:
I was thinking the other day that we need something for this!

dalathegreat (approved):
Cool improvement! Let's try it out!

PR #2789, opened 2026-08-09
Move BMS Power to GPIO45 on T-2CAN FD to avoid pin conflict

brynrichards:
### What
This PR moves the BMS Power pin from GPIO3 to GPIO45 on T2-CAN FD devices.

### Why
Currently the BMS Power pin is not usable on the T2-CAN FD due to a pin conflict on IO3.   Attempting to use it gives the following error:

       0.225 2CAN FD detected (ret=0)
       0.225 GPIO conflict for pin 3 between CANFD and BMS power.
       0.226 Event: GPIO Pin Conflict: The pin used by 'CANFD' is already allocated by 'BMS power'.

Even if the second CAN is configured off and this error is bypassed, the BMS Power pin on GPIO3 never signals high and is unusable.

GPIO45 was chosen because it is unused by any other code on the T2-CAN FD (Chademo support uses a number of the other pins not currently labelled on the wiki).

### How
There is existing code to detect FD devices is_fd(), this code reuses that method to assign GPIO45 rather than GPIO3 on FD devices only.

Tested locally using an SSR connected to GPIO45 and confirmed it signals correctly.

Note: The wiki will need to be updated to reflect the pin assignment on the FD here https://github.com/dalathegreat/Battery-Emulator/wiki/Hardware:-LilyGo-T%E2%80%902CAN#expansion-header,  and in the second bullet point here https://github.com/dalathegreat/Battery-Emulator/wiki/Hardware:-LilyGo-T%E2%80%902CAN#hardware-basics to remove the TODO.    Happy to do this if/when this is merged in and released.

### DIFF
--- a/Software/src/devboard/hal/hw_lilygo2can.h
+++ b/Software/src/devboard/hal/hw_lilygo2can.h
@@ -81,10 +81,11 @@ class LilyGo2CANHal : public Esp32Hal {
     if (user_selected_gpioopt1 == GPIOOPT1::ESTOP_BMS_POWER) {
       return GPIO_NUM_2;
     }
-    return GPIO_NUM_3;
+    // GPIO3 is used as INT1 on a T-2CAN FD (MCP2518), so use GPIO45 instead.
+    return is_fd() ? GPIO_NUM_45 : GPIO_NUM_3;
   }
-  // Pins to be latched across a reset/OTA reboot (RTC-capable pins only): BMS_POWER is either GPIO2, GPIO3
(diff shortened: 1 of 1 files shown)


jonny5532:
Looks sensible, IO45 seems as reasonable a choice as any.

I can redo that pinout image when this gets merged (or if it gets too confusing, make another one for the FD, but that's probably not necessary for a single pin difference).

dalathegreat (approved):

dalathegreat:
Massive thanks for the patch @brynrichards :raised_hands:

PR #2790, opened 2026-08-10
Nissan LEAF🔋: fix Hx on ZE0/AZE0; expose Hx over MQTT; add QC & AC charge count; remove BMS ID; faster startup; cosmetics

nagyrobi:
### What
This PR fixes the Nissan LEAF HX value, which always reported 0 on ZE0 and AZE0 packs, and exposes it as a new Leaf-specific MQTT sensor with Home Assistant autodiscovery.

image

Home Assistant:
image

Charge counts:
image
image

### Why
The decoder only ever implemented the ZE1 layout of the LBC group 1 reply. ZE0 (24 kWh) and AZE0 (30 kWh) answer with a two-byte-shorter layout, so those packs read the high half of the following field - a value small enough that the `/102.4` scaling always truncated to 0. LeafSpy reports a real Hx on the same packs.

### How
The announced ISO-TP length of the group 1 reply identifies the layout (`0x29` ZE0, `0x2B` AZE0, `0x35` ZE1), so the offset and scale are picked from that: payload 26..27 already in hundredths of a percent for ZE0/AZE0, payload 28..29 scaled by `125/128` for ZE1. The value is stored as `battery_HX_pptt` to keep LeafSpy's two decimals, rendered with a `%` on the battery info page, and published as `leaf_hx` (`state_class: measurement`, precision 2). Unknown reply lengths and the pre-first-poll state leave the sensor unpublished rather than reporting a false 0.

Renamed 'HX' to 'Hx' for consistency and parity with Leafspy ui.

> [!TIP]

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -1125,7 +1125,10 @@ void NissanLeafBattery::transmit_can(unsigned long currentMillis) {
     }
 
     //Send 10s CAN messages
-    if (currentMillis - previousMillis10s >= INTERVAL_10_S) {
+    //The first pass through the group list runs at the faster burst interval, so the battery info
+    //page is populated within seconds of startup rather than over the following minute.
+    if (currentMillis - previousMillis10s >=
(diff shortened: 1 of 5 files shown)


dalathegreat (approved):
Tested OK on double 30kWh

Image

Image

dalathegreat (approved):

PR #2791, opened 2026-08-10
Change identify_build.py to generate a header instead of using defines

jonny5532:
### What
The recent change here (https://github.com/dalathegreat/Battery-Emulator/pull/2788) causes a full rebuild every time you compile, due to the version vars being passed in as defines.

Instead, the script now generates a temporary header file, which is included only where needed, so only a few TUs get rebuilt.

Does rely on `__has_include` (and my favourite `#pragma once`) but I think that's OK. Initially AI'd and then rewritten a few times...

@ngehrsitz - does this seem OK? It is now possible for a stray autogenerated header, and the script not running for some reason, to result in the wrong version number in builds, but that seems unlikely.

### Why
Losing the ability to do incremental rebuilds makes testing things very slow!

> [!TIP]


Software/src/devboard/utils/version.h
```diff
@@ -1,23 +1,13 @@
...
-#if defined(GIT_TAG)
-#define BUILD_VERSION GIT_TAG
-
-#elif defined(GIT_ANCESTOR_TAG) && defined(GIT_SHORT_SHA)
-#if defined(GITHUB_PR) && defined(GITHUB_PR_HEAD_SHORT_SHA)
-#define BUILD_VERSION GIT_ANCESTOR_TAG "dev-" GITHUB_PR_HEAD_SHORT_SHA " (#" GITHUB_PR ")"
-#elif defined(GIT_BRANCH)
-#define BUILD_VERSION GIT_ANCESTOR_TAG "dev-" GIT_SHORT_SHA " (" GIT_BRANCH ")"
-#else
-#define BUILD_VERSION GIT_ANCESTOR_TAG "dev-" GIT_SHORT_SHA
```
ngehrsitz, line 16: I purposefully kept the logic how the version is defined in the code. In an earlier version I also had it defined in the script.
The tooling should only take care of fetching the constants, otherwise the version definition is detached in python code.

ngehrsitz (changes requested):
My suggested changes: 8907803226e66201ea249c445134d43b47def494

Software/src/devboard/utils/version.h
jonny5532: That makes sense, I've applied your changes (I couldn't figure out how to fetch that commit).

ngehrsitz: If you do it over git manually you need to have the other fork defined as an origin to cherry pick from. Alternatively if you have a Github plugin in your IDE you can usually directly jump to that commit in your Git view.

ngehrsitz (approved):
Let´s accelerate rebuilds

dalathegreat (approved):

PR #2795, opened 2026-08-10
MQTT: hw name and sw version, mark emulator entities as diagnostic type

nagyrobi:
### What
This PR adds the board name and running firmware version to the MQTT `/info` payload and to the Home Assistant discovery device block, and publishes the version as its own recorded sensor.
It also marks the emulator-level entities as `entity_category: diagnostic` and renames the restart button to "Reboot Emulator" - to be consistent with the UI.

image

image

image

### Why
User requested on discord to save the software version in a Home Assistant sensor, which can be tracked over time. This is useful to keep track of when the software was upgraded, and easily watch for behavior changes on the other sensor reports.
The emulator's own status entities were mixed in with the battery sensors on the device page `entity_category` declutters the list.

### How
`set_common_discovery_attributes()` now emits `hw_version`/`sw_version` from `esp32hal->name()` and `version_number`; the same two values go out as `hardware` and `software_version` keys on `<hostname>/info`, backing a new `Emulator Version` sensor.

Both values are const char* literals with static storage duration, so ArduinoJson stores them zero-copy and the document pool does not grow.

`publish_sensor_discovery()` gained a `diagnostic` flag, set for every `globalSensorConfigTemplate` entry and applied separately to the Event sensor and the reboot button.

> [!TIP]

### DIFF
--- a/Software/src/devboard/mqtt/mqtt.cpp
+++ b/Software/src/devboard/mqtt/mqtt.cpp
@@ -199,7 +199,8 @@ static const SensorConfig globalSensorConfigTemplate[] = {
     {"event_level", "Event Level", "", "", always},
     {"emulator_status", "Emulator Status", "", "", always},
     {"emulator_uptime", "Emulator Uptime", "s", "duration", always},
-    {"cpu_temp", "CPU Temperature", "°C", "temperature", always}};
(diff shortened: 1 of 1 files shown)


ngehrsitz (approved):

dalathegreat (approved):
Excellent improvement 👍

PR #2797, opened 2026-08-10
Wi-Fi 🛜: always associate with the strongest AP, log the BSSID

nagyrobi:
## What
This PR makes the Wi-Fi station always associate (at boot and on reconnect) with the strongest access point in a mesh/multi-AP network where the APs broadcast the same SSID.
Reworks the "Wi-Fi connected" log line: the IP address is dropped and the BSSID added.

## Why
The driver defaults to `WIFI_FAST_SCAN`, which stops at the first matching AP it hears, so a device in such a network can latch onto a distant one and stay on a weak link.
The IP address is already printed by the following "Wi-Fi Got IP" line, while the BSSID - the only way to tell which AP was joined - was missing.

## How
`WiFi.setScanMethod(WIFI_ALL_CHANNEL_SCAN)` + `WiFi.setSortMethod(WIFI_CONNECT_AP_BY_SIGNAL)` before the first `WiFi.begin()`; both land in the STA config, so `WiFi.reconnect()`, `FullReconnectToWiFi()` and the driver's own auto-reconnect reuse them.
SSID and BSSID are read from the `STA_CONNECTED` event payload instead of `esp_wifi_sta_get_ap_info()`, dropping two `String` allocations; the inverted RSSI sign is fixed at the same time.

> [!TIP]

### DIFF
--- a/Software/src/devboard/wifi/wifi.cpp
+++ b/Software/src/devboard/wifi/wifi.cpp
@@ -364,10 +364,8 @@ void onWifiConnect(WiFiEvent_t event, WiFiEventInfo_t info) {
   clear_event(EVENT_WIFI_DISCONNECT);
   set_event(EVENT_WIFI_CONNECT, 0);
   connected_once = true;
-  // SSID and BSSID are taken straight from the event payload: no esp_wifi_sta_get_ap_info()
-  // round trip and no String allocation. The BSSID identifies which AP we landed on when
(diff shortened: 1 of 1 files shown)


ngehrsitz (approved):
👍

dalathegreat (approved):
👍 LGTM, nice find!

PR #2799, opened 2026-08-11
Safety⚠️: Report temperature events for every battery, with correct values

nagyrobi:
### What
This PR fixes the three temperature checks in `update_machineryprotection()`: event data is widened to `int16_t`, `EVENT_BATTERY_TEMP_DEVIATION_HIGH` is no longer latched, and battery 2/3 are checked as well.
Events tied to one pack now render as `" (Battery N)"` in the log, syslog, events page, MQTT and ESP-NOW.

### Why
`data` was `uint8_t` while temperatures are `int16_t` deci-Celsius, so OVERHEAT reported 50.1 °C as 245 and FROZEN reported -25.1 °C as 5; the deviation event was latched but its `else` branch called `clear_event()`, which only acts on `EVENT_STATE_ACTIVE` and could never release it.
Battery 2 and 3 had no temperature checks at all, and with the three enums being global there was no way to tell which pack raised one.

### How
All configured packs are evaluated in one pass in `check_battery_temperatures()` — required because the enums are shared, so a per-battery set/clear pair would let a healthy pack clear another pack's event (same reasoning as the existing `EVENT_CAN_CORRUPTED_WARNING` block); deviation is computed per pack and the worst offender is reported.
`EVENTS_STRUCT_TYPE` gains a `battery` field and keeps its 16-byte size, so there is no RAM cost; on ESP-NOW key `0xA5` is retired rather than redefined and `ESPNOW_KEY_EVENT_DATA_I16 = 0xAA` allocated, per the compatibility rules in `espnow.h`.

> [!TIP]


ekholm:
The `uint8_t data` truncation is a good find — 50.1 °C arriving as 245 and -25.1 °C as 5 is the kind of thing people would have squinted at in the log for a long time without connecting it to the field width.

Worth flagging an overlap: #2775 changes the same function, though different events — it fixes `EVENT_CELL_DEVIATION_HIGH` and `EVENT_SOH_DIFFERENCE` being set *and cleared* inside the per-pack blocks, so the last block to run decided the outcome for all of them. `safety.cpp` is the only file we both touch. Mine is the smaller diff, so if yours lands first I'll rebase onto it rather than the other way round.

One direct contradiction, better named now than found by CI: #2775 adds a test `TempDeviationIsLatched` asserting that `EVENT_BATTERY_TEMP_DEVIATION_HIGH` survives the condition going away. That is a characterization test — it pins what main does today — and you are deliberately changing it. I think you are right to: `set_event_latched()` paired with a `clear_event()` in the else branch is self-contradictory, and un-latching is the coherent reading. If yours merges first, that test goes with the rebase.

A question about the shared events rather than the temperature ones. Carrying `battery` as a single number reports the worst offender, which is exactly right when the value belongs to one pack — the hottest pack's temperature is one pack's reading. But a few of these events are about a *set*: cell deviation and SOH difference can be true of several packs at once, and naming only the worst one leaves the log implying the others are fine. Would a bit pattern suit those better — one bit per pack in `data`, cleared when the mask empties? It would also remove the "which pack cleared whose event" problem at the root, which is what #2775 is working around with set-if-any / clear-only-when-all-clean — really just the degenerate case of a mask that tracks whether rather than which.

Software/src/devboard/safety/safety.cpp
```diff
@@ -58,6 +58,84 @@ static void check_can_component_alive(uint8_t& still_alive_counter, bool& detect
...
+    }
+    if (deviation_dC > widest_deviation_dC) {
+      widest_deviation_dC = deviation_dC;
+      widest_battery = packs[i].number;
+    }
+  }
+
+  // Battery is overheated!
+  if (hottest_dC > BATTERY_MAXTEMPERATURE) {
+    set_event(EVENT_BATTERY_OVERHEAT, hottest_dC, name_the_battery ? hottest_battery : 0);
```
dalathegreat: We have the flash to spare, so why not use more Event types?

EVENT_BATTERY1/2/3_OVERHEAT
EVENT_BATTERY1/2/3_FROZEN
EVENT_BATTERY1/2/3_TEMP_DEVIATION_HIGH

Would make troubleshooting way easier IMO

ekholm: Works for me — and you're right that a log line naming the pack is what someone troubleshooting actually wants. I'll follow the same shape in #2775: `EVENT_CELL_DEVIATION_HIGH_BAT2`/`_BAT3` alongside the existing one, matching how `EVENT_VOLTAGE_DIFFERENCE_BAT2` and `EVENT_CAN_BATTERY2_MISSING` already read, rather than introducing a third naming style. It also makes that PR simpler: with an event per pack, a healthy pack physically cannot clear another pack's warning, so the masking fix stops being a guard and becomes structural.

One note for later rather than a counter-proposal: this scales as instances × event kinds, so it stays comfortable while the counts are fixed at three packs and four CAN devices, and it's the piece that would need rethinking if either ever became configurable. If that day comes, encoding the affected instances as a bit pattern in `data` would collapse it back to one event per kind — a mask can also say "packs 1 and 3" in a single line, which naming can't. Not something to change now; just flagging where the seam is.

nagyrobi: @dalathegreat this could scale per-event-type, and there are ~30 battery-specific events in the enum. Should I take the three temperature ones and the argument for like EVENT_BATTERY2_OVERVOLTAGE, EVENT_BATTERY2_UNDERVOLTAGE, EVENT_BATTERY2_EMPTY, EVENT_BATTERY2_FULL, EVENT_BATTERY2_ISOLATION, EVENT_BATTERY2_FUSE, EVENT_BATTERY2_CHG_STOP_REQ… 
???

dalathegreat: Yup, looks like there are 21x _BATTERY_ events, that could absolutely be scaled up to 1-2-3 to make it more clear for user what is going on!

nagyrobi: 17 of the 21 are raised from inside battery drivers, which don’t know which slot they’re in. 

Eg. set_event(EVENT_BATTERY_CHG_STOP_REQ, ...) in NISSAN-LEAF-BATTERY.cpp runs from the same .cpp whether the instance is battery 1, 2 or 3 — 29 call sites across 11 driver files. There’s no static way to pick between three enums there. Only TEMP_DEVIATION_HIGH and a few others are safety.cpp-only, where the pack is known at the call site. OVERHEAT and FROZEN are raised from both places.

We need the driver instance to know its index, which it currently doesn’t — it only holds datalayer_battery, a pointer to its slot. Two ways:

1.	Add uint8_t battery_index to the Battery base class, set to 1 in the default constructor and passed explicitly in the battery2/battery3 constructors in BATTERIES.cpp.

2.	Derive it in the base class by comparing datalayer_battery against &datalayer.battery / battery2 / battery3. No constructor churn, slightly uglier.

Which would you consider the better one?

dalathegreat: void set_event(EVENTS_ENUM_TYPE event, int16_t data, uint8_t battery); sounds like a good solution then!

nagyrobi: Done

dalathegreat (approved):
LGTM 👍 Signinficantly improves safety handling for double/triple setups

PR #2803, opened 2026-08-12
MQTT: run Home Assistant autodiscovery only on demand and after firmware updates

nagyrobi:
## What
- Publishes the button discovery configs from the MQTT task instead of the esp-mqtt event handler.
- Renames the setting to "Home Assistant autodiscovery at next boot:", unticks it automatically once every discovery config has been published.
- Runs discovery automatically at boot whenever the running firmware version changes, ticked or not.

## Why
- The event handler runs on the esp-mqtt task and used the same `shared_doc` / `mqtt_msg` as the publish cycle on the MQTT task, so the two could race.
- Discovery configs are retained, so the broker replays them to Home Assistant on every restart; republishing all of them at each boot costs airtime, heap and broker storage for nothing.
- Version-driven republishing keeps `sw_version` and any new or changed entities correct without the user having to remember to tick anything.

## How
- `MQTT_EVENT_CONNECTED` now only sets a pending flag; `mqtt_client_loop()` publishes and retries it, restoring single-threaded use of the shared document and buffer.
- One `uint32` NVS key (`HADISCFW`) holds an FNV-1a hash of the version string; a mismatch at boot forces discovery on, and the hash is stored only after a full round went out.
- Completion is gated on all four discovery flags, with cell voltages counted only when `MQTTCELLV` is enabled, so an incomplete round retries at the next boot instead of clearing the setting.

> [!TIP]


nagyrobi:
image

With the setting enabled or after a sw upgrade:
image
At simple reboots no more autodiscovery:
image

nagyrobi:
Reworked the UI: \"Home Assistant autodiscovery\" is now a switch that stores nothing of its own (it derives from the options below it), with \"Publish at firmware updates\" and \"Publish at next boot\" as the two triggers. Topic field and both checkboxes only show while the switch is on. Publish logic in mqtt.cpp unchanged.

image

nagyrobi:
[Doc updated](https://dalathegreat.github.io/Battery-Emulator-Wiki/setup/software/mqtt/)

dalathegreat (approved):

PR #2806, opened 2026-08-12
Github: Speedup by removing compiler build, move werror to all builds

dalathegreat:
### What
This PR speeds up the build system by removing the specific compiler build

### Why
Slow to build

### How
The only thing it did was use -Werror , we can now run -Werror on all builds and skip a separate compiler warning build

> [!TIP]


ngehrsitz (approved):
Great if we can get rid of this target. Instead of disabling the errors on the S3 we could also limit the check to our own code: 3e2bc471750cedd0bc8386f15b46d847be125750
That has surfaced that there was an issue on the waveshare.

PR #2807, opened 2026-08-12
Tesla: Fix BMS signal values

greenoem:
### What
This PR implements a fix for incorrect decoding of BMS_state and BMS_uiChargeStatus (since renamed by Tesla as BMS_userChargeStatus but not changed in this PR) - verified against both live car data and firmware investigation from 2021 through 2026, and running on multiple live BE instances for 10 months.

### Why
The corresponding states/status are listed wrongly in BE (e.g. SUPPORT instead of DRIVE).

### How
Fixes the target bit extraction.


dalathegreat (approved):
Thanks for the fix @greenoem 🙌

PR #2809, opened 2026-08-12
MG: Make all Gen1 batteries do 10/20ms pings

jonny5532:
### What
It seems that some MG Gen1 batteries are happy with a slower ping rate (100ms) but some are not and need the 10/20ms rate. Make all MG Gen1 batteries use the 10/20ms rate to avoid future issues.

### Why
Should solve contactor closing issues on MG Gen1 batteries.

(The original 100ms rate was a seemingly-ill-advised optimization to reduce CPU and bus load)

> [!TIP]


vtamm:
On the mainline 12.5.0 firmware contactors did not close on my two MG 69.9 kW batteries connected via Stark CMR v2. I can confirm that with the fix from this PR applied to 12.5.0 (you provided a .bin file to us on Discord) contactors now open and close properly.

PR #2811, opened 2026-08-13
MQTT: optional heap metric diagnostics for Home Assistant

nagyrobi:
## What
Adds four optional Home Assistant diagnostic sensors over MQTT: Heap Free, Heap Max Block, Heap Min Free and Heap Fragmentation.

image

They are published on the `/info` topic and in the autodiscovery pack, behind a new "Publish heap metric diagnostics:" checkbox in the MQTT settings.

## Why
Makes long-term memory behaviour visible in Home Assistant, so slow leaks and fragmentation build-up can be spotted on a graph instead of only in a live web UI reading.
Off by default, so nothing changes for existing users and no payload or flash cost is paid unless enabled.

## How
Four entries in `globalSensorConfigTemplate` gated by a `heap_metrics_enabled()` condition, using the same `heap_caps_*` calls and fragmentation formula as ESPHome.
The global discovery loop now honours `config.condition` (previously all entries were `always`), and the values are appended to the battery&nbsp;1 `/info` payload.
New `MQTTHEAP` bool setting wired through `comm_nvm`, `webserver` and `settings_html`.

Of course the values are useful for any other platform other than Home Assistant (eg Grafana). Autodiscovery is only helping Home Assistant showing the entities.


nagyrobi:
image

dalathegreat (approved):
LGTM 👍 Care to update the MQTT wiki page for this new feature?

PR #2813, opened 2026-08-13
Experimental ESP-IDF based Native CAN driver

jonny5532:
### What
Replaces ACAN_ESP32 with a largely-AI'd implementation that uses the ESP-IDF `twai_*` calls instead. These implement the various errata that ACAN_ESP32 doesn't.

Don't merge this as-is, just for testing!

### Why
The MG integration has had issues with bogus data received over Native CAN during OTA (tested with an ESP32 T485).

Surprisingly, switching to this has completely eliminated such errors in testing. Here for others to try before we decide what to do next.

> [!TIP]


nagyrobi:
Is this intended for ESP32 built-in CAN, or the external ones?

jonny5532:
> Is this intended for ESP32 built-in CAN, or the external ones?

Just the built-in CAN (a SJA1000-like internal peripheral). The ESP32 version has some quite nasty errata, the ESP32S3 one supposedly doesn't - but I am keen to see if it makes a difference on the ESP32S3 also, there could be some undisclosed weirdness that ACAN_ESP32 doesn't handle.

nagyrobi:
could you please update the branch to the latest main so I can test against the newest?

nagyrobi:
Testing on Waveshare ESP32-S3-RS485-CAN, Leaf AZE0 onboard CAN. Works.

Heap usage difference compared to default driver:

kép

Minimal loss on free, but gains on fragmentation.

nagyrobi:
FYI
image

It's good.

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
```diff
@@ -0,0 +1,360 @@
...
+  tseg1 -= prop ;
+  uint16_t sjw = tseg2 >> 1 ;
+  if (sjw < 1) sjw = 1 ;
+  if (sjw > kSjwMax) sjw = kSjwMax ;
+  mBitRatePrescaler = (uint8_t) preDiv ;
+  mTimeSegment1 = (uint8_t)(prop + tseg1) ; // Propagation + phase segment 1
+  mTimeSegment2 = (uint8_t) tseg2 ;
+  mRJW = (uint8_t) sjw ;
+  return true ;
+}
```
ngehrsitz: Do we really need to recompute the bitrate. This is only used for logging and not setting an invalid bitrate. But our bitrates should bitrates should be fixed known-good values or am I missing something?

Software/src/lib/esp32-twai/TWAI_ESP32.h
```diff
@@ -0,0 +1,237 @@
...
+    const uint32_t TQCount = 1 + mTimeSegment1 + mTimeSegment2 ;
+    if (TQCount == 0) return 0 ;
+    return ((1 + mTimeSegment1) * 100) / TQCount ;
+  }
+
+  // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+  //   Driver instance
+  // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+  public: static TWAI_ESP32 can ;
```
ngehrsitz: Newer chips like the C6 have two can interfaces. We should move away from the antipattern of static singleton instances for new drivers.

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
```diff
@@ -0,0 +1,360 @@
+#include "TWAI_ESP32.h"
+
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+
```
ngehrsitz, line 6: What about the IRAM compiler options: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/twai.html#performance
If we enable that we also need to switch to the isr callbacks.

Software/src/communication/can/comm_can.cpp
```diff
@@ -406,24 +400,23 @@ static void
...
-      CAN_frame rx_frame;
-      rx_frame.ID = frame.id;
-      rx_frame.ext_ID = frame.ext;
-      rx_frame.DLC = frame.len;
-      rx_frame.FD = false;
-      for (uint8_t i = 0; i < frame.len && i < 8; i++) {
-        rx_frame.data.u8[i] = frame.data[i];
-      }
+  int count = 0;
+  while (count++ < 16 && TWAI_ESP32::can.receive(frame)) {
```
ngehrsitz: Why the hardcoded `16`? `kRxBufferSize = 32`

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
```diff
@@ -0,0 +1,360 @@
...
+    }
+    if (!recovered) {
+      // Bus is still broken (e.g. shorted); keep the (bus-off) node so a later
+      // begin() can retry. Any new node creation will fail until it recovers.
+      return ;
+    }
+    twai_node_disable (mNode) ; // Should succeed once recovered
+  }
+  mNodeEnabled = false ;
+  twai_node_delete (mNode) ; // Requires the node to be disabled (bus-off state)
```
ngehrsitz: AI found that this could block the main task:
`teardownNode()` calls `twai_node_recover()` then spins up to ~100 ms with `vTaskDelay(1)` waiting to leave bus-off (`TWAI_ESP32.cpp:131-141`). This is reached from `receive_frame_can_native()` → `change_can_speed()` → `begin()` → `teardownNode()` on the `TWAI_BUS_OFF_ST` path — i.e. from the main loop.

The problem: bus-off recovery *by CAN spec* requires observing 129×11 consecutive recessive bits on a **healthy, idle** bus. On a genuinely broken bus (shorted, or we're the only node) recovery can never complete — so the loop burns the full 100 ms on exactly the failure case we most want to handle gracefully. And because the flag is still set on the next iteration, it repeats every loop pass → the main loop is throttled to ~10 Hz for as long as the bus stays broken.

Per the IDF docs, `twai_node_recover()` is **async and non-blocking**, the node **stays enabled** through recovery (no `twai_node_enable()` needed afterwards — pending TX resumes automatically), and an `on_state_change` callback fires on the `BUS_OFF → ERROR_ACTIVE` transition. So the cleaner model is:

- Register `on_state_change`; on bus-off detection call `twai_node_recover()` **once**, set a latch, and return — don't tear down and recreate the node.
- Let the driver rejoin the bus on its own once the bus is physically healthy again; clear the error flag from the state-change callback.
- Keep the delete/recreate path (with a bounded, ideally shorter, wait) only for `end()` and real speed changes — the rare cases where the bus is presumably fine.

This also dissolves the awkward `ESP_ERR_INVALID_STATE` / `mNodeEnabled` dance in `teardownNode()`: that only exists because the code insists on deleting a node that's in bus-off, which requires forcing it out of bus-off first. Not deleting it on recovery means the "disable() refuses until it recovers" constraint is never hit on the hot path.

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
jonny5532: Indeed this was entirely pointless - I hadn't looked closely at the code (I have now!).

Have removed this entirely (the logs don't need the receiver timing printed every time anyway).

Software/src/communication/can/comm_can.cpp
jonny5532: This is arbitrary, but deliberate - the tradeoff is between emptying the buffers fast enough, and trying to keep to the 1ms core loop period. 16 seems to work OK with the other drivers (the constraint is the frame processing time of the various battery/inverters).

Trying to avoid TASK_OVERRUNs has been a process of endlessly tweaking things - it has got better (the main culprit was the WiFi stack randomly making flash writes), but still happens occasionally. Perhaps equipping an agent with some real hardware would get the problem solved quicker, I may try that (although it's still very intermittent).

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
jonny5532: I've revisted the bus-off stuff, it wasn't actually working very well on real hardware. It does seem to now - rather than trying to tear it down and restart (which was the only option with ACAN_ESP32), it uses the bus-off recovery feature in ESP-IDF.

jonny5532: (Notably, bus-off recovery dosen't seem to work with the ACAN1517FD driver, that's something to fix too...)

Software/src/lib/esp32-twai/TWAI_ESP32.h
jonny5532: Good point, this was copied from how ACAN_ESP32 did it. Now more conventional.

One headache is that you can't reliably tear down the underlying ESP-IDF 'node', since you can't delete it if it is in bus-off, you have to recover first. This makes the destructor unreliable (it'll free its own memory, but may leave a stuck ESP-IDF node and make reinit impossible).

Software/src/communication/can/comm_can.cpp
ngehrsitz: Then I would declare it as `kRxBufferSize / 2` if we want to avoid runtime tracking

jonny5532: It's not specifically half-the-buffer though, it's just an arbitrarily chosen number of packets which the current integrations seem to be able to process in a millisecond or two max.

It should probably indeed be a const though (and the same across all the integrations), to make it less arbitrary...

Software/src/lib/esp32-twai/TWAI_ESP32.cpp
jonny5532: I've now made it forcibly add `IRAM_ATTR` if `CONFIG_TWAI_ISR_CACHE_SAFE` is set (which allows flash-disabled operation), and conditionally add it if you set a new define, `TWAI_CALLBACKS_IN_IRAM` (currently unset).

Indeed it'd be interesting to experiment with, firstly to see how much IRAM enabling it would use up. I'm not sure how significant any benefit might be - most batteries send <500 msgs/sec, and in theory the core loop runs at 1000/sec, so each received message triggers a lot of not-in-IRAM code anyway. The benefit would be on bursts, where the messages might arrive faster than the flash-cache can reload the callback/frame-storage (but even then, it'd be cached for subsequent messages...)

jonny5532:
There were reports of occasional panics with this branch - I got one myself (apparently, never actually saw it but a separate experiment  (recovering the coredumps from flash) showed a panic) from:

`assert(false && "should always get frame at this moment");`

in `_node_isr_main` in `esp_twai_onchip.c`

After some digging (and AI re-discovering it!), this seems to be a now-fixed, but not-yet-released, race condition (https://github.com/espressif/esp-idf/commit/10273d3c6b6c4fbbb1244c67280adcf970d6b16d).

The options for workaround are:
- Avoid enqueuing more than one transmit frame at a time, which requires more driver code, and hurts performance
- Patching ESP-IDF, given that we build it ourselves anyway.

I've added a crude patching script and a folder with a patch file in, which seems to work. This is at least transparent, and will also fail when the patch no longer applies (after eventual upgrade) - at which point the file (and potentially this whole mechanism) can be removed.

nagyrobi:
> a now-fixed, but not-yet-released, race condition 

Does it affect the original driver too, or just this one?

jonny5532:
The consequences of the race condition were an assert failure (and subsequent panic) in the driver - the patching effectively just removes the assert. 

The original driver doesn't have this mechanism at all (but also doesn't have fixes for the many ESP32 TWAI errata). I'm now pretty sure that the primarily problem is the overflow one - if the TWAI RX buffer overflows, it then starts producing corrupt frames (despite the original frames passing CRC). This has been a long term issue with BE, with lots of integrations needing workarounds for bad data, it would be remarkable if the underlying cause was this issue.

One option would be attempting to roll the many errata fixes back into ACAN_ESP32. However, looking at the ESP-IDF git commit history, there is an endless stream of fixes, with new ones appearing regularly - I think the only way we can stay on top of the ESP32's TWAI hardware issues is just to use the vendor library.

This supposedly is only an issue for the original ESP32 - the ESP32S3 has no published TWAI errata. However this PR seems to reduce corruption issues with an ESP32S3, implying that there may be other issues in ACAN_ESP32 (or perhaps, unpublished errata due to the official ESP-IDF library not hitting them due to the extant ESP32 workarounds, who knows...).

Software/src/lib/esp32-twai/TWAI_ESP32.h
jonny5532: It turns out this complicates things quite a bit - to be able to safely destruct, you need to keep some statics to store the node (which may not be deleteable), as well as a static transmit buffer (as it may still be in use), and then to unset the callbacks during destruction, incase the controller tries calling them after, eg, bus-off recovery.

Thus i'm slightly inclined to go back to the old singleton behaviour to completely avoid all this. It might be possible to use template instantiation to allow multiple CAN interface support (on the C6/P4 etc) to avoid needing the extra complexity...

jonny5532: OK, I have switched back to static singletons, allocated on first access, with one-per-hardware-controller. The more complex approach of an intermediate static layer pretty much doubled the size of the class, which seems excessive.

jonny5532:
I did some more experiments, including getting Deepseek to make a new SDK-less TWAI driver which implements all the the errata, is functionally equivalent to the ESP-IDF version (for the functionality it exposes), and is in the style of MCP2515_Lite (so FreeRTOS queues and no `node_*` stuff). (No idea if we want to consider this for merge, I have looked through it but am not going to detect subtle register read/write ordering issues etc - maybe sufficient testing would add confidence).

(which is here: https://github.com/jonny5532/Battery-Emulator/pull/20)

The first attempt still had the same CAN corruption issues under heavy load (OTA), it was only moving the ISRs and dependent code to IRAM that seems to solve it. ESP-IDF will automatically disable non-IRAM interrupts during flash-disable, so the issue wasn't that these interrupts were being called and faulting, but that corruption seems to appear if long stalls happen during CAN reception.

The robot's best guess as to what this might be is that the "drop-the-next-frame-too on any bus error" errata workaround fails to drop the correct frame if there's a huge stall in between, so a frame corrupted by the errata issue sneaks in. 

(The current sdkconfig does have TWAI-in-IRAM enabled, so ESP-IDF would similarly avoid these long stalls).

This doesn't explain any ESP32S3 issues - this has been briefly tested without problems on ESP32S3 but unclear if it would fix (ESP32S3 theoretically doesn't have the problem errata, but maybe the long stalls cause some unrelated undocumented issue).

jonny5532:
Some user feedback from running this on an ESP32S3 is that there have been no corrupt messages since (previously one every 1-2 days, has been fine for a week).

(My only theory for it improving the supposedly-errata-free ESP32S3, is that the ESP32S3 has unpublished issues with error handling, which have never been detected because everybody else just uses the ESP-IDF driver).

Still deliberating on whether to recommend this driver or the alternate mostly-AI'd raw one. Despite the undesirable flash hit, going with the ESP-IDF code is probably going to save us future headaches.

jonny5532:
I note that the currently patched fix should be in ESP-IDF v6.1, which will be out in 2-3 weeks, we should probably just wait rather than merging  this patching stuff.

ekholm:
Two things that may be useful, plus one question.

**The source confirms the overflow diagnosis.** ACAN_ESP32 only ever enables two interrupts:

```
ACAN_ESP32.cpp:274:  TWAI_INT_ENA_REG () = TWAI_TX_INT_ENA | TWAI_RX_INT_ENA ;
```

`TWAI_OVERRUN_INT_ST` and `TWAI_BUS_ERR_INT_ST` are declared in `ACAN_ESP32.h` (lines 343 and 346) but never enabled and never handled, and the ISR dispatches RX and TX only. The command register is only ever written with `ABORT_TX`, `RELEASE_BUF` and the TX request; `CLR_DATA_OVERRUN` is never issued. So once the RX FIFO overruns there is nothing that clears the condition, and the message counter stays desynced. That is `TWAI_ERRATA_FIX_RX_FIFO_CORRUPT` and `TWAI_ERRATA_FIX_RX_FRAME_INVALID` both missing, which matches what you are seeing.

For completeness, `comm_can.cpp` does react to `TWAI_ERR_ST` (line 408) and to bus-off (line 403), but there is no overrun path at all.

**BE has run the ESP-IDF driver before.** `esp_twai.h` / `esp_twai_onchip.h` went in with 62486d9b (Lilygo T-2CAN, 2025-08-01) and were replaced by ACAN_ESP32 in eb047bad sixteen days later. Both first appear in v9.0.RC4, so the IDF version never reached a release, and the swap commit gives no reason. Worth recovering that reason before going back, in case it was something this PR would hit again.

**The IRAM result looks like the more general finding.** If long stalls during flash-disable are what let a corrupted frame through, that is independent of which driver is in use, and it would explain the ESP32-S3 reports without needing unpublished errata. #2652 (IRAM_ATTR on MCP2518FD) is arguably the same bug on the other controller. It might be worth treating the IRAM placement as its own change that can land ahead of, and independently of, the driver decision.

Question on that: can you confirm `CONFIG_TWAI_ISR_IN_IRAM` is actually set in the pioarduino 55.03.39 prebuilt libs? I could not verify it locally, and the stall explanation depends on it.

Agreed on waiting for the released IDF fix rather than merging the build-time patching. Separately, the +12.7 KB puts T-CAN485 at 95.4% flash, which looks like the thing that has to be answered before this can land on the 2 MB boards at all.

PR #2828, opened 2026-08-17
BYD-Modbus: fix inverted charge/discharge sign in two registers

nagyrobi:
## What
Register 303 now sends 128/129/130 for idle/charging/discharging, and registers 310/312 negate `active_power_W` so charging is reported as negative.

## Why
Both registers carried the opposite direction to the inverter, so every charge session looked like a discharge to a polling Gen24 - plausibly why it ignores a zero charge limit and keeps pushing its own minimum charging power.

Based on https://gitlab.com/pelle8/inverter_resources/-/blob/main/byd_registers_modbus_rtu.md

## How
`128 + bms_char_dis_status` used the DISCHARGING(1)/CHARGING(2) defines, which are the inverse of the register's bit order. The new `byd_power_W()` helper also clamps to int16 - the old direct cast wrapped above 32.7 kW.

> [!TIP]


dalathegreat:
Tested OK on the Discord!

dalathegreat (approved):

PR #2831, opened 2026-08-17
Glob the test sources recursively so new test directories are picked up too

ekholm:
Follow-up to #2829, which was too narrow. The patterns there named the directories that existed when it was written, so a test file added in a *new* subdirectory is silently never compiled — which is the failure that change set out to prevent. `emul/` also stayed hand-listed, for no reason other than the patterns not reaching it.

`GLOB_RECURSE` over `test/` covers both. The filter drops the leftovers of an in-source `cmake -B build` inside `test/`, which are not sources and would otherwise be compiled.

The firmware list stays explicit, and must. It is not "every file under `Software/src`" — it is the subset that links in a host build. `Software.cpp`, the webserver and anything reaching for `FS.h`, `ip_addr` or the async server are deliberately absent, and globbing there would drag them in and silently change what is under test whenever a driver is added. Keeping the two lists separate makes that distinction visible instead of hiding it among the tests.

One file changes state: `test/emul/Logging.cpp` has been in the tree but not in the build, and the recursive glob picks it up. It defines the `logging` object, which the test build does not otherwise need — under `UNIT_TEST` the `Logging` methods are static, so nothing emits a reference to the object — so compiling it adds an unused global and nothing else. There is no second definition in the test build; `Software.cpp` is not part of it.

Verification: the source sets were expanded and compared, 135 before and 136 after, the difference being exactly the file above — nothing dropped. The suite itself is what CI runs here.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -65,22 +65,23 @@ include_directories(emul)
 # For eModBus
 add_compile_definitions(ESP32 HW_LILYGO COMMON_IMAGE)
 
-# Test sources are globbed so that adding a test does not mean editing this
-# list. Naming every file here means two branches that each add a test conflict
-# on the same line, always with the same resolution. CONFIGURE_DEPENDS re-globs
-# at build time, so a new file is picked up without re-running cmake by hand.
-# The production sources below stay explicit: they are a curated set, and not
-# everything under Software/src builds against the host emulation.
-file(GLOB TEST_SOURCES CONFIGURE_DEPENDS
-    "${CMAKE_CURRENT_SOURCE_DIR}/*_tests.cpp"
-    "${CMAKE_CURRENT_SOURCE_DIR}/battery/*.cpp"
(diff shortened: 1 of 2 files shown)


ekholm:
CI caught something the narrower glob was hiding, and it is worth more than the change itself: `test/emul/Logging.cpp` does not compile. Its include is `../../src/devboard/utils/logging.h`, missing the `Software/` segment that `Logging.h` next to it gets right, so it has been broken since it was added in 825db356 last August. Nothing noticed because it was never in the CMake source list.

I have removed it rather than fixing the path. It is not needed: under `UNIT_TEST` the `Logging` methods are static, so nothing emits a reference to the object it defines, and the suite has been building and passing without it for a year. Fixing the include would add an unused global instead.

So the file set is now the same 135 sources as before this PR, with the difference that a new test file — or a new test directory — is picked up automatically.

dalathegreat (approved):

PR #2833, opened 2026-08-17
Add Ethernet support

ngehrsitz:
### What
This PR builds on #2827 and adds ethernet for the DFRobot board. This board uses the built in EMAC chip. I also considered adding support for W5500 based boards like #2234 or `3LB`. But I decided against adding it since the pins used for the [W5500 in the schematic](https://github.com/malcolmputer/3lb/blob/main/exports/pdfs/3lb.pdf) are currently used for other peripherals on the `3LB`. Also since it´s not part of the CI it looks like it´s not a target where somebody else could test.
But the only hurdle after this PR lands would be configuring the SPI bus and then conditionally compiling in `custom_component_remove = espressif/w5500`.
eth+wifi

### How
The stack is already dual-interface by default. I only had to wire up ethernet and prevent the duplicate start of mdns. You can have both interfaces connected at the same time, with the exception that you will get a hostname collision since both interfaces currently advertise the same hostname. This means only the last connected interface will show with the correct hostname in your router.

> [!TIP]


nagyrobi:
Wonderful!

A quick pour of thoughts:
- Both DCHP?
- How is default gateway be handled if in different networks? 
- Static IP for Ethernet?
- Hostname could be different automagically as it includes last octets of MAC, however when overriding with user-set hostname that poses questions.  Plus, hostname is used for a multitude of stuff to uniquely identify the device.

To avoid over-complication of the implementation, I would only allow only one interface at the time. User would simply have to choose in settings which interface to configure at boot. Default to WiFi as it has AP fallback etc. 
Long-press of BOOT button step 2 should wipe this setting (+any other Ethernet setting) too, and reboot as usual defaulting to WiFi with AP.
ESPNow has to be checked too, will it work without a WiFi connection, at least channel number will have to be set manually.

ngehrsitz:
> Wonderful!
> 
> A quick pour of thoughts:
> 
>     * Both DCHP?
> 
>     * How is default gateway be handled if in different networks?
> 
>     * Static IP for Ethernet?
> 
>     * Hostname could be different automagically as it includes last octets of MAC, however when overriding with user-set hostname that poses questions.  Plus, hostname is used for a multitude of stuff to uniquely identify the device.
> 
> 
> To avoid over-complication of the implementation, I would only allow only one interface at the time. User would simply have to choose in settings which interface to configure at boot. Default to WiFi as it has AP fallback etc. Long-press of BOOT button step 2 should wipe this setting (+any other Ethernet setting) too, and reboot as usual defaulting to WiFi with AP. ESPNow has to be checked too, will it work without a WiFi connection, at least channel number will have to be set manually.

With this implementation ethernet mirrors the WiFi station mode. That means DHCP by default with the option to set an ethernet specific static configuration.
grafik
All of the networking is abstracted by the Arduino ethernet stack, so I would have to test what the default gateway selection behaves.
I also considered switching the default hostname over to the ethernet mac if that is active. The reason why I decided against it is that it´s probably confusing to a user if they connect to a network over one and then switch and suddenly the name changes.
From a complexity and user experience standpoint I would also argue against any explicit interface selection. Currently you can connect either one without any configuration beforehand and it will just work. The framework does all the heavy lifting for us. I suspect that there might be some disadvantages from connecting both like higher RAM usage, or routing issues, but I don´t think they warrant the complexity of adding an extra selection. Maybe we could display a warning that this is not recommended if both are detected.
I don´t have an ESPNow setup is there an easy way to test this without an LILYGO T-Display-S3?

nagyrobi:
With any other ESP32. What do you have at hand?

ngehrsitz:
I let AI build an ESPHome configuration for my Espressif ESP32-C3-DevKitM-1 and that did indeed reveal a flaw in my startup logic. With #2843 both the ethernet only and the WiFi AP+STA disabled case correctly bring up the interface for ESPNow.

ngehrsitz:
I merged the latest changes in. Any remaining concerns?

nagyrobi:
Implementing at least taking down automatically the WiFi STA interface when Ethernet Link+IP is obtained should be considered (and also the WiFi AP interface if the Ethernet obtains an address via DHCP from AP's own 192.168.4.0/24 domain) to avoid increase of support load caused by network conditions falling outside BE.
(see discord dev-thread discussion)

ngehrsitz:
I tested with an MQTT server in a non local network and this is indeed an issue. When you misconfigure an interface there is currently no guarantee that the working one is choosen. I will try if I can find a better fix than just disabling the second interface.

dalathegreat:
@ngehrsitz , wanna make some changes to that in this PR still?

ngehrsitz:
Yes, I have some changes prepared but haven´t had time to test them extensively. I will get to it next week.

dalathegreat:
Just checking in @ngehrsitz , any progress on this front? :)

PR #2834, opened 2026-08-18
Nissan LEAF: transmit `CHG_STA_RQ=01b`, and re-read the poll-once groups after a BMS reset

nagyrobi:
## What
Transmits `CHG_STA_RQ=01b` in `0x1F2` from init instead of leaving it at `00b`. Re-reads groups `0x62`, `0x84` and `0x83` after a BMS reset, not only at boot.

## Why
`01b` ("Normal charge") is what the battery control spec has the VCM hold for all of normal operation, so the LBC sees what a running vehicle would send. Difference has been observed in the way cell balancing is being carried out on AZE0, compared to how it was so far with `00b`. A wider user base would be needed to test this.

The charge counts, serial and part number were previously stuck on values read before the reset, and the LBC's new session was never queried for them. Read all of them to verify. AC charge counts will increase each time we init the pack with `CHG_STA_RQ=01b`.

## How
Byte 2 of the `0x1F2` initializer gains the signal; the 20 precomputed byte-7 values shift with it, since that nibble is the message checksum. No runtime cost.
The BMS reset branch arms a one-pass re-poll and re-arms the startup burst; old values stay on display until each fresh reply lands. Costs one bool.

> [!TIP]


nagyrobi:
@dalathegreat instead of changing this definitively from what we had earlier `CHG_STA_RQ=00b` to what the spec recommends `CHG_STA_RQ=01b`, maybe it would be wiser to add a selection box to the users to open the change fpr experimentation with different packs.

So I'd say we add a dropdown box with two options:
- `other (default)` - for 00b
- `normal charge (experimental)` - for 01b

(naming from the spec, see 1st_gen_CAN_Message_List.xls or 2nd_gen_CAN_Message_List.xls)

Maybe worth trying with the quick charge bit too.

- `quick charge (experimental)` - for 10b

To change the value the BMS must be reset the usual way (changing the value will perform a BMS reset, are you sure kind of stuff). And this way we could ask more people to share their experience, if they see SOH / Hx changes and how balancing behaves differently.

On AZE0, the observation is that the pack inited at boot with CHG_STA_RQ=01b improves balancing A LOT:
Me:
image
User Frans from discord:
image

What do you think?

dalathegreat:
Yes, having a dropdown setting for this would be smart. Then we could merge it as is (with the default being the same as before), and ask for feedback. This is as you noticed further complicated as all different LEAF generations seem to behave a bit differently...

nagyrobi:
Done:
image
image

dalathegreat (approved):

PR #2839, opened 2026-08-19
Settings page: render and save the Tesla GTW settings with the defaults they boot with

ekholm:
Follow-up to #2724, which moved the boot fallbacks for the five Tesla gateway keys to the driver globals (country 17477 "DE", right-hand-drive `true`, map region 2, chassis 2, pack 1). The render and save sides did not move with it: the settings page still read the keys with a `0`/`false` default, so a device that never saved them showed country 0 / unchecked RHD while the firmware was running the driver defaults.

The right-hand-drive checkbox was also sticky: the save loop compared against `getBool("GTWRHD", false)`, so leaving the box unchecked and pressing Save wrote nothing at all, and the page kept disagreeing with the firmware.

Fix: all three surfaces (boot fallback, page render, save comparison) now read the same `user_selected_tesla_GTW_*` globals, so a future default change moves them together. A host test pins the five defaults so changing one is a deliberate act.

Devices with stored settings are unaffected (stored keys win, as before).

### DIFF
--- a/Software/src/devboard/webserver/settings_html.cpp
+++ b/Software/src/devboard/webserver/settings_html.cpp
@@ -329,20 +329,23 @@ String settings_processor(const String& var, BatteryEmulatorSettingsStore& setti
                             name_for_comm_interface);
   }
 
+  // The GTW keys must render with the same fallbacks init_stored_settings()
+  // boots with (the driver globals), or a device that never saved them shows
+  // values the firmware is not running.
   if (var == "GTWCOUNTRY") {
-    return options_from_map(settings.getUInt("GTWCOUNTRY", 0), tesla_countries);
+    return options_from_map(settings.getUInt("GTWCOUNTRY", user_selected_tesla_GTW_country), tesla_countries);
   }
 
   if (var == "GTWMAPREG") {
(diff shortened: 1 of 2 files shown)


test/tesla_gtw_defaults_tests.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "../Software/src/battery/BATTERIES.h"
+
+// The Tesla gateway settings have three surfaces that must agree on what an
+// unset key means: the boot fallback in init_stored_settings(), the settings
+// page render defaults (settings_html.cpp), and the save-loop comparison
+// default for GTWRHD (webserver.cpp). All three read the globals pinned here,
+// so they cannot drift from each other — but a change to a driver default
+// must still be a deliberate act. Whoever changes one updates this list, and
+// the render/save surfaces follow automatically.
+TEST(TeslaGtwDefaults, BootDefaultsArePinned) {
```
dalathegreat: Nice improvement, BUT, I really don't think we need a unit test for this.

ekholm: Dropped the test, thanks. All three surfaces still read the same globals so they can't drift from each other; the pin on the default values is gone with it.

dalathegreat (approved):

PR #2840, opened 2026-08-19
Refuse FD-length frames on the classic CAN transmit paths

ekholm:
The settings UI allows configuring a CAN-FD battery on a classic CAN interface. When that happens, the battery driver hands 32/48-byte frames to `transmit_can_frame_to_interface()`, and the `CAN_NATIVE` case copies `DLC` bytes into `CANMessage`'s 8-byte on-stack data array — memory corruption on every poll, plus malformed DLCs on the wire. On my bench this produced `CAN_NATIVE_BUS_ERROR` + `CAN_NATIVE_BUFFER_FULL` from one second after every boot and made the channel unusable for every node on the bus. The MCP2515 path has no overflow (fixed-size memcpy) but forwarded the oversize DLC to the controller.

Both classic paths now refuse frames whose DLC exceeds the target frame's data buffer (`sizeof` of the actual member, not a literal) and raise the existing send-fail flag, so the misconfiguration surfaces as a persistent CAN-failure event instead of corruption.

### DIFF
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -324,6 +324,13 @@ void transmit_can_frame_to_interface(const CAN_frame* tx_frame, CAN_Interface in
 
   switch (interface) {
     case CAN_NATIVE: {
+      if (tx_frame->DLC > sizeof(CANMessage::data)) {
+        // An FD-length frame cannot be sent on a classic CAN interface (a CAN-FD
+        // battery configured on it produces these), and copying it below would
+        // overflow frame.data on the stack.
+        datalayer.system.info.can_native_send_fail = true;
+        break;
+      }
       CANMessage frame;
       frame.id = tx_frame->ID;
       frame.ext = tx_frame->ext_ID;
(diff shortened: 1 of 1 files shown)


Software/src/communication/can/comm_can.cpp
```diff
@@ -324,6 +324,13 @@ void transmit_can_frame_to_interface(const CAN_frame* tx_frame, CAN_Interface in
 
   switch (interface) {
     case CAN_NATIVE: {
+      if (tx_frame->DLC > sizeof(CANMessage::data)) {
```
dalathegreat, line 327: Can't we get better performance by just checking if > 8 ?

ekholm:
Ran an A/B on the bench after filing (Stark, MEB battery on CAN 1 (Native), inverter on the CAN-FD channel as an ACK partner over a wired loopback): before the fix the oversize frames actually cross the wire truncated — a 48-byte 0xFC is delivered to the partner as a valid-looking 8-byte classic frame carrying the first 8 bytes. So besides the stack overflow, the peer receives silently corrupted data. With the fix, the oversize frames are refused, the CAN-failure event makes the misconfiguration visible, and the remaining ≤8-byte frames flow clean.

Software/src/communication/can/comm_can.cpp
ekholm: The binary is identical either way — `sizeof(CANMessage::data)` resolves to a compile-time 8, so the emitted check is the same instruction. I kept `sizeof` because it ties the guard to the buffer it protects: the MCP2515 site guards a different struct that also happens to be 8 today, and if either array ever changes, the checks follow without anyone remembering they exist. Happy to switch to the literal if you prefer it for readability.

dalathegreat:
Great, if the binary is the same, let's go with this one :+1:

dalathegreat (approved):

PR #2844, opened 2026-08-20
Add a settings descriptor table, and audit the stored settings against it at boot

ekholm:
Every persisted setting currently states its NVS key, storage type and default wherever it is touched — boot load, settings-page render, save comparison. Those copies can disagree, and when they do the failure is silent: NVS enforces its type tag, so a read with the wrong type or default does not error, it just returns something other than what the firmware runs.

This is not hypothetical. #2697 fixed six render-vs-boot default mismatches found by a hand audit. #2839 is fixing five more that appeared afterwards — introduced by #2724, one of them a checkbox whose boot default is `true` while the page rendered it unchecked, and whose save loop compared against `false`, so saving the page could not correct it. Hand audits catch these after the fact, one at a time.

This PR adds the structural half:

**1. A settings descriptor table** (`settings_table.{h,cpp}`): one row per persisted setting — key, on-flash kind (BoolU8/U32/I32/Str, matching the NVS type tag), default, range, flags. 145 rows. Nothing reads it in this commit, and because it is `constexpr` with no consumer the linker drops it: **the firmware binary is byte-identical**. What the commit adds in practice is the tests: key length ≤ 15 (the NVS limit, now assertable in one place), key and placeholder uniqueness, defaults within range, and the signed-key set pinned. The five defect classes the validation claims to catch were each injected and confirmed to fail the build. A further test scans the loader's own source, so adding a setting without a row fails the suite and names the key.

**2. A boot-time audit** that walks the table after the normal loads and compares each row's kind against the type tag NVS actually recorded for stored keys. A disagreement raises one info-level event naming the setting — the wrong-type → silent-default path made visible. The audit changes no behaviour: the existing loads always win, never-stored keys are skipped, and it runs once per boot. Cost: ~4.7 KB (devkit +4,704 B, stark +4,832 B, from wiped build directories).

Run on hardware with 95 of the 145 keys stored, the audit reports zero disagreements, and reports the agreement rather than staying silent, so a clean device and a device with nothing stored cannot be confused. Deliberately mis-typing one row was confirmed to produce exactly one event naming that setting, with the log showing the table reading `0` where `1883` is stored — which is the failure this exists to catch, reproduced.

**On the frontend work.** #1544 already carries per-kind settings tables of the same shape — the name doubling as flash key and API field, per-row setter lambdas, a secret flag, a default-true list, server-side range validation. That is the same conclusion reached from the other end, and this PR is deliberately the half those tables do not cover: they serve the web API and the live updates, while `comm_nvm.cpp` still hand-loads at boot with its own copy of every default — which is where the mismatches above come from. A backend table carrying the on-flash kind per row is what the boot side needs, and it composes with a frontend table rather than competing with it: whichever frontend lands, the key, the type and the default it reads can come from one place.

A follow-up (prepared, not included here) uses the same rows to deduplicate the ~95 plain loads in `init_stored_settings()` so key/type/default exist in exactly one place; it has passed an on-device migration test (all values preserved across the flash). Kept out of this PR so the table and audit can be judged on their own: pure addition, zero behaviour change, and a class of bug that stops being invisible.


ngehrsitz (commented):
I agree that we should improve the settings storage.
But is a settings table and a startup check really required?
Why not just keep it simple and have getters and setters for every setting? These would be a wrapper around `getInt`/`getBool`/`...` so the compiler should optimize them away. The current design where the frontend directly accesses settings by their key is an antipattern in my opinion.
Access to settings should be abstracted so that all callers never see the key in the background but only access a named function and get a concrete type back.
In the future when we switch to a JSON API this would also be the cleaner approach.
In the interim for the migration this would look like
```c
if (p->name() == "inverter") {
  storeEnum<InverterProtocolType>(p->value().c_str(), settings.setInverter);
} else if (p->name() == "INVCOMM") {
```
Where `storeEnum` is a helper that only calls the set function passed in as a parameter if the value is actually is in integer in enum range. You could of course then have other helpers with more in-depth checking like IP validation....
`getInverter` now also returns a `InverterProtocolType` so callers can directly use the value. If there are restrictions on a type like min/max value of an int then this could be part of a generic type.
But now `setInverter` and `getInverter` are the only way to access a setting and we can change how that is stored independently from the consumers and thus migrate keys or introduce other changes to the settings storage.

ekholm:
Agreed on the accessors — callers shouldn't see keys, and a `getInverter()` returning the enum is the right shape. That layer is planned as a follow-up, and it is generated *from* these rows (one X-macro list → enum + table + accessors), so the key, NVS type and default still exist once and the getters can't disagree with the loader.

What per-setting functions can't give on their own is what this PR is for: invariants across the whole set. Key ≤ 15 chars, key uniqueness, default inside range, "every key the loader reads has a row" — the tests here iterate the table; there is nothing to iterate over 145 hand-written getters, and a `getUInt`-vs-`getInt` slip inside one of them is as silent as today. A JSON API needs the same data — type, default, range, enum options — and #1544 ended up with per-kind row tables for that reason. Your `storeEnum` helper is the row's min/max applied on save; that consumer is the next step.

On the boot audit: fair challenge. It exists for the migration case you describe — change the storage, and a type-tag mismatch is exactly what falls back to defaults silently. It costs ~4.7 KB; the table alone is byte-identical. If the cost isn't worth it on the 4 MB boards I'm fine gating it behind a debug define or dropping it — your/Dala's call.

ngehrsitz:
The main reason for the getter/setter aproach is that you can then do whatever you like in the storage layer. A table, a autognerated byte structure, whatever.
The macro table in the current PR is built built based on the requirement that access by key has to be possible.
That requirement is something I would call in question. Because if we drop it it´s not only cleaner architectually, it also allows us to use a simple key structure while still avoiding the issues the current design has.
In regards to the key length that is also something you can easily enforce in a test. Let AI generate a test for every getter and setter and then mock the nvs access methods. You can then check if you see any len(keys)>15 and if the access types are the same per key. If you trade in some simplicity you could also generate getters and setters with macros.

ekholm:
I think we're aligned on the destination, so let me state the roadmap explicitly. Two follow-ups are planned on top of this PR: first the generated accessor layer — one X-macro list producing the enum, this table, and typed getters/setters, so they cannot disagree with the loader by construction — and second, **prohibiting direct key access outside that layer**, enforced the same way this PR's completeness guard works: a test that scans the tree and fails naming the offender. At that point no call site touches a key, which I believe is the architecture you're describing.

The mocked NVS you suggest testing against already exists — #2778's Preferences emulation is the harness this PR's table tests and boot audit run on.

Where by-key deliberately survives is the storage itself: NVS is key-value, and the fleet's flash already contains these keys. Changing the storage format means a migration for every deployed device — a risk this PR intentionally doesn't take. The table makes the *existing* storage checkable without touching it, and if a byte-structure backend ever happens, it consumes these same rows: kind, default, range is exactly the metadata it would need.

On generated per-accessor tests: they verify each accessor in isolation, but the defects that motivated this (#2697's six, #2839's five) were cross-copy disagreements — boot, render and save holding different defaults for one key. Catching those needs the whole set declared in one place; once it is, length/uniqueness/range are three assertions over 145 rows rather than 145 generated pairs.

jonny5532:
I've been iterating on something similar - this attempts to unify the settings init and update (for my new frontend thing): https://github.com/dalathegreat/Battery-Emulator/blob/9baed92a0b23673ab82035dbeb5975c3770c9d7f/Software/src/devboard/webserver/webserver_settings.cpp

(It's not especially beautiful, but the alternative of a unified settings list with type info and getters/setters takes up several more kb of memory)

Edit: I note that you mention that branch - it does now remove all the inits from `comm_nvm.cpp` and perform them with loops instead.

Also it effectively achieves the static type checking by ensuring that the references passed match the expected type.

ekholm:
Measured, since the cost concern was the one thing without a number: on the devkit env the typed getters/setters are 0 B (−80 B in fact — they compile to the same store calls and delete repeated literals); the table is +4.7 KB including the boot audit, constexpr in flash, no RAM. The metadata cost is one both shapes pay — your arrays carry name/range/pointer per row too.

I think the designs compose rather than compete: your reference-binding kills the loader and webserver call sites; the table's kind/default/range is what the audit and compile-time validation need. One generated list with a storage-pointer column could emit both — your loops unchanged, drift impossible. Two corners you may want regardless: BATTERY_WH_MAX / TARGETCHVOLT / TARGETDISCHVOLT lose their stored-zero "never set" guard in the loop path (min=1 only guards the web input — a legacy device with stored 0 boots with 0 capacity), and WEBAUTH + the BYD autocal keys load twice (table loop, then comm_nvm by hand with a different default expression).

ekholm:
For the record, both follow-ups named above are built, so the shape can be judged on code rather than description:

- #2851 (draft): the typed accessors, generated from the rows.
- The guard and the first migrations, stacked on it: https://github.com/ekholm/Battery-Emulator/compare/refactor/settings-accessors...refactor/settings-direct-access — a tree-scan test that fails, naming the offender, on any new key-addressed access outside the layer (shrink-only allowlist; the store and the loader are exempt), `webserver.cpp` 44 → 5 literal-key sites, the mqtt saves, and the BYD routes off their own raw `Preferences` handles. A call site now reads `setting_save<Sid::BYDAUTOCALEN>(enabled);` — no key, no store; single reads open a read-only session.

Not filing that as a PR while #2844 and #2851 are open — three deep is noise. What it leaves un-migrated is the five runtime-key save loops, which is exactly the ground jonny5532's reference-binding covers; the two designs can meet there.

PR #2852, opened 2026-08-21
BYD-MODBUS: implement Fronius ControlData block and inverter settings UI organization

nagyrobi:
### What
This PR makes the BYD-Modbus driver act on three ControlData registers the inverter already writes: WatchDogTimeout (402), RebootCommand (407) and UTC (403-406). Splits the "Inverter config" settings card into inverter-specific settings and settings common to all inverters.

kép

### Why
Inverter-missing detection assumed a 60s watchdog period instead of the one the inverter itself declares, so it could not follow a potential period change from inverter side. Two further registers the inverter writes on every ControlData update were stored and ignored: a restart request, and a wall clock the emulator currently has no other source for.
Settings that apply to every inverter were interleaved with driver-specific ones, making the card hard to scan and the relationship between options unclear.

### How
Register 402 drives the sampling interval in `verify_inverter_modbus()`, restored from NVM at boot and re-persisted only when the inverter changes it; 407 triggers the existing pause/open-contactors restart path behind a new opt-out checkbox; 403-406 are decoded as big-endian epoch seconds and shown on the settings page. The driver flags a changed watchdog period and the core loop performs the write, so no inverter protocol gains a dependency on storage.
The card now holds two grids: protocol, interface and driver-specific settings first, then the common ones (ramp up, SOC taper, longer CAN timeout, offgrid). 

> [!TIP]


nagyrobi:
kép

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -150,6 +150,9 @@ void init_stored_settings() {
   user_selected_tesla_GTW_chassisType = settings.getUInt("GTWCHASSIS", user_selected_tesla_GTW_chassisType);
   user_selected_tesla_GTW_packEnergy = settings.getUInt("GTWPACK", user_selected_tesla_GTW_packEnergy);
   user_selected_primo_gen24 = settings.getBool("PRIMOGEN24", false);
+  user_selected_accept_inverter_reboot = settings.getBool("INVACCREB", true);
+  // Watchdog period the inverter last told us about, or the default if it never has
+  inverter_modbus_watchdog_timeout_s = settings.getUInt("INVWDTMO", MODBUS_INV_WATCHDOG_DEFAULT_S);
```
ngehrsitz, line 155: How often does the inverter send the timeout? Do we need to persist it in flash, or would it be good enough when we just adapt at runtime?

nagyrobi: At reboot/coldstart of inverter only. We reboot BE much more often…

ngehrsitz: Unfortunate, then we have to persist it

ekholm:
Heads-up relevant to the section split: `POST /saveSettings` currently treats every absent boolean as unchecked-and-save-false, which is correct for the one full-page form but means a per-section POST would wipe every checkbox outside its section the moment it saves. #2857 makes that safe ahead of you: the full settings form marks itself with a hidden `FULLFORM` field and keeps today's behaviour exactly, while marker-less posts apply booleans present-only — so your sectioned forms need no changes and no coordination; they just post without the marker and can't clear what they don't carry.

nagyrobi:
This should be addressed in a separate PR.

nagyrobi:
@dalathegreat added a PR to the docs with the new features explained: https://github.com/dalathegreat/Battery-Emulator-Wiki/pull/37 
We can merge the docs PR only when we merge this, or even later, when the next release will be out - so the docs always remains in sync with the released functionality.

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -150,6 +150,9 @@ void init_stored_settings() {
   user_selected_tesla_GTW_chassisType = settings.getUInt("GTWCHASSIS", user_selected_tesla_GTW_chassisType);
   user_selected_tesla_GTW_packEnergy = settings.getUInt("GTWPACK", user_selected_tesla_GTW_packEnergy);
   user_selected_primo_gen24 = settings.getBool("PRIMOGEN24", false);
+  user_selected_accept_inverter_reboot = settings.getBool("INVACCREB", true);
```
dalathegreat: IMO this reboot parameter should default to FALSE, to avoid any unforseen consequences

Software/src/communication/nvm/comm_nvm.cpp
```diff
@@ -317,6 +320,15 @@ void store_settings_equipment_stop() {
   settings.saveBool("EQUIPMENT_STOP", datalayer.system.info.equipment_stop_active);
 }
 
+void store_settings_inverter_watchdog() {
+  if (!inverter_modbus_watchdog_changed) {
+    return;
+  }
+  inverter_modbus_watchdog_changed = false;
+  BatteryEmulatorSettingsStore settings(false);
+  settings.saveUInt("INVWDTMO", inverter_modbus_watchdog_timeout_s);
```
dalathegreat: Note, calling saveUint on the fly always risks TASK_OVERRUN (missed CAN messages), since accessing the settings. halts the critical core. I am always a bit wary of functions that can call save settings at any time during runtime ⚠️

nagyrobi: Well, this likely occurs very rarely. Only once when the inverter boots up. And write would only happen if value changes.

Software/src/communication/nvm/comm_nvm.cpp
nagyrobi: Let’s raise an event if disabled then.

nagyrobi:
- Tightened saveUint so it really, really occurs only if this value is changed by the inverter. I expect they probably never ever change it.
- reboot parameter defaults to FALSE, if a command falls in when FALSE an event is raised with an explanation. Otherwise it prints a log entry.
- for monitoring, a log entry is emitted when DarkstartEnable message arrives. Nothing else happens, we just want to have a mark for future reference

Software/src/inverter/BYD-MODBUS.cpp
```diff
@@ -166,11 +170,13 @@ void BydModbusInverter::verify_temperature() {
...
-  // Every 60 seconds, the Gen24 writes to this 401 register, alternating between 00FF and FF00.
-  // We sample the register every 60 seconds. Incase the value has not changed for 5 minutes, we raise an event
+  // Once per watchdog period, the Gen24 writes to this 401 register, alternating between 00FF and FF00.
+  // We sample the register at the same period. In case the value has not changed for HISTORY_LENGTH
+  // periods, we raise an event. The period defaults to 60s and follows register 402 from the moment
+  // the inverter declares a different one.
   unsigned long currentMillis = millis();
 
-  if (currentMillis - previousMillis60s >= INTERVAL_60_S) {
+  if (currentMillis - previousMillis60s >= (inverter_modbus_watchdog_timeout_s * 1000UL)) {
```
dalathegreat, line 179: I don't think this brings a real improvement/gain here. We have a set timeout that we use, what real value does it bring to use a timeout set by inverter? Just makes the system harder to troubleshoot IMO

nagyrobi: On the contrary. My view is the following:
- if the timeout of 60 seconds would have been an always fixed value, never intended to be changed, they wouldn't publish it in a write register towards the battery when the inverter boots. It would be enough as observed by sniffing the modbus data.
- the fact that they publish it at inverter statrup very likely means that Fronius devs already reserved the possibility to change this value whenever they want. It may be possible that they chance it by varying circumstances, or by a firmware update or whatever other reason.
- the system startup sequence described both by the Fronius ("Suitable batteries" - page 28) and BYD battery ("Operation - Switch on the Battery System" - page 40) operating manual is always turn on battery first, and turn on inverter following the battery. This means Fronius counts on this (and likely other) parameters to be accounted by the battery when it boots up, it's not something just shot in the wild.

Based on these observations I'd say we should respect this, exactly to avoid a possible discrepancy between what the inverter wants and what the Emulator offers. Currently if Fronius unilaterally would change this value, despite writing it to our register we wouldn't respect that, keep 60s - will break running installations. By taking this value into account, we automagically follow the inverter's request - if there's any, so not harder to troubleshoot, on the contrary - one less point of failure removed, which would need an immediate firmware change of the Emulator in case of a hardcoded timeout.

And since we likely reboot much more often than the inverter (which only writes this when it boots) we need to persist it.

dalathegreat: Okidoki! Let's get this PR merged and let's observe with a broader audience the new Fronius registers :raised_hands:

dalathegreat (approved):

PR #2853, opened 2026-08-21
Repair a mistagged setting instead of skipping the write that would fix it

ekholm:
**Stacked on #2778: the diff carries its commit until it merges; this PR is the two commits above it. #2778 is the emulation this is tested on — merging that first collapses this PR to its two top commits.**

NVS records the type a value was written under, and a typed read returns the caller's default when that tag does not match the getter. `BatteryEmulatorSettingsStore::saveX()` skipped the write whenever the current read equalled the new value — without checking the tag.

On a key stored under the wrong type, `getX()` reports the default rather than what is stored. So saving exactly the value the mistagged read reports compares equal, and the write is skipped. That write is the only thing that would have repaired the tag: `nvs_set_*` finds the existing entry whatever its type, writes the new one and deletes the old. The key stays mistagged, and every later boot reads the row default instead of the user's choice.

One value per type is affected — `0`, `false`, and `""`.

**Why it is worth fixing rather than noting:** `WIFIAPENABLED` defaults to **true**. A user switching the access point off writes `false`, which is exactly the value a u32-tagged key reports to `getBool()`. The save is skipped, and the device keeps booting with the AP on, indefinitely, with no symptom pointing at the cause.

A save is now skipped only when the stored tag matches the type being written, for all four savers: `saveBool`→`PT_U8` (`putBool` forwards to `putUChar`), `saveUInt`→`PT_U32`, `saveInt`→`PT_I32`, `saveString`→`PT_STR`.

### What this adds to the emulation, and why

Stacked on #2778, whose emulation already stores real values in real namespaces. Two things were still missing, and the fix needs both:

- **`getType()`.** The tag is precisely what a getter cannot tell you — on a mismatch it answers with the caller's default, so the mistagged case and the correct case are indistinguishable from the outside. `saveX()` has to ask directly.
- **A write counter.** The skip-identical optimisation is invisible from the stored value, so a test otherwise cannot tell "already correct, left alone" from "rewritten with the same value" — which is exactly what the two preservation tests below must distinguish.

### Tests

All four savers, each on its dangerous value, plus both directions of what must not change:

- an unchanged, correctly-tagged value is still skipped — otherwise every save cycle rewrites every key and the flash wears for nothing;
- the first save of a falsy value into a missing key is still written, which is what the existing `isKey()` guards are for.

Verified by reverting the four tag checks: the four repair cases fail, the two preservation cases pass.

No behaviour change for correctly-tagged keys — which is every key on a device that has only ever been written by this firmware; the repair path only engages where a tag is already wrong. How a key becomes mistagged in the first place is out of scope and not claimed to be fixed: this makes the condition self-healing on the next save rather than permanent.

### DIFF
--- a/Software/src/communication/nvm/comm_nvm.h
+++ b/Software/src/communication/nvm/comm_nvm.h
@@ -44,7 +44,7 @@ void clear_wifi_sta_settings();
 // runs them automatically (via constructor/destructor).
 class BatteryEmulatorSettingsStore {
  public:
-  BatteryEmulatorSettingsStore(bool readOnly = false) {
+  BatteryEmulatorSettingsStore(bool readOnly = false) : readOnly(readOnly) {
     if (!settings.begin("batterySettings", readOnly)) {
       set_event(EVENT_PERSISTENT_SAVE_INFO, 0);
     }
@@ -53,6 +53,9 @@ class BatteryEmulatorSettingsStore {
   ~BatteryEmulatorSettingsStore() { settings.end(); }
 
   void clearAll() {
+    if (readOnly) {
+      return;
+    }
     settings.clear();
     settingsUpdated = true;
   }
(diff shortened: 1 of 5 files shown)


jonny5532:
Neat solution, makes sense to me.

(Claude's comments are a bit OTT - it could just be:

`// Save setting if it doesn't exist yet, or overwrite if the existing type or value doesn't match`

although that is somewhat stylistic...)

ekholm:
Fair — adopted your wording.

Issue #2687, opened 2026-07-27
kia 64 kWh FD battery


QmatiQ:
Please add an option to enable the use of two batteries:
Battery 1 - CAN-A FD (MCP2515)
Battery 2 - CAN FD (MCP2518 add-on)
Inverter: Pylontech HV battery via CAN bus - CAN B (native)
LilyGo T-2CAN FD
Contactors controlled via GPIO - No
MQTT - Yes

QmatiQ:
Hello, there will be a chance to add a second battery as mentioned above in the topic....

nagyrobi:
Please try the testing binary in PR https://github.com/dalathegreat/Battery-Emulator/pull/2855
Post your test results there.

QmatiQ:
The batteries have already been tested individually. Now I'll test them in a dual configuration. Thank you very much for adding the option of a second battery.

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/ada6ead6-7aed-43a3-bedd-00068196c5a5" />

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/a9447251-073e-4360-89f5-f06c3e4d3fd8" />

nagyrobi:
> The batteries have already been tested individually. 

The reason fir separate individual test from dual setup is that the driver was modified in BE for single setup too.

QmatiQ:
everything is fine, then I will also test the batteries individually..

nagyrobi:
Also, what batteries are these so small?

QmatiQ:
kia niro z 2024r 11 kwh a can FD

nagyrobi:
Give the part number please so we can add to wiki

QmatiQ:
Kia Niro Hybrid 37501-AT510 11 kWh

QmatiQ:
<img width="1049" height="1031" alt="Image" src="https://github.com/user-attachments/assets/9d0554da-5447-4a2d-9764-5258292ce85e" />
<img width="1080" height="1062" alt="Image" src="https://github.com/user-attachments/assets/9f61397e-52f9-4e06-b2f0-37e2a780b49b" />

PR #2855, opened 2026-08-21
Add double battery support for Kia 64kWh FD

nagyrobi:
# Add double battery support for Kia 64kWh FD

## What
Enables the double battery option for the Kia 64kWh FD integration, so two FD packs can run in parallel on separate CAN-FD buses. Triple battery is deliberately not offered for this integration.

## Why
User request in https://github.com/dalathegreat/Battery-Emulator/issues/2687. The FD integration was the only 64kWh Kia variant without parallel battery support, unlike the Kia/Hyundai 64/40kWh integration next to it. Triple is excluded because only two physically distinct CAN-FD buses exist.

## How
Adds the second-battery constructor and moves all pack state behind a `datalayer_battery` pointer, matching the pattern used by the other double-capable integrations. Registers the type in `battery_supports_double()` and the `battery2` switch. Contactor permission now goes through the `allows_contactor_closing` pointer instead of the hardcoded global.

## Notes

- `CANFD_NATIVE` and `CANFD_ADDON_MCP2518` are served by the same MCP2518 controller (see `transmit_can_frame_to_interface()` and `receive_canfd()` in `comm_can.cpp`), leaving `CANFD_ADDON_MCP2518_2` as the only separate FD bus. Hence double, not triple. This is documented in a comment above `battery_supports_triple()`.
- Requires hardware with a second MCP2518 (`MCP2517_CS2()` / `INT2()`): Becom, Stark, and the FD variant of LilyGo 2-CAN.
- `write_cell_voltages()` becomes a member function; it previously wrote to `datalayer.battery` directly as a free function. The loop counter changes from `size_t` to `int` to silence the signed/unsigned comparison.
- Two global reads are intentional and match `CMFA-EV-BATTERY.cpp` and `KIA-HYUNDAI-64-BATTERY.cpp`: the charge/discharge overrides are only populated from NVM on the primary datalayer, and `battery_protocol` is a single system-wide string.

## Testing
- `clang-format` clean (v22.1.8, matching `.pre-commit-config.yaml`).
- Not yet tested on hardware with two physical FD packs.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -432,8 +432,7 @@ void setup_battery() {
         // Double only: needs a CAN-FD bus of its own, and only two exist.
         // See the comment on battery_supports_triple() above.
         case BatteryType::Kia64FD:
-          battery2 = new Kia64FDBattery(&datalayer.battery2, &datalayer_extended.Kia64FD_2, can_config.battery_double,
-                                        &datalayer.system.status.battery2_allowed_contactor_closing);
+          battery2 = new Kia64FDBattery(&datalayer.battery2, &datalayer_extended.Kia64FD_2, can_config.battery_double);
           break;
         case BatteryType::KiaHyundai64:
           battery2 = new KiaHyundai64Battery(&datalayer.battery2, &datalayer_extended.KiaHyundai64_2,
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -370,14 +370,6 @@ bool battery_supports_double(BatteryType type) {
 
 // The integrations that can be instantiated a third time on a separate
 // interface. Must match the switch in setup_battery() below.
-//
-// CAN-FD integrations must NOT be listed here. Only two physically distinct
-// CAN-FD buses exist: CANFD_NATIVE and CANFD_ADDON_MCP2518 are served by the
-// same MCP2518 controller (see transmit_can_frame_to_interface() and
-// receive_canfd() in comm_can.cpp), leaving CANFD_ADDON_MCP2518_2 as the only
-// separate one. A third FD pack would therefore have to share a bus with one
-// of the first two, which collides on both the diagnostic requests and the
-// periodic broadcasts. Such integrations are limited to double battery.
 bool battery_supports_triple(BatteryType type) {
   switch (type) {
     case BatteryType::NissanLeaf:
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -351,6 +351,7 @@ bool battery_supports_double(BatteryType type) {
     case BatteryType::CmfaEv:
     case BatteryType::CmpSmartCar:
     case BatteryType::StellantisEcmp:
+    case BatteryType::Kia64FD:
     case BatteryType::KiaHyundai64:
     case BatteryType::MgGen1:
     case BatteryType::Pylon:
@@ -369,6 +370,14 @@ bool battery_supports_double(BatteryType type) {
 
 // The integrations that can be instantiated a third time on a separate
 // interface. Must match the switch in setup_battery() below.
+//
+// CAN-FD integrations must NOT be listed here. Only two physically distinct
+// CAN-FD buses exist: CANFD_NATIVE and CANFD_ADDON_MCP2518 are served by the
+// same MCP2518 controller (see transmit_can_frame_to_interface() and
+// receive_canfd() in comm_can.cpp), leaving CANFD_ADDON_MCP2518_2 as the only
+// separate one. A third FD pack would therefore have to share a bus with one
+// of the first two, which collides on both the diagnostic requests and the
+// periodic broadcasts. Such integrations are limited to double battery.
 bool battery_supports_triple(BatteryType type) {
   switch (type) {
     case BatteryType::NissanLeaf:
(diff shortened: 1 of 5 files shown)


nagyrobi:
@QmatiQ please test stable operation with both single and double battery setup.

dalathegreat (approved):

Software/src/battery/KIA-64FD-BATTERY.cpp
```diff
@@ -415,11 +432,14 @@ void Kia64FDBattery::transmit_can(unsigned long currentMillis) {
...
-  datalayer.battery.info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
+  datalayer_battery->info.number_of_cells = 96;
+  datalayer_battery->info.max_design_voltage_dV = MAX_PACK_VOLTAGE_DV;
+  datalayer_battery->info.min_design_voltage_dV = MIN_PACK_VOLTAGE_DV;
+  datalayer_battery->info.max_cell_voltage_mV = MAX_CELL_VOLTAGE_MV;
+  datalayer_battery->info.min_cell_voltage_mV = MIN_CELL_VOLTAGE_MV;
+  datalayer_battery->info.max_cell_voltage_deviation_mV = MAX_CELL_DEVIATION_MV;
+
+  if (allows_contactor_closing) {
+    *allows_contactor_closing = true;
```
jonny5532, line 445: There seems to be confusion about how this should work, between different batteries.

My understanding is:
- The 2nd/3rd batteries get passed a pointer to the datalayer `battery[23]_allowed_contactor_closing`  in their constructors, and keep it internally.
- This is an INPUT, to the battery, indicating whether it should close or open contactors (see `parallel_safety.cpp`)
- Therefore the battery itself should never write to it.

This battery doesn't seem to allow contactors to be controlled anyway, so people will be using GPIO ones, and writing to `battery[23]_allowed_contactor_closing` may result in contactor closing (see `handle_contactors_battery2`) earlier than expected.

Software/src/battery/BATTERIES.cpp
```diff
@@ -369,6 +370,14 @@ bool battery_supports_double(BatteryType type) {
 
 // The integrations that can be instantiated a third time on a separate
 // interface. Must match the switch in setup_battery() below.
+//
+// CAN-FD integrations must NOT be listed here. Only two physically distinct
```
jonny5532: I suspect it is only a matter of time until someone makes a 3xFD board, will you go through and remove all these comments then?

Software/src/battery/KIA-64FD-HTML.h
```diff
@@ -0,0 +1,41 @@
...
+    content += "<h4>SOC (display): " + String(data.SOC_Display / 10.0f, 1) + " %</h4>";
+    content += "<h4>SOC (estimated from lowest cell): " + String(data.SOC_estimated_lowest / 100.0f, 2) + " %</h4>";
+    content += "<h4>SOC (estimated from highest cell): " + String(data.SOC_estimated_highest / 100.0f, 2) + " %</h4>";
+    content += "<h4>SOH: " + String(data.batterySOH / 10.0f, 1) + " %</h4>";
+    content += "<h4>Allowed charge power: " + String(data.allowedChargePower / 100.0f, 2) + " kW</h4>";
+    content += "<h4>Allowed discharge power: " + String(data.allowedDischargePower / 100.0f, 2) + " kW</h4>";
+    content += "<h4>Highest cell: no " + String(data.CellVmaxNo) + "</h4>";
+    content += "<h4>Lowest cell: no " + String(data.CellVminNo) + "</h4>";
+    content += "<h4>12V voltage: " + String(data.leadAcidBatteryVoltage / 10.0f, 1) + " V</h4>";
+    content += "<h4>Inverter voltage: " + String(data.inverterVoltage) + " V</h4>";
```
jonny5532, line 28: Very minor, but the appending trick using a short operator template:

```C++
template <typename T>
inline String& operator<<(String& str, const T& value) {
  str += value;
  return str;
}
```

Will reduce the code size and impact of this string concatenation (see `String CmfaEvBattery::get_uds_info_html()`). Given we keep repeating the template, maybe it should be in a util (although it's inlined anyway), although I quite like having it near the code in question so people can see what is going on. Doesn't help with the `String(val, 2.0)` formatting ones though.

nagyrobi: I'd say let's handle this in the new UI

Software/src/battery/KIA-64FD-BATTERY.cpp
nagyrobi: The confusion you point at is real and it appears to reflect from the codebase. There are two conventions sharing one name:

- `allows_contactor_closing` as an **output**: the battery decides and writes.
  NissanLeaf, TestFake, BMW i3 and KiaHyundai64 all set it to `nullptr` in the
  second-battery constructor and guard every write.
- KiaHyundai64 and BMW i3 then take the **input** through a separately named
  `contactor_closing_allowed` pointer, which is what actually receives
  `&battery2_allowed_contactor_closing`.
- Relion is the odd one out: it takes `battery2_allowed_contactor_closing` into a
  pointer named `allows_contactor_closing` and only ever reads it. Same name,
  opposite direction.

Claude had followed the Relion spelling while writing to it, which combines the worst of both.

I had no chance to catch this. Sorry for that.

nagyrobi:
@QmatiQ please retest

QmatiQ:
@nagyrobi  On one as well as on a double battery the data is read correctly, only on the battery information the inverter voltage shows too high.

QmatiQ:
Screenshot_20260824_111610_Chrome
Screenshot_20260824_111604_Chrome

nagyrobi:
@QmatiQ did you try the new binary 48e81e1?
Screenshots are not visible

QmatiQ:
@nagyrobi Yes I tried the new binary 48e81e1

QmatiQ:
Screenshot_20260824_112138_Chrome
Screenshot_20260824_111801_Chrome

QmatiQ:
would it be possible for the BE to control the contactors via BMS? Kia Niro II Hybrid 37501-AT510
Screenshot_20260826_083656_WhatsApp (1)
Screenshot_20260826_083712_WhatsApp (1)

nagyrobi:
Make a new issue please as feature request

QmatiQ:
ok, thank you for your answer

Issue #1863, opened 2025-12-22
[BUG] Chademo is not working properly and directly closing all contactors without chademo and/or inverter connected


HansNetcom:
**Describe the bug**
The CHAdeMO configuration is instantly closing all contactors without being connected to a car or inverter

load the latest build

**Expected behavior**
Wait for CHAdeMO plug connected, hide undervoltage message if no car connected and closed contactors only after all checks are ok.

ekholm:
Traced this in the source — the mechanism is a floating input pin, and the "instant closing" is explainable:

**The plug-detect is a raw read of an unbiased input.** `setup()` configures the CHAdeMO sense pins as plain `INPUT` (no pull-up/-down), and the IDLE state exits on `plug_inserted = digitalRead(pin7)` alone. With nothing plugged in, pin 7 (proximity, GPIO47 on the T-2CAN) **floats** — and a floating ESP32 input near switching supplies and CAN transceivers happily reads HIGH.

**What happens next matches your report:** IDLE → CONNECTED on the floating read, and CONNECTED immediately drives pin 2 HIGH — the d2 EVSE relay physically closes, with the lock actuation following in later states. So relays audibly/visibly close "instantly" with no car present. (Full traction-contactor closure additionally requires vehicle CAN — the `EVSE_START` transition needs `x102` session content — so the instant clicking should be the EVSE-side relays/lock; still clearly wrong unplugged.)

**Suggested fix, two layers:**
1. `pinMode(pin7, INPUT_PULLDOWN)` (and same for pin 4) — the T-2CAN is an ESP32-S3, so internal pull-downs exist on both GPIOs; unplugged then reads a defined LOW. If your board/harness provides external biasing per the CHAdeMO proximity circuit, this becomes a polarity/threshold question instead — worth stating which hardware and wiring you're running.
2. Debounce the IDLE-exit: require the plug-present read stable across several cycles before leaving IDLE — the code's own comment in CONNECTED already admits the "racy realities of the real world."

**On the undervoltage message:** with no vehicle there's no valid pack voltage, so the plausibility events fire on 0 V. Gating voltage-related events until the handshake has at least reached NEGOTIATE (i.e., a vehicle actually answered on CAN) would silence that cleanly.

Caveat: analysis from source only — no CHAdeMO rig here to reproduce. Happy to put the pull-down + debounce into a PR if this direction looks right; confirmation of your hardware (T-2CAN? external proximity circuit?) would firm up the pin-biasing part.

PR #2860, opened 2026-08-24
CHAdeMO: bias the plug-detect and vehicle-permission inputs

ekholm:
The CHAdeMO plug-detect (pin 7) and vehicle-permission (pin 4) inputs are configured as plain `INPUT`. With nothing plugged in they float, and a floating ESP32 input next to CAN transceivers reads HIGH often enough to walk the state machine off IDLE with no vehicle present — part of the #1863 picture. This enables the internal pull-down on both where the silicon has one (T-2CAN, GPIO 47/4). On the 3LB and LilyGo the pins are GPIO 34/35, input-only with no internal pulls: there the call degrades to a plain `INPUT` and those boards need external pull-downs, which the comment now says.

Polarity is unchanged (HIGH = asserted), so existing installs keep working. #2124 takes the other route — `INPUT_PULLUP` with inverted reads, i.e. an active-low convention — which is a valid design if the interface hardware is wired for it. I don't have a rig to settle which convention the deployed boards actually use; if the hardware owners prefer the active-low design this PR should be dropped in favour of that, and the 3LB/LilyGo external-pull note is the part worth keeping either way.

Refs #1863.

### DIFF
--- a/Software/src/battery/CHADEMO-BATTERY.cpp
+++ b/Software/src/battery/CHADEMO-BATTERY.cpp
@@ -904,8 +904,15 @@ void ChademoBattery::setup(void) {  // Performs one time setup at startup
   digitalWrite(pin10, LOW);
   pinMode(pin_lock, OUTPUT);
   digitalWrite(pin_lock, LOW);
-  pinMode(pin4, INPUT);
-  pinMode(pin7, INPUT);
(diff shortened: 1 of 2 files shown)


dalathegreat:
No need to fix anything related to Chademo. User "hansol"'s branch  will become the main Chademo interface when it is done

PR #2861, opened 2026-08-24
BYD Atto 3: retry contactor close when the BMS wakes after the inverter's permission

ekholm:
The wiki documents a startup-order restriction for the Atto 3: if the BMS comes up after the inverter has already granted permission to close, the emulator sits in standby until a reboot. The cause is in the state machine: the close is requested only on the permission edge; when the close-confirm window times out (BMS not awake yet), the state machine falls back to standby with the edge already consumed — nothing re-requests, so it stays there until a reboot re-creates the edge.

This adds a feedback-gated retry: after a failed confirm, the close is re-requested only when battery feedback (0x344) is newer than the give-up — a dead bus cannot loop, and a permission withdrawal disarms the retry. With a live but refusing BMS the worst case is one visible MISMATCH per 15-second window, which matches the keep-trying-while-permitted semantics elsewhere in the machine.

Host-verified: a regression test drives the state machine through transmit_can and fails without the fix; a second test pins that stale feedback must not retry. Full suite + shuffle seeds green; lilygo_330 builds with -Werror. I don't have an Atto 3 pack, so this is host/bench evidence only — a confirm from an Atto 3 owner would close the wiki restriction for good.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -762,9 +762,21 @@ void BydAttoBattery::handle_contactor_control(unsigned long currentMillis) {
     previousContactorsAllowedClosed = contactorsAllowedClosed;
     if (!contactorsAllowedClosed) {
       requestContactorOpen = true;
+      closeRetryArmed = false;  // Permission withdrawn - the abandoned close must not resurrect
     } else {
       requestContactorClose = true;
     }
+  }
+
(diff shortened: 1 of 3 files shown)


dalathegreat:
@demon1300 please approve/reject this PR!

demon1300:
@dalathegreat sure thing.

My only comments would be:

1. The late-start retry is a good fix, but it appears that pressing open contactors does not cancel a pending retry. When the battery next sends the relevant CAN message, it may try to close the contactors again.

2. Could the open command cancel any pending close retry, with a test confirming the contactors remain open afterward?

PR #2863, opened 2026-08-24
Fix: second CAN FD interface (BECom) fails to start when used without the first

ekholm:
On the BECom, the second MCP2518 is clocked from the first chip's CLKO output — `hw_becom.h` sets `MCP2517_CLKODIV()` to divide-by-1 for exactly this (aebce6da). That divider is programmed by the first FD interface's `begin()`, and the chip powers up at divide-by-10. So configuring only "CAN FD Battery 2" leaves the second chip on a 4 MHz clock while the driver assumes 40 MHz, and init fails with a CAN-FD-2 configuration error. Both interfaces together work, which is why this only bites standalone use.

Fix: when the second FD interface starts without the first, program the first chip's OSC register directly (reset + one register write over the already-initialized shared SPI bus) so CLKO carries the declared frequency. Boards with the default divider (their second chips have their own clocks) are unaffected — the gate excludes them.

Verified: host tests for the gate and the exact SPI command sequence (192/192 pass); BECom_330 / stark_330 / lilygo_2CAN_330 compile. I don't own a BECom, so a hardware confirm from someone who does would be very welcome — the CLKO wiring is inferred from the CLKODIV override and its commit message, not a schematic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,3 @@
+cmake_minimum_required(VERSION 3.16.0)
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+project(fd2-standalone)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,0 @@
-cmake_minimum_required(VERSION 3.16.0)
-include($ENV{IDF_PATH}/tools/cmake/project.cmake)
-project(fd2-standalone)
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -257,11 +257,9 @@ bool init_CAN() {
       SPI2517_2->begin(sck_pin, sdo_pin, sdi_pin);
     }
 
-    // On boards that clock the 2nd chip from the 1st chip's CLKO output (a
-    // non-default MCP2517_CLKODIV), the divider is normally programmed when
-    // the 1st interface starts. Running the 2nd interface alone leaves the
-    // 1st chip at its divide-by-10 power-up default, clocking this chip 10x
-    // slower than MCP2517_FREQ2() declares, so program it here.
+    // Boards that clock the 2nd chip from the 1st chip's CLKO output normally get
+    // the divider programmed when the 1st interface starts. Running the 2nd alone
+    // leaves the 1st chip at its divide-by-10 default, so program it here.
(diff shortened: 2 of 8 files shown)


dalathegreat:
Review comments from RJSC on Discord

The AI went overboard:
New files: mcp2517fd_clko.cpp/h
many new functions in SPI.h
Only the same code already initializing the clock in canfd1 would be needed 

I think a simple bit flag like CANFD1_CLK initilazied with 0 and set to 1 in CAN1 initialization would be enough.
Then just put the can1 clock initialization inside an if statement in CAN2 init.

ekholm:
Agreed the ratio was wrong for the size of the fix. Cut down and pushed.

Where I would push back is on losing the test. My rule is that where a regression test is cheap, it gets written, and this one is about as cheap as they come: a pure predicate plus a single function taking `SPIClass&`, so the suite can assert the exact sequence (reset, then a write of `(clkodiv & 3) << 5` to OSC at 0xE00) against a recording fake.

That is also what decides where the code can live. `comm_can.cpp` is not in the host test build: `test/CMakeLists.txt` lists its production TUs explicitly and comm_can.cpp is not one of them, and neither are the ACAN libraries. So moving this into `comm_can.cpp` as a static helper, or onto the vendored ACAN2517FD as a method, does not just shrink the diff, it converts a cheap-to-test change into an untestable one. The saving is one-off; the loss lands on whoever next touches CLKODIV handling or the OSC bit layout. A bench confirm on real hardware would not substitute either, since it shows the sequence is right today, not that it stays right.

So I kept the separate TU and cut everything around it: comment blocks gone from the helper and the call site, header down to the two declarations, SPI fake down to only what the assertions use. I also dropped the test that pinned the 800 kHz transaction speed, since that asserted a preference rather than a correctness property. 176 added lines down to 131, with the gate and the SPI sequence still covered.

On the flag, `fd1_in_use` is already that, just derived from the receiver config rather than a mutable global.

The one direction I would rule out regardless is calling `begin()` on chip 1 to reuse its clock setup. The OSC write sits mid-`begin()` at `ACAN2517FD.cpp:399`, after `initCS()`, the configuration-mode request, `reset2517FD()` and the RAM connectivity check, and before the FIFO setup and `attachInterrupt`. Calling it would attach an ISR to chip 1's INT pin and bring up an interface nobody configured.

PR #2867, opened 2026-08-26
Make the Lilygo T-2CAN use INT rather than INT0/INT1

jonny5532:
### What
Due to me misreading the schematic, the previous LilyGo T-2CAN FD integration had a modified HAL and driver to use the `INT0`/`INT1` pins instead of `INT` (since `INT` wasn't very obvious).

This rolls back those changes, which also fully restores the use of `IO3` for `BMS_POWER`

Well done @ekholm for spotting it!

### Why
Simpler code, and means people won't have to mess with DIP switches.

> [!TIP]

### DIFF
--- a/Software/src/communication/can/comm_can.cpp
+++ b/Software/src/communication/can/comm_can.cpp
@@ -190,24 +190,12 @@ bool init_CAN() {
 
     auto cs_pin = esp32hal->MCP2517_CS();
     auto int_pin = esp32hal->MCP2517_INT();
-    auto int0_pin = esp32hal->MCP2517_INT0();
-    auto int1_pin = esp32hal->MCP2517_INT1();
 
-    if (!esp32hal->alloc_pins("CANFD", cs_pin)) {
+    if (!esp32hal->alloc_pins("CANFD", cs_pin, int_pin)) {
       return false;
     }
-    if (int_pin != GPIO_NUM_NC) {
-      if (!esp32hal->alloc_pins("CANFD", int_pin)) {
-        return false;
-      }
-    } else {
-      if (!esp32hal->alloc_pins("CANFD", int0_pin, int1_pin)) {
-        return false;
-      }
-    }
 
-    canfd = new ACAN2517FD(cs_pin, *SPI2517, int_pin != GPIO_NUM_NC ? int_pin : 255,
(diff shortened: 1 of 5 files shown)


ekholm:
Tested — this is the same change I measured before you filed it, so here are the numbers as
test evidence.

Rig: two T-2CAN FD boards back-to-back, classic 500k, peer offering ~425 frames/s,
**receiving board fully silent** (important — see below). Same image, one define apart,
A/B on the HAL half of exactly this PR's change
([our equivalent branch](https://github.com/ekholm/Battery-Emulator/commit/a5f8a2303308ce80b92fd5ae30baa1557bae3065),
which this PR supersedes and cleans up properly):

| configuration | RX at the application | interrupts |
|---|---|---|
| stock pins (`INT0=9/INT1=3`, DIP switches as shipped) | 55–56 frames/s | **0** |
| `INT` on GPIO8 (this PR's pin map) | **292 frames/s** | **exactly 1 per frame** (14,630 : 14,632) |

The stock 55/s is the driver's 500 ms timeout drain doing all the work (27-deep FIFO / 0.5 s
= 54 predicted), reproduced to the frame across three independent measurements. With the
dedicated INT the ceiling is gone; the residual gap to line rate is per-frame SPI servicing
cost, not interrupts — a separate topic.

Two notes for anyone else testing:

- **Loopback cannot verify this.** In loopback the same broken pin map measures a
  healthy-looking ~480 frames/s, because every received frame was just transmitted and the TX
  interrupt (which *is* wired, GPIO9) drains the RX FIFO as a side effect. You need a peer and
  a silent DUT.
- **DIP switches stop mattering with this PR.** As shipped they're open (under the Kapton
  tape), which is why the old INT1-on-IO3 path never delivered an RX interrupt out of the box;
  with the dedicated INT they can stay untouched in either position — though if GPIO3 is used
  for BMS_POWER (restored by this PR), they should stay open so the chip's INT1 pad and
  BMS_POWER never share the net.

Also checked: this composes with #2652 (the driver-file edits merge cleanly; only
`platformio.ini` needs a trivial resolution, whichever lands second).

dalathegreat (approved):

dalathegreat:
Unit test PrechargeControlTests.StepSizeFollowsTheDistanceBands fails due to unrelated changes in another PR. Merging this one!

PR #2868, opened 2026-08-26
Precharge: reset the last-voltage-reading when starting

jonny5532:
### What
This was initially a test fix because the test-shuffle was causing the precharge test to fail.

It turns out that the precharge (used by MEB) doesn't reset the `prev_external_voltage` last voltage reading to the initial bogus 2000V value when it starts, meaning that a repeat precharge doesn't start in exactly the same way. This now resets it, which will change the behaviour and fix the test.

**CAVEAT:** - I'm not familiar enough with the MEB precharge to know if this is now the correct behaviour, or the test was wrong. It looks like this will have hardly any effect, as the chance of the new reading exactly matching the previous value is very small, but would be good to be sure.

> [!TIP]


dalathegreat (approved):
Yup, agreed 👍

PR #2879, opened 2026-08-28
Feature: BYD cell balance timers page

demon1300:
### What
Adds a **Cell Balance Timers** page for the BYD integration, opened from a button on More Battery Info.

### Why
The BMS keeps a lifetime balancing-hours counter for every cell, but nothing exposed it. You can now see which cells the BMS is working hardest on, and which ones it bled recently.

### How
Pressing **Read Timers** queries the pack over UDS and returns the per-cell hours as JSON. The page compares your last two readings to work out which cells were balanced in between — the pack has no live balancing signal, so a difference in the totals is the only reliable indicator.

The page is stored on the device already compressed, so it takes about a third of the space (4.9 KB instead of 13.5 KB) and doesn't have to be built up again every time someone opens it. If you change the HTML, run tools/gen_balance_page.py to refresh it.

Scans are only manual on this initial release.

image

> [!TIP]


nagyrobi:
Can you post a screenshot please for us who don't own a BYD battery?

demon1300:
> Can you post a screenshot please for us who don't own a BYD battery?

done, check the original post.  This was done using dummy data :)

dalathegreat:
@demon1300 , I just merged your other PR #2880 which caused a merge conflict in the unit tests

demon1300:
@dalathegreat should be clear to merge now!

dalathegreat (approved):

PR #2880, opened 2026-08-28
Fix: Re-disable the BYD insulation monitor when it re-arms after a contactor open

demon1300:
### What
Re-send the insulation monitor disable when the BMS re-arms it on its own after contactor opening.

### Why
Opening the contactors makes the BMS re-arm its monitor about 30s later, which isn't a BMS start so the disable never got re-sent — the pack then stores an insulation fault and derates discharge to ~7% until it's physically power cycled.  This fix is important for the balancing update, which requires contactors to be opened and re-close if the full discharge limit is to be kept.

### How
Arm the re-assert on the 0x35E measurement-active edge as well as the existing BMS-alive edge.

> [!TIP]

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -346,6 +346,12 @@ void BydAttoBattery::
       }
     }
     bms_was_alive = bms_alive;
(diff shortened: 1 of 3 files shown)


dalathegreat (approved):
Sensible change, merging this!

PR #2884, opened 2026-08-29
SOLAX-CAN: questions about four fields, based on two real captures

cringvig:
While bringing up an AKASOL pack against a SolaX X3-Hybrid G4 I ended up comparing `SOLAX-CAN.cpp` byte for byte against two independent real-world references:

- a CAN log from a working SolaX installation (343.7 V pack)
- a DBC from an Orion BMS2 driving a SolaX, in service for over two years (453 V pack)

In four places the two references agree with each other and differ from the driver. I changed them locally to match, but **I am raising these as questions, not as bug reports** — if there is a reason behind the current behaviour that I have not understood, I would much rather hear it than have anything changed.

Worth saying up front: **none of these were needed to make my system work.** The fault that blocked me for three weeks turned out to be corrupted DSP firmware in the inverter, cleared by reflashing the same version number over itself. These are just differences I noticed along the way.

### 1. `0x1873` bytes 6-7 — remaining energy unit

Currently `reported_remaining_capacity_Wh / 10`. One capture reads 419 here alongside a total of 944 in `0x1878`, which is consistent as 41.9 kWh of 94.4 kWh, i.e. 0.1 kWh units. With `/10` my 33 kWh pack transmits 1650 — read as 0.1 kWh that would be 165 kWh remaining out of a 33 kWh pack.

### 2. `0x1878` bytes 4-7 — total capacity layout

Currently a 32-bit raw Wh value across bytes 4-7. Both captures instead use a 16-bit value in bytes 4-5, byte 6 a constant `04`, byte 7 zero:

    capture A:  B0 03 04 00     ->  944 = 94.4 kWh
    capture B:  B0 00 04 00     ->  176 = 17.6 kWh

I do not know what byte 6 signifies — the 32-bit form leaves it at zero.

### 3. `0x1874` bytes 4-7 — the 3.0→3.5 V rescale

The captures carry the actual cell voltage in 0.1 V units (`41` = 4.1 V) with no compression, and `0x1876` already sends the real value — so the two frames report different figures for the same cells. The rescale may well exist for a reason I have not found, which is why this one is a question rather than a patch.

### 4. `0x1877` bytes 6-7

Both captures send `0x1D` / `0x10`; the driver sends `0x22` / `0x02`. The comment in the source marks these as uncertain, so matching the captures seemed the safer choice for my setup.

---

### Separately: `0x1877` byte 1 is a BMS alarm register

Not a change — the driver correctly leaves this byte alone. But I mapped it bit by bit against a live inverter and it seems worth documenting, because I wasted days on it by writing to it:

| Bit | Value | Fault | Bit | Value | Fault |
|---|---|---|---|---|---|
| 0 | 1 | BE09 CellImbalance | 4 | 16 | BE13 BMS_VolSen |
| 1 | 2 | BE10 BMS hardware | 5 | 32 | BE14 BMS temp sen |
| 2 | 4 | BE11 BMS circuit | 6 | 64 | BE15 BMS cur sen |
| 3 | 8 | BE12 BMS iso fault | 7 | 128 | BE16 BMS relay |

The display shows the **lowest set bit** — `0x50` (bits 6+4) renders as BE13, `0xFF` renders as BE09. Twelve data points, no exceptions, and the last two rows were predicted before being measured.

I had put the battery type into this byte on a hunch from a spreadsheet. 131 is `0b10000011` — bit 0 set — so I was literally commanding the inverter to report cell imbalance, then chasing the BE09 it dutifully displayed. If anyone reports BE09 through BE16 on a Battery-Emulator setup, it is worth checking that nothing is writing to `0x1877` byte 1 before looking at the battery.

Builds clean against current `main` (`lilygo_2CAN_330`).

Happy to split this into separate PRs, drop any of the four, or just close it if the current behaviour is deliberate.

### DIFF
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -212,8 +212,7 @@ void solax_pack_identity_ascii(const uint8_t mac[6], uint8_t slot, uint8_t half,
   serial[0] = '6';
   serial[1] = 'S';
 
-  uint32_t seed = ((uint32_t)mac[0] << 24) | ((uint32_t)mac[1] << 16) | ((uint32_t)mac[2] << 8) |
-                  (uint32_t)mac[3];
+  uint32_t seed = ((uint32_t)mac[0] << 24) | ((uint32_t)mac[1] << 16) | ((uint32_t)mac[2] << 8) | (uint32_t)mac[3];
   seed ^= (((uint32_t)mac[4] << 8) | (uint32_t)mac[5]);
   seed += (uint32_t)slot * 7919u;
 
--- a/Software/src/inverter/SOLAX-CAN.cpp
+++ b/Software/src/inverter/SOLAX-CAN.cpp
@@ -11,6 +11,33 @@
 #define Contactor_Open_Payload __builtin_bswap64(0x0200010000000000)
 #define Contactor_Close_Payload __builtin_bswap64(0x0200010001000000)
 
+// ---------------------------------------------------------------------------------------
+// SolaX assumes a fixed number of cells per configured module. Derived from the published
+// type table, not guessed: the type-131 row for 4 modules reads 179.2 / 204.8 / 233.6 V
+// (min / nominal / max). Per module that is 44.8 / 51.2 / 58.4 V, and per 16 cells it lands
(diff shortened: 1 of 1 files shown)


dalathegreat:
Seems like the S/N from MAC ID should be actually merged. Care to make a PR only contain that @cringvig ?

cringvig:
@dalathegreat sure, I will revert.....

PR #2885, opened 2026-08-29
AKASOL AKASYSTEM  battery driver 

cringvig:
Adds support for the AKASOL AKASYSTEM 15 OEM 50 PRC — a single-tray, 655 V nominal NMC pack (180 cells, 50 Ah, ~33 kWh) with one BMM01 controller. Running here on a LilyGo T-2CAN against a SolaX X3-Hybrid-6.0-D, charging and discharging normally.

### Wake-up sequence

The BMM01 will not come up if its discrete inputs are asserted together, so the driver runs a timed state machine that follows the order the AKASOL manual specifies, holding each step long enough for the previous one to settle:

| State | Action | Hold |
|---|---|---|
| `INIT_HW` | all GPIOs low | – |
| `KL30_SAFE_ON` | KL30_safe alone | 500 ms |
| `KL30_ON` | KL30 added | 1000 ms |
| `REQUEST_USE` | Wake added, then `req_batuse=1` once Standby is seen | until Standby |
| `RUNNING` | battery reported Operational | – |

A `VCU1_to_BMM01` frame goes out every 100 ms with a rolling alive counter and the validation code `0x55AA`. CAN is 250 kbit/s with 29-bit IDs; BMU address `0xF3` (Tray01), our VCU address `0xE3`.

### Current limiting

AKASOL reports what the pack can do (~137 A charge, ~180 A discharge). That is far beyond what a typical inverter battery port accepts, so the driver clamps to 30 A before the value reaches the datalayer.

### Diagnostics

A status page shows BMM01 state, contactor and safety-loop flags, and all 64 error/alarm/warning bits by name. The names come from AKASOL's own PublicCAN.sym. Descriptions were deliberately left out to keep the flash cost to about 1.2 kB — flash only, nothing in RAM.

### Notes

- Only the single-tray BMM01 variant is implemented. Multi-tray systems use a different addressing scheme and are not covered.
- `AKASOL_ISOLATED_BMU` is 0. The isolated-BMU path is described in the manual but is not implemented and was not needed here.
- Tested against a SolaX X3-Hybrid G4. Other inverters are untested.
- Builds clean against current `main` (`lilygo_2CAN_330`).

Happy to adjust naming, structure or anything else to fit the project's conventions.


Software/src/battery/AKASOL-BATTERY.h
```diff
@@ -0,0 +1,221 @@
...
+// board (Stark/LilyGo/etc). Pick pins that are free on your hardware.
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+// >>> SSR WIRING LIVE <<<
+// -1 test build ruled out (no functional difference vs 16/47/15, as expected
+// since nothing was physically wired to the GPIOs at the time). SSRs are now
+// soldered and confirmed wired to IO16/IO47/IO15 (GPIO sources +3.3V on
+// assert, common LilyGo GND is the SSR's "-" reference) - pins restored so
+// the driver's bounce-free timed state machine (KL30_safe -> settle -> KL30
+// -> settle -> Wake, see transmit_can()) actually drives the real signals.
+#define AKASOL_PIN_KL30 16       // Expansion header IO16 - drives KL30 SSR
```
dalathegreat: Can you change these pin definitions to be same as Pre-Pos-Neg contactor definitions? That way it will work on any BE board, and not only the LilyGo board you have been using.

dalathegreat (changes requested):

Software/src/battery/AKASOL-BATTERY.h
cringvig: Happy to do that. Just to make sure I do it the way you want:

Do you mean reusing PRECHARGE_PIN() / POSITIVE_CONTACTOR_PIN() /
NEGATIVE_CONTACTOR_PIN() directly for KL30_safe / KL30 / Wake, since
those are already defined on every board? Or adding three new
AKASOL_* entries to hal.h in the same style, defaulting to
GPIO_NUM_NC?

Reusing the contactor pins looks like the better fit to me, but it
does mean the driver and GPIO contactor control can't both allocate
them. For Akasol that seems fine — the battery closes its own
contactors over CAN, so contactor control should be off anyway — but
I wanted to check you're happy with that before I change it.

dalathegreat: Reusing the pins definitions directly is fine!

cringvig: Done — pin defines replaced with the HAL contactor pins:

  PRECHARGE_PIN()          -> KL30
  NEGATIVE_CONTACTOR_PIN() -> KL30_safe
  POSITIVE_CONTACTOR_PIN() -> Wake

I rewired my board to the new pins and flashed it — the startup
sequence runs and the battery closes its contactors exactly as before.

Two notes on the diff:

- I used alloc_pins_ignore_unused() rather than alloc_pins(), so a board
  that leaves any of these at GPIO_NUM_NC still comes up CAN-only
  instead of refusing to start.

- The mapping isn't in the order you'd expect. Losing KL30 drops the BMU
  supply, and losing KL30_safe is an eStop that latches an Error the
  battery only clears through a full restart sequence, while dropping
  Wake for a moment is harmless. So the two critical signals get the
  pins that can be held across a reboot and Wake takes the leftover —
  on the T-2CAN that puts KL30 and KL30_safe on RTC-capable GPIOs.

dalathegreat:
Failing unit test can be fixed like this: https://github.com/dalathegreat/Battery-Emulator/pull/2887/changes/491597295c1d47f5f62e9a25c7fbd19d53dfb856

cringvig:
**Failing unit test**

Thanks for the pointer — same cause. AKASOL-BATTERY.cpp wasn't in
test/CMakeLists.txt. Added, and merged current main so your BYD fix is
in as well.

dalathegreat:
Nice work!

I added Akasol to the Wiki page, feel free to continue editing this page with more info that makes it easier for next person trying it out! https://github.com/dalathegreat/Battery-Emulator-Wiki/pull/43

dalathegreat (approved):

PR #2889, opened 2026-08-30
Nissan LEAF 🔋: apply Hx raw-to-percent scale on ZE0/AZE0 too

nagyrobi:
### What
Applies the raw -> percent conversion to Hx for ZE0/AZE0 packs, not only ZE1.

### Why
"More battery info" reported Hx 100.00 % on my AZE0 pack that LeafSpy reads as 97.66 %. All generations likely report Hx in the same 1/102.4-per-percent unit; only the payload offset differs, so the ZE0/AZE0 branch would also be in parity with LeafSpy. 

kép
kép

### How
Reads the raw value per generation, then converts once. Verified on my 30 kWh AZE0: group 1 frame `24 5c 27 10 ...` gives 0x2710 = 10000 -> 97.66 %, matching LeafSpy. SOC at the adjacent offset decodes to 95.95 %, matching `0x55B` in the same capture.

> [!TIP]

### DIFF
--- a/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
+++ b/Software/src/battery/NISSAN-LEAF-BATTERY.cpp
@@ -585,18 +585,21 @@ void NissanLeafBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
         }
 
         if (rx_frame.data.u8[0] == 0x24) {  // Fifth frame
-          // Hx sits at a different payload offset and uses a different scale depending on which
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):

nagyrobi:
Hmm just tested with SafetyUggs's [Leaf-BMS-Report](https://github.com/SafetyUggs/Leaf-BMS-Report---ZE0-AZE0-ZE1):
kép

PR #2898, opened 2026-08-31
Fix: Modbus FC06 response should echo the register address

dalklein:
### What
This PR implements correctly echoing the address of a register write command
A Modbus FC06 response is a byte-for-byte ECHO of the request:
     [slaveID][0x06][addrHi][addrLo][valHi][valLo]
The previous version omitted the address and returned only
     [slaveID][0x06][valHi][valLo]
which is 4 bytes instead of 6 and is not a valid FC06 response.

### Why
At least my Delta inverter & LG Resu Prime rs485 comms fail without it.  Not sure that anyone else had used this functionality yet.  Searching project shows that `BYD-MODBUS` is currently the only Modbus RTU inverter protocol, and BYD never writes registers — so `FC06` has likely not been run.

### How
 Add the addr to the response, looks like it was missing.  The existing response left out the addr.

### Types of changes

### Checklist:

  - [x] The code change is tested and works locally.    Needed for my inverter & battery startup, several successful power up sequences (only happens on startup) where inverter configures the battery with FC06 commands, and several days now of running nominally.  BE rs485 is communicating with inverter, and separate RPi rs485 talking to battery, mqtt between them to form MITM.

Verified against a real device: an LG RESU10H Prime answering a Delta E6-TL-US
echoes the full request on every fn6 write (21 writes captured 2026-08-26 across
four inverter power cycles, e.g. request 0f0604 4e10 04.. -> identical response).
Masters that validate the echo -- the Delta does -- reject the short form, which
breaks the boot handshake and leaves the battery unconfigured.

### DIFF
--- a/Software/src/inverter/ModbusInverterProtocol.cpp
+++ b/Software/src/inverter/ModbusInverterProtocol.cpp
@@ -83,8 +83,8 @@ ModbusMessage ModbusInverterProtocol::FC06(ModbusMessage request) {
   // Do the write
   mbPV[addr] = val;
 
-  // Set up response
-  response.add(request.getServerID(), request.getFunctionCode(), mbPV[addr]);
+  // Set up response, including addr
+  response.add(request.getServerID(), request.getFunctionCode(), addr, mbPV[addr]);
   return response;
 }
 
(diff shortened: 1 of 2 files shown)


dalathegreat:
Successfully deployed on my Fronius system which uses BYD-Modbus :+1:

dalathegreat (approved):

PR #2901, opened 2026-08-31
Nissan LEAF🔋: read SOH, Hx, capacity Ah and VBAT level from the LBC health block

### LINKED ISSUE #1822: [BUG] Nissan LEAF SOH% calculation stuck
**Describe the bug**
SOH% never updating is a bug. The Nissan documentation on how to re-use a LEAF battery is wrong, and simply following that documennt will lead to SOH% being stuck on whatever value it was before it was taken into stationary storage use.

To fix this, you need to send more of the full vehicle towards the battery. I remember that this PR on the LEAF-battery-upgrade solved it for ZE0 https://github.com/dalathegreat/Nissan-LEAF-Battery-Upgrade/pull/28

So to proceed here, we need to read DTCs with Leafspy, andn see what codes are active. Then start sending more vehicle CAN messages to get the DTCs to clear. After no DTCs are active, SOH% calculation is resumed. 

**To Reproduce**
Use a LEAF battery. Observe how SOH% never goes down, even after years of use

**Expected behavior**
SOH% goes down as battery ages

**Screenshots**

- Software version: X.Y.Z
- Battery used: [Nissan LEAF]
- Inverter communication protocol: [BYD Modbus]
- Hardware used for Battery-Emulator: [LilyGo T-CAN/Stark CMR/LilyGo T-2CAN]
- GPIO controlled contactors: [yes/no]
- MQTT: [yes/no]

nagyrobi:
### What
Reads SOH, Hx, pack capacity and the 12 V level from the LBC health block (group 0x61) and shows capacity, derived kWh and VBAT on More Battery Info. SOH reports as unknown until the pack supplies one. Cells the LBC returns as 0xFFFF are reported as unknown rather than 0 V.

image

### Why
SOH came from `0x5BC` at whole-percent resolution and started at a hardcoded 99 % that looked like a reading. Capacity in Ah and the 12 V level were never decoded at all, and sentinel cells were charted as 0.000 V.

### How
Group 0x61 joins the polling rotation; its reply exceeds 255 bytes, so the group latch is masked and the length widened to the full 12-bit ISO-TP field. A per-reply frame counter replaces the sequence number for locating payload offsets. `soh_available` gates display and MQTT while `soh_pptt` keeps its safe default for the inverter and safety paths.

### Types of changes

- fixes https://github.com/dalathegreat/Battery-Emulator/issues/1822

- wiki PR <link to PR>

### Checklist:

> [!TIP]
> 


nagyrobi:
The capacity calculated from Ah matches reality a whole lot closer than what we had before:
image
(this is a SoH resette pack)

Question is, now, how should we use the new value?

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -291,6 +308,8 @@ void NissanLeafBattery::
     datalayer_nissan->MaxPowerForCharger = battery_MAX_POWER_FOR_CHARGER;
     datalayer_nissan->Interlock = battery_Interlock;
     datalayer_nissan->Insulation = battery_insulation;
+    datalayer_nissan->CapacityCAh = battery_capacity_cAh;
+    datalayer_nissan->VBAT_mV = battery_vbat_mV;
```
dalathegreat, line 328: Nice! Now when battery_vbat_mV has been read (larger than 0), please if you can, add a safety check for low 12V

```
  if (battery_vbat_mV < 11000) {  //11.000V
    set_event(EVENT_12V_LOW, battery_vbat_mV);
  }
```

nagyrobi:
We could estimate a SOH value based on the Actual capacity we can get from Ah compared to the "real" capacity.
On my AZE0 real is 28.9, the one derived from 60.64 Ah using nominal voltage is 21.83. The proportionally estimated SOH is 75.53 % which corresponds to reality.

Before SOH reset I had:
image

Let's add a checkbox in Battery Settings like "Use estimated SOH and Capacity" and treat the values calculated from Ah as real capacity and SOH?

dalathegreat:
Yes we could estimate a SOH value, but what is better;
- We see if actual SOH% moves (which signifies a proper shutdown procedure)
- We get a good estimated SOH% (withouth needing the proper shutdown procedure)

I think we should not add too many checkboxes, we should just decide to use either the actual SOH or the estimated from capacity SOH.

nagyrobi:
Well since SOH freezes in stationary use as we know it, estimated SOH at least would follow the polled Ah value, which seems to self-calibrate after SOH reset. Will do a new SOH reset again to confirm.

nagyrobi:
I tried resetting SOH again on the same pack. Well, couldn't see much difference, as it's still showing the same 100% for Hx and SOH. But Actual capacity is yet 60.64 Ah (21.83 kWh) so I guess that doesn't really reset I guess.

But still seems to do tiny decreases - look at the LeafSpy row shot above - was taken before the first reset I made months ago. Lost 0.01 Ah since.

nagyrobi:
We can leave this question to be decided in a future PR.

Until then, this one is open to review/consideration.

Software/src/battery/NISSAN-LEAF-BATTERY.cpp
```diff
@@ -576,12 +627,25 @@ void NissanLeafBattery::handle_incoming_can_frame(CAN_frame rx_frame) {
...
 
-        if (rx_frame.data.u8[0] == 0x23) {  // Fourth frame
+        if (rx_frame.data.u8[0] == 0x23) {  // Fourth frame, payload[20..26] in u8[1..7]
           battery_insulation = (uint16_t)((rx_frame.data.u8[5] << 8) | rx_frame.data.u8[6]);
           if (battery_insulation > 0) {
             datalayer_battery->status.insulation_resistance_kOhm = battery_insulation;
             datalayer_battery->status.insulation_resistance_available = true;
           }
+          //12 V accessory battery level at payload[22..23], in mV. Documented on the ZE0/AZE0
+          //layout only, and read at the same offset on ZE1 on the chance that its layout carries
```
dalathegreat: Should we update this ZE1 comment

nagyrobi: done

nagyrobi:
image

nagyrobi:
Thanks to @safetyuggs, now we have a SOH raw value too.
kép

Pushing it out through MQTT for longer term monitoring and comparison with normalized SOH value
kép

dalathegreat (approved):

PR #2908, opened 2026-09-02
fix(safety): only run the CAN liveness watchdog on CAN batteries

dalklein:
### What

update_machineryprotection() guards the inverter liveness check on inverter->interface_type() == Can, but ran the battery check unconditionally.

A battery on a non-CAN interface never refreshes CAN_battery_still_alive, so the counter sits at 0 and EVENT_CAN_BATTERY_MISSING is raised on every cycle. That event is ERROR level, so system_status latches to FAULT, and the block at the top of the same function then zeroes max_charge_power_W and max_discharge_power_W. The result is a healthy battery reported to the inverter as accepting and delivering 0 W, with the status page pointing at CAN wiring that is not part of the setup. Clearing the event does not help, because the watchdog raises it again on the next cycle.

Battery gains interface_type(), mirroring InverterInterfaceType. It is defaulted to Can rather than pure virtual so every existing battery stays correct without a change, and only a non-CAN battery has to declare itself. RS485Battery declares Rs485, which covers DalyBms.

Adds two tests: a non-CAN battery is not policed by the CAN watchdog and keeps its power limits, and a CAN battery still is, so the guard is not a blanket disable.

### Why
Testing with an nonCAN data driven battery, a MQTT input data battery for testing, found the error.  Other existing battery types would be similarly affected.  Clearly most batteries use CAN data, hah!

### How
Only check CAN liveness for CAN batteries

### Types of changes

### Checklist:

  - [x] The code change is tested and works locally.      Using a mqtt fed battery in BE found this bug and needed a fix. Claude finds that it would affect other existing battery types, so this might be of interest.

### DIFF
--- a/Software/src/battery/Battery.h
+++ b/Software/src/battery/Battery.h
@@ -75,6 +75,9 @@ extern bool user_selected_triple_battery;
 
 extern battery_chemistry_enum user_selected_battery_chemistry;
 
+// The type of comm interface the battery is using. Mirrors InverterInterfaceType.
+enum class BatteryInterfaceType { Can, Rs485, Modbus };
+
 // Abstract base class for next-generation battery implementations.
 // Defines the interface to call battery specific functionality.
 class Battery {
@@ -84,6 +87,9 @@ class Battery {
 
   // The name of the comm interface the battery is using.
   virtual const char* interface_name() = 0;
+
+  // The type of the comm interface. Defaults to CAN; override in batteries that use another.
+  virtual BatteryInterfaceType interface_type() { return BatteryInterfaceType::Can; }
 
   // These are commands from external I/O (UI, MQTT etc.)
   // Override in battery if it supports them. Otherwise they are NOP.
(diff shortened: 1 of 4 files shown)


dalklein:
Closing this — on a closer look the premise is wrong and the change would be a regression.

`DalyBms`, the only other non-CAN battery, already refreshes `CAN_battery_still_alive` itself
([DALY-BMS.cpp#L129](https://github.com/dalathegreat/Battery-Emulator/blob/main/Software/src/battery/DALY-BMS.cpp#L129)),
so no battery in the tree is actually affected. The counter is really a generic
"the battery answered recently" heartbeat rather than a CAN-specific one, and every shipped
driver takes part in it.

Worse, guarding the watchdog on the interface would **remove** liveness monitoring from
`DalyBms`, which has it today — if a Daly BMS went silent, nothing would raise an event any more.
That is a clear regression and I did not spot it before opening this.

The real problem was in my own out-of-tree battery driver, which sources its state over MQTT and
simply never refreshed the counter. Fixed there in one line, which also gives that driver the
liveness watchdog it was missing.

Apologies for the noise, and thanks for the CI — the order-independence check correctly failed
this at seed 7.

PR #2913, opened 2026-09-03
Add Growatt LV (GBLI-series) battery support

schmellic:
## Summary
- Adds a `GROWATT-LV-BATTERY` decoder for Growatt's low-voltage BMS CAN protocol (GBLI6532 and siblings). Existing Growatt support here targets the ARK HV pack; the `GROWATT-LV-CAN` inverter driver emulates this protocol toward a real Growatt inverter but doesn't consume it from a real battery - this fills that gap on the battery side.
- Outputs through the existing `PYLON-LV-CAN` inverter driver, unchanged - no new inverter-side code needed.
- This is a port of the decode side of [schmellic/growatt2solis](https://github.com/schmellic/growatt2solis), a project that's been running this exact protocol translation live against a real GBLI6532 and a Solis hybrid inverter. Frame layout, byte order, and the specific bits used for charge/discharge-enable and pack count were confirmed against real capture data rather than the spec alone - including one case (`0x319`'s own enable bits) where the spec-plausible reading doesn't match what a real pack actually does. See the header comment in `GROWATT-LV-BATTERY.h` for details.
- Second commit fixes the outgoing `0x301` keepalive payload, which was initially ported using Growatt's protocol-doc *example* value rather than the real payload captured off a genuine SPH6000 - this exact mistake caused a real ~594s enable-then-disable bug in the upstream project for its entire history until it was found and fixed today. Now uses the real captured value.
- Third commit fixes three more bugs found running this live end-to-end for the first time (real GBLI6532 -> this driver -> `PYLON-LV-CAN` -> a real Solis S6-EH1P8K-L-PLUS):
  - `max_charge_power_W`/`max_discharge_power_W` were never set by this driver. The generic safety layer in `safety.cpp` zeroes the *current* fields back to 0 whenever the *power* fields read 0, so every current value this driver computed was silently discarded every cycle - the Solis saw a battery permanently reporting 0A in both directions despite `chg_en`/`dis_en` being correctly decoded as enabled.
  - `total_capacity_Wh`/`remaining_capacity_Wh` were left at the datalayer's generic 30kWh/0Wh defaults. Now decoded from Growatt's `0x314` frame (received but previously discarded) and converted to Wh using the real pack voltage.
  - **`PYLON-LV-CAN`'s own `MANUFACTURER_NAME` (`"BatEmuLV"`) caused a hard "Battery Name Failure" on the real Solis.** The Solis validates this field strictly on the `PYLON_LV` battery profile and expects the literal string a genuine Pylontech battery sends. Changed to `"PYLON   "`, matching what growatt2solis already proved works against this same real inverter. **This is a default-behavior change for every inverter using `PYLON-LV-CAN`, not just this new Growatt LV battery** - flagging it explicitly since it's outside this PR's own battery driver and may be worth a second look/its own discussion rather than folding in silently.

## Test plan
- [x] Builds clean on `esp32devkit_330`
- [x] Real-world validated in `growatt2solis`: with the corrected `0x301` payload, a real GBLI6532 stayed enabled 50+ minutes with zero disables (vs. reliably disabling at ~594s with the placeholder payload) - see that project's `CLAUDE.md` open question 8 for the full investigation trail
- [x] Now also validated running inside Battery-Emulator itself, end-to-end, against real hardware: real GBLI6532 -> this driver -> `PYLON-LV-CAN` -> a real Solis S6-EH1P8K-L-PLUS. Confirmed correct SOC/voltage/current/temperature, correct real capacity (Ah-derived kWh), correct charge/discharge current up to the real BMS ceiling (208.3A on this 2-pack system) once the manual cap was raised above it, and the Solis's own display cleared its "Battery Name Failure" and showed normal generating status once the manufacturer-name fix landed.

### DIFF
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -27,6 +27,7 @@
 #include "GEELY-GEOMETRY-C-BATTERY.h"
 #include "GEELY-SEA-BATTERY.h"
 #include "GROWATT-HV-ARK-BATTERY.h"
+#include "GROWATT-LV-BATTERY.h"
 #include "HYUNDAI-IONIQ-28-BATTERY.h"
 #include "IMIEV-CZERO-ION-BATTERY.h"
 #include "JAGUAR-IPACE-BATTERY.h"
@@ -130,6 +131,8 @@ const char* name_for_battery_type(BatteryType type) {
       return GeelyGeometryCBattery::Name;
     case BatteryType::GrowattHvArk:
       return GrowattHvArkBattery::Name;
+    case BatteryType::GrowattLv:
+      return GrowattLvBattery::Name;
     case BatteryType::HyundaiIoniq28:
       return HyundaiIoniq28Battery::Name;
     case BatteryType::OrionBms:
@@ -282,6 +285,8 @@ Battery* create_battery(BatteryType type) {
       return new GeelyGeometryCBattery();
     case BatteryType::GrowattHvArk:
       return new GrowattHvArkBattery();
+    case BatteryType::GrowattLv:
+      return new GrowattLvBattery();
     case BatteryType::HyundaiIoniq28:
       return new HyundaiIoniq28Battery();
     case BatteryType::OrionBms:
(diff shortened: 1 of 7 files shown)


dalathegreat (approved):
Nice work! 

Could you update the Wiki page I made for Growatt LV Batteries? https://dalathegreat.github.io/Battery-Emulator-Wiki/battery/growatt_lv/

Merging this 🙌

PR #2917, opened 2026-09-04
Tesla: Fix 0x2B4, 0x392 and 0x332 and HVP_debugMessage 0x7AA decoding on newer pack firmware

greenoem:
### What

This PR decodes PCS_dcdcRailStatus (0x2B4), BMS_packConfig (0x392) and BMS_bmbMinMax (0x332) correctly for both older and newer pack firmware, instead of only pre-2023 as is currently the case. It also corrects the mux 0 decode of HVP_debugMessage (0x7AA), reads the shunt current debug value from mux 7 where it actually is, and removes shuntRefVoltageMismatch, shuntThermistorMia and shuntHwMia as they will be wrongly flagged as being active when the pack is running >2025.8.4 firmware (which some users on Discord have reported).

As a side note, 2019 firmware findings should not be used as any kind of reference nowadays - they are firmly out of date.

### Why

Tesla changed these frames during 2023 and the BE code parsed the old layouts. On any pack firmware since then, the PCS LV/HV bus voltages and 12V current were wrong on More Battery Info, as well as via MQTT/ESP-NOW - CAN logs showed Platform Max Bus Voltage was 24 V low for some packs, and the Brick Temp Max/Min values were swapped. With regards 0x7AA, the mux 0 layout in the BE code matches no firmware version since 2022.40. HVP_hvp1v5Ref was read from bit 28 and showed 0.3 V for a 1.5 V reference, gpio bits 22 to 25 had incorrect names, and the shunt current and MIA flags were read from bits 40 to 61 (where there is nothing defined).

### How

0x2B4 changed from DLC 5 to DLC 6 between firmware 2023.2.12 and 2023.6.11 (new bit positions and scaling: LV bus 0.01 V, HV bus 0.1 V, LV current signed).
0x392 platformMaxBusVoltage went from x0.1+375 to x1+300 with firmware 2023.2.12 (no change to the frame itself). The 0x2B4 frame length is used as the indicator for later firmware, and the datalayer now holds volts.
0x332 mux 0 swapped NumTMax/NumTMin and they added thermistorTAvg in byte 4 between firmware 2023.7.30 and 2023.12.1 - an average value that sits between TMin and TMax selects the newer firmware order.

HVP_debugMessage 0x7AA:

Bits 22 to 25 are gpioFcContPowerEnable, gpioHvilEnable, gpioPortSelSpiRdy and gpioPyroUnlock - bit 20 is gpioFcContFlywheelEnable.
HVP_hvp1v5Ref is at bit 25 up to 2023.7.30 and bit 26 from 2023.12.1, with the MIA flags shifting as well. The 0x332 thermistorTAvg check marks >2023.12.1, so we use the same decode flag.
HVP_shuntCurrentDebug is decoded from mux 7 bit 24 (signed).
packCurrentMia, auxCurrentMia and currentSenseMia are kept. shuntRefVoltageMismatch, shuntThermistorMia and shuntHwMia moved again at 2025.8.4 and 2025.44 (!) with nothing in the frame to identify those exact releases, so they have been removed rather than show wrongly.

As above, scaling changes also needed to propagate through to MQTT and ESPNow.

Verified against DBCs from firmware versions 2021.24.4 through to 2026.8.3, and across 5 different packs.

### Types of changes

Bugfix (non-breaking change which fixes an issue)


dalathegreat (approved):
Nice work as always @greenoem , approved and merged!

PR #2920, opened 2026-09-05
Fix PYLON-LV-CAN 0x359 over-current sign bug

### LINKED ISSUE #2919: PYLON-LV-CAN: 0x359 over-current error/warning has inverted sign, causes false alarms
## Summary

`PYLON-LV-CAN.cpp`'s `0x359` over-current error and warning checks compare `reported_current_dA` against the wrong sign of their own limit, so the *charge*-current check actually fires on ordinary *discharge* and vice versa.

## Where

`Software/src/inverter/PYLON-LV-CAN.cpp`, `update_values()`, the ERROR and WARNING blocks populating `PYLON_359`.

```cpp
// ERRORS
if (datalayer.battery.status.reported_current_dA >= (datalayer.battery.status.max_discharge_current_dA + 10))
    PYLON_359.data.u8[0] |= 0x80;   // "Discharge over current"
...
if (datalayer.battery.status.reported_current_dA <= -1 * datalayer.battery.status.max_charge_current_dA)
    PYLON_359.data.u8[1] |= 0x01;   // "Charge over current"
```

The datalayer's own current-sign convention is **positive = charging** (confirmed via `reported_current_dA`'s usage elsewhere, e.g. directly written into `PYLON_356` as the reported pack current). Under that convention:
- Discharge current is **negative**, so the discharge-overcurrent check should compare against `-max_discharge_current_dA`, not `+max_discharge_current_dA`.
- Charge current is **positive**, so the charge-overcurrent check should compare against `+max_charge_current_dA`, not `-max_charge_current_dA`.

As written, both checks are effectively testing the wrong direction. The bug is most visible the instant `max_charge_current_dA` reaches 0 (e.g. a battery that's reached 100% SOC and correctly stops accepting charge) - at that point **any** discharge current at all (negative) satisfies `current <= -0`, permanently firing a false "charge over current" error and warning on every single cycle for as long as the battery is full and discharging, which is completely normal operation.

The same inversion exists in the WARNING block a few lines below (bytes 2/3).

## Real-world impact

Reproduced live against a real Solis S6-EH1P8K-L-PLUS paired with a Growatt GBLI6532 (via the new `GROWATT-LV-BATTERY` driver in #2913): the moment the battery reached 100% SOC and began a normal discharge to serve house load, the Solis raised a genuine "BMS error" alarm, even though the real BMS itself (`0x311`/`0x312`) reported a completely healthy state (`chg_en=1 dis_en=1`, no protection/warning bits set, correct CCL/DCL). The alarm cleared immediately after correcting the sign.

This affects every battery type routed through `PYLON-LV-CAN`, not just Growatt LV - any battery/inverter pairing using this driver will see the same false alarm the first time the battery reaches full charge while discharging.

## Fix

PR incoming - flips the sign comparisons for both the error and warning over-current checks to match the datalayer's actual current-sign convention, with a comment explaining why.

🤖 Found and reported with [Claude Code](https://claude.com/claude-code)

schmellic:
Fixes #2919.

## Summary
- `PYLON-LV-CAN.cpp`'s `0x359` over-current error/warning checks compared `reported_current_dA` against the wrong sign of their own limit. The datalayer's convention is **positive = charging**, so:
  - The discharge-overcurrent check needs to compare against `-max_discharge_current_dA` (discharge current is negative), not `+max_discharge_current_dA`.
  - The charge-overcurrent check needs to compare against `+max_charge_current_dA` (charge current is positive), not `-max_charge_current_dA`.
- As written, both checks tested the wrong direction. Most visible the instant `max_charge_current_dA` reaches 0 (e.g. battery at 100% SOC) - at that point any discharge current at all satisfies the old (broken) charge-overcurrent condition, firing a false error and warning on every cycle during completely normal operation.
- Same inversion existed in the WARNING block, fixed identically.
- Affects every battery type using `PYLON-LV-CAN`, not just any one driver.

## Test plan
- [x] Builds clean on `esp32devkit_330`
- [x] Reproduced and fixed live against a real Solis S6-EH1P8K-L-PLUS + Growatt GBLI6532 pairing (via #2913): confirmed the real BMS reported a fully healthy state (`chg_en=1 dis_en=1`, zero protection/warning bits, correct CCL/DCL) while the Solis raised a genuine "BMS error" purely from this sign bug. Alarm cleared immediately once the sign was corrected, confirmed on the real inverter's own display.

### DIFF
--- a/Software/src/inverter/PYLON-LV-CAN.cpp
+++ b/Software/src/inverter/PYLON-LV-CAN.cpp
@@ -50,7 +50,15 @@ void PylonLvInverter::update_values() {
   PYLON_359.data.u8[6] = 0x4E;  //N
 
   // ERRORS
-  if (datalayer.battery.status.reported_current_dA >= (datalayer.battery.status.max_discharge_current_dA + 10))
+  // reported_current_dA follows the datalayer's "+ = charging" convention, so
+  // discharge current is negative and charge current is positive - the two
+  // over-current checks below must compare against opposite signs of their
+  // own limit, not the same sign. (Previously both compared as if positive
+  // current meant discharge, which made a charge-current check trigger on
+  // ordinary discharge instead - most visible as a false "charge over
+  // current" error the instant max_charge_current_dA reaches 0, e.g. a full
+  // battery discharging normally.)
+  if (datalayer.battery.status.reported_current_dA <= -1 * (datalayer.battery.status.max_discharge_current_dA + 10))
     PYLON_359.data.u8[0] |= 0x80;
   if (datalayer.battery.status.temperature_min_dC <= BATTERY_MINTEMPERATURE)
     PYLON_359.data.u8[0] |= 0x10;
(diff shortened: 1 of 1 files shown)


dalathegreat (approved):
Massive thanks for the patch @schmellic 🙌

PR #2928, opened 2026-09-06
Stellantis Pro One: report SOH from 0x359 and add BPCM DTC descriptions

TPPS999:
Follow-up to #2926, on the same branch. Two independent changes.

## 1. SOH and pack energy from `0x359` bytes 2-3

Bytes 2-3 of `0x359` hold the pack capacity in 0.1 Ah.

Evidence from the captures:

- **Constant within every session**, in all 25 logs.
- **Differs per pack**: 3340 and 3339 on two packs at nominal, 3235 and 3052 on two aged ones.
- **Independent of SOC**: unchanged over a 13-point SOC span on one pack and 8.6 points on another.
- **Matches UDS `DA75`** in every log where both are present.
- Present in the battery-only captures too, so it is BPCM-sourced rather than vehicle data.

| b2b3 | Ah | soh_pptt | total_capacity_Wh |
|---|---|---|---|
| 3340 | 334.0 | 100.00 % | 110 620 |
| 3339 | 333.9 | 99.97 % | 110 587 |
| 3235 | 323.5 | 96.85 % | 107 143 |
| 3052 | 305.2 | 91.37 % | 101 082 |

**Caveat worth flagging in review:** the two denominators are datasheet nominals, not bus readings —
3340 = 334.0 Ah cell, 33120 = 90 × 3.68 V = 331.2 V. The 0.1 Ah reading itself is what the evidence
supports; the conversion to a *percentage* rests on the assumption that a healthy pack reports 3340. The
resulting 110.6 kWh against the pack's nameplate 110.55 kWh is a consistency check on those nominals, not
an independent confirmation of them. Happy to drop the percentage and keep only the raw capacity if you
would rather not bake in the assumption.

## 2. DTC descriptions for the BPCM

The integration already reads DTCs through `UdsCanBattery` and already sets `dtc = &datalayer_battery->dtc`,
but with no description file the web UI renders every code as "Unknown". This adds the 112 BPCM codes from
the 2024 Ram ProMaster EV service documentation, which uses the same pack.

- Keyed by the 7-character SAE string the renderer emits, so `P0AA1-00` becomes `"P0AA100"`.
- **BPCM codes only.** The integration queries `0x7E7`, so the IDCM and EVCU codes in the same index can
  never come back here; leaving them out keeps the file lean as `web_data/dtc/README.md` asks.
- Descriptions are left in the documentation's original uppercase rather than title-cased, to avoid
  mangling `HV`, `EV`, `HYBRID/EV` and the `A`/`B` designators.
- Passes `tools/validate_dtc_json.py` with 0 errors and 0 warnings.

Same pattern as `mg_dtc.json` and `renault_zoe_gen1_dtc.json`: one data file plus a
`get_dtc_json_filename()` override.

Note the loader fetches from `raw.githubusercontent.com/dalathegreat/Battery-Emulator/main/`, so the
descriptions only reach users once this lands on `main`; until then the page's file picker loads a local
copy.

## Testing

Decoding was derived and cross-checked against ~1.5 GB of CAN captures (bench, vehicle, CATL 4-channel,
and a complete 80-minute charge to termination). The DTC JSON is validated by the repo's own validator.
Not yet run on hardware — worth a check on a live pack before merge.

### DIFF
--- a/Software/src/battery/STELLANTIS-PRO-ONE-BATTERY.cpp
+++ b/Software/src/battery/STELLANTIS-PRO-ONE-BATTERY.cpp
@@ -41,8 +41,7 @@ void StellantisProOneBattery::
     datalayer.battery.status.soh_pptt =
         (uint16_t)((uint32_t)pack_capacity_ah_tenths * 10000u / NOMINAL_CAPACITY_AH_TENTHS);
     //Energy = capacity x nominal pack voltage. (tenths/10) Ah x (centivolt/100) V = tenths * cV / 1000
-    datalayer.battery.info.total_capacity_Wh =
-        (uint32_t)pack_capacity_ah_tenths * NOMINAL_PACK_VOLTAGE_CV / 1000u;
+    datalayer.battery.info.total_capacity_Wh = (uint32_t)pack_capacity_ah_tenths * NOMINAL_PACK_VOLTAGE_CV / 1000u;
     datalayer.battery.status.remaining_capacity_Wh =
         (uint32_t)((uint64_t)datalayer.battery.status.real_soc * datalayer.battery.info.total_capacity_Wh / 10000u);
   }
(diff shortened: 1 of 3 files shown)


dalathegreat:
Nice work @TPPS999 :raised_hands:

PR #2929, opened 2026-09-06
Fix SOL-ARK-LV-CAN 0x359 over-current sign bug

MaxFreedomPollard:
### What
This PR fixes the two over-current bits in the Sol-Ark LV 0x359 frame, which were compared against the wrong sign of their own limit, and adds unit tests covering them.

### Why
`datalayer.h` says the current sign is positive while charging and negative while discharging. Both checks in `SolArkLvInverter::update_values()` read it the other way round:

`Software/src/inverter/SOL-ARK-LV-CAN.cpp:57` raised discharge over-current (byte 0, bit 7) when `reported_current_dA` rose above `+max_discharge_current_dA`, which is a charging current, so heavy charging was reported as a discharge over-current and heavy discharging was never reported at all.

`Software/src/inverter/SOL-ARK-LV-CAN.cpp:67` raised charge over-current (byte 1, bit 0) when `reported_current_dA` fell below `-max_charge_current_dA`, which is a discharging current. A full pack reports `max_charge_current_dA` 0, so that check reduced to `reported_current_dA <= 0` and stayed true for the whole time the pack sat idle or supplied the house. The Sol-Ark reads that bit as the battery having entered self-protection.

`SOL-ARK-LV-CAN.cpp` was copied from `PYLON-LV-CAN.cpp` and still carries its "See Pylon protocol for example integration" comment. Pylon had the identical inversion and #2920 fixed it two days ago; the Sol-Ark copy did not get the fix.

### How
The discharge check now compares against the negated discharge limit and the charge check against the plain charge limit, matching what #2920 did for Pylon. The +50 dA margin this file already had on the discharge side is kept, and the charge side gets the same margin, so a full idle pack no longer sits exactly on the `0 >= 0` boundary. `test/solark_lv_overcurrent_tests.cpp` builds one 0x359 frame per case through `update_values()` and `transmit_can()` and asserts the two bits directly, including the full-pack-discharging case, the idle case and both cross-direction cases.

### Types of changes

- same bug as #2919, in the Sol-Ark copy of the Pylon code

- not applicable, no user facing behaviour or setting changes

### Checklist:

I do not have a Sol-Ark, so this is verified in the host unit test suite rather than on hardware. Built and run with `cd test && mkdir build && cd build && cmake .. && cmake --build .` then `ctest`, the same commands as `.github/workflows/unit-tests.yml`. On unmodified `main` six of the seven new cases fail. With the fix the whole suite is 263 tests, 262 passed, 1 pre-existing skip, and `ctest` reports `100% tests passed, 0 tests failed out of 263`. Also ran the workflow's order independence step, `./tests` plus `--gtest_shuffle` with seeds 1, 7, 42, 1234 and 99999, all green, and `pre-commit run --files` on both changed files, which clang-format passes without reformatting.

### DIFF
--- a/Software/src/inverter/SOL-ARK-LV-CAN.cpp
+++ b/Software/src/inverter/SOL-ARK-LV-CAN.cpp
@@ -54,7 +54,10 @@ void SolArkLvInverter::update_values() {
   SOLARK_359.data.u8[7] = 0x00;  //Unused, should be 00
 
   // Protection Byte 1 Bitfield: (If a bit is set, one of these caused batt self-protection mode)
-  if (datalayer.battery.status.reported_current_dA >= (datalayer.battery.status.max_discharge_current_dA + 50))
+  // reported_current_dA is positive while charging and negative while discharging, so the discharge
+  // over-current check has to compare against the negated discharge limit. Comparing against the
+  // positive limit tested a charging current against the discharge limit instead.
+  if (datalayer.battery.status.reported_current_dA <= -1 * (datalayer.battery.status.max_discharge_current_dA + 50))
     SOLARK_359.data.u8[0] |= 0x80;
   if (datalayer.battery.status.temperature_min_dC <= BATTERY_MINTEMPERATURE)
     SOLARK_359.data.u8[0] |= 0x10;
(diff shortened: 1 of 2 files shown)


dalathegreat (approved):
Massive thanks for the contribution 🙌

PR #2930, opened 2026-09-06
Support DNS lookup for Syslog servers

ngehrsitz:
### What
I noticed that syslog unlike MQTT does not support hostnames. This PR fixes that. The main downside of using a DNS record is that the lookup takes time and can fail. But on a healthy network this should not be an issue since there is a built-in DNS chache.
While testing I also noticed that syslog does not retry send failures. I will open a follow-up PR for that after this one.

### Types of changes

### Checklist:

  - x] The code change is tested and works locally.

> [!TIP]
> 

### DIFF
--- a/Software/src/devboard/utils/logging.cpp
+++ b/Software/src/devboard/utils/logging.cpp
@@ -2,8 +2,8 @@
 #include "../../datalayer/datalayer.h"
 #include "../sdcard/sdcard.h"
 
+#include <NetworkUdp.h>
 #include <WiFi.h>
-#include <WiFiUdp.h>
 #include "../network/hostname.h"        // active_hostname()
 #include "../network/network_status.h"  // network_connected()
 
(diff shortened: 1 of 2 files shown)


dalathegreat:
@nagyrobi any input?

nagyrobi:
I left out DNS on purpose to steer people into using IP addresses instead of DNS on a local network, which is just waste of CPU cycles imho. But it can be added, I'd just discourage anybody taking a step forward to put such resources on public internet.

> While testing I also noticed that syslog does not retry send failures. I will open a follow-up PR for that after this one.

It does backlog the messages that couldn't be sent, and will send them (prepended with ms elapsed since boot) as soon as network is up again (similar to what it does at boot before network being up). There's no wallclock in BE so the date/time is added by the syslog server when UDP packet arrives there.

ngehrsitz:
I use internal DNS names in my network just so that I don´t have to remember/type IPs.
In terms of performance only the error case will be worse. If DNS resolves correctly you will permanently hit the DNS cache thus there won´t be any performance penalty in working setups that use a DNS name.
What syslog currently does is check for `syslog_online`
https://github.com/dalathegreat/Battery-Emulator/blob/6bc71761dcc377599327c75860cba9f6abdbff9b/Software/src/devboard/utils/logging.cpp#L170-L193
But if there is an error inside of `syslog_send` then the send won´t be retried and we will drop the logs. That´s what I have another PR prepared for.

nagyrobi:
Actually, very happy to see I’m not the only one using the Syslog

PR #2932, opened 2026-09-07
Optional Inverter Modbus gateway (RS485 ⇄ Modbus-TCP)

hack-un:
# Optional Inverter Modbus gateway (RS485 ⇄ Modbus-TCP)

## Summary
When the inverter link uses **CAN**, the board's RS485 port sits idle. This PR adds an
optional feature that turns that idle RS485 into a **transparent Modbus gateway** to the
device wired to it (typically the inverter's Modbus/COM port): an eModbus **RTU master**
owns the bus, and a **Modbus-TCP server** forwards network requests to it. This gives
local Modbus access to the inverter (Home Assistant, custom dashboards, local control)
without any extra hardware.

Per review feedback, the feature is gated on `#ifndef SMALL_FLASH_DEVICE`: it is
compiled **out entirely on 4 MB / SMALL_FLASH_DEVICE boards** (zero flash cost there)
and compiled **in by default on boards with enough flash**. It is still **inactive at
runtime** until enabled in settings, so there is no behavior change out of the box —
until a build enables it, and it never activates at runtime unless the RS485 bus is free.

## Design / principles
- **Protocol-agnostic passthrough** — the gateway forwards function codes (read FC03/04,
  write FC06) to the RTU device and returns the raw response. No per-inverter register
  knowledge lives here; meaning stays with the client.
- **Single bus owner** — one `ModbusClientRTU` owns UART2 (reuses the existing RS485
  init / DE handling). The TCP server and any future on-board logic submit through it.
- **RS485-only** — the gateway activates solely on boards that expose an RS485
  interface (`RS485_RX_PIN()/RS485_TX_PIN()` not `GPIO_NUM_NC` in the HAL); on
  boards without RS485 it's inert even where it is compiled in.
- **Mutual exclusion with the battery-slave role** — the gateway only starts when the
  active inverter interface is **not** RS485/Modbus (i.e. a CAN link, RS485 idle). It
  never contends with `ModbusInverterProtocol`.
- **Secure by default** — disabled out of the box; reads over TCP are **read-only**
  (write FCs are not registered); writes require an authenticated HTTP endpoint.
- **Fail-safe** — TCP server binds only after WiFi is up; on fault the bus goes quiet.

## Security model (all on-device, no external infra)
- **Off by default**, read-only by default.
- **Writes** go through `POST /modbus/write` (behind the existing web auth), signed with
  **HMAC-SHA256(secret, "id:addr:val:nonce")** + a monotonic **nonce** (anti-replay).
  The raw Modbus-TCP port stays read-only. Secret is auto-generated on first boot.
- **IP allowlist** for writes; **audit ring** of the last write attempts (`GET /modbus/audit`).
- Value/register whitelisting is intentionally left to the client (it owns the register map).

## Configuration
Native section **“Inverter Modbus gateway”** on the standard `/settings` page, stored in
`BatteryEmulatorSettingsStore` (keys `GWENAB / GWWRITE / GWPORT / GWBAUD / GWSLAVES /
GWSECRET / GWIPALLOW`) — so it's saved by `/saveSettings`, cleared by factory reset, and
applied on reboot like every other setting. No separate config page.

## Footprint
Measured on **LilyGo T-CAN485** (`lilygo_330`, esp32dev, min_spiffs, `SMALL_FLASH_DEVICE`):
baseline **95.1%** flash → with the feature **~96.3%** (**+~24 KB**), static RAM +~0.8 KB.
**Zero cost on SMALL_FLASH_DEVICE / 4 MB boards** (compiled out). On roomier boards the
code is compiled in but stays inactive until enabled in settings.

## eModbus additions
The vendored `lib/eModbus-eModbus/` is trimmed to the RTU server. This PR adds the
**master** (`ModbusClient.*`, `ModbusClientRTU.*`) and the **sync TCP server**
(`ModbusServerTCPtemp.h`, `ModbusServerWiFi.h`) from **eModbus v1.7.4** (byte-identical to
upstream), plus a **no-op `Logging.h` shim** so those files compile with logging stripped
to nothing — keeping the vendored copy's existing zero-logging style. Happy to change the
vendoring approach to whatever you prefer.

## Testing
Verified on hardware: LilyGo T-CAN485 emulating a battery over CAN, RS485 wired to a
**GoodWe GW10KN-ET** Modbus port. Reads through the gateway match direct Modbus-TCP to the
inverter (grid voltage, work mode, PV, etc.). HMAC-signed writes succeed; bad signature →
401, replayed/stale nonce → 409, raw-TCP writes rejected.

## Files
- `Software/src/communication/modbus_gateway/modbus_gateway.{h,cpp}` (new module)
- `Software/src/lib/eModbus-eModbus/{ModbusClient,ModbusClientRTU}.{h,cpp}`,
  `ModbusServerTCPtemp.h`, `ModbusServerWiFi.h`, `Logging.h` (eModbus 1.7.4 master + TCP server + shim)
- `Software/Software.cpp` — `modbus_gateway_init()` after `setup_inverter()`, `modbus_gateway_loop()` in `core_loop`
- `Software/src/devboard/webserver/webserver.cpp` — register `/modbus/write` + `/modbus/audit`; add GW* names to the save lists
- `Software/src/devboard/webserver/settings_html.cpp` — settings-card + processor cases

## Notes for review
- Built against the **12.4.0** release snapshot; the three edited files may need a trivial
  re-apply onto current `main`. The new files apply cleanly.
- Feature name / setting keys / default port (502) — open to your conventions.

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -653,7 +653,7 @@ void core_loop(void*) {
       previousMillisUpdateVal = currentMillis;  // Order matters on the update_loop!
       START_TIME_MEASUREMENT(values);
       modbus_gateway_loop();  // Bind Modbus-TCP gateway once WiFi is up (no-op otherwise)
-      update_pause_state();  // Check if we are OK to send CAN or need to pause
+      update_pause_state();   // Check if we are OK to send CAN or need to pause
 
       // Fetch battery values
       if (battery) {
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -12,6 +12,7 @@
 #include "src/communication/can/comm_can.h"
 #include "src/communication/contactorcontrol/comm_contactorcontrol.h"
 #include "src/communication/equipmentstopbutton/comm_equipmentstopbutton.h"
+#include "src/communication/modbus_gateway/modbus_gateway.h"
 #include "src/communication/nvm/comm_nvm.h"
 #include "src/communication/precharge_control/precharge_control.h"
 #include "src/communication/rs485/comm_rs485.h"
@@ -651,6 +652,7 @@ void core_loop(void*) {
     if (currentMillis - previousMillisUpdateVal >= INTERVAL_1_S && loopPhase == 1) {
       previousMillisUpdateVal = currentMillis;  // Order matters on the update_loop!
       START_TIME_MEASUREMENT(values);
+      modbus_gateway_loop();  // Bind Modbus-TCP gateway once WiFi is up (no-op otherwise)
       update_pause_state();  // Check if we are OK to send CAN or need to pause
 
       // Fetch battery values
@@ -780,6 +782,11 @@ void setup() {
   setup_charger();
   setup_inverter();
   setup_battery();
+
+  // Optional Modbus gateway (build-gated). Must run after setup_inverter() so the
+  // free-RS485 gate can read the active inverter interface type. No-op if the bus
+  // is busy or the feature is compiled out.
+  modbus_gateway_init();
 
   /* Some battery types mandate the SOC-based charge power taper. Enforce at
      runtime regardless of stored settings, and restrict the start SOC to
--- a/Software/src/communication/modbus_gateway/modbus_gateway.cpp
+++ b/Software/src/communication/modbus_gateway/modbus_gateway.cpp
@@ -94,12 +94,14 @@ static std::vector<uint8_t> parse_slave_ids(const String& csv) {
   int start = 0;
   while (start < (int)csv.length()) {
     int comma = csv.indexOf(',', start);
-    if (comma < 0) comma = csv.length();
+    if (comma < 0)
+      comma = csv.length();
     String tok = csv.substring(start, comma);
     tok.trim();
     if (tok.length()) {
       int v = tok.toInt();
-      if (v >= 1 && v <= 247) ids.push_back((uint8_t)v);
+      if (v >= 1 && v <= 247)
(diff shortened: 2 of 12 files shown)


dalathegreat:
We recently closed a similar build: https://github.com/dalathegreat/Battery-Emulator/pull/2330

The comment still stands; _a Modbus bridge isn't really directly related to Battery Emulator's role (you could just use an extra hardware gadget), but enough people seem interested that it is probably worth it, as long as we can keep the code size and performance impact low enough._

- Get the implementation to not consume flash for 4MB devices
- Remove the MODBUS_GATEWAY define, replace it with SMALL_FLASH_DEVICE is most likely a sensible idea

hack-un:
Thanks for the review! Made both changes:

- **Removed the `MODBUS_GATEWAY` define** and gated the whole feature on `#ifndef SMALL_FLASH_DEVICE` instead. It's now **compiled out entirely on 4 MB / `SMALL_FLASH_DEVICE` boards** (zero flash cost there) and compiled in on boards with enough flash.
- On the boards where it is compiled in, it stays **inactive at runtime** until an admin enables it on the `/settings` page, so there's no behaviour or performance change out of the box.

Additional safety gates so it never interferes: it only activates when the board actually has an RS485 interface (`RS485_RX/TX_PIN()` present in the HAL) **and** the active inverter link is not RS485/Modbus (i.e. CAN, RS485 idle) — so it never contends with `ModbusInverterProtocol`.

The size-report bot should now show ~0 delta on the 4 MB boards. Let me know if you'd like anything else adjusted.

PR #2933, opened 2026-09-07
Stellantis eCMP multi-battery support fix + optional BMS contactor interlock

pauLTU3:
> **Note:** most of this code was written with the help of Claude (AI), guided
> and tested by me on my own hardware. It builds and works on my 3-battery
> setup, but please review it carefully - I do not fully trust AI-written code
> yet and I could easily have missed something. I am happy to change anything.

### What

Makes the Stellantis eCMP battery work properly when you run 2 or 3 eCMP packs
at the same time. The project already supports multi-battery; the eCMP part
only half worked. Each point below is a separate problem and its fix.

---

**1. "More Battery Info" page only worked for battery 1**

Problem: the eCMP page used one shared data block. Battery 2 and 3 got a NULL
pointer instead, so they never filled it in, and the page showed "Advanced
detailed info is limited to the Main Battery".

Fix: three separate data blocks (`stellantisECMP` / `_2` / `_3`), same as how
the Nissan LEAF already does it. Battery 2 and 3 each get their own, so their
data can't overwrite battery 1's.

**2. "Contactor opened" event didn't tell you which battery**

Problem: all 3 packs used the same event.

Fix: added `EVENT_CONTACTOR2_OPEN` and `EVENT_CONTACTOR3_OPEN`. The condition
that triggers it is unchanged.

**3. Battery 2 and 3 cards showed battery 1's numbers**

Problem: on the main page, Scaled SOC / scaled total capacity / scaled
remaining capacity were calculated once for battery 1 and then just copied onto
the battery 2 and 3 cards. Battery 1's card showed the combined
(battery1 + battery2) capacity as if it were its own. Battery 3's card showed a
leftover default of 30 kWh.

Fix: each card now calculates its own numbers from that battery's own real SOC
and capacity. This is display only - nothing the inverter uses is changed here.

**4. Cards didn't show the configured (Settings) limits**

Fix: added a "Configured max charge / discharge" line to every battery card. It
shows the value from Settings, the same on every card, always visible even
during a fault. The existing live "Max ... power" lines stay.

**5. No contactor state on the main page**

Fix: added "BMS contactors: CLOSED / OPEN (reason: N) / UNKNOWN" to each eCMP
card. It reads the battery's own "Contactor positive" and "Contactor negative"
feedback (PID 0xD44D / 0xD44C, 1 = closed, 0 = open). Shows UNKNOWN if the CAN
is missing or the value is old - it never shows a stale CLOSED.

**6. Contactor state got stuck during a fault**

Problem: the eCMP diagnostic polling stops completely when the system is in
FAULT. Contactors usually open *because* of a fault, so the contactor readout
froze on its old value ("CLOSED") right when it mattered.

Fix: during a fault the driver keeps asking for just 3 PIDs (contactor positive,
contactor negative, opening reason). Everything else stays stopped. These are
read-only requests - they don't command anything.

**7. Battery 3 was left out of the inverter's capacity total (bug)**

Problem: in `update_calculated_values()`, when SOC scaling is on, the code adds
battery 2 into the total the inverter sees but there was no code for battery 3.
So with 3 packs the inverter was told it had less capacity than it really did.

Fix: added the missing battery 3 block, same as the battery 2 one. SOC %,
current and power limits are not changed.

**8. New optional Setting: "Require battery BMS contactors closed"**

eCMP only, off by default, not built on 4 MB boards (they can't run 3 CAN buses
anyway).

- Won't start the GPIO contactor precharge until all eCMP packs report their
  BMS contactors closed, and have stayed closed for about 8 seconds. This stops
  a cold 3-pack start from starting precharge too early and tripping straight
  back off.
- If a pack opens its BMS contactors while running, the emulator pauses (power
  to 0), waits for the current to drop, then opens the GPIO contactors. It's a
  warning, not a fault, so no reboot is needed - it recovers on its own once
  every pack is closed again.
- The "waiting" event disappears from the events page once it clears, instead
  of staying as an old entry.

Side effect I noticed while testing: starting all 3 packs is easier now. Before,
after the contactors had opened I usually had to clear events and power-cycle
the whole emulator, and it wasn't obvious why it eventually came back. Now it is
usually just: Open Contactors, Close Contactors, reboot the emulator, wait - and
the system comes up. No full power cycle needed most of the time.

---

### Why
With 3 eCMP packs: battery 2 and 3 diagnostics were missing, the dashboard
showed battery 1's numbers everywhere, the inverter was told the wrong total
capacity, and there was no way to see which battery opened its contactors.

### How
See the list above. The CAN/UDS messages, the DID decoding, precharge timing,
welding protection, the equipment-stop handling and the 60 s CAN timeout are
not touched. With the new Setting off, and on any non-eCMP battery, GPIO
contactor control works exactly as before. Tested on a real system with 3 eCMP
packs. Builds clean for `lilygo_2CAN_330` and the 4 MB `esp32devkit_330`.

### Types of changes

- [ ] Breaking change
- [ ] Code quality improvements

### Checklist:

### DIFF
--- a/Software/Software.cpp
+++ b/Software/Software.cpp
@@ -506,6 +506,27 @@ void update_calculated_values(uint32_t currentMillis) {
           datalayer.battery2.status.reported_remaining_capacity_Wh;
     }
 
+    if (battery3) {
+      // If battery info is valid
+      if (datalayer.battery3.info.total_capacity_Wh > 0 && datalayer.battery.status.real_soc > 0) {
+
+        datalayer.battery3.info.reported_total_capacity_Wh = scaled_total_capacity;
+        // Scale remaining capacity based on scaled SOC
+        datalayer.battery3.status.reported_remaining_capacity_Wh = (scaled_total_capacity * scaled_soc) / 10000;
+
+      } else {
+        // Fallback if scaling cannot be performed
+        datalayer.battery3.info.reported_total_capacity_Wh = datalayer.battery3.info.total_capacity_Wh;
+        datalayer.battery3.status.reported_remaining_capacity_Wh = datalayer.battery3.status.remaining_capacity_Wh;
+      }
+
+      //Triple battery: fold battery3 into the battery1 aggregate too, so the inverter sees all
+      //three packs as one large battery (matches the soc_scaling_active == false path below).
+      datalayer.battery.info.reported_total_capacity_Wh += datalayer.battery3.info.reported_total_capacity_Wh;
+      datalayer.battery.status.reported_remaining_capacity_Wh +=
+          datalayer.battery3.status.reported_remaining_capacity_Wh;
+    }
+
   } else {  // soc_scaling_active == false. No SOC window wanted. Set scaled SOC & capacity to same as real.
     datalayer.battery.status.reported_soc = datalayer.battery.status.real_soc;
 
--- a/Software/src/battery/BATTERIES.cpp
+++ b/Software/src/battery/BATTERIES.cpp
@@ -460,7 +460,8 @@ void setup_battery() {
           battery2 = new CmpSmartCarBattery(&datalayer.battery2, can_config.battery_double);
           break;
         case BatteryType::StellantisEcmp:
-          battery2 = new EcmpBattery(&datalayer.battery2, can_config.battery_double);
+          battery2 =
+              new EcmpBattery(&datalayer.battery2, &datalayer_extended.stellantisECMP_2, can_config.battery_double);
           break;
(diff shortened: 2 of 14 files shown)


dalathegreat:
@pauLTU3 , I just started on refactoring the eCMP for DTC Read/Clear, and I think I can fix this at the same time while SAVING flash space, instead of adding 3100 bytes. Let me have a go before we decide how to proceed

dalathegreat:
@pauLTU3 , #2939 has been merged, feel free to rebase this PR

PR #2936, opened 2026-09-07
Tesla: Update Model 3/Y DTC json

greenoem:
### What
This PR updates the Model 3/Y "DTC" json list based on the alert matrices from early Tesla firmware all the way to mid-2026.

### Why
The existing json looks to have been based on old 2019 firmware, and therefore most packs running BE could be showing incorrect or incomplete alerts.

### How
Keeps the existing order/IDs, notes both the new and old firmware definition (if applicable), and adds new alerts at the bottom of the file. The BE Tesla code needs updating separately to pick up the new alerts, which this PR does not cover.

### Types of changes


dalathegreat (approved):
Nice work! 🙌

PR #2947, opened 2026-09-09
Fix phantom BYD Atto 3 temperature sensor 13

tlm:
### What

Fix the BYD Atto 3 advanced page displaying a phantom temperature
sensor 13 at a constant -40°C.

### Why

The driver allocated 13 temperature slots, but the 0x43C decoder only
populates 12: six readings from each of two multiplexer values.
The thirteenth slot therefore retained its initial value of -40°C.

This was previously hidden by filtering out -40. When the renderer
changed to filter 215, the decoded 0xFF unused-channel marker, the
unpopulated slot became visible as an apparent temperature reading.

### How

- Reduce the driver and BYD extended datalayer arrays to 12 entries.
- Replace individual temperature assignments with memcpy.
- Render temperatures in a loop bounded by the array size.
- Stop at the first unused-channel marker, assuming unused channels
  are trailing entries.

Temperature initialisation, CAN decoding and the separate battery
minimum/maximum temperature reporting are unchanged.

### Types of changes

No existing issue found. Observed on a live BYD Atto 3 installation.

Not applicable; this corrects existing diagnostic display behaviour.

### Checklist:

### Validation

- Confirmed the original symptom on a live BYD Atto 3 installation.
- Traced the cause through the decoder, datalayer and renderer.
- Whitespace checks passed.
- Firmware build and runtime validation of this patch are pending.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -318,7 +318,8 @@ void BydAttoBattery::
     datalayer_bydatto->insulation_valid = battery_insulation_valid;
     datalayer_bydatto->iso_status_valid = (last_35E_ms != 0) && ((millis() - last_35E_ms) < 3000);
     datalayer_bydatto->iso_measurement_active = battery_iso_measurement_active;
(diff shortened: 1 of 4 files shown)


dalathegreat (approved):
Massive thanks for the contribution @tlm 🙌

PR #2951, opened 2026-09-09
Add BYD Atto 3 external DC voltage telemetry

tlm:
## Summary

Add external DC voltage telemetry for BYD Atto 3 batteries, showing the
voltage on the connected equipment side separately from pack voltage.

Decode bytes 3–4 of CAN frame `0x35E` as a little-endian value in
decivolts, based on captured readings.

## Changes

- Display **External DC voltage** on the battery status page.
- Publish `external_dc_voltage` over MQTT in volts.
- Add an **External DC Voltage** sensor to Home Assistant discovery.
- Treat readings as stale after three seconds without a checksum-valid
  `0x35E` frame. The status page shows unavailable, and MQTT omits the
  value so Home Assistant shows unknown.
- Preserve valid zero-volt readings and independent data for dual BYD
  battery setups.

The value is telemetry-only. Pack voltage reporting, precharge and
contactor control are unchanged.

Screenshot 2026-09-10 at 9 10 53 am

## Testing

Manual testing on the battery confirms that opening the contactors makes
external DC voltage drop low, and closing them shows the voltage rising
again as precharge completes.

Added unit tests covering:

- Captured voltage samples and zero-volt readings.
- Status-page output.
- Checksum rejection, stale readings and recovery.
- Independent readings for two batteries and timer wraparound.

`git diff --check` passed. Unit tests and firmware builds have not been
run in the local agent environment because CMake and PlatformIO were
unavailable. MQTT discovery and dual-battery publishing still require
runtime verification.

### DIFF
--- a/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
+++ b/Software/src/battery/BYD-ATTO-3-BATTERY.cpp
@@ -314,6 +314,9 @@ void BydAttoBattery::
     datalayer_bydatto->SOC_polled = BMS_SOC;
     //Resolved voltage, not raw 0x438, so the page matches the main view when 0x438 is rejected
     datalayer_bydatto->pack_voltage_dV = datalayer_battery->status.voltage_dV;
+    datalayer_bydatto->external_dc_voltage_dV = battery_external_dc_voltage_dV;
+    datalayer_bydatto->external_dc_voltage_valid =
+        last_35E_ms != 0 && static_cast<uint32_t>(millis() - last_35E_ms) < 3000;
     datalayer_bydatto->insulation_ohm_per_volt = battery_insulation_ohm_per_volt;
     datalayer_bydatto->insulation_valid = battery_insulation_valid;
     datalayer_bydatto->iso_status_valid = (last_35E_ms != 0) && ((millis() - last_35E_ms) < 3000);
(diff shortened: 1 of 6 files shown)


dalathegreat:
@demon1300 might be the best to review this change

Software/src/battery/BYD-ATTO-3-BATTERY.cpp
```diff
@@ -314,6 +314,9 @@ void BydAttoBattery::
     datalayer_bydatto->SOC_polled = BMS_SOC;
     //Resolved voltage, not raw 0x438, so the page matches the main view when 0x438 is rejected
     datalayer_bydatto->pack_voltage_dV = datalayer_battery->status.voltage_dV;
+    datalayer_bydatto->external_dc_voltage_dV = battery_external_dc_voltage_dV;
+    datalayer_bydatto->external_dc_voltage_valid =
+        last_35E_ms != 0 && static_cast<uint32_t>(millis() - last_35E_ms) < 3000;
     datalayer_bydatto->insulation_ohm_per_volt = battery_insulation_ohm_per_volt;
     datalayer_bydatto->insulation_valid = battery_insulation_valid;
     datalayer_bydatto->iso_status_valid = (last_35E_ms != 0) && ((millis() - last_35E_ms) < 3000);
```
demon1300, line 322: Isn't this the same check `iso_status_valid` does a couple of lines below? Could probably just share a local between them. Don't think the cast is needed either, `millis()` is already 32-bit.

demon1300 (approved):
Had a look at this against some vehicle logs and the decode holds up — reads near zero with the contactors open, follows the precharge ramp, and sits about a volt above pack when closed, which is about what you'd expect across the contactor and cable. Doesn't just mirror 0x441 either, so it looks like a real BMS-side measurement rather than an echo of what we're sending.

Couple of things, neither blocking:

- I saw the odd 457.8 V sample in the logs while the pack was shutting down/waking. Rare, but the page shows the raw value so it might flash something silly occasionally.
- This adds to /advanced, which has a history of quietly truncating when it gets long. Only a few bytes here so probably fine, just flagging it.

Only other thought — we already call this node link voltage in the 0x441 transmit code, and I think the i3 driver uses DC link voltage too. Wonder if it's worth matching that here, mostly because the MQTT key and HA entity name are painful to change once people have history against them. Not fussed either way if you prefer external.

Tests are thorough, nice work.

tlm:
Ta, i'll get this feedback done in the next few days.

PR #2963, opened 2026-09-13
Fix Tesla extended data sharing between double batteries

nizmot:
### What
Double-battery Tesla setups no longer share one extended-data struct between the two packs.

### Why
`datalayer_extended.tesla` was a single global struct, so both battery instances overwrote each other's "More Battery Info" data (contactor state, serial numbers, faults, etc.). Also fixed: battery 2's Advanced page showed "limited to Battery 1" instead of its own data.

### How
Added `tesla_2`, routed the driver and HTML renderer through a per-instance pointer instead of the shared global. Same pattern already used for BYD Atto 3 / Kia64FD.

> [!TIP]


nizmot:
Friendly ping on this one 🙂

I know the diff looks big (955 insertions / 973 deletions in `TESLA-BATTERY.cpp` alone), but it's almost entirely only line changes in that file `datalayer_extended.tesla.X` → `datalayer_tesla->X` (the new per-instance pointer), nothing else.  Same for `TESLA-HTML.cpp`: it's the exact body that used to live in `TESLA-HTML.h`, moved out and pointer-routed, no logic changes.

Happy to split it into smaller commits or add more detail if that'd help review.

PR #2964, opened 2026-09-13
Fix Renault Zoe Gen2 extended data sharing between double batteries

nizmot:
### What
Double-battery Renault Zoe Gen2 setups no longer share one extended-data struct between the two packs.

### Why
Both battery instances wrote SOC, SOH, temps, balance data, and the NVROL reset flag into the single shared `datalayer_extended.zoePH2`, silently overwriting each other's data. Also fixed: battery 2's Advanced page showed "limited to Battery 1" instead of its own data.

### How
Added `zoePH2_2`, routed reads/writes through a per-instance pointer instead of the shared global, and enabled `renders_own_battery_data()` so battery 2's page renders instead of showing the notice. Same pattern already used for BYD Atto 3 / Kia64FD.

> [!TIP]


dalathegreat (approved):
Nice, thanks for the fix @nizmot 🙌

PR #2968, opened 2026-09-13
Stellantis Pro One: read the charge and discharge limits from 0x281, 0x285 and 0x359

TPPS999:
Follow-up to #2926 and #2928. Two files, no new dependencies.

## What the frames carry

`0x281` and `0x285` each hold **three 16-bit big-endian currents in 0.1 A**, bytes 6-7 always zero, `0x1FFF` as the invalid sentinel. Read that way they taper in opposite directions with SOC:

| SOC | `0x285` charge | `0x281` discharge |
|---|---|---|
| 49.8 % | 476.0 A | 436.9 / 273.9 / 270.8 A |
| 68.6 % | 367.2 A | 458.1 / 287.7 / 282.7 A |
| 86.7 % | 219.0 A | 486.2 / 303.2 / 299.2 A |
| 92.5 % | 37.0 A | 490.4 / 306.3 / 302.6 A |
| 98.4 % | 0 | 492.2 / 308.2 / 304.8 A |

Holds across **four physical packs and seven captures**. Charge falls to zero as the pack fills, discharge rises and saturates — a charge/discharge pair.

The `0.1 A` scale is corroborated by the traction motor, rated **270 KM ≈ 199 kW**, which is a hard ceiling since shaft output cannot exceed battery output. At 0.1 A/bit the peak lands at 183.6 kW = 92 % of that; 0.05 would give 125 KM, too little to feed the motor, and 0.2 would give 499 KM. A residual 8 % gap is unexplained and noted in the code.

Timing the wake ramp per field shows **one allowance rising at 9.47 ± 0.43 A/s and clipping at three different ceilings**, not three independent signals — while still rising, B/A = 0.990 and C/A = 0.980; the final 0.625 / 0.617 appears only after each clips. That is the shape of a limit family at increasing time horizons, as in `ECMP` `0x554` and `STELLANTIS-SMALL-WIDE-4x4` `0x305`. **The durations themselves are unknown** and are not claimed anywhere in the code.

`0x359` bytes 0-1 are a third charge limit, granted only after the car asks on `0x2A5` and withdrawn when it stops asking.

## None of the three is ever the binding constraint

Worth stating plainly, because it is the main weakness of this decode. Through a complete charge to termination:

| t (s) | SOC | charge current | `0x359` limit | utilisation |
|---|---|---|---|---|
| 0 – 2871 | 92.5 → 96.5 % | 14.6 A | 59.0 A | **~25 %** |
| 3837 | 98.0 % | 14.6 A | 36.7 A | 40 % |
| 4460 | 98.4 % | 5.2 A | 30.4 A | 17 % |
| 4683 | 98.4 % | 3.5 A | 29.3 A | 12 % |

`0x359` clearly responds — it tapers 59 → 52 → 37 → 29 → 0 A as the session ends — but the current sits at about a quarter of it throughout, and near the end **the current falls faster than the limit does**, so the taper is coming from somewhere else. The charging here went through the OBC and was limited on the AC side at about 5.5 kW, which is also why `0x285` never binds. On the discharge side the largest sustained current in any capture is **3.3 A** against a ~490 A limit.

So this is a well-supported hypothesis, not a confirmed decode. A driving log would settle it, and would also name the horizons.

## Three truncating reads fixed on the way

| Field | Was | Result |
|---|---|---|
| `0x281` field 1 | byte 1 alone, byte 0 discarded | `0x1329` = 4905 came out as **41** |
| `0x281` field 2 | masked to 12 bits | exposed |
| `0x285` all fields | masked to 12 bits | 4760 observed with bit 12 set → **664** |

The `0x1FFF` sentinel alone shows the fields are wider than 12 bits. This is the same class of silent truncation as the cell-voltage bug in #2926, and the `0x285` case is happening in real captures, not merely possible.

## How the limits are used

`max_charge_power_W` and `max_discharge_power_W` now come from the **lowest** limit being reported — index 2 of each triple, the long-term field — falling back to the existing user override when nothing usable is on the bus.

Deliberate choices, since this decode is still a hypothesis:

- **Zero means "not reported", not "no power".** `0x285` reads zero above roughly 97 % SOC while charging is still in progress, so a zero drops out of the comparison rather than forcing the result to zero.
- **`LIMIT_MAX_DA` caps anything derived from these frames**, so a bad read cannot ask for more than the pack could plausibly deliver.
- **`mandatory_charge_taper()` is left in place.**
- The contactor gate is unchanged: still zero while open or precharging.

## Testing

Not compiled locally — no toolchain on this machine; CI covers the file via `test/CMakeLists.txt`. The decode and the power arithmetic were replayed against real captures: during in-vehicle charging the mapping yields 21.9 kW charge, from the `0x359` grant of 59.5 A at 367.5 V, and 110 kW discharge from `0x281` index 2; with contactors open both read zero; with no `0x2A5` present the charge limit falls back to the traction field as intended, and the sentinel never leaks through.


dalathegreat:
Not sure I wanna merge this until we get more CAN logs

dalathegreat:
Closing as stale, already implemented!

PR #2970, opened 2026-09-14
Egmp checksum, counter and message order

kukumagi:
### What
This PR implements startup-to-work counter continuity for the Kia/Hyundai E-GMP emulated CAN traffic.

The EGMP battery emulation reused the same CAN IDs in both the startup sequence and the steady-state work loop, but the work loop could restart from the template base counter instead of continuing from the last valid startup counter. This change keeps the last observed counter per ID and resumes from that value when the same ID is transmitted in work mode.

### Why
The vehicle expects the counter byte to remain consistent across the startup-to-working transition. If the same ID restarts from a lower or stale counter value after the startup burst, the bus can appear invalid and the contactors may briefly close before dropping again.

This was especially relevant for repeated IDs such as 0x10A, 0x120, 0x19A, and the other EGMP traffic IDs that appear in both startup and steady-state traffic.

### How
- Added tracked per-ID last-counter state in the EGMP battery class.
- On startup frame transmission, the current counter value for any counter-bearing ID is saved.
- In the work loop, if that ID has already been seen, the next counter is generated as the previous value + 1 instead of resetting to the template value.
- This preserves counter continuity across the startup-to-working boundary while keeping the existing checksum logic intact.

### Types of changes

- fixes issue related to EGMP startup/work counter continuity and contactor sequencing

- wiki PR not yet created

### Checklist:

> [!TIP]
>


kukumagi:
Forgot that i had this PR. Since this didnt fix much and this change is also present in a PR that actually fixes something, I am closing this PR.
Check this pr instead: https://github.com/dalathegreat/Battery-Emulator/pull/3007

PR #2969, opened 2026-09-14
Bugfix: Give every battery-driver event its own pack 1/2/3 variant

nagyrobi:
## What
Extends the per-pack event triplets to the 23 remaining events that battery drivers raise, so each one names the pack it came from.
Pack 1 keeps its existing name; packs 2 and 3 take a `_BAT2` / `_BAT3` suffix.

With single battery:
image

With multiple batteries:
image

## Why
Worst case today is a shared clear_event(): a healthy pack clears a warning another pack raised a few lines earlier.
Driver code is per-instance, so in double/triple setups all three packs raised the same shared event — the events page, MQTT and log backends never said which battery.

## How
Moved the 23 events into the contiguous per-battery block and passed `battery_index` at 97 driver call sites.
Level init and the message switch now work off the triplet layout, so neither grew by 46 lines.
New `test/per_battery_events_tests.cpp` pins per-pack resolution, message suffixes and cross-pack independence.

### Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before) 
- [x] Code quality improvements to existing code or addition of tests

### Related issue or feature (if applicable):

- fixes <link to issue>

### Pull request in [Battery Emulator Wiki](https://github.com/dalathegreat/Battery-Emulator-Wiki) updating documentation (if applicable):

- wiki PR <link to PR>

### Checklist:

  - [x] The code change is tested and works locally.

> [!TIP]
> Don't worry if you haven't submitted the Wiki docs PR yet! You can still submit this PR, but please don't forget to also document your changes in the [docs](https://github.com/dalathegreat/Battery-Emulator-Wiki) as the next step. Then, come back here and edit this description with the link of the PR you just created there.
> 
> [You can help test this PR using these steps](https://dalathegreat.github.io/Battery-Emulator-Wiki/setup/contributing/contributing/#downloading-a-pull-request-build-to-test-locally)

PR #2974, opened 2026-09-14
Give every battery pack its own numbers and the installation its own aggregate

nagyrobi:
## What

Adds `datalayer.aggregate`, the whole installation seen as one pack, and feeds every inverter protocol, the safety layer, MQTT and ESP-NOW from it.
`datalayer.battery`, `.battery2` and `.battery3` now hold only their own pack, and the system-wide settings move out of them into `datalayer.battery_settings`.
The main page leads with a combined card; the pack cards below show nothing but their own pack:

image

## Why

`datalayer.battery` was doing two jobs at once — pack 1's data store and the "one big battery" the inverter reads — so it was rewritten in place every cycle and nobody could tell which number belonged to what.
A double setup reported its capacity as pack 1 counted twice, a pack 2 fault never reached the inverter, and any driver reaching its settings through its own pack pointer silently got dead defaults.
Splitting the two roles makes each number answerable to one owner, which is what let the bugs below be found at all.

## How

`update_aggregate_values()` rolls the packs up once per second; `update_aggregate_limits()` runs last, after the safety layer, the SOC taper and the low pass filter, so nothing they zero can be walked back.
Three gates decide what counts: energy on *configured*, measurements on *detected*, the SOC hand-over on *joined* — so a pack that has not closed its contactor never over- or under-reports the installation.
Aggregation rules follow #2464 where it had already solved the same physics: emptiest-pack SOC blended toward the fullest above 90%, weakest-pack health, tightest design-voltage bounds.

---

### Bugs found and fixed

| | Was | Now |
|---|---|---|
| Scaled capacity | Pack 2 was assigned **pack 1's** scaled capacityand pack 3's was never written at all, reading 0 Wh whenever SOC scaling was on — so the installation's scaled total was pack 1 counted twice: 30.5 kWh for a 21.8 + 17.9 kWh pair | With several packs nothing is scaled per pack. The packs report real figures and the SOC window is applied once, to the summed installation: 39.7 kWh × 70% = 27.8 kWh. A single pack is the installation, so it keeps its own scaled figures |
| Pack 2/3 faults | The limit cap ran *before* the safety layer, so a pack 2 or 3 fault that zeroed that pack's limits never reached the inverter | The cap runs last; any pack's zero propagates |
| Zero power, non-zero current | Safety zeroed `max_charge_power_W` but never the matching current — with both filters off the inverter could be told 0 W and 30 A in one frame | Current is derived from the final power |
| Per-pack current limits | Never computed for packs 2 and 3, so the cards and the ESP-NOW pack frame carried 0.0 A. MQTT had no per-pack current-limit entity at all | Derived from each pack's own voltage, and published over ESP-NOW |
| Current rounding | A 19.0 A ceiling became 6697 W at 352.5 V and truncated back to 18.9 A | All conversions round to nearest |
| Per-pack power limits | `max_charge_power_W` on a pack is rewritten in place by safety, taper and filter, so MQTT and ESP-NOW published the system's decision for pack 1 next to raw BMS values for packs 2 and 3 — three kinds of number under one key | Each pack publishes what its BMS asked for; the installation's limits live on the aggregate topic and frame |
| Dead settings copies | `battery2.settings` / `battery3.settings` were never read: BYD Atto 3 got the default charge ceiling as pack 2, Tesla's forced balancing, isolation-clear, BMS-reset and SOC-reset requests wrote to nothing, and MQTT/ESP-NOW limiting-factor flags were always defaults for packs 2 and 3 | One copy at `datalayer.battery_settings` |
| Design voltages | Pack 1's only, so a mismatched pack could be handed a charge ceiling it will not tolerate | Lowest ceiling and highest floor across talking packs |
| SOH, cells, temperatures | Taken from pack 1 alone | SOH from the weakest pack; cell and temperature extremes across every talking pack |
| BYD-MODBUS cold cap | Clamped the datalayer temperature in place to work around the Gen24 refusing to run below −10 °C, so one inverter's workaround rewrote what the web page, MQTT, ESP-NOW and every other inverter read | The cap applies only to the registers this inverter writes |
| Capacity at empty | The SOC window fell back to unscaled capacity at a real 0%, so the reported total jumped from the window to the whole pack | Stays scaled |

### Notes

- Inverter drivers: 857 mechanical replacements to `datalayer.aggregate`, plus 280 for the settings move. Nothing else in those files changed.
- Per-pack MQTT and ESP-NOW now carry unscaled figures, since the SOC window is the installation's aggregate. MQTT drops the duplicate per-pack scaled entities and gains `<name>/info_agr`; ESP-NOW gains `ESPNOW_FRAME_AGGREGATE` with its own keys at `0xB0..0xC3`. A single battery is unchanged throughout.
- The power limit is still capped to the weakest pack rather than #2464's *lowest × pack count*.
- 316 host tests, 315 pass, order-independent. `mqtt.cpp` and `espnow.cpp` are not in the host build and were type-checked separately.
- Depends on #2969 (uses `datalayer.system.info.configured_batteries`).

### Types of changes

- fixes <link to issue>

- wiki PR https://github.com/dalathegreat/Battery-Emulator-Wiki/pull/58

### Checklist:

> [!TIP]
> 


nagyrobi:
Draft because https://github.com/dalathegreat/Battery-Emulator/pull/2969 has to preceede this.

nagyrobi:
> Nice! Love seeing flash save PRs! <3

Well this one is also flash saving in the aspect of how it renders the main page - for multi-batteries cards I removed the triplicated html sections... true that it cancels that out with new MQTT possibilities...

ekholm:
This matches what I've seen working on multi-pack support: `datalayer.battery` doing double duty as pack 1 and as the whole installation is behind several of these bugs, and splitting the roles makes them visible.

One question from reading `battery_aggregate.cpp`. The description lists three gates (energy on configured, measurements on detected, SOC hand-over on joined), but the code has two: the aggregate SOC takes the lowest and highest `real_soc` over every *detected* pack. On main, pack 2/3 only influenced the reported SOC while `battery2/3_allowed_contactor_closing` was true (`Software.cpp:537`/`:542`). With this PR, a pack that is on the bus but not on the DC link (held out by the 1.5 V check, or opened after a fault) drives the SOC the inverter sees: an empty detached pack makes the installation read empty, and a full one lifts the blend above 90%.

Was dropping the join condition for SOC intended? If yes, the "Three different gates" comment and the description could say so. If not, `NotYetJoinedPackStillCounts` has a natural SOC sibling.

nagyrobi:
Hmmm right lemme check

nagyrobi:
Addressed in latest commit

Software/src/inverter/VCU-CAN.cpp
```diff
@@ -30,8 +30,8 @@ void VCUInverter::update_values() {  //Called every 1s
   // dislimit = ((byte0<<2) + (byte1>>6)) * 0.25
   // chglimit = (((byte1&0x3F)<<4) + (byte2>>4)) * 0.25
   // so the raw value = power_in_watts / 250.
-  uint16_t dislimit_raw = static_cast<uint16_t>(datalayer.battery.status.max_discharge_power_W / 250) & 0x3FF;
-  uint16_t chglimit_raw = static_cast<uint16_t>(datalayer.battery.status.max_charge_power_W / 250) & 0x3FF;
+  uint16_t dislimit_raw = static_cast<uint16_t>(datalayer.aggregate.max_discharge_power_W / 250) & 0x3FF;
```
dalathegreat, line 33: VCU is a special case that will only ever use battery1, but I guess no harm in using the aggregate also here

Software/src/datalayer/battery_aggregate.cpp
```diff
@@ -0,0 +1,257 @@
...
+      agg.temperature_max_dC = MAX(agg.temperature_max_dC, pack->status.temperature_max_dC);
+      agg.temperature_min_dC = MIN(agg.temperature_min_dC, pack->status.temperature_min_dC);
+      /* Only a pack that is actually on the link gets to move the installation's SOC */
+      if (pack_joined[i]) {
+        lowest_soc = MIN(lowest_soc, pack->status.real_soc);
+        highest_soc = MAX(highest_soc, pack->status.real_soc);
+      }
+      /* Health follows the weakest pack, like every other limit here. A pack reporting zero
+         has not decoded one yet and is skipped rather than zeroing the installation. */
+      if (pack->status.soh_pptt > 0) {
```
dalathegreat: Is this not a problem since most packs report 99% SOH by default in the datalayer?

Software/src/devboard/espnow/espnow.h
```diff
@@ -132,10 +132,13 @@ enum espnow_type_t {
 #define ESPNOW_LEN_CODE_MAX_INLINE 29
```
dalathegreat, line 132: Does these changes break existing ESPNow installs?

Software/src/devboard/mqtt/mqtt.cpp
```diff
@@ -255,6 +275,35 @@ static const SensorConfig batterySensorConfigTemplate[] = {
     {"charge_term_cell_max_num", "BYD Charge: Termination High Cell #", "", "", supports_byd_autocal_metrics},
     {"charge_term_cell_min_num", "BYD Charge: Termination Low Cell #", "", "", supports_byd_autocal_metrics}};
 
+// The installation as the inverter sees it, published on its own topic when more than one
+// battery is configured. Entity ids get "_multi" where batteries 1, 2 and 3 get "", "_2" and
+// "_3"; the names carry no suffix at all, because the unqualified "SoC" sitting beside "SoC 1"
+// and "SoC 2" is the installation. With a single pack this is never published: it would only
+// repeat battery #1.
+static const SensorConfig aggregateSensorConfigTemplate[] = {
```
dalathegreat, line 283: I guess these MQTT changes are so massive that we need a new major version number after this PR is merged?

dalathegreat (approved):

Software/src/devboard/espnow/espnow.h
nagyrobi: No, they don’t, it’s by design backwards compatible. New fields added, for multibatt totals.
Which worth being added to displays, to show up.

Software/src/devboard/mqtt/mqtt.cpp
nagyrobi: Yes, they break a bit but only for multibatt installs (new entities appear for the aggregated pack, pack 1 values will show its own values). Single batt unchanged.

Software/src/inverter/VCU-CAN.cpp
nagyrobi: Guess so.

dalathegreat (approved):
LGTM 👍

PR #2983, opened 2026-09-16
Locked ecmp temp

TamaMcGlinn:
### What

[This old PR](https://github.com/dalathegreat/Battery-Emulator/pull/2226) is still needed for testing, we have at least 3 different users with the same fault.

### Why

The build artifacts were expired, hence I rebased and resubmitted.


dalathegreat:
@TamaMcGlinn , please test the DTC Read PR up here: https://github.com/dalathegreat/Battery-Emulator/pull/2939

paulheystee:
The DTC read PR is great, but I also want to test this workaround, to see if just setting temp to 0 will bypass the fault. However, building in VSCode gives me errors, I hoped I could just test this PR bin under checks > compile common images. Can that be run, please?

PR #2986, opened 2026-09-17
Streamed logging. Add seconds parameter to log longer, default 10s

kukumagi:
### What
Streamed logging since the 10kb log was kinda useless

### Why
Why does it do it?
I wanted longer log. Currently the streamed logging can have missing frames if the can transmission is too fast, but thats a sacrifice im willing to make.

### How
How does it do it?
Download now streams for default 10s. export_can_log?seconds=120 if you want 2 minute log. And yes you will have to wait for full 2 minutes in that case.

### Types of changes

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as before) 

- fixes <link to issue>

- wiki PR <link to PR>

### Checklist:

  - [ x] The code change is tested and works locally.

> [!TIP]
> 


dalathegreat:
I think we can close this, since we have dump_can available https://dalathegreat.github.io/Battery-Emulator-Wiki/setup/can_related/can_logging/#dump-can

(We just should add a button so people can discover this page without needing to read the Wiki)

This PR also adds +1,884 flash, so not so nice, since we already have dump_can

jonny5532:
Using AsyncWebServer's chunked response and pointers into the existing buffer is a kinda neat way of doing this, probably a bit smaller than the dump_can method. However it's not as fast, AsyncWebServer seemingly doesn't fire the callback often enough - logging the TESLA output, dump_can returns 358 frames/sec, this one 326 frames/sec. Plus the dump_can approach indicates buffer overruns (with blank lines) whereas this drops them silently.

Still pretty tidy though!

PR #2978, opened 2026-09-15
Improvement: Nissan Leaf current measurement precision by averaging and peak safety

nagyrobi:
### What
This PR implements 1-second averaging of Nissan Leaf battery current telemetry.
It also preserves peak current for safety-limit checks.

image

This way we can mitigate the 0.5A resolution problem, we'll have 0.1A precision which will be far better for the taper, and produce less oscillations.

### Why
The Leaf current is received much more frequently than the 1-second datalayer update.
Averaging improves reported current precision, while peak tracking prevents short overcurrent spikes from being hidden.

### How
Current samples from CAN ID `0x1DB` are accumulated in the receive handler and averaged in `update_values()`.
The peak current is tracked separately and used by the machinery protection power checks.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before) 
- [x] Code quality improvements to existing code or addition of tests

### Related issue or feature (if applicable):

- fixes <link to issue>

### Pull request in [Battery Emulator Wiki](https://github.com/dalathegreat/Battery-Emulator-Wiki) updating documentation (if applicable):

- wiki PR <link to PR>

### Checklist:

  - [x] The code change is tested and works locally.

> [!TIP]
> Don't worry if you haven't submitted the Wiki docs PR yet! You can still submit this PR, but please don't forget to also document your changes in the [docs](https://github.com/dalathegreat/Battery-Emulator-Wiki) as the next step. Then, come back here and edit this description with the link of the PR you just created there.
> 
> [You can help test this PR using these steps](https://dalathegreat.github.io/Battery-Emulator-Wiki/setup/contributing/contributing/#downloading-a-pull-request-build-to-test-locally)

PR #2990, opened 2026-09-18
Add generic lifetime Wh/Ah counters with optional NVS persistence

hughes4262:
### What

This PR moves cumulative charged/discharged energy accounting into the generic Battery-Emulator datalayer. Calculated Wh/Ah counters run in RAM for batteries that do not provide native lifetime counters.

Saving calculated counters to NVS is optional and disabled by default. Users can enable it through a web UI checkbox that warns about possible CAN communication interruptions during flash writes. Changes to this setting take effect after a restart.

It adds:

- Generic calculated charged/discharged lifetime energy counters in Wh.
- Generic calculated directional charged/discharged capacity counters in 0.1 Ah.
- Independent handling of native Wh and native directional-Ah capabilities.
- Optional, low-frequency NVS persistence for calculated counters.
- Native directional-Ah mapping for supported BYD and Kia/Hyundai batteries.
- FoxESS EP consumption of the generic counters instead of maintaining separate lifetime counters inside the inverter driver.
- A ±0.5 A deadband for calculated counters, retained in this PR pending further evaluation following the LEAF current-averaging fix in #2978.

### Why

The FoxESS EP implementation aims to provide more complete OEM-like battery integration, including energy statistics and battery cycles in the FoxESS app, rather than basic charging and discharging alone.

Hardware testing established that FoxESS uses cumulative energy and capacity values for:

- 0x1878: absolute charged-plus-discharged energy throughput.
- 0x1879: independent directional charged/discharged capacity.
- 0x187A: cumulative charged/discharged energy.
- 0x1875: equivalent battery cycles.

Previously, calculated lifetime counters were maintained inside the FoxESS EP driver and were lost on reboot or reflash. Restarting them from zero can disrupt the app's daily energy figures, reactivate the Total Charged fallback and reset the reported cycle count.

Moving the accounting to the generic datalayer makes it reusable by other inverter protocols. Optional persistence allows users who need continuity of lifetime statistics to preserve calculated totals across restarts.

Persistent energy statistics are not required for basic FoxESS inverter charging and discharging.

### How

Calculated counters are updated from the central 10 ms path using generic battery voltage and reported current, subject to the existing battery-readiness and contactor-engagement checks.

Positive Battery-Emulator current accumulates charged totals; negative current accumulates discharged totals.

Native Wh and native directional-Ah sources independently bypass the corresponding local calculation, restoration and checkpointing paths.

When persistence is enabled, calculated totals are restored at startup and checkpointed approximately every 24 hours of runtime using the existing batterySettings NVS namespace:

- ENERGY_CHG_WH
- ENERGY_DIS_WH
- CAP_CHG_DAH
- CAP_DIS_DAH

Unchanged values are not rewritten. When persistence is disabled, calculated totals operate in RAM only, without counter-related NVS restore or checkpoint writes.

The persistence setting is read at startup, so enabling or disabling it requires a restart.

The current implementation retains a ±0.5 A deadband for calculated counters only: reported current from -5 dA through +5 dA is excluded from calculated Wh and directional-Ah accumulation. At ±6 dA and above, the full reported current is integrated. Native counters are unaffected.

The deadband predates the LEAF current-averaging change in #2978. Its continued necessity is being assessed separately rather than assuming the earlier idle-current observations still apply.

### NVS write risk and limitations

Writing to internal flash can briefly interrupt processing and may cause CAN overruns on heavily loaded buses. The web UI warns about this risk, and counter persistence is disabled by default.

The 24-hour checkpoint interval limits write frequency but does not eliminate the operational risk or guarantee that all energy accumulated since the previous checkpoint will survive an unexpected restart.

The four totals are currently saved as separate NVS values. Recovery from an interruption partway through a checkpoint has not been proven transaction-safe; values from different checkpoints could potentially be restored together.

These limitations should be considered before enabling persistence on a particular installation.

### Validation

Automated checks on the current PR revision:

- Unit-test job: passed.
- Pre-commit: passed.
- All seven common firmware compilation targets: passed, including Stark CMR (stark_330).
- Firmware-size check: passed.

Energy-counter tests cover calculated and native capability combinations, charge/discharge direction, readiness gating, fractional accumulation, checkpoint timing, persistence disabled by default, restart-required setting changes, restoration, and counter boundaries.

Earlier hardware testing on a FoxESS KH9 with a Stark CMR Battery-Emulator and Nissan LEAF battery confirmed:

- Cumulative charged and discharged energy reporting.
- Independent Total Charged reporting and its observed fallback behaviour.
- Battery Cycles increasing and retaining continuity when saved counters were restored.
- Normal charging and discharging with the generic energy-counter implementation.

The current opt-in persistence revision has passed automated checks but has not yet completed its own hardware-validation sequence. In particular, safe transition of existing saved counters and behaviour during NVS checkpoint writes remain to be verified.

### Types of changes

- [x] New feature
- [x] Code quality improvements and additional tests
- [ ] Breaking change

### Related work

- #2978: LEAF current averaging. The calculated-counter deadband is under separate review in light of this change.

### Documentation

Protocol findings and hardware evidence are maintained separately in the FoxESS EP protocol documentation. Upstream documentation can be updated following review of this implementation.


nagyrobi:
> Hardware testing also showed that the Nissan LEAF can report a sustained one-LSB idle current of ±0.5 A.

This has been addressed in #2978.

hughes4262:
> > Hardware testing also showed that the Nissan LEAF can report a sustained one-LSB idle current of ±0.5 A.
> 
> This has been addressed in #2978.

Thanks — I hadn't realised #2978 had addressed the LEAF 0.5 A quantisation issue when I originally developed and hardware-tested the deadband.

This PR branch is based on current main and already includes #2978, so the averaging fix is present here.

The ±0.5 A deadband predates that change. I'll re-evaluate that part separately; if it is no longer necessary with the averaged LEAF current, I'll remove the deadband from this PR and keep the generic Wh/Ah accounting and persistence changes independent.

nagyrobi:
Are these Wh counters regularly saved to NVS for everybody? Every kind of inverter?

Writing to NVS causes intreruptions, doing so regularly is not beneficial on long term stable operation, which will cause issues and support load.

Imho this should be only implemented for inverters that have this as an absolutely, mandatory requirement. Inverters able to work without unlimited increase of Wh counters should avoid saving any such data to NVS.

hughes4262:
Yes, as it stands now "store_energy_counters()" is called from the central loop.

It doesn't write to NVS for every battery though. If a battery already provides its own lifetime Wh/Ah counters, the calculated persistence path is bypassed. For batteries that don't, the checkpoint is only about once every 24 hours of accumulated runtime, and the existing settings code also avoids rewriting values that haven't changed.

That said, your point is fair: at the moment a battery without native lifetime counters could still end up using the NVS persistence even with an inverter protocol that doesn't actually need it.

The reason I moved the counters into the generic datalayer was based on Dala's earlier direction, so the Wh/Ah accounting wasn't tied specifically to FoxESS EP.

The persistence requirement itself is really coming from FoxESS EP, because those cumulative counters need to stay monotonic across reboots for the daily/lifetime energy, fallback behaviour and cycle count.

So I think the better split is to keep the counters generic in RAM, but make the NVS restore/store opt-in for inverter protocols that actually need persistent lifetime counters. FoxESS EP would be the current one enabling it, and unrelated inverter protocols wouldn't get the extra NVS writes.

Software/src/datalayer/energy_counter.cpp
```diff
@@ -0,0 +1,111 @@
...
+  if (battery->supports_charged_energy() && battery->supports_directional_capacity()) {
+    return;
+  }
+  // Use the existing NVS wear levelling and changed-value-only settings writes.
+  // Fractions and the checkpoint timer are intentionally volatile. A reboot
+  // restores the last checkpoint and starts a new 24-hour runtime interval.
+  BatteryEmulatorSettingsStore settings;
+  const auto& status = datalayer.battery.status;
+  if (!battery->supports_charged_energy()) {
+    settings.saveInt("ENERGY_CHG_WH", status.total_charged_battery_Wh);
```
dalathegreat, line 111: Performing NVM writes can affect things like CAN overruns on heavy bus loads. It essentially halts the core to do the flash, and this needs to be behind a config option that warns the user that saving persistent Wh counters every 24h can be done, but at the risk of disturbing operataion

dalathegreat (changes requested):

Software/src/datalayer/energy_counter.cpp
nagyrobi: I just weigh on what happens if there's no Wh reported to FoxESS? Blocks operation? Is this a must-have for it? How does it behave without?

Or is it just something "nice to have" to show in their app?

Not against, I just wonder the price to pay for this...

hughes4262: The Wh counters aren't essential for basic inverter operation. FoxESS can charge and discharge without persistent lifetime energy statistics.

But the whole point of developing the EP protocol was to emulate an OEM FoxESS battery more fully, rather than just getting the inverter to charge and discharge. That includes the daily and lifetime energy figures, independent Total Charged reporting and battery cycles in the FoxESS app. I've spent a lot of time getting those features working and proving them on real hardware. If someone only needs basic inverter operation, the older FoxESS protocol is still available.

The issue is what happens when Battery-Emulator restarts. FoxESS retains its previous energy history, but if our cumulative counters restart from zero, the daily figures can go negative, the Total Charged fallback can reactivate and the reported cycle count loses continuity. That's why I added persistence.

I understand the concern about NVS writes interrupting CAN communication, especially on hardware with heavier bus loads. That's why I've now made persistence opt-in and disabled by default, with a warning explaining the risk. The generic Wh/Ah counters still run in RAM regardless of that setting. Users who want their FoxESS energy statistics to remain consistent across restarts can choose to enable persistence; everyone else avoids those counter-related NVS writes.

So I agree that persistence shouldn't be imposed on everyone. The new setting gives users that choice while retaining the more complete OEM-like battery reporting that I developed the EP implementation to provide.

I'd appreciate your thoughts on whether there's a better way to preserve these counters while minimising the risk of interrupting CAN communication.

nagyrobi: Reasonable, thank for the explanation.

hughes4262:
Follow-up on the ±0.5 A deadband and #2978:

I've now hardware-tested v71 on my FoxESS KH9 / Stark CMR setup with the LEAF current-averaging fix from #2978 present, but with the calculated-counter deadband removed.

During a period when the battery was expected to be idle, the Battery-Emulator web UI repeatedly reported approximately 196 W of charging. The FoxESS app also showed a charging current of approximately 0.5 A.

Between 15:26 and 17:26 on 20 September, FoxESS Total Charged increased from 64.4 to 64.7 kWh. Approximately 0.1 kWh was genuine charging immediately after the first screenshot, leaving approximately 0.2 kWh of additional reported charging during the idle period. SOC remained at 83%, and Total Discharged remained at 55.3 kWh.

At approximately 392 V, a sustained 0.5 A reading corresponds to about 196 W, or 0.196 kWh per hour. This closely matches the additional reported energy.

Although I cannot establish from the app readings alone that every increment was false, this test shows that unwanted near-zero energy accumulation remains a concern even with #2978 included.

I'll therefore retain the existing ±5 dA deadband in this PR. It applies only to calculated lifetime Wh and directional-Ah counters; native battery counters remain unaffected.

PR #3010, opened 2026-09-22
Bugfix: e-NV200 fall back to the LBC's SOH when no capacity can be read

nagyrobi:
## What
When a pack yields no measured capacity, SOH falls back to the value the LBC publishes, and total capacity is worked out from it as before. The fallback engages only once every group has been polled since the pack came up.

## Why
On packs whose group 0x01 layout the capacity decode doesn't cover (reported on e-NV200), v12.6.0 shows SOH as Unknown and sends the 30 kWh placeholder to the inverter as total capacity.
Waiting for the first poll pass stops degradation-reset packs from briefly publishing their 100% before the measured capacity arrives.

## How
In the no-capacity branch, once `poll_burst_remaining` is 0 and the LBC has reported an SOH, `soh_pptt` takes it and `total_capacity_Wh` = max GIDs × `WH_PER_GID` × SOH.
Remaining capacity stays on the GID count, and a capacity that arrives later takes over.
Two new unit tests cover the wait, the fallback values and the hand-over.

### Types of changes

- fixes <link to issue>

- wiki PR <link to PR>

### Checklist:

> [!TIP]
> 


WJDijkstra:
tested, working,  thanks @nagyrobi !

Software/src/devboard/webserver/webserver.cpp
```diff
@@ -1250,7 +1250,13 @@ String processor(const String& var) {
       else
         content += "<h4 style='color: white;'>SOC: " + String(socRealFloat, 2) + "&percnt;</h4>";
 
-      content += "<h4 style='color: white;'>SOH: " + String(sohFloat, 2) + "&percnt;</h4>";
+      //Unknown until the integration has decoded a state of health, rather than the soh_pptt default
+      //shown as if it had been read from the pack.
+      if (datalayer.battery.status.soh_available) {
```
dalathegreat, line 1274: No other integration than Nissan LEAF sets .soh_available to TRUE, wont this break all other integrations?

nagyrobi: No, it doesn't break them, because the field defaults to true in the struct itself:

bool soh_available = true;   // datalayer.h:156

Other integrations don't need to set it — they get it from the declaration.

Intention is to prepare the possibility for them too, to properly show "Unknown" on frontends until real data arrives on the bus.

nagyrobi: Added a test to ensure everything's alright

dalathegreat (approved):

PR #3013, opened 2026-09-23
Leaf: per-pack current measurement offset

nagyrobi:
Some LBCs report a small non-zero current with the contactors open. Add a "Current measurement offset, dA" setting (-100..+100, default 0) that is added to the averaged 0x1DB current of each pack before it is published.

Battery 1 shows it under "BMS starting sequence request", the 2nd/3rd pack next to their interface selector. Stored in NVM as LEAFCUROFS/2/3; 0 is the same as a missing key, so the key is removed instead of written. A saved value is taken into use at once. The safety extremes stay untrimmed.

### What
Adds a per-pack "Current measurement offset, dA" setting (-100…+100, default 0) to the Nissan LEAF integration.
Battery 1's field sits under "BMS starting sequence request:"; 2ⁿᵈ/3ʳᵈ packs get theirs next to their interface selector.

### Why
Some LBCs report a small current with the contactors open, which makes idle packs look like they're charging or discharging. Trimming it out gives clean readings to assure taper functionality matches reality. Values reflect in the UI, MQTT and inverters.

### How
The offset is added to the 1 s averaged 0x1DB current of each pack, applies immediately on save, and is stored in NVS only when non-zero.
Safety-layer current extremes stay untrimmed; unit tests cover the offset math, per-pack selection and no double-application.

### Types of changes

- [ ] Breaking change (users may need to reconfigure their setups) 

- fixes <link to issue>

- wiki PR <link to PR>

### Checklist:

> [!TIP]
> 


jonny5532:
BTW, a simpler approach might just be a checkbox for 'automatic current offset correction' or something. Then, keep a moving average of the current whilst the contactors are open. When closed, use that as the offset. Thus it is self-calibrating every time you open/close contactors.

(we do this on the Tesla retrofit BMS, works great)

Also these packs presumably have terrible drift, since their internal coulomb count is using the errant current?

nagyrobi:
How come this didn't cross my mind?

PR #3016, opened 2026-09-23
Nissan LEAF🔋automatic current offset correction

nagyrobi:
### What
Adds an "Automatic current offset correction" checkbox to the Nissan LEAF integration, working independently for every pack.
Each pack's More Battery Info Status card shows the offset learned for it.

image

On a double setup:
image

image

### Why
Some LBCs report current with the contactors open, up to 2.5 A seen in the field, which skews power, SOC tracking and inverter data.
The pack calibrates itself every time its contactor opens, with no values to tune.

### How
After boot and from 30 ms after a pack's contactor opens, raw 0x1DB samples are averaged over the last 10 s and subtracted from the published current.
Learns only where contactor control drives that pack's contactor; the value holds while closed and is re-measured on every opening.

### Types of changes

- [ ] Breaking change (users may need to reconfigure their setups) 

- fixes <link to issue>

- wiki PR <link to PR>

### Checklist:

> [!TIP]
> 


nagyrobi:
Suggestion for Wiki:

### Automatic current offset correction

Some Leaf battery packs report a small current even when none can flow; in some cases it even reaches a few amperes. This shows up as phantom charging or discharging, and skews the power and SOC figures passed to the inverter. With **Automatic current offset correction** enabled under Battery config, Battery-Emulator measures this error itself. Whenever a pack's contactors are open, any current the BMS reports can only be sensor offset. After a short settling time, those readings are averaged over up to the last 10 seconds and subtracted from that pack's current from then on. The offset is measured at every boot and again each time the contactors open. It is kept unchanged while they are closed, including during a periodic BMS reset. A second ord third pack that stays disconnected, for example because its voltage differs too much from the other packs, keeps updating its offset. Battery Emulator can only tell that contactors are open when it drives them itself, so this needs **Contactor control via GPIO** enabled, plus **2ⁿᵈ / 3ʳᵈ battery contactor control via GPIO** in double or triple battery setups. Each pack is measured separately, using its own contactors. The value found for each pack is shown as **Automatic current offset** at the bottom of the Status panel on the More Battery Info page. "Unknown" means that pack has not been measured yet, for example because its contactors are not controlled by Battery-Emulator. The setting takes effect after a reboot. Put it simple, to re-calibrate, just reboot Battery Emulator.

nagyrobi:
image

dalathegreat (approved):

